PR #8, opened 2016-04-27
Move `print_err` macro to a common file

Connorcpu:
In order to avoid duplicating macros like that, you can simply put them in a module that gets added to your root first, and has a `#[macro_use]` annotation.


Connorcpu:
I signed it!

raphlinus:
Thanks!

PR #9, opened 2016-04-27
Fix warnings in Xcode 7.3,Swift 2.2

chancyWu:


chancyWu:
I signed it!

raphlinus:
Thanks for the PR. Unfortunately, this breaks building in Xcode 7.2.1, which is what I have installed. At some point it'll make sense to roll it forward, but for now I think warnings are better than forcing a version of Xcode that not everyone might have.

supermarin:
@raphlinus Xcode 7.3.1 is the newest one and there's no reason for people not to update to the newest Xcode / Swift IMO

raphlinus:
Fair enough. I just upgraded my Xcode to 7.3 and this patch makes the build warning-free. Thanks!

Issue #10, opened 2016-04-27
[feature] regular text shortcuts


Haroenv:
some examples: 
- [ ] <kbd>⌘</kbd><kbd>a</kbd> should select everything
- [ ] <kbd>⌘</kbd><kbd>⌫</kbd> removes until the start of the line

cmyr:
Basic keyboard commands are implemented in xi-mac.

PR #12, opened 2016-04-27

rtsuk:
AppKit has facilities for interpreting key events. Here's a partial implementation for discussion. I think this is the right aproach but how to plumb it through to the core deserves some consideration.

This would partially address https://github.com/google/xi-editor/issues/10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,9 @@ If you’re not on a Mac, you can build just the core:
 ```
 
 See the [xi_glium](https://github.com/potocpav/xi_glium) project for an
-experimental front-end in Rust.
+experimental front-end in Rust. There are notes (I wouldn’t call it
+documentation at this point) on the protocol at
+[frontend.md](doc/frontend.md).
 
 ## Design decisions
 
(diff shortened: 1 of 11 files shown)


raphlinus:
Yes, this is the direction I'd like to go. The JSON language to the core should be expanded with lots of small commands generally following the "Responding to Action Messages" methods in NSResponder.

There is some subtlety. In a hypothetical Vim personality, while in command mode, keys should _not_ be interpreted to an "insert" command and then have the letters in that insert reinterpreted as vi-compatible commands.

rtsuk:
Would the Vim personality be implemented in the core? If so, there's a conflict here.

raphlinus:
No, my thinking is that the first line of dispatch of the vim personality is in the front-end. The core will of course do the heavy lifting, and perhaps also the second line of dispatch. This seems broadly compatible with your approach, but will almost certainly require some additional logic in the interpretation of key events. I'm pretty convinced by now that such additional logic is required to handle a vim personality and do input methods correctly in any case.

rtsuk:
Then I will complete this implementation, filling in the core implementation when it is obvious but otherwise leaving them as TBD.

rtsuk:
A rebase went awry. I will restore this tonight.

rtsuk:
Let's see if I can re-open.

rtsuk:
I will have to open a new PR.

PR #13, opened 2016-04-27
`.init` is not needed in Swift and is discouraged

kmikael:


kmikael:
I signed it! My google mail is "mkonutgan@gmail.com”.

PR #16, opened 2016-04-28
Add Project Icon

facekapow:
This PR adds an icon to the project, included in the Xcode build and the readme. The icon design is just an idea I had for an icon.


raphlinus:
Thanks! Not gonna promise this'll be the final icon, but better to have something.

PR #17, opened 2016-04-28
Fix some low-hanging lints from Clippy

shssoichiro:
Focused on objective performance-centric lints, while leaving more
style-based lints for later if the original author approves.


raphlinus:
Thanks. I admit I haven't been running Clippy. I was going to, but just haven't gotten around to it yet.

PR #20, opened 2016-04-28
Small improvements

Pearapps:
Making CoreConnection final, so it can utilize static dispatch which is faster.

Makes the state private


raphlinus:
Closing due to lack of CLA :(

PR #24, opened 2016-04-28
Various code cleanup

shssoichiro:
Move default fns into trait impls
Derive Copy where possible
Clean up a couple of TODOs
Other various clippy fixes


raphlinus:
Thanks, looks good!

PR #30, opened 2016-04-29
Handling Keyboard Actions and Inserting Text, take 2

rtsuk:
New pull request since my rebase attempt went awry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
 <h1 align="center">
-  <br>
   <a href="https://github.com/google/xi-editor"><img src="icons/xi-editor.png" alt="Xi Editor" width="256" height="256"/></a>
-  <br>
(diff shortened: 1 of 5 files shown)


raphlinus:
This looks like a great direction. I'm not going to merge it now because it breaks a lot of existing stuff, but it'll smooth the way to doing lots more editing operations in a clean way.

One question: I notice that F-keys map to "noop", which is not ideal (and I'm using them during development). Is there a relatively simple way to either not map it, or custom map it to the debug commands?

rtsuk:
Yep, I still need to hook up the commands that are implemented already.

We could intercept F-keys in keydown before calling process.

rtsuk:
Closed to try for a cleaner PR.

PR #34, opened 2016-04-29
Set the dirty bit on scroll

potocpav:
The dirty bit should be updated in that case, right?

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -246,6 +246,7 @@ impl Editor {
         if let Some(array) = args.as_array() {
             if let (Some(first), Some(last)) = (array[0].as_i64(), array[1].as_i64()) {
                 self.view.set_scroll(max(first, 0) as usize, last as usize);
+                self.dirty = true;
             }
         }
     }


potocpav:
I signed it!

raphlinus:
No, the dirty bit explicitly represents changes in content, which doesn't happen on scroll. Think of it as invalidating a cache of what the front-end might have drawn. If you're getting incorrect rendering which is fixed by invalidating more I think it's a problem in the front-end.

potocpav:
After a scroll, I want to display a different part of a file. So the stuff that was previously drawn is invalidated and I need to get new data from the back-end to display. I can immediately request it, but this is the job for the dirty bit, no?

raphlinus:
No, the concept is that it's not invalidating the large scrollable surface. Conceptually, that could be rendered as one big texture and the scroll implementation just blits from that into your window. In practice, that should be a cache - if you have the region drawn, blit it without involving the back-end. If you don't, do a `render_lines` RPC to fetch the parts you want.

This is the way NSScrollView works, and I think it's very effective. They do a lot of clever things to ensure fast, smooth scrolling (such as prioritizing the visible region, then rendering to offscreen textures above and below the visible region after that's updated). Setting invalidate on scroll would degrade scrolling performance on Cocoa quite a bit. The direction it's going is even finer grained invalidation, so, eg, you can just redraw a single line if that's all that's changed.

potocpav:
Ok, thanks for the info.

raphlinus:
I recommend taking a look at 33f7ac5, which makes the nature of caching more explicit, and is probably more efficient in terms of minimizing RPC's. You probably want to implement something like that internally.

PR #36, opened 2016-04-29
Record opened files in the "Open Recent" menu

rtsuk:
Should make testing quicker.


raphlinus:
Seriously? I was figuring I'd have to maintain the list, persist it to disk, wire up the menu item for clearing it, all that.

PR #37, opened 2016-04-29

rtsuk:
AppKit has facilities for interpreting key events, integrated here. This would partially address #10.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -21,6 +21,10 @@ func eventToJson(event: NSEvent) -> AnyObject {
         "flags": flags]]
 }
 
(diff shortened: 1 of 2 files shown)


raphlinus:
This looks great. The only thing I see missing is `page_down_and_modify_selection`, which was working before.

Too bad implementing "insert" is not enough to get emoji working, oh well.

rtsuk:
Thanks for the catch. I'll fix that sometime today.

https://developer.apple.com/library/mac/documentation/TextFonts/Conceptual/CocoaTextArchitecture/TextEditing/TextEditing.html#//apple_ref/doc/uid/TP40009459-CH3-SW24 lays out all of what our editor view needs to do. I'm guessing implementing handleEvent would get us emoji. Accented characters will likely require the whole shooting match.

PR #39, opened 2016-04-30
Add depth to clone command in README

steelbrain:
Just like the title says

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ You need [Xcode 7](https://developer.apple.com/xcode/) (only on Mac) and [Rust](
 recommended and supported). You should have `cargo` in your path.
 
 ```
-> git clone https://github.com/google/xi-editor
+> git clone https://github.com/google/xi-editor --depth=1
 > cd xi-editor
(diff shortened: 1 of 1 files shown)


raphlinus:
Curious about the motivation. The repo is small enough I wasn't able to see any measurable difference in clone time (about 2s in both cases). Sometimes history is useful.

steelbrain:
Most of the times, history is not as useful and even if it's required, one can easily fetch the history on a shallow clone. This is just to make it copy/paste friendly for the majority.

shssoichiro:
I am personally against using `--depth=1`, because it actually creates more work for git if you later try to pull, if any of the newer commits being pulled in are merges onto an earlier commit than your HEAD. This project is so small at the moment that it probably doesn't matter, but it's caused significant issues for some larger repos (CocoaPods being the most notable).

leavengood:
I think this can be closed since generally it seems most of us don't believe adding the `--depth` param adds much value at this time.

raphlinus:
Agreed.

PR #41, opened 2016-04-30

rtsuk:
AppKit has facilities for interpreting key events, integrated here. This would partially address #10.

Sorry for yet another pull request but there's something screwed up with the history in my fork.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -21,6 +21,10 @@ func eventToJson(event: NSEvent) -> AnyObject {
         "flags": flags]]
 }
 
+func insertedStringToJson(stringToInsert: NSString) -> AnyObject {
+    return ["insert", ["chars": stringToInsert]];
+}
+
 // compute the width if monospaced, 0 otherwise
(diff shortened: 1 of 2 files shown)


raphlinus:
Looks good. Thanks for your work on this!

Sineaggi:
So now there are two entry points for those functions you modified?

I have code currently trapping directional arrows and sending over the OSX key equivalent. Should all input be handled through a different json command?

rtsuk:
It's not so much a function for every combination of modifiers as much as a function for distinct text editing actions, such as deleting a word backwards or extending the selection by moving to the right. It's worked well for AppKit, I think it will work well here too.

Sineaggi:
Sounds good. But I still wonder, should we have the two entry points defined? One in the json call, one in the do_key switch statement. I can handle one or the other, but having multiple options is bothering me. :)

raphlinus:
I would lean towards doing things the most semantic way, ie `move_up_and_modify_selection` rather than trying to pass in a shift.

rtsuk:
As soon as Raph no longer needs the do_key interface I'd be happy to delete it.

PR #43, opened 2016-04-30
Notes on plugin architecture, for discussion

raphlinus:
Ideas on building an architecture for plugins. Comments and suggestions welcome!

### DIFF
--- a/doc/plugin.md
+++ b/doc/plugin.md
@@ -1,0 +1,243 @@
+# Plugin architecture
+
+## Philosophy
+
+All serious programming editors have a mechanism for extensibility,
+often in the form of plugins, also often in terms of a scripting
+language with bindings to editor objects.
+
+Xi follows the philosophy that plugins should be asynchronous, and it
+should be possible to write them in any language. Thus, plugins are
+invoked through RPC, and no language bindings are provided inside the
+front-end or back-end process. A slow plugin should not interfere with
+typing or other editing operations, and a crashing plugin should not
+lead to data loss.
+
+The idea of asynchrony is appealing, but actually implementing it is
+challenging. The fundamental problem is that the plugin may provide
+edits (for example, inserting indentation in a language auto-indent
+mode) that happen at the same time as additional edits by the user.
+These edits must be reconciled somehow. In some cases, specificially
+where the user edits are to the text and not the rich-text annotations,
+and the plugin edits are to the annotations but not the text (as will
+often be the case for syntax highlighting), there is no fundamental
+conflict and the overall system will eventually converge. (In this
+case, it converges to the same state as running the plugin in batch
+mode over the text input, and this should be taken as a correctness
+criterion for plugins that do incremental comptuation). In xi, I plan
+to address the challenge of parallel edits head-on, using some form
+of operational transforms or differential synchronization. One
+potential advantage of this approach is that it may make collaborative
+editing practical.
+
+In xi, not everything is a plugin. Even though the front-end to
+back-end communication is similarly mediated by RPC, the protocol is
+quite different.
+
+It is expected that most plugins will be written on top of a
+convenience library. This library will provide caching for access to
+buffer contents, and in general abstract away low-level details of the
+RPC protocol. Even so, it should stay small and simple, so it is
+practical to provide for a number of languages. Initially, I will
+probably develop the plugin protocol and library using Python for
+rapid iteration, and then do Go and Rust, for higher performance.
+
+Xi will not provide a package manager, but rather will defer to
+existing mechanisms. Using, for example, apt-get, brew, or chocolatey
+to install plugins should work well. Alternatively, others may choose
+to package a _distribution_ containing the xi editor (front-end and
+back-end) along with a curated collection of plugins, as, for example,
+[Anaconda](https://www.continuum.io/downloads) does for IPython.
+
+## Basic architecture
+
+### Invocation; config files
+
+Deciding when to invoke a plugin, and how, is non-trivial. This starts
+with a configuration file, which represents a _trigger_ of when to
+invoke the plugin, as well as a path to the plugin and some options. A
+trigger can be a keyboard and/or menu command, a programming language
+(so basically a selector based on file extension), or hooks for other
+events (an example would be running gofmt before every save).
+
+There are three levels of invocation: one-shot, per-buffer, and
+editor-global. In a one-shot invocation, the editor starts the child
+process, performs the RPC, and shuts down the process when the RPC
+is complete. In per-buffer invocation, the process stays open for the
+life of the buffer. If more than one buffer requires the use of a
+plugin, xi will invoke multiple instances. In editor-global, a single
+process is expected to accept RPCs involving multiple buffers, and
+RPC requests are annotated with the buffer id.
+
+The config file also indicates the protocol version expected by the
+plugin, and xi will attempt to conform to a range of versions in
+actual use.
+
+Loading plugin info potentially has huge impact on startup time. Xi
+will load all config files at startup, but will attempt to defer
+executing the plugin binaries. Thus, the format of the config files
+needs to be quick to parse. I am leaning toward TOML as providing a
+good balance. (YAML is an alternative, I'm considering it because it's
+required to process new-format
+[Sublime Text syntax definitions](https://www.sublimetext.com/docs/3/syntax.html).)
+
+For developing plugins, the config file can direct the plugin to be
+compiled at invocation time ("go run" or "cargo run", for example).
+This mechanism should not be used for publishing plugins and
+distributing them to users.
+
+I'm thinking that config files will be able to to "include" another
+one. This would be the preferred way to handle optional plugins; they
+would be stored in a directory that would not by default get processed
+on editor startup, but a config file in a user-editable space can
+point to another one.
+
+### Read access to the buffer
+
+When attaching a buffer (ie, on startup of one-shot or per-buffer
+plugins), xi starts by sending a one-megabyte window of the buffer,
+centered around the cursor. The plugin may request additional
+substrings of the buffer through RPC. Note that such requests access
+a _snapshot_ of the buffer, even if the user is concurrently editing.
+As mentioned above, buffer access is one of the functions provided by
+a convenience library - the actual plugin logic should be able to
+request an arbitrary substring, or iterate through all lines, and have
+that served by cache and RPC on cache miss.
+
+When the RPC to the plugin completes, the snapshot is released. Any
+edits to the buffer are then sent as _deltas_ to all plugins
+subscribing to that buffer. Again, a major function of the convenience
+library is to apply these deltas. The deltas may also, of course,
+trigger computation, such as reapplying syntax coloring.
+
+### Write access to the buffer
+
+The plugin can also send deltas back to the core, either in the course
+of RPC processing or spontaneously. These deltas can be to the text
+buffer (for example, for indentation and electric brackets) and as
+rich text spans (for syntax highlighting).
+
+These deltas are suggestions; the core may need to reconcile them with
+other edits, and may possibly discard them. Xi will communicate back
+to the plugin to indicate whether the delta was accepted as-is or
+modified. A sophisticated plugin may attempt to retry, based on more
+up-to-date information about edits to the buffer. This seems like a
+reasonable approach to implementing differential synchronization.
+
+Other responses from the plugin are expected to include:
+
+* Populating a completion menu.
+
+* Displaying status messages.
+
+* Popping up modal dialogs?
+
+* What else?
+
+### Asynchrony modes
+
+Three asynchrony modes are anticipated. I might not implement all of
+them.
+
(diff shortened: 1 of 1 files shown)


raphlinus:
Thought this would be automatic, but I guess I have to add it by hand: [Rendered](https://github.com/google/xi-editor/blob/plugin-rfc/doc/plugin.md)

leavengood:
Nice document, I didn't know the syntax highlighting line state technique, but it makes a lot of sense.

My gut feeling is that syntax highlighting might be worth putting into the core, but a reasonable step to experiment with plugin-based syntax highlighting would be to develop a Rust crate with most of the functionality and then expose that in a simple plugin binary. If later it was deemed that syntax highlighting is indeed better in the core, the crate could be used directly there (maybe with some small changes as needed.) One advantage of the plugin approach is that different methods of syntax highlighting could be tried (such as LPeg vs Sublime's YAML files with regexp) without having to change the core or add dependencies to Lua or whatever. This could also allow for different approaches to syntax themes if that was deemed interesting, with all syntax highlighting plugins just sending spans with color information that the front-end faithfully renders. The other option is to leave all theming in the front-end with some sort of established naming convention for syntax scopes (identifier, string, etc) which are mapped to colors.

As for the configuration language I would lean towards TOML but as you note the fact that YAML is also needed to parse Sublime's new syntax definitions does make things more complicated. Though I doubt supporting both would add too much bloat to the core.

Beyond this I need to do more research on RPC approaches and differential synchronization before commenting.

Thanks for writing this document!

doc/plugin.md
```diff
@@ -0,0 +1,243 @@
...
+implemented simply as a choice of what the plugin chooses to do with
+delta notifications - if it batches them up until the RPC completes,
+then it is effectively normal mode. This also seems like a good
+function for the convenience library. Synchronous mode, however,
+requires explicit cooperation from the editor, to prevent concurrent
+edits while an RPC is in flight.
+
+## Security
+
+Plugins can potentially
```
davelab6: Fill this in before merge :)

disordinary:
@leavengood you'd want to decouple colour from the backend and as you said just send through scopes with the content, similar to CSS - the benefit obviously being theming which can take advantage of whatever niceties the platform has.

fjbfy51888-github:
Loading latest commit…

fjbfy51888-github:
#44 Breaking changes coming up

chriskrycho:
Major upside to TOML over YAML is its readability, of course. I'd also note that doing anything for the sake of compatibility with Sublime is… maybe not a great plan. ST is incredibly popular, of course, but I think you're probably better with doing whatever's actually best and providing a conversion mechanism. I like YAML in principle, but it is incredibly hard to write *correctly* in my experience.

raphlinus:
Since posting the above, I've done a lot more reading, and am now convinced that OT's are too complicated, differential synchronization relies on a poorly defined "patch" operation, and CRDT's are the One True Way. I'll update the concrete proposal after I've had a time to refine the details a bit, but in the meantime here are some readings I highly recommend for those who want to follow along:

I recommend starting with [Marc Shapiro's 2011 talk](http://research.microsoft.com/apps/video/default.aspx?id=153540&r=1) at Microsoft Research explaining the CRDT concept and the underlying mathematical concept of monotonic semi-lattice.

The concept of CRDT actually started with the collaborative editing use case, before they generalized it to more general cloud-flavored problems (it now powers eventual consistency in Riak, among other things). The [WOOT paper](https://hal.inria.fr/inria-00071240/document) (2006) is very good at comparing the new approach to operational transforms, but I'm definitely not going to adapt the details of that implementation.

Lastly, I recommend reading [A comprehensive study of Convergent and Commutative Replicated Data Types](http://hal.upmc.fr/inria-00555588/document) (2011), which surveys many of the use cases and goes into more detail about the duality of the state-based and operation-based approaches to implementing CRDTs. It also provides a very useful retrospective on the application to collaborative text editing (WOOT and TreeDoc), placing the somewhat complex implementation details into a clean mathematical framework.

Lastly, yes, I'll definitely want to flesh out the security section. A situation like [neovim](/neovim/neovim/issues/4443) is completely unacceptable imho.

disordinary:
As someone who has only had a limited need to develop for mobile platforms I'm wondering if the system of communicating between the view layer and the editor itself via stdin and stdout rather than ffi means that there might be issues with mobile platforms?

chriskrycho:
That's been a point of some curiosity on my part as well.. I don't believe you can do XPC this way on iOS, because of the sandboxing rules and restrictions on running things like Python.

leavengood:
Personally I think mobile platforms are the last target for something like a programmer's editor, which as far as I know is what Xi is meant to be.

Therefore I think for the moment mobile should not be a concern. But long term the core could probably be used as a library on mobile rather than through the stdin/stdout mechanism.

chriskrycho:
Define 'mobile platform', though: I'm typing this from an iPad Pro, which has astounded me with how capable it is *even for programming tasks*. I think the question of whether that's in line with @raphlinus' goals is a totally fair one, but distinct from whether the platform is capable of supporting it (ad that one has a clear answer: it is). here are of course plenty of reasons why mobile devices (whether Android, iOS, or other) are not a go-to yet, some of them technical and some of them cultural, but I wouldn't count on that remaining the case over even the medium term.

To be clear, I think opting not to support mobile devices is totally fair and reasonable! I nite it not because it's a worry for me with regard to this project—about which, honestly, I'm still just kind of giddy—but more just as a point of curiosity and consideration.

I'll leave my thoughts there, as I don't want to derail this particular issue, and like I said I'm mostly just curious.

disordinary:
@leavengood I don't know, like @chriskrycho I use a fully functioning bluetooth keyboard on an ipad pro and could easily see myself hacking on it while on the road. I've also used a keyboard with android in the past for writing.

The other thing is that while there is Sublime, Atom, Visual Studio Code, etc. on PC platforms there isn't a decent cross platform text editor that can work on mobile platforms which are increasingly becoming productivity tools. I think it's reasonable to think about it, and it could be that the core is able to communicate with a client both using stdin/out or embeddable as a library (although that would likely break all plugins and syntax highlighting as currently discussed).

So very likely the goal of this project is not to support mobile platforms at all, which as @chriskrycho mentioned is perfectly fine, it's just worth thinking about at this early stage.

raphlinus:
Yes, I am thinking about mobile. If anything, I think asynchrony is even more important to get right because the input method is in a separate process and generally does more complex edits than a typical hardware keyboard. Figuring out the interaction between that and, say, undo, is not simple.

Subprocesses work great on Android, but (as I understand it) are not allowed on iOS App Store apps. Running the front-end in core in separate threads should be fine. The plugin story will be more restrictive, but I think at least some subset should still be doable (say, ones written in Rust and bundled with the app, probably also JavaScript). I believe that JSON for the IPC is still viable on mobile, but I can imagine replacing or augmenting it with a binary transport (maybe even using shared memory buffers to cut down on copies).

I definitely want to make xi good on desktops _first_ before getting too distracted by mobile, but I totally agree it's worth thinking about and planning for.

To the RFC, I'm planning on updating it soon. I've explored CRDT-land and do believe I can come up with a design that's a lot simpler than the fully distributed peer-to-peer protocols, yet allows high degrees of asynchrony from the plugins and input method. More soon.

leavengood:
You make fair points guys. I'm still using my MacBook Pro as my development machine and my iPad Air as a consumption machine, but I can see how we are moving toward mobile devices being productivity machines too.

In the time I've thought about this project I think the safest and most flexible approach is to build all the various pieces as standalone parts that have clear interfaces and that can be used as either a library or through the stdin/stdout mechanism (which really would just be a sort of "frontend" on the library part.) If most of the main components are in Rust it would always be an option to compile them into the core. For example I'm starting work on a Rust syntax highlighting crate which hopefully can prove useful and it could serve as a testbed for trying out SH in a plugin. If that proves too slow or otherwise is not working well, the same code should be usable directly by the core.

And to bring this back to the topic of mobile, if a lot of the pieces of Xi are like that then they could be compiled into the core for mobile. But overall I think we just need to see what happens.

rkusa:
@raphlinus What use cases do you have in mind regarding the OT/CRDT stuff? Multiple frontends connected to one backend? Or also multiple backends connected together? Or is it mostly about asynchronous updates from plugins?

raphlinus:
@rkusa What I plan on implementing in short order is a simple engine that evaluates the CRDT serially, for asynchronous updates from plugins. However, I'm imagining that the engine could be replaced with one that actually does distributed CRDT without affecting the rest of the editor.

rkusa:
@raphlinus Thanks for the info. Since in the plugin case, the backend is the single point all plugins report to, OT would work perfectly fine. I read the papers you linked some time ago, but if I remember correctly using CRDTs for collaborative/concurrent text updates introduces some overhead (e.g. in data structure) compared to OT. Anyway, I think most CRDTs are easier to understand/implement and since CRDTs are better with regards to additional future features: +1 for CRDTs

trishume:
@raphlinus I just read through this and thought I might weigh in on syntax highlighting specifically:

There's a possible middle ground between including syntax highlighting in the core, and having it be just another plugin that uses a JSON RPC API to place edits on the rich text annotations.

As I calculated earlier, syntax highlighting on large files can be quick if the serialization overhead is low. But I worry about using the normal JSON API, especially if the API is adding edits to the rich text sounds like it would have high overhead. With parse state caching, streaming and incremental rendering you could make it work for common cases but there would still be laggy cases like changing the syntax highlighting mode at the bottom of a long file. Or scrolling quickly soon after opening a file.

I'm thinking that perhaps there could be a special API for plugins that want to annotate buffers with Sublime-style scopes that uses a custom binary protocol (perhaps with Cap'nProto/Thrift). This could have low enough overhead that RPC is not an issue.

Alternatively, just a special JSON protocol optimized for low overhead might work. Perhaps something like sending large batches of updates to text styles as JSON arrays to avoid re-sending key names a billion times.

Also I bet you'll have to have a special case for the CRDTs for syntax highlighting so as not to overload them with millions of edits per second. An easy way is not to store past highlighting and just re-highlight on undo. Also if you plan on heavily relying on caching to make undo work fast that you may also have to cache parse states in the past (or worse, rely on plugins to cache past parse states).

Another thing to think about (can be combined with any of the above) is representation of semantic annotations. I really like Sublime/Textmate scopes and that is what syntect outputs. The thing is without a good representation they're super costly because if you use strings that can be like a 50 character string every 5 characters of text. I use a string interning method that also allows fast prefix checks for super fast colouring, you might want to borrow that from syntect (or just use it as a library, it is public). When giving data to plugins you could can just send them the numbers, which are all you need to check equality and prefixes, and then give them an API to access the interning table so they can stringify them and intern their own scopes/selectors.

My only non-highlighting related comment is that 20 undos is way too few for me. I use undo all the time. I undo over 50 things more often than I open large files even, so I would take an editor that offered long undo history over large file support (although I rarely use both at the same time, so that might be an out).

raphlinus:
@trishume Sorry for taking so long to respond. Life has been busy, it's been hard to find time for xi. I'm basically going to start coding on a prototype to get myself unstuck. I think I have a pretty good design. It's more line oriented than what I was thinking before (

To address some of the things you raised: I'll definitely send deltas, sending the whole file goes against the spirit of the project. I'll avoid difficulties with undo and the like by having syntax highlighting a transient layer, basically downstream of all edits to the buffer. There's a small amount of logic needed to handle deltas (you move the spans along with the underlying text they annotate, then the plugin may or may not replace the spans on any given line later). I'll have a "get" RPC from the plugin to the core which is about sending chunks of up to a megabyte or so (but aligned on line boundaries, unlike my previous ideas). This lets the plugin keep as much or as little state as it likes; in the latter extreme, it can just scan through the whole file, fetching chunks and replacing the spans line-by-line.

The prototype is an experiment. If performance is good, I'll stick with it, and this will make things very good for people wanting to experiment with alternative techniques for syntax highlighting (PEG parsers, etc). If performance is not awesome, we very much _could_ move to a more optimized interprocess communication model, maybe with binary serialization, maybe with shared memory buffers. But at that point, I think it would be simpler (and more performant) to just bring it into the core and share ropes across thread boundaries. But we'll see.

I hear you regarding the undos. In fact, a large part of what's been distracting me has been researching Operational Transforms and CRDT's. The current design is very much n^2 in the number of undos, so increasing it too much is expensive. I have some ideas how to reduce that, but unfortunately it's pretty complicated, so I'd like to get back to implementing basic functionality (including highlighting!) before spending too much more time on it.

raphlinus:
A heads-up for people following this issue: I've made considerable progress on the plugin protocol recently, to the point where it wouldn't be crazy for someone to start prototyping a plugin for syntax highlighting. I've written a bit more at a [thread on the xi_editor subreddit](https://www.reddit.com/r/xi_editor/comments/4t4vz6/plugin_progress/).

raphlinus:
I'm going to merge this stuff even though it leaves the docs in a far-from-perfect state. Having the content in the repo and iterating from there seems better than holding PR's open for a long time without activity.

PR #46, opened 2016-05-01
Add a debug menu

rtsuk:
In preperation for removing F-key handling from core.


XiEditor/Base.lproj/MainMenu.xib
```diff
@@ -649,6 +649,25 @@
                         </items>
                     </menu>
                 </menuItem>
+                <menuItem title="Debug" id="r1r-EE-uzA" userLabel="Debug">
+                    <modifierMask key="keyEquivalentModifierMask"/>
+                    <menu key="submenu" title="Debug" id="V9K-nn-2xz">
+                        <items>
+                            <menuItem title="Rewrap" keyEquivalent="" id="eyx-NK-xxg">
```
raphlinus: Is it cool to put pseudo-unicode in here? I think I'd be slightly more comfortable with &#xF704;

rtsuk:
Whatever the Interface Builder portion of Xcode will accept is cool here. I'm not certain exactly what you're asking for, though.

raphlinus:
Updated my comment, might be clearer now. Also see [tabs branch](/google/xi-editor/commits/tabs) for insight into what I'm doing now. Some overlap, but merging should be pretty simple.

raphlinus:
Ok, I experimented with this, even after setting changing the XML to the hex entity, Xcode changes it back to (private use) Unicode. I don't think there's anything better to be done, so I'm merging. Thanks!

PR #49, opened 2016-05-01
Fix for #47

### LINKED ISSUE #47: Can't backspace over text
Occasionally, I end up in a state where pressing backspace doesn't seem to work.

I'm able to reproduce it pretty consistently by doing this:

1. Open a reasonably large file (> 30 K)
2. Scroll to the very bottom
3. Quickly scroll to the middle of the file
4. Position the cursor at the end of a line
5. Try to backspace twice

The first backspace works, but then it stops working. I'm still able to insert text, and I can even delete the text if I make a visual selection and *then* press backspace. But after deleting it, I still can't backspace over the point where the text was.

Thourum:
Appears to fix #47

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -125,7 +125,11 @@ impl Editor {
         } else {
             if let Some(bsp_pos) = self.text.prev_codepoint_offset(self.view.sel_end) {
             // TODO: implement complex emoji logic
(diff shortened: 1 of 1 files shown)


Thourum:
I signed it!

Thourum:
I realized it fixed just the delete

PR #50, opened 2016-05-01
Refactor front-end/back-end protocol to allow tabs

### LINKED ISSUE #29: 8 bytes
Why do we send the 8 bytes back and forth when dealing with stdio? We certainly don't do that when dealing with with stderr.

I understand the 8 bytes are to tell the client/server how much data we're sending over, but if we're using json and prepending a custom binary sequence to the beginning of the transmission, why not go full binary for the interface? Or the opposite question, why not go full json? We already push an extra '\n' at the end.

### LINKED ISSUE #44: Breaking changes coming up
I want to make some big changes, especially in the FE/BE protocol, after which hopefully it will stabilize for a while. A main motivation is to support multi-window / multi-tab. While I'm making those changes, I'd like to roll in some other stuff to make it cleaner. So:

* Get rid of the 8 byte framing (issue #29) and move to something closer to JSON-RPC (see issue #2). When I look closely at JSON-RPC, I see some fragmentation, and there's nothing that's exactly what I need, so I'm inclined to mix and match. I do want to unify the syntax of sync and async calls (and a bunch of calls that are now async should become sync so they can return errors, like open/save). The details don't matter so much for FE/BE, but they do matter a lot for plugins (see #43), and it would be ugly for the two to use different low-level mechanism. So what I'm looking at is basically JSON-RPC 2 syntax, minus the `"jsonrpc": "2.0"` (there's no need to version-negotiate, and signalling strict compatibility with the spec is misleading), with requests allowed in both directions. I don't want to rule out binary stuff in the future, but it's better to start with something clean.

* All requests in both directions grow a "tab" field, which identifies a tab with a unique identifier. I'm thinking this should be a string derived from the file name (but guaranteed unique) to make debugging more friendly. I've also considered having a two-level structure, with the outer layer identifying the tab and having only a small number of commands, and dispatching the inner message, but I don't think that's better. Obviously `main.rs` will no longer plumb directly to the editor, but will need to go through a lookup based on tab before dispatching to the editor.

* `do_key` should probably go away, and I should just add menu items (maybe in a Developer tab which would be optional for end users).

I know people are working on stuff this might break, which is one reason I'm discussing it before diving in.

raphlinus:
A significant restructuring, motivated to allow editing of more than
one tab at a time. With this commit, the front-end/back-end protocol
supports multiple tabs, but that functionality is not implemented in
the front-end yet.

The RPC mechanism has been reworked to be much more like JSON-RPC 2.
Thus, this patch closes #29.

This also closes #44, which describes the motivation for these
changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ recommended and supported). You should have `cargo` in your path.
 Or `open XiEditor.xcodeproj` and hit the Run button.
 
 It will look better if you have [InconsolataGo](http://levien.com/type/myfonts/inconsolata.html)
-installed, a customized version Inconsolata tuned for code editing. To choose other
+installed, a customized version of Inconsolata tuned for code editing. To choose other
 fonts, edit the `CTFontCreateWithName()` call in EditView.swift.
 
 ### Building the core
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -26,19 +26,28 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
         let corePath = NSBundle.mainBundle().pathForResource("xicore", ofType: "")
         if let corePath = corePath {
-            coreConnection = CoreConnection(path: corePath) { [weak self] data -> () in
-                self?.handleCoreCmd(data)
+            coreConnection = CoreConnection(path: corePath) { [weak self] json -> () in
+                self?.handleCoreCmd(json)
             }
         }
         appWindowController?.coreConnection = coreConnection
 
         appWindowController?.showWindow(self)
     }
-    
+
     func handleCoreCmd(json: AnyObject) {
-        if let response = json as? [AnyObject] where response.count == 2, let cmd = response[0] as? NSString {
-            if cmd == "settext" {
-                self.appWindowController?.editView.updateSafe(response[1] as! [String: AnyObject])
+        if let obj = json as? [String : AnyObject], let method = obj["method"] as? String, let params = obj["params"] {
+            handleRpc(method, params: params)
+        } else {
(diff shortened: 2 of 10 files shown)


raphlinus:
Sending as a PR both for comment and because this has a lot of potential to break what people are working on. Even so, I like the direction. The RPC stuff feels less messy, and I'm comfortable that the communication with plugins can be harmonious with this.

I believe this is a good step in the direction of a stable FE/BE protocol, but there are steps remaining. For example, getting minimal invalidation _right_ is going to require some tricky async logic, not unlike the plugin discussion (in particular, the FE should be able to request rendered lines from a particular snapshot, to avoid races).

Sineaggi:
After a first glance, the biggest thing that's tripped me up so far is having two `sendRpcAsync` methods, one that calls the other. No wonder the core was telling me unknown method `"open"`, I have to wrap the `"open"` command in an `"edit"` method.
Along that same thought, there are also two `unknown method` responses. Which also tripped me up :)

raphlinus:
@Sineaggi If you have suggestions of how I could make that clearer in the doc, they'd be appreciated.

leavengood:
@Sineaggi I don't know, I think the two `sendRpcAsync` methods is reasonable approach where one does most of the work and the other makes a slight change (the editor wrapping the command in the edit method.)

@raphlinus FWIW I've been reading commits on the project since last week and overall this approach looks good. So 👍 to merging.

raphlinus:
Thanks for the feedback. Merging this. I have some ideas for fine-tuning a bit, but they won't cause conflicts, so going forward.

Sineaggi:
Looks great! I've updated my frontend to support the new protocol.

PR #54, opened 2016-05-03
Fix page_up and page_down

potocpav:
`self.view.vertical_motion()` should not probably be called twice. This causes jumps of the view that are too big.

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -133,7 +133,7 @@ impl Editor {
     fn insert_newline(&mut self) {
         self.insert("\n");
     }
-    
+
     fn move_up(&mut self, flags: u64) {
         let old_offset = self.view.sel_end;
         let offset = self.view.vertical_motion(&self.text, -1, self.col);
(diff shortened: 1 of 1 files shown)


raphlinus:
I'm basically trying to mimic what Sublime Text does, but I might have gotten details wrong. Say your screen is 20 lines high, and the cursor is at the top, so it's displaying lines 0-19. After pressing "page down", I want to see lines 19-38 on the screen. Sublime Text also places the cursor at line 19. Note, TextEdit on Mac scrolls by one screenful but doesn't move the cursor.

So I don't think this patch does what I want.

potocpav:
OK, so if the cursor is on the screen, the screen should move down by one height and the cursor position on-screen should be preserved. What should be the behavior if the cursor is off-screen?

This patch is mimicking what Atom does, that is move the cursor by one height, and move the view minimally to display the cursor.

raphlinus:
Ok, I'll take a look at Atom and see which behavior I like better. Thanks for the explanation!

raphlinus:
I finally got around to reviewing this, and you're right, my old logic got some important cases wrong. I'll go with the much simpler approach you suggest.

I'd merge this but there are conflicts, easiest just to apply it locally. Thanks!

PR #55, opened 2016-05-03
Refactor some bits of xi-rope

killercup:
I was interested in how you implemented the rope data structure, so I read through the code and noticed some things.

This is just me fiddling with the code a bit. If you don't like the changes, feel free to close this PR 😄

### DIFF
--- a/rust/rope/src/lib.rs
+++ b/rust/rope/src/lib.rs
@@ -64,6 +64,38 @@ fn is_char_boundary(s: &str, index: usize) -> bool {
 /// Also note: in addition to the `From` traits described below, this module
 /// implements `From<Rope> for String` and `From<&Rope> for String`, for easy
 /// conversions in both directions.
+///
+/// # Examples
+///
+/// Create a `Rope` from a `String`:
+///
+/// ```rust
+/// # use xi_rope::Rope;
+/// let a = Rope::from("hello ");
+/// let b = Rope::from("world");
+/// assert_eq!("hello world", String::from(a.clone() + b.clone()));
+/// assert!("hello world" == a + b);
+/// ```
+///
+/// Get a slice of a `Rope`:
+///
+/// ```rust
+/// # use xi_rope::Rope;
+/// let a = Rope::from("hello world");
+/// let b = a.slice(1, 9);
+/// assert_eq!("ello wor", String::from(&b));
+/// let c = b.slice(1, 7);
+/// assert_eq!("llo wo", String::from(c));
+/// ```
+///
+/// Replace part of a `Rope`:
+///
+/// ```rust
+/// # use xi_rope::Rope;
+/// let mut a = Rope::from("hello world");
+/// a.edit_str(1, 9, "era");
+/// assert_eq!("herald", String::from(a));
+/// ```
 #[derive(Clone)]
 pub struct Rope {
     root: Node,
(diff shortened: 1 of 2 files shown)


rust/rope/src/lib.rs
```diff
@@ -730,8 +779,9 @@ impl Node {
         (node.get_leaf(), offset)
     }
 
-    // precondition: offset > 0 and in range
     fn prev_codepoint_offset(&self, offset: usize) -> usize {
+        debug_assert!(offset > 0 && offset <= self.len());
```
shssoichiro: Should this be `offset < self.len()`? (It's entirely possible that it's correct as is)

rust/rope/src/lib.rs
```diff
@@ -1101,30 +1153,6 @@ impl<'a> PartialEq<Rope> for Cow<'a, str> {
...
-    assert_eq!("llo wo", String::from(c));
-}
-
-#[test]
-fn replace_small() {
-    let mut a = Rope::from("hello world");
-    a.edit_str(1, 9, "era");
-    assert_eq!("herald", String::from(a));
-}
-
```
shssoichiro: Is there a reason for removing these tests?

killercup: I moved them to be doc tests (see above)

Josh Holmer <notifications@github.com> schrieb am Di., 3. Mai 2016 um 19:23:

shssoichiro: I see. TIL that examples in rustdoc get run as tests.

rust/rope/src/lib.rs
raphlinus: No, `<=` is valid, the return value should then be the index of the last codepoint in the buffer.

rust/rope/src/tree.rs
```diff
@@ -179,6 +179,14 @@ impl<N: NodeInfo> Node<N> {
     fn height(&self) -> usize {
         self.0.height
     }
+    
+    fn is_leaf(&self) -> bool {
+        if let NodeVal::Leaf(_) = self.0.val {
```
raphlinus: Might be easier / cleaner to say height == 0

raphlinus:
Generally looks good. Some notes, the version in lib.rs is basically going away, I'm trying to move toward the tree.rs / rope.rs one. I'd definitely appreciate even more tests.

rust/rope/src/tree.rs
killercup: Oh, I was about to do that, and even wanted to debug_assert that the height is 0 when the val is a Leaf. I'll just change it to height == 0, because it's so short :)

killercup:
Rebased. Don't know when I have the time to do more here. The version in lib.rs contains some changes the version in tree.rs does not have.

rust/rope/src/lib.rs
```diff
@@ -474,38 +512,45 @@ impl Node {
...
-                return Node::merge_nodes(children1, &[rope2]);
-            }
-            let lastix = children1.len() - 1;
-            let newrope = Node::concat(children1[lastix].clone(), rope2);
-            if newrope.height() == h1 - 1 {
-                Node::merge_nodes(&children1[..lastix], &[newrope])
-            } else {
-                Node::merge_nodes(&children1[..lastix], newrope.get_children())
+
+        match h1.cmp(&h2) {
```
faizaanceg: This part is so good. :D Definitely learned a new way to handle comparisons,

raphlinus:
Thanks for the contribution!

PR #56, opened 2016-05-03
Fix leaf optimization path in Node replace

lukehoban:


raphlinus:
Nice catch! I should point out that the rope impl in lib.rs is basically going away, replaced by the more generic one in tree.rs and rope.rs. The main reason I haven't deleted it already is that it has some tests and some more aggressive optimizations which haven't yet been duplicated in the more generic version.

Other than that, tests and improvements very welcome!

lukehoban:
Thanks @raphlinus.  I did notice there were two copies of a lot of the Rope functionality.  The `lib.rs` Rope implementation is working for my needs for a side project now, but I'll take a look at moving to the `rope.rs`/`tree.rs` impl instead.

It looks like the optimization that this PR fixes is one of the ones not yet ported to the new impl.

raphlinus:
Makes sense, thanks!

Issue #52, opened 2016-05-02
Request: Core Text-friendly keyboard shortcuts


chriskrycho:
Core Text has built-in support for the standard Emacs navigation actions and the simplest of its text manipulation actions:
- `^f`: cursor forward
- `^b`: cursor backward
- `^p`: cursor up one line
- `^n`: cursor down one line
- `^a`: cursor or to start of line
- `^e`: cursor to end of line

Text manipulation: 
- `^k`: "kill" – cut to end of line and add to OS X "kill ring" (a secondary clipboard, essentially)
- `^y`: "yank" – paste the last item in the kill ring
- `^t`: "transpose" – swap characters on either side of cursor and advance cursor by one

Support for these would immediately make the editor feel much more native on OS X to me (and presumably a lot of others).

(If I get some time I'll see what would be needed to implement, but I'm in a very busy season and we have some family health t now, so I wouldn't hold your breath for a PR for me.)

raphlinus:
Note that `^n` and `^p` already work. The others are interpreted into actions, it's just that the actions aren't implemented.

Also note that the difference between, say, `^f` (`move_forward`) and right arrow (`move_right`) is only relevant for BiDi. The former always advances one grapheme cluster forward, so is simpler. The latter moves forward in LTR and backward in RTL. Since we don't support BiDi yet (though I'd like to at some point), it's fine to just implement both as move_forward.

Tagging this as easy in case someone would like to contribute.

chriskrycho:
Awesome. Can you point me to the relevant implementation area? I'll at least take a look and see if it's something I *can* do.

(PS: much :heart: for the project.)

raphlinus:
@chriskrycho Thanks! You basically want to add more cases to [do_rpc()](https://github.com/google/xi-editor/blob/master/rust/src/editor.rs#L326) in editor.rs. Don't stress if you don't get around to it, I have confidence someone will soon :)

rtsuk:
The Cocoa bits are all done. 

When your run Xi in Xcode and hit ^f, you see in the console:

```
to core: ["move_forward",[]]
unknown cmd move_forward
```

in rust/editor.rs there is a function called do_cmd that matches agains the command string and calls an implementation function. The task for these control keys is to write the implementation function for those commands not currently implemented.

eestrella198:
Thank you

gsquire:
I would like to try and add the rest of the cursor functionality if nobody else has started working on it.

[...]

PR #59, opened 2016-05-05
Cursor control

gsquire:
@raphlinus I added functionality for Control-{b, f, a, e}. However my end-of-paragraph movement (e) isn't working as I had hoped. I thought the `is_boundary` function for `Cursor` would advance until a newline. Can you give some advice on that or tell me it's the wrong approach? :)

For #52

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -17,9 +17,10 @@ use std::fs::File;
 use std::io::{Read,Write};
 use serde_json::Value;
 
-use xi_rope::rope::{Rope,RopeInfo};
+use xi_rope::rope::{LinesMetric,Rope,RopeInfo};
 use xi_rope::interval::Interval;
 use xi_rope::delta::Delta;
+use xi_rope::tree::Cursor;
 use view::View;
 
 use tabs::update_tab;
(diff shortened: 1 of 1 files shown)


rust/src/editor.rs
```diff
@@ -177,6 +178,28 @@ impl Editor {
...
+        self.set_cursor(start, true);
+    }
+
+    fn cursor_end(&mut self) {
+        let current = self.view.sel_max();
+        let rope = self.text.clone();
+        let mut cursor = Cursor::new(&rope, current);
+        let mut new = current;
+        // Advance this cursor as this is considered a boundary.
+        if new == 0 {
```
raphlinus: The return value of `cursor.next()` is the offset in the buffer of the next line (paragraph) boundary. Note, that'll be after the `'\n'`.

I think the logic you want is:

* Do `cursor.next()` from current.
* If you're not on a boundary (`is_boundary`), you're at the end of the file and there's no newline at the end. Use that.
* If you are on a boundary, go back one grapheme (see end of `line_col_to_offset` in view.rs)

Does that help?

gsquire: Thanks so much for the pointer. I misunderstood the `is_boundary` function. I believe it should be good to go now.

gsquire:
If it looks good, I would be happy to squash my commits before you merge too.

rust/src/editor.rs
```diff
@@ -177,6 +178,29 @@ impl Editor {
...
+    fn cursor_end(&mut self) {
+        let current = self.view.sel_max();
+        let rope = self.text.clone();
+        let mut cursor = Cursor::new(&rope, current);
+        match cursor.next::<LinesMetric>() {
+            None => { self.set_cursor(current, true); },
+            Some(offset) => {
+                if cursor.is_boundary::<LinesMetric>() {
+                    if let Some(new) = rope.prev_grapheme_offset(offset) {
+                        self.set_cursor(new + 1, true);
```
raphlinus: Pretty sure `new + 1` should just be `new`. Other than that it looks good. I don't have any strong feeling on squashing, I know Rust itself tends not to.

gsquire: I thought so too, but `new` would advance to right before the last character. For example if I had the cursor before the `e` in `test` it would advance the cursor right before the last `t`. Is that desired behavior? I thought it would go to after the last `t`?

raphlinus: This is strictly for going back before the newline at the end of the line, as it is ctrl-e will advance to the beginning of the next line. I'll push the change to `is_boundary` now.

gsquire:
Ok @raphlinus I fixed the offset to just be `new`. Thanks for your help, I enjoyed the challenge of understanding some parts of the backend code!

raphlinus:
Thanks for the contribution, glad you enjoyed it!

PR #61, opened 2016-05-07
Use consistent naming scheme xi-* for xi-core

JesseFarebro:
`xi-rope`'s `Cargo.toml` has the naming scheme xi-* and the core's `Cargo.toml` uses xi*. This PR is just a simple naming scheme change to make the core module follow the ropes naming convention.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "xicore"
+name = "xi-core"
 version = "0.0.0"
 license = "Apache-2.0"
 authors = ["Raph Levien <raph@google.com>"]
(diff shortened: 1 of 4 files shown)


JesseFarebro:
I signed it!

rust/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "xicore"
+name = "xi-core"
```
raphlinus: I'm fine with this, but AppDelegate.swift will need to have its reference to this binary changed as well, and, I'm guessing, the build script that copies the binary into the app.

JesseFarebro:
I amended the remaining changes onto my first commit. Renaming `xicore.sh` -> `xi-core.sh` as well. Everything builds fine and the new xcode build files are generated!

Let me know if you need anything else.

raphlinus:
Thanks!

PR #67, opened 2016-05-10
Implement kill ring

canaltinova:
Implement kill ring mentioned in #52. I had to change `Default::default()` back in Tabs. Default is not implemented for Mutex in current stable release of rust yet (1.8.0). I get a bound error in 1.9 but it seems default working fine with nightly(1.10). What do you think @raphlinus?

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -15,6 +15,7 @@
 use std::cmp::max;
 use std::fs::File;
 use std::io::{Read,Write};
+use std::sync::Mutex;
 use serde_json::Value;
 
 use xi_rope::rope::{LinesMetric,Rope,RopeInfo};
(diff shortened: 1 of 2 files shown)


canaltinova:
I signed it!

rust/src/editor.rs
```diff
@@ -368,12 +376,39 @@ impl Editor {
...
     }
 
-    pub fn do_rpc(&mut self, method: &str, params: &Value) -> Option<Value> {
+    fn delete_to_end_of_paragraph(&mut self, kill_ring: &Mutex<Rope>) {
+        let current = self.view.sel_max();
+        let offset = self.cursor_end_offset();
+        let mut val = String::from("");
+
+        if current != offset {
+            val = self.text.slice_to_string(current, offset);
```
raphlinus: This feels like unneeded duplicate code. But I'll merge this because overall it looks quite good, and this can be a cleanup for later. Thanks!

raphlinus:
I'm not seeing any errors on freshly-updated 1.9 beta. If they pop up, probably worth investigating whether it's an issue with the code or with rustc.

canaltinova:
Thanks for your help! :)

PR #70, opened 2016-05-15
IME Support

pocket7878:
Trying to support MacIME.

Currently I handled IME translating range on frontend, 
But I'm worried whether it should be managed by the backend..
And replacing IME markedRange is very rush job (replacing by delete and inserting..)

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -62,7 +62,7 @@ func camelCaseToUnderscored(name: NSString) -> NSString {
     return underscored;
 }
 
-class EditView: NSView {
+class EditView: NSView, NSTextInputClient {
     var tabName: String?
     var coreConnection: CoreConnection?
 
@@ -96,6 +96,9 @@ class EditView: NSView {
     var pendingUpdate: [String: AnyObject]? = nil
 
     var currentEvent: NSEvent?
+    
+    var _selectedRange: NSRange
+    var _markedRange: NSRange
 
     override init(frame frameRect: NSRect) {
         let font = CTFontCreateWithName("InconsolataGo", 14, nil)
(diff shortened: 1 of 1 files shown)


raphlinus:
I'll take a closer look later, but I tried it and it seems to work. I also saw these 5 warnings, and I'm trying to keep compilation warning-free.

```
/Users/raph/Documents/xcode/xi-editor/XiEditor/EditView.swift:201:25: warning: variable 'markRangeStart' was never mutated; consider changing to 'let' constant
                    var markRangeStart = cix - markedRange().length
                    ~~~ ^
                    let
/Users/raph/Documents/xcode/xi-editor/XiEditor/EditView.swift:209:25: warning: variable 'selectedRangeStart' was never mutated; consider changing to 'let' constant
                    var selectedRangeStart = cix - markedRange().length + selectedRange().location
                    ~~~ ^
                    let
/Users/raph/Documents/xcode/xi-editor/XiEditor/EditView.swift:305:17: warning: variable 'i' was never used; consider replacing with '_' or removing it
        for var i in 0..<aRange.length {
                ^
                _
/Users/raph/Documents/xcode/xi-editor/XiEditor/EditView.swift:318:13: warning: variable 'effectiveRange' was never mutated; consider changing to 'let' constant
        var effectiveRange = self.replaceCharactersInRange(self.replacementMarkedRange(replacementRange), withText: aString)
        ~~~ ^
        let
/Users/raph/Documents/xcode/xi-editor/XiEditor/EditView.swift:331:21: warning: variable 'i' was never used; consider replacing with '_' or removing it
            for var i in 0..<_markedRange.length {
                    ^
                    _
```

Thanks very much for this contribution! Input method is high on my list of things I wanted to support, but I didn't get around to it yet because it's not simple.

XiEditor/EditView.swift
```diff
@@ -233,14 +260,126 @@ class EditView: NSView {
...
+            len = attrStr.string.characters.count
+        } else if let str = aString as? NSString {
+            len = str.length
+        }
+        if (replacementRange.location == NSNotFound) {
+            replacementRange.location = 0
+            replacementRange.length = 0
+        }
+        for var i in 0..<aRange.length {
+            sendRpcAsync("delete_backward", params: [])
```
raphlinus: This is probably the part of the PR I'm having the most trouble with. Right now it should work most of the time, but that will change. The issue is that NSRange counts UTF-16 code units, where "delete_backward" currently does one codepoint, but that's a placeholder for more sophisticated logic. Even now it wouldn't be right for non-BMP characters.

I think the best way to fix this is to have an RPC for exactly "replaceCharactersInRange", and pass in a precise count. Currently counts passed over the wire are all utf-8, so the front-end has to do the conversion, but I'm open to changing that.

There are other more subtle issues with perhaps the cursor position not aligning because of concurrent edits (say, by a plugin). That I'm prepared to deal with but will likely require some rocket science (like passing revision id's across the wire and doing transforms on the edit operations to effectively "rebase" them onto a potentially newer revision).

More generally, I'm willing to take your patch now, and refine it later, either myself or as an open issue, because I think having something working is more important than having it perfect.

pocket7878:
@raphlinus Thank you for your review! I just make my code warning-free!

raphlinus:
Thanks!

PR #74, opened 2016-05-15
Add support for typing umaults and such

modelorganism:
Currently in is not possible to type umaults in Xi.  U-umault (ü) is normaly typed on MacOSX with a 2 keys <option-u-> <u>. This adds support to Xi for listening for dead keys allowing umaults and other things on vowels. 
The code to track dead keys may seem to duplicate some of the work that Cocoa is doing for us, but there does not appear to be a simpler way to handle this standard behavior.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 import Cocoa
+import Carbon
 
 func eventToJson(event: NSEvent) -> AnyObject {
     let flags = event.modifierFlags.rawValue >> 16;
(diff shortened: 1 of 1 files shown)


raphlinus:
I'll take a more careful look at these soon, but in the meantime this seems quite similar in scope to pull request #70, which uses NSTextInputClient. Do you think this approach is worth pursuing if there's input method support?

raphlinus:
This should be fixed by #70.

modelorganism:
Looks like #70, IME Support, dose everythiong that this would do, and also gives visual feedback about the pending dead key.

PR #76, opened 2016-05-17
Implement Transpose Shortcut

hjylewis:
As specified in #52:
> `^t:`  "transpose" – swap characters on either side of cursor and advance cursor by one

Relatively new to Rust so any tips/advice are appreciated :)

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -484,9 +484,9 @@ impl Editor {
 
         let end = end_opt.unwrap_or(self.view.sel_end);
         let (start, middle) = if end_opt.is_none() && start_opt.is_some() { // if at the very end, swap previous TWO characters (instead of ONE)
-            let middle_temp = start_opt.unwrap();
-            let start_temp = self.text.prev_grapheme_offset(middle_temp).unwrap_or(middle_temp);
-            (start_temp, middle_temp)
+            let middle = start_opt.unwrap();
+            let start = self.text.prev_grapheme_offset(middle).unwrap_or(middle);
+            (start, middle)
         } else {
             (start_opt.unwrap_or(self.view.sel_end), self.view.sel_end)
         };
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -218,7 +218,7 @@ impl Editor {
         self.this_edit_type = EditType::InsertChars;
         self.insert("\n");
     }
-    
+
     fn move_up(&mut self, flags: u64) {
         let old_offset = self.view.sel_end;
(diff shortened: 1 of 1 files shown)


hjylewis:
I signed it!

hjylewis:
Also, I has curious about the the open and closed endpoints of Intervals.  What are they needed for?  It seems as though that information is lost when `start_end()` is called.

raphlinus:
Thanks for the contribution! Generally looks pretty good, but a couple of points.

1. While Sublime Text swaps codepoints (as does your patch), Mac TextEdit swaps grapheme clusters. I think the latter is more correct. This wouldn't really affect the current code as I haven't implemented grapheme boundaries yet. We could consider this a cleanup for when grapheme boundaries are implemented.

2. It seems like Sublime does nothing when the cursor is at the end of the buffer. TextEdit behaves as yours does, but also has special handling of the end of the line - at EOL, it basically swaps the two characters before EOL. Not sure what the "right" answer is, just noting what I observe in other editors.

I'm willing to take your patch as-is, but figured I'd give you a chance to address any of the above if you wanted.

The open/closed interval stuff basically doesn't make a difference for text, but has some effect on line breaks (you need an interval to be closed at the end to pick up a line break at the end of the file), and will have a significant effect on rich-text spans - if an interval is closed at one end and text is inserted at that location, it will be included in the span. If closed, then not. Sorry if it's a bit confusing, I should probably document it more carefully as part of the rope crate.

hjylewis:
1. It seems like this would look like swapping `next_codepoint_offset()` for `next_grapheme_offset()` (as well as the prev offset functions).  Also, using a iterator for grapheme clusers (Rust recommends this [library](https://crates.io/crates/unicode-segmentation)) instead the one for characters.  Does that sound right?

2. I was using Emacs as a reference. Changing this wouldn't be too hard if we wanted to.

Thanks for the interval info!  I think it makes more sense but I'm sure the the rope crate documentation will help too.

I can make the 1. changes if they look okay to you.

raphlinus:
@hjylewis I think a simpler and more reliable approach than using a grapheme cluster iterator would be to just take the substrings and concatenate them, basically `s[curs:next] + s[prev:curs]`. The problem is that the concept of grapheme cluster boundaries is a bit fuzzy and two implementations may disagree.

Emacs is a fine reference for the behavior of `^t`, you can make an argument the other two are trying to copy it and not getting the edge cases perfect. It never ceases to amaze me how much implementations can differ once you get past the basics.

hjylewis:
Good idea.  Implemented it in the new commit. Thanks.

raphlinus:
Looks great, thanks for the revisions!

PR #77, opened 2016-05-23
some hotkey develop

subumm1:
I develop below functions.

[move to left end of line]
[move to right end of line]
[move to left end of line and modify selection]
[move to right end of line and modify selection]
[delete forward]
[delete beginning of line]

------------ opinion ----------
I think add function of getting end of line offset to  "view.rs" is better.

Because in move_left() / move_right(),
[calculate end of line] part's source is looks dirty.

I think

    let offset = self.view.offset_to_offset_of_end_of_line(&self.text, self.view.sel_end);

is more beautiful than

        let line_col = self.view.offset_to_line_col(&self.text, self.view.sel_end);
        let mut offset = self.text.len();

        // calculate end of line
        let next_line_offset = self.view.line_col_to_offset(&self.text, line_col.0 + 1, 0);
        if offset > next_line_offset {
            if let Some(prev) = self.text.prev_grapheme_offset(next_line_offset) {
                offset = prev;
            }
        }

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -345,6 +345,18 @@ impl Editor {
         }
     }
 
+    fn move_to_beginning_of_document(&mut self) {
+        let offset = 0;
+
+        self.set_cursor(offset, true);
+    }
+
+    fn move_to_end_of_document(&mut self) {
+        let offset = self.text.len();
+
+        self.set_cursor(offset, true);
+    }
+
     fn scroll_page_up(&mut self) {
         let scroll = -max(self.view.scroll_height() as isize - 2, 1);
         let old_offset = self.view.sel_end;
@@ -376,7 +388,7 @@ impl Editor {
                     self.move_up();
                 }
                 "\u{F701}" => {  // down arrow
-                    self.   move_down();
+                    self.move_down();
                 }
                 "\u{F702}" => {  // left arrow
                     self.move_left();
@@ -606,6 +618,10 @@ impl Editor {
             "move_to_right_end_of_line_and_modify_selection" => async({ self.modify_selection(); self.move_to_right_end_of_line() }),
             "move_to_beginning_of_paragraph" => async(self.cursor_start()),
             "move_to_end_of_paragraph" => async(self.cursor_end()),
+            "move_to_beginning_of_document" => async(self.move_to_beginning_of_document()),
+            "move_to_beginning_of_document_and_modify_selection" => async({ self.modify_selection(); self.move_to_beginning_of_document() }),
+            "move_to_end_of_document" => async(self.move_to_end_of_document()),
+            "move_to_end_of_document_and_modify_selection" => async({ self.modify_selection(); self.move_to_end_of_document() }),
             "scroll_page_up" |
             "page_up" => async(self.scroll_page_up()),
             "page_up_and_modify_selection" => async({ self.modify_selection(); self.scroll_page_up() }),
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -199,7 +199,25 @@ impl Editor {
         }
     }
 
+    fn delete_forward(&mut self) {
+        if self.view.sel_start == self.view.sel_end {
+            self.move_right(0);
+        }
+
+        self.delete();
+    }
+
     fn delete_backward(&mut self) {
+        self.delete();
+    }
+
+    fn delete_to_beginning_of_line(&mut self) {
+        self.move_left(MODIFIER_SHIFT | MODIFIER_COMMAND);
+
+        self.delete();
+    }
+
+    fn delete(&mut self) {
         let start = if self.view.sel_start != self.view.sel_end {
             self.view.sel_min()
         } else {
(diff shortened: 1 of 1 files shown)


subumm1:
I signed it!

rust/src/editor.rs
```diff
@@ -392,7 +448,7 @@ impl Editor {
         }
     }
 
-    fn do_scroll(&mut self, args: &Value) {
+    fn do_scroll(&mut self, args: &Value) {
```
raphlinus: extra space

rust/src/editor.rs
```diff
@@ -29,6 +29,7 @@ use view::View;
 use tabs::update_tab;
 
 const MODIFIER_SHIFT: u64 = 2;
+const MODIFIER_COMMAND: u64 = 4;
```
raphlinus: I want to move away from interpreting modifiers in the back-end (do_key is going away), so have been thinking of changing from testing MODIFIER_SHIFT to a bool, say `extend_selection`. I think it makes sense to have a separate `move_beginning_of_line`, as opposed to yet another option for `move_left`.

raphlinus:
I'm on vacation and away from my Mac, but I'll try to get access to one and then will be able to merge. Overall looks pretty good. I'm very happy to have these commands added, and thanks for your contribution!

subumm1:
Thank you for code review!

I understand back-end design what you want.

I'll convert my code to separate functions.

And sorry for bother your vacation.

Enjoy your vacation! :)

rust/src/editor.rs
```diff
@@ -322,22 +385,22 @@ impl Editor {
                     self.delete_backward();
                 }
                 "\u{F700}" => {  // up arrow
-                    self.move_up(flags);
```
raphlinus: With these changes, `flags` becomes a dead variable. But for clients that still send raw keys, the modify-selection behavior is broken. Perhaps it should say, `if flags & MODIFIER_SHIFT { self.modify_selection(); }`

rust/src/editor.rs
```diff
@@ -94,8 +95,8 @@ impl Editor {
         self.add_delta(sel_interval, Rope::from(s), new_cursor);
     }
 
-    fn set_cursor_impl(&mut self, offset: usize, set_start: bool, hard: bool) {
```
raphlinus: I'm not sure whether I prefer this, though it does work. I was thinking of keeping this.

rust/src/editor.rs
```diff
@@ -530,27 +597,37 @@ impl Editor {
             "render_lines" => Some(self.do_render_lines(params)),
             "key" => async(self.do_key(params)),
             "insert" => async(self.do_insert(params)),
+            "delete_forward" => async(self.delete_forward()),
             "delete_backward" => async(self.delete_backward()),
+            "delete_to_beginning_of_line" => async(self.delete_to_beginning_of_line()),
             "delete_to_end_of_paragraph" => async(self.delete_to_end_of_paragraph(kill_ring)),
             "insert_newline" => async(self.insert_newline()),
-            "move_up" => async(self.move_up(0)),
-            "move_up_and_modify_selection" => async(self.move_up(MODIFIER_SHIFT)),
```
raphlinus: I was thinking of keeping an argument, just having it a bool for `modify_selection`. So `move_up` would call `self.move_up(false)` and `move_up_and_modify_selection` would call `self.move_up(true)`. In general, I like less statefulness and more of a functional style.

rust/src/editor.rs
```diff
@@ -198,7 +191,26 @@ impl Editor {
         }
     }
 
+    fn delete_forward(&mut self) {
+        if self.view.sel_start == self.view.sel_end {
+            self.move_right();
+        }
+
+        self.delete();
```
raphlinus: When we actually implement grapheme clusters, this won't be quite the correct behavior. I'd prefer it to determine `next_grapheme_cluster` and delete that explicitly.

rust/src/editor.rs
```diff
@@ -198,7 +191,26 @@ impl Editor {
...
+
     fn delete_backward(&mut self) {
+        self.delete();
+    }
+
+    fn delete_to_beginning_of_line(&mut self) {
+        self.modify_selection();
+        self.move_to_left_end_of_line();
+
+        self.delete();
```
raphlinus: This has interesting behavior when the cursor is already at the beginning of the line - it deletes the newline at the end of the previous line. That could actually be useful behavior but doesn't match the Cocoa model (TextEdit does no-op in that case).

raphlinus:
Thanks for the revisions! I've added detailed comments, I hope you find them useful, and I appreciate your work in refining the patch.

rust/src/editor.rs
subumm1: On Note.app this behavior is wrong. But on "Sublime Text"  this behavior is right. How you think about it?

rust/src/editor.rs
subumm1: I think self.move_up(false) is not explicit express.
But... it just taste.

I'll change it if you want.

subumm1:
I revision it to function style.

raphlinus:
Awesome, thanks very much for accommodating my preferences!

PR #82, opened 2016-06-04
Separate edit command reading (from JSON) from command execution

bch29:
Previously each of the edit commands parsed its own parameters before executing. A lot of `unwrap()` was used, along with returning dummy values like `Value::Null` that should be error values instead. This PR improves that situation by separating out method parsing logic into a new module. Now all of the `do_*` functions in the `editor` module are much simpler and can't fail because of invalid parameters.

This approach also lends itself to easier testability since the parameter parsing and command execution can be tested separately.


bch29:
I signed it.

raphlinus:
Thanks for sending this. I definitely see the benefits, but one hesitation is that overall I generally prefer more concise code. In any case, this would have significant conflicts with #77 and that's ahead in line. So I'll think on this a bit while that's landing.

Sorry for the slow response to your PR, I've been traveling for family and work reasons.

bch29:
I'm happy to do the heavy lifting in merging this with #77 if you decide to use this.

In general I'm a big fan of using more precise types where possible. That is, data with the right type should always (or usually) be valid. Rust gives us a lot of great tools to achieve this, and using `Value` everywhere doesn't take advantage of Rust's type system at all.

Another advantage of this approach is that the module with the `EditCommand` enum in it could be factored out into a library. If I provide a `to_json` method, plugins can use this library too. That way, any plugins written in Rust could make use of this and be guaranteed to produce correct JSON data to send to the core.

raphlinus:
Sorry for the long delay. Yes, this seems like a good thing, happy to merge it after it's been updated.

bch29:
The merge is done. I haven't tested it with the new shortcuts since I'm not on a mac, so I can't run the up-to-date Cocoa interface. Let me know if anything is broken, but I don't think anything is.

raphlinus:
Thanks for this, and thanks for revising the patch. I tested most command bindings on Cocoa, and everything seemed to work. If anything else turns up, I'm sure it'll be easy enough to fix.

Issue #51, opened 2016-05-02
A better icon.


xpol:

raphlinus:
I was hoping this was a pull request, not just an issue :)

What do you think of this? I just dashed it out in Inkscape, could certainly polish it more.

![xi-logo](https://cloud.githubusercontent.com/assets/242367/14947778/3beeba54-0fed-11e6-82e7-e273a1fb903c.png)

xpol:
Sorry, I'm just a programmer, has no gift in art. 😭

rkusa:
I am also just a programmer and my graphical skills leave much to be desired. However, I'd still like to contribute at least some general ideas. So please consider the following not as final proposals, but just as ideas.

![xi1](https://cloud.githubusercontent.com/assets/409021/14951206/86a4d7ee-1057-11e6-8120-8f3dcad54492.png)
![xi2](https://cloud.githubusercontent.com/assets/409021/14951207/88358ba8-1057-11e6-89ec-7c584ab3d927.png)

Credits: Seen the idea of using highlighted code represented as blocks on a replacement Sublime Text icon some time ago. The colors are from the base16 ocean color scheme. Instead of drawing the blocks by hand, I've used the BLOKK Neue font. "Xi" is written in Source Code Pro. The code used is:

```rust
impl Default for Editor {
    fn default() -> Editor {
        Editor {
            text: Rope::from(""),
            view: View::new(),
            dirty: false,
            delta: Delta::new(),
            new_cursor: None,
            scroll_to: Some(0),
            col: 0
        }
    }
}
```

@raphlinus Does the name *Xi* has a special meaning? If so, the icon could be based around this meaning.

disordinary:
According to Wikipedia Xi is the Greek letter Ξ or ξ it's also the Romain numerals for 11.

The thing to think about is what will set Xi apart from the other editors that exist; as far as I can gather it's mainly being platform native, the client/server architecture, and the performance goals. And from an end users perspective its the native ui and performance that will set it apart from something like atom.

With the above in mind and considering raphlinus's background I'd suggest simplicity with a beautiful font.

leavengood:
[...]

PR #87, opened 2016-06-28
Added redesigned logo.

dragostis:


dragostis:
@googlebot I've signed it.

yaskhan:
I do not like. There is a better icons:
https://github.com/google/xi-editor/issues/51

dragostis:
@yaskhan Not trying to be mean, but try to consider your reply a bit.

> I do not like.

Okay, I totally respect that. Someone on the internet does a language for a project for free, for fun, and you simply say that you don't like it without backing up your feedback without any constructive criticism. I may not take offense, but you might want to try a different approach when criticizing someone's work the next time because other people might. 😃 

> There is a better icons:

I did not know about the issue. It is my fault for not searching, but I was pretty sure there wouldn't be any because Xi is a very new project. The issue here is that, again, you're not backing up your assessment. If you think other icons are better, make sure you say why.

PixelPirate:
[OS X Human Interface Guidelines - Icons](https://developer.apple.com/library/prerelease/content/documentation/UserExperience/Conceptual/OSXHIGuidelines/Designing.html#//apple_ref/doc/uid/20000957-CH87-SW1)

[Google Material Design Guidelines - Icons](https://material.google.com/style/icons.html#icons-system-icons)

One of the goals of this project is to use native UI on each platform, so I guess the main critique on your work is that it does not seem to fit with the general ascetic most icons have. (e.g.: The green rectangle background feels strange for a icon, at least for me)

dragostis:
@PixelPirate Sure! The actual idea is not to use the as an icon, per se, but as a logo. Of course, you can change the shape/background. What I want to emphasize on is the idea to use part of the "X" to symbolize the cursor.

dragostis:
@PixelPirate Here's an example of usage.

os x icon template 1 0

raphlinus:
Thanks for the contribution! However, the best process for this is to get feedback on #51 and see if it's well-loved.

beefsack:
Perhaps this one should be closed as a dupe in the meantime.

raphlinus:
Kinda dropped the ball on this, sorry.

There's obviously not consensus to use this logo; I think we're gonna have to come back to #51 and have more discussion there.

@yaskhan I really want this to be a pleasant community to work in. So I do value your feedback, but ask that in the future you try to be a little more constructive and respectful of the effort our contributors put in. Thanks!

PR #94, opened 2016-07-17
Changed to use a more reliable path for plugin loading

drPytho:
A better way to acquire the executable's path.

[`pub fn args_os() -> ArgsOs`](https://doc.rust-lang.org/std/env/fn.args_os.html)
> Returns the arguments which this program was started with (normally passed via the command line).
>The first element is traditionally the path of the executable, but it can be set to arbitrary text, and it may not even exist, so this property should not be relied upon for security purposes.

[`pub fn current_exe() -> Result<PathBuf>`](https://doc.rust-lang.org/nightly/std/env/fn.current_exe.html)
Will get the path and provide error message if it fails.


raphlinus:
Thanks, that's helpful!

PR #97, opened 2016-07-20
use arrays for JSON-RPC request params

rkusa:
I was updating a test plugin to use [golang.org/pkg/net/rpc/jsonrpc](http://golang.org/pkg/net/rpc/jsonrpc). However, Go always dropped the `ping` and `ping_from_editor` messenges without even raising an error. While digging into the Go source I found out that it expects an array for the `params` property. This is also part of the [specification](http://json-rpc.org/wiki/specification):

> *params* - **An Array** of objects to pass as arguments to the method.

That being said, I changed the `ping` param from null to an empty array, and I changed the `ping_from_editor` param from just `buf_size` to an array containing `buf_size`.

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -693,7 +693,7 @@ impl Editor {
 
     pub fn on_plugin_connect(&mut self, peer: &PluginPeer) {
         let buf_size = self.text.len();
-        peer.send_rpc_async("ping_from_editor", &Value::U64(buf_size as u64));
+        peer.send_rpc_async("ping_from_editor", &Value::Array(vec![Value::U64(buf_size as u64)]));
     }
 
     // Note: the following are placeholders for prototyping, and are not intended to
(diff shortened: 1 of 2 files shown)


rkusa:
I signed it!

raphlinus:
Yup, I was sloppy about this. Thanks!

rkusa:
I just noticed that this is also the case for `get_line`, `set_line_fg_spans` and also `alert` that I myself added recently. The question is probably if Xi should adhere to to the *params must be an array for requests* rule, since it is neither a strict v1 nor a strict v2 implementation?

raphlinus:
It doesn't have to be an array, it can also be an object. It just can't be null or a scalar. So I think we're fine, and I don't think that's too onerous a requirement.

I took a look at `net/rpc/jsonrpc` a month or two ago, and am not surprised you concluded it's not easy to adapt for xi. I think some off-the-shelf implementations will be, others won't, but in the latter case it's reasonably simple to roll one from scratch.

PR #100, opened 2016-07-21
remove unnecessary lifetime annotation from MainPeer

rkusa:
I think the lifetime annotation for `MainPeer` is, after your recent refactoring, not necessary anymore, isn't it?

Side question: what is your feeling about using rustfmt?


raphlinus:
Thanks for the cleanup.

I'm fine with rustfmt. I haven't been too careful about following a style guide (or using clippy for that matter). Probably a good time to add those tools to my daily toolbox.

Issue #99, opened 2016-07-21
RPC notification from backend to frontend


rkusa:
If I understood the backend correctly, the current message flow is as follows:

Text Edit → RPC notification from frontend to backend  → Tabs::do_rpc -> Editor::do_rpc → Editor::render – [if is dirty] → RPC update notification from backend to frontend

Are there already plans to where to plug in other notifications from backend to frontend, e.g. alerts? Since the editor does not have direct access to the `rpc_peer` currently, the editor can only send notifications from its render method (which receives the `rpc_peer` from the tab context). I am considering doing a PR, but want to make sure that it goes into the right direction.

raphlinus:
You're quite right; since the plugin sends notifications asynchronously, the editor needs a way to notify the front-end other than inside a call from Tabs. So you're right, it needs to hold a reference to the `rpc_peer`.

I think it needs to be weak reference, because the peer's handler holds a strong reference to the tabs, which in turn owns the editor. The same will probably be true of the link between the editor and the plugin (the editor will need to hold a reference to the plugin to notify it when the content changes).

A PR would be welcome!

rkusa:
Thanks for the feedback! I'll try to integrate it

rkusa:
I played around with different ideas of where and how to implement this.

My feeling is to try to avoid using `Arc<T>`, `Rc<T>` and Co if possible. This is why I actually like the current implementation where a reference to the `rpc_peer` is only handed out purposely.

That being said, my current proposal would be to handle this similar to how receiving the RPC messenges work. I'll try to illustrate the general outline:

- each tab has a queue of pending notifications
- `main.rs` spawns a thread that is waiting for content in these queues
- if there is content, it will hand out the `rpc_peer` for sending out the notification

This would also allow to prioritize notifications to the active tab and maybe even suppress (or delay) notifications to tabs in the background.

However, maybe it is not possible to completely avoid using `Arc<T>` or something similar, because all editors propably need to share the same `Condvar` to notify the thread of new content in the queue.

(Since I am not really experienced in Rust, it is possible that I am totally wrong—if this is the case, please tell me! 😉)

raphlinus:
[...]

PR #101, opened 2016-07-22
RPC notification from backend to frontend

rkusa:
This PR implements RPC notification from backend to frontend, as already discussed in #99.
It also had to already contain the PR #100.

I went for using `Arc<T>` instead of the discussed queue approach. Mainly because it is simpler and also has, as @raphlinus mentioned, slightly better performance.

Some points for discussion:

- I've used `Arc<T>` instead of `Arc<Mutex<T>>` because we do not need write access to the `rpc_peer` object and `RpcPeer.send` already has uses a mutex when writing data (`self.0.writer.lock()`). But maybe I've missed something?
- We now call `self.rpc_peer.send_rpc_async` from both the editor itself and also from the `TabCtx`. Maybe `TabCtx` can be removed and integrated into editor, now?

If wished, I can also remove the `alert` plugin method from this PR.

screen shot 2016-07-22 at 10 16 14

### DIFF
--- a/XiEditor/AppDelegate.swift
+++ b/XiEditor/AppDelegate.swift
@@ -44,11 +44,23 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     func handleRpc(method: String, params: AnyObject) {
-        if method == "update" {
+        switch method {
+        case "update":
             if let obj = params as? [String : AnyObject], let update = obj["update"] as? [String : AnyObject] {
                 // TODO: dispatch to appropriate editView based on obj["tab"]
                 self.appWindowController?.editView.updateSafe(update)
             }
+        case "alert":
+            if let obj = params as? [String : AnyObject], let msg = obj["msg"] as? String {
+                dispatch_async(dispatch_get_main_queue(), {
+                    let alert =  NSAlert.init()
+                    alert.alertStyle = .InformationalAlertStyle
+                    alert.messageText = msg
+                    alert.runModal()
+                });
+            }
+        default:
+            print("unknown method from core:", method)
         }
     }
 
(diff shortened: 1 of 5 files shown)


raphlinus:
Thanks for doing this! I have no problem with an alert method, it's a good placeholder for plugins starting to drive the UI in various ways.

I see this PR didn't plumb updates from the plugin back through the core to the frontend. Were you planning to do that, or would you like to hand that off to me as I continue building out the plugin infrastructure?

rkusa:
True, I forgot about that. But since I am currently finishing some changes on the lint plugin, which requires this functionality, I should have a PR ready shortly.

raphlinus:
Just some quick notes as I'm diving into this code:

I was mistaken when I suggested that the `RpcPeer` contains a reference to the tabs state. The _handler_ does, but `RpcPeer` doesn't own the handler, it's passed it as an argument to `mainloop`. Thus, we _don't_ need to break reference cycles here. I also think we can make RpcPeer cloneable, which would avoid having two nested `Arc`s. There will still be reference cycles between Editor and plugin, but I'm not sure whether you're going to run into that before I add notifications from Editor to plugin.

Much of `TabCtx` can move into `Editor`, certainly `update_tab` and the state it needs (tab name). But the kill ring is shared across all editors. Also keep in mind that there will at some point be a one-to-many relationship between editor buffers and tabs.

rkusa:
Thanks for the input! Unfortunately, I run out of time while finishing changes to the xi-golint plugin. I do not have a PR ready yet and I am also not able to continue work on this until starting next week. If you have the desire to continue work on this in the meantime, it is of course fine for me.

Issue #95, opened 2016-07-19
split the update method


eyelash:
As not all fields of the update method are always set (if you send an undo, you'll get an update without scrollto for example) I propose to split the update method into 4 methods for height, lines, scrollto and cursors.

For example like this:

```
update_height 12
```

```
update_lines {
    "first_line": 0,
    "lines":[["hello",["sel",4,5]]]
}
```

```
update_scrollto [0,4]
```

```
update_cursors [[0,4]]
```

eyelash:
The current design is also not optimal for multiple cursors (the front-end would need to store the cursors per line to know which ones are added and which are removed) so what about just splitting out the cursors from the update?

raphlinus:
I think what I'd like to do for that is send the line's cursors for each line which has a change of cursors. So it'd look something like `"cursors": [{"line": 3, "cursors": []}, {"line": 4, "cursors": [1]}]` (that's something you'd see when moving the cursor from line 3 to 4). Each of the elements of height, lines, cursors, scrollto would be optional. Would that meet your needs?

Splitting off cursors into a separate method still has the problem I mentioned above; the front-end would draw with stale cursor info (the cursor might not even address a valid location), then draw correctly.

eyelash:
This would be a field in the update method, right? So why not just `"cursors": [[4,1]]` and only send the cursors that are inside the visible area? Or why separate them per line? The front-end needs to draw and therefore store them separately anyway (to support blinking and stuff).

raphlinus:
Yes, a field in the update method.

The core doesn't know the exact visible area, it's just a guess. On Cocoa, the UI toolkit prerenders well outside the window, to make scrolling smoother.

I think you're right though that storing them per-line is over-complicating things. Probably best to just send all the cursors every time they change.

eyelash:
I just realized my idea wouldn't work so well with the render_lines call so per lines cursors is probably still the best approach. Sending all the cursors means the front-end would need to decide which cursors are visible and I'm not sure how well this would scale with lots of cursors (for example searching for something common inside a large file and inserting cursors for each result).

eyelash:
So then what about sending each line like this:
```json
{
	"index":3,
	"text":["hello",["sel",1,3],["fg",2,4,255]],
	"cursors":[]
}
```
Either text or cursors would be optional and the lines field of update would be an array of such lines. This way you could also send one update if for example only lines 3 and 5 changed.

DivineDominion:
[...]

PR #102, opened 2016-07-22
Typed events

DivineDominion:
**This is a work-in-progress mainly to open the topic for discussion before I move along.**

I propose that RPC commands be typed properly so that we don't run into the usual "stringly" typing problems and let the compiler do most of the work.

- [ ] Don't dispatch from `AppWindowController` into its `editView` directly (like `editView.sendRpcAsync("save", params: ["filename": filename!])`)
- [ ] Type `EditView` events for text editing

What does the community think about this approach?

---

I don't like the `EventDispatchMethod` enum very much to distinguish between sync/async sending of RPC requests. Maybe changing the method signature to `Event.dispatch(dispatcher: Dispatcher, method: EventDispatchMethod = .sync) -> Output` will be necessary, too, so the dispatching method can vary on a per-call basis instead of per-event.

### DIFF
--- a/XiEditor.xcodeproj/project.pbxproj
+++ b/XiEditor.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		50370A2B1D4209F300EA1B18 /* Dispatcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 50370A2A1D4209F300EA1B18 /* Dispatcher.swift */; };
 		AE228F5D1C897CE7000E9B0F /* xi-core in Resources */ = {isa = PBXBuildFile; fileRef = AE228F5C1C897CE7000E9B0F /* xi-core */; };
 		AE499DD01C82BB2B002D68AF /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE499DCF1C82BB2B002D68AF /* AppDelegate.swift */; };
 		AE499DD21C82BB2B002D68AF /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = AE499DD11C82BB2B002D68AF /* Assets.xcassets */; };
@@ -19,7 +20,6 @@
 		AED23F151C83C689002246CE /* AppWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = AED23F131C83C689002246CE /* AppWindowController.swift */; };
 		AED23F161C83C689002246CE /* AppWindowController.xib in Resources */ = {isa = PBXBuildFile; fileRef = AED23F141C83C689002246CE /* AppWindowController.xib */; };
 		AED23F181C83CBED002246CE /* EditView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AED23F171C83CBEC002246CE /* EditView.swift */; };
-		DBB50DF11CDE819000B4BBBE /* xi-core.sh in Resources */ = {isa = PBXBuildFile; fileRef = DBB50DF01CDE819000B4BBBE /* xi-core.sh */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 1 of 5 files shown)


DivineDominion:
Ha, this is weird, but hey. "I signed it!" :)

raphlinus:
I took a quick look at the patch but want to study it a bit more, perhaps have some comments.

I'm sympathetic to stronger typing, though as I'm prototyping I'm finding value in a more dynamic (or "stringly typed") approach - the total size of a patch to add or change an RPC is larger as it goes through conversions to a typed value.

One comment, in the json-rpc work going on for plugins, it was suggested to use terminology consistent with json-rpc, namely "request" for a synchronous RPC and "notification" for an asynchronous one. Maybe a better idea to use that terminology in enums, etc.

DivineDominion:
Didn't know about these names, but I like how much more expressive they are. "Notification" sounds strange at first (is the client notifying the service or is the resulting callback the notification?) but works well. 👍

Sure the drawback is that you have to introduce a new type for each new kind of RPC call. I find the cost negligible. Stringly typing is all good and well when your code is backed by tests. (That's why Rubyists tend to be very strong test advocates.) But introducing tests is just as much an impediment to changes as types are, only that the production code becomes a bit "safer" directly. Ultimately, it's your call, of course. Prototyping and solidifying features can be done in two steps as well, I think.

Regarding prototyping: this seems to imply the RPC interface isn't fully planned, yet. Is there an issue for discussions or something to participate?

raphlinus:
> Regarding prototyping: this seems to imply the RPC interface isn't fully planned, yet. Is there an issue for discussions or something to participate?

#43 is where a bunch of that discussion is happening, and also on the [reddit thread](https://www.reddit.com/r/xi_editor/comments/4t4vz6/plugin_progress/) linked from there.

raphlinus:
After thinking about it a couple days, I'm going to merge this because I think it's progress - typed is better, I agree. I'm still not 100% convinced it's what I want the final state to look like. I'd be interested in exploring something that looks like more like my recent commits on the plugin, where there's a layer (plugin_base) that wraps the RPC protocol and exports a plain procedural interface, no explicit dispatch calls.

In its current state that layer has flaws, mostly not handling errors well. I think it could go further in a couple of ways. First, I might have a trait (protocol) for incoming requests, rather than a callback that takes a request object. That would cause the return type of the response to be typed as well, which I think would be a win. Second, and I think this is of particular interest for notifications (async rpc's) from front end to core, there might be an intermediate generic RPC layer, so (say) a method like `move_right` calls a generic "command" method with `"move_right"` as a string parameter. Then, if the transport changes from json-rpc to something else, only the latter has to change. Right now, a lot of that stuff goes through `doCommandBySelector`, which is expedient (there's no essentially code on the front-end side for implementing a new command), but definitely not consistent with the goal of having RPC's be more strongly typed.

This area is going to be a bit fluid (I know people are talking about motion-based editing), so I want to let people try things. If a strong consensus emerges on the Right Way to do things, we'll make everything consistent with that. Thanks for contributing!

DivineDominion:
Ah, great. I planned to convert other events before you merged, but doing it afterwards will work just as well.

I agree that the RPC stuff could be further encapsulated. Right now, UI objects directly talk to the RPC service. From an app architecture standpoint, a cleaner approach would be to further tear these two apart.

My usual go-to-approach is to define a UI event interface (be it typed NSNotifications or event handler protocols). A mediating layer then takes care of mutating state/the model. Now the fact that the state isn't part of the Cocoa app, the distinction is easier to keep up conceptually.

Have you looked into a unidirectional data flow approach, or how to make the interface more reactive?

Usually, smaller apps of overwrite the UI contents with new state once the state successfully incorporates the changes made. I imagine an editor backend that pushes all the file's contents to the app can get kinda clumsy when syntax highlighting is added.

raphlinus:
A couple of thoughts.

First, there's strong interest in making commands more structured. Factoring the movement out has been suggested (this would be very much in line with vi). You might not be keen to go through the effort of making an enum with dozens of no-argument actions, just because that's the way Cocoa does it, only to have them significantly refactored later.

I have no idea what a unidirectional data flow or a reactive approach would even mean in the context of xi. In any case, we are _absolutely_ not pushing the entire state to the front-end. Rather, there's a protocol where the core pushes an update that contains _part_ of the state (currently invalidating all the rest, but that's going to change), and any additional state that the front-end requires, it issues a synchronous RPC request to get. The front-end is effectively holding a cache, and there's a cache invalidation protocol. That protocol needs to become even more sophisticated (see #95 for more discussion), but it works reasonably well for now. I haven't done anything remotely resembling careful measuring yet, but I doubt that syntax coloring spans will be a problem.

I'm sure this is different from most apps. In any case, I'm letting performance drive the design.

I agree separating out the cache from `EditView` would be a good move. I'll probably move in that direction when I do more work on minimal invalidation.

Happy to discuss more - I'm not sure we're quite connecting here.

DivineDominion:
I'd have to understand what the Rust code does, first, to make a better judgement it seems :)

The unidirectional data flow thing struck me to be similar, but if the RPC back-end already sends the changes instead of the whole state, you don't need any of the stuff Flux/Redux/ReSwift worries about. The approach is similar, though: the "truth" comes from a single state representation for which changes are pushed to the UI. Only xi is doing the diffing/change computation in a service instead of another layer in the same app.

What I understood of your thoughts on minimal invalidation sounds good. But I cannot be of much help as long as I can only grasp the Swift side. Porting all of `NSResponder` to actions can become very unwieldy indeed. Movement events needn't be `struct`s; they can be `enum` cases, too, with a bit of extra care. But still, it's quite something.

PR #103, opened 2016-07-23
Make the mouse cursor become an IBeamCursor on mouse over

Gbear605:
This is the standard for all text editing fields on MacOS.


raphlinus:
Thanks for doing this! I was thinking about it recently, in the context of what would be required for modal editing (it should be an IBeam in insert mode, probably a regular arrow in command mode). But this looks right for what we're doing now.

PR #104, opened 2016-07-25
Propagate Plugin changes to the frontend

rkusa:
When thinking about my implementation of #101 during the last couple of days, I came to the conclusion that I prefer the idea of having the `TabCtx` as some kind of glue instead of making the editor to have references to everything.

With that in mind, the idea was now to introduce a `PluginCtx`, which is basically a `TabCtx` with an additional `PluginPeer`, even though this `PluginPeer` reference is not yet used. But it should pave the way for getting change notifications from the editor to its plugins. I also got rid of the nested `Arc` for the `MainPeer` (at least if I understood this one right).

Things that could be improved:

- [ ] I made the `TabCtx` cloneable, to be able to add a clone to the `PluginCtx`. Therefore, I had to wrap the `TabCtx` `kill_ring` into an `Arc`. Maybe there is better way to solve this?
- [ ] Maybe it is possible to change the `rpc_peer` of the `PluginCtx` from `Option<PluginPeer>` to `PluginPeer`, but I wasn't able to achieve this because of the lifetime requirements of the plugin thread.
- [ ] Do not run `editor.render()` after each plugin command, but rather collect some changes and send them in bulk. Not yet sure how to do implement this in a nice way ...
- [ ] Implement notifications form editor to its plugins.

For the latter, it is probably necessary that either the tabs or the editor keeps a reference to the `PluginCtx`. However I am not yet sure how to implement this; e.g. should the `PluginCtx` be cloned or an `Arc` be used for the `PluginCtx`. Up to now, I am not quite happy with both, but I'll keep thinking about it.

Please understand this PR mainly as a basis for discussions. If you prefer the implementation without the `PluginCtx`, I am totally fine if we close this PR. Also, if some parts are hard to understand, please tell me and I'll try to rephrase them.

### DIFF
--- a/rust/src/tabs.rs
+++ b/rust/src/tabs.rs
@@ -28,20 +28,20 @@ use MainPeer;
 pub struct Tabs {
     tabs: BTreeMap<String, Arc<Mutex<Editor>>>,
     id_counter: usize,
-    kill_ring: Mutex<Rope>,
+    kill_ring: Arc<Mutex<Rope>>,
 }
 
-pub struct TabCtx<'a> {
-    tab: &'a str,
-    kill_ring: &'a Mutex<Rope>,
-    rpc_peer: &'a MainPeer,
+#[derive(Clone)]
+pub struct TabCtx {
+    tab: String,
+    kill_ring: Arc<Mutex<Rope>>,
+    rpc_peer: MainPeer,
     self_ref: Arc<Mutex<Editor>>,
 }
 
 pub struct PluginCtx {
-    main_peer: MainPeer,
-    plugin_peer: Option<PluginPeer>,
-    editor: Arc<Mutex<Editor>>,
+    tab_ctx: TabCtx,
+    rpc_peer: Option<PluginPeer>,
 }
 
 impl Tabs {
@@ -49,7 +49,7 @@ impl Tabs {
         Tabs {
             tabs: BTreeMap::new(),
             id_counter: 0,
-            kill_ring: Mutex::new(Rope::from("")),
+            kill_ring: Arc::new(Mutex::new(Rope::from(""))),
         }
     }
 
@@ -64,7 +64,7 @@ impl Tabs {
                 None
             },
 
-            Edit { tab_name, edit_command } => self.do_edit(tab_name, edit_command, &rpc_peer),
+            Edit { tab_name, edit_command } => self.do_edit(tab_name, edit_command, rpc_peer),
         }
     }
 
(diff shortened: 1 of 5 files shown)


rust/src/tabs.rs
```diff
@@ -116,7 +122,53 @@ impl<'a> TabCtx<'a> {
...
-        self.self_ref.clone()
+    pub fn to_plugin_ctx(&self) -> PluginCtx {
+        PluginCtx {
+            tab_ctx: self.clone(),
+            rpc_peer: None,
+        }
+    }
+}
+
+impl Clone for TabCtx {
```
raphlinus: Can you derive clone?

raphlinus:
Thanks for digging in to this! I had a quick scan over the patch and it looks good to me, but I'll review it more carefully this evening.

We will probably end up with a fully wired object graph, for the editor to propagate changes to the plugin. Another complication to consider is that we may end up multiplexing a single plugin to handle multiple tabs (this has the potentially significant advantage of eliminating the plugin startup cost after the first tab). I'm not sure the best way to do this - I'll try to think on it some more as I review your patch more carefully.

rust/src/tabs.rs
rkusa: I've updated the PR

rkusa:
> Another complication to consider is that we may end up multiplexing a single plugin to handle multiple tabs (this has the potentially significant advantage of eliminating the plugin startup cost after the first tab).

This is a good point! Instead of keeping plugins and the editor tightly coupled (as still the case in this PR), moving the plugin handling (startup, communication, ...) out of the editor, sounds actually quite good. The plugin could then maybe even considered–at least at the technical point of view–to be equal to the actual front-end. So in other words, we don't have one front-end and many plugins, we have just many front-ends.

rkusa:
I couldn't resist to do a [proof of concept implementation](https://github.com/rkusa/xi-editor/blob/plug-main/rust/src/main.rs) of moving the plugins into the `main.rs`. Doing so definitely comes with some challenges. Since it is only a proof of concept implementation, the code is ugly and there are many parts that would require improvement. That is, if the discussion of how to implement this, tends to go into this direction of moving plugins into the the top level, I can highlight the major challenges/issues that I had with doing so.

Having live syntax highlighting was nice anyway 😎

output

raphlinus:
Thanks for prototyping that, it's something good to look at. I was tempted to also build a prototype based on this PR, adding mutual references between front-end and core, and core and plugin, but didn't have the energy last night.

It's very tempting to just merge the functionality, so we can start moving the plugin protocol forward to accommodate OT / CRDT, among other things. Making the communication cleaner (and maybe more efficient) can happen in parallel. I'm ok merging this PR, but I can see that your fork is not quite ready yet.

What do you think? Merge this PR, then we can both play with fully wiring the event propagation, and then combine the best features of both?

rkusa:
Merging this PR and iterating on it sounds good!

On 26 Jul 2016, at 16:24, Raph Levien <notifications@github.com> wrote:

Thanks for prototyping that, it's something good to look at. I was tempted
to also build a prototype based on this PR, adding mutual references
between front-end and core, and core and plugin, but didn't have the energy
last night.

It's very tempting to just merge the functionality, so we can start moving
the plugin protocol forward to accommodate OT / CRDT, among other things.
Making the communication cleaner (and maybe more efficient) can happen in
parallel. I'm ok merging this PR, but I can see that your fork is not quite
ready yet.

What do you think? Merge this PR, then we can both play with fully wiring
the event propagation, and then combine the best features of both?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/google/xi-editor/pull/104#issuecomment-235283551>, or mute
the thread

Issue #23, opened 2016-04-28
Syntax highlighting via LPeg


leavengood:
One possible suggestion for the method of syntax highlighting is via [LPeg](http://www.inf.puc-rio.br/%7Eroberto/lpeg/). This is used by the [Vis editor](https://github.com/martanne/vis), which borrowed it from the Scintillua project as described in this [README](https://github.com/martanne/vis/blob/master/lexers/README.md).

You can see all the languages supported in the [Vis lexers directory](https://github.com/martanne/vis/tree/master/lexers).

If you wanted to embed this in xicore there seem to be multiple Rust Lua bindings, such as these two:

https://github.com/kballard/rust-lua
https://github.com/jcmoyer/rust-lua53

Or maybe LPeg could be use from a Lua-based external tool using whatever future RPC mechanism you intend for "plugins". But this may make sense to be done inside the core.

raphlinus:
I am very uninclined to link a scripting language into the core, I feel that goes against the philosophy of the project. The plugin interface should facilitate this kind of thing well, over RPC.

To get this written down, my original plan was to implement TextMate-compatible syntax highlighting, which is why I did the fancy-regex library (to get Oniguruma-compatible regex parsing). That I was going to put in the core, so that plugins wouldn't be required for basic highlighting. However, I am very open to evidence about other systems which might be better (performance, accuracy, availability of high-quality syntax definitions which can be packaged with appropriate licensing). Thanks for the pointer to the LPeg ones, they very much look worth evaluating.

leavengood:
I understand your feelings on embedding a scripting language in the core. Lua is probably one of the best options though, and LPeg and the infrastructure provided by the [lexer.lua file](https://github.com/martanne/vis/blob/master/lexers/lexer.lua) is pretty nice. I'm not actually a Lua programmer (yet), but this is making me want to learn it a bit.

It is worth trying, but I don't know how well syntax highlighting over RPC would work.

TextMate-compatible is probably good too, as you get to inherit all the existing syntax definitions for TextMate and of course Sublime Text (before they changed the file format.) Sublime's new format is another option I guess.

The main point is to reuse another format so as not to have to reinvent all the syntax definitions, but I'm sure you have already planned for that.

leavengood:
I guess one nice benefit of LPeg is the parsers are composable, so it is easier to support things like CSS or JavaScript inside HTML or HTML and Ruby together in erb files, etc. I don't know how that is handled with the TextMate format.

[...]

PR #110, opened 2016-07-30
Plugin notify

raphlinus:
This is an experiment in wiring up the object graph with `Arc` references, and plumbing mutual update notifications between editor and plugin. As a result, it does implement live syntax highlighting (although currently with pretty poor performance, because I haven't started optimizing that yet).

@rkusa What do you think?

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -16,6 +16,7 @@ use std::cmp::max;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::collections::BTreeSet;
+use std::sync::{Arc, Mutex};
 use serde_json::Value;
 
 use xi_rope::rope::{LinesMetric, Rope};
@@ -28,7 +29,7 @@ use view::{Style, View};
 
 use tabs::TabCtx;
 use rpc::EditCommand;
-use run_plugin::start_plugin;
+use run_plugin::{start_plugin, PluginRef};
 
 const FLAG_SELECT: u64 = 2;
 
@@ -58,6 +59,9 @@ pub struct Editor {
     dirty: bool,
     scroll_to: Option<usize>,
     col: usize, // maybe this should live in view, it's similar to selection
+
+    tab_ref: Arc<Mutex<TabCtx>>,
+    plugins: Vec<PluginRef>,
 }
 
 #[derive(PartialEq, Eq, Clone, Copy)]
@@ -69,7 +73,7 @@ enum EditType {
 }
 
 impl Editor {
-    pub fn new() -> Editor {
+    pub fn new(tab_ref: Arc<Mutex<TabCtx>>) -> Editor {
         let engine = Engine::new(Rope::from(""));
         let last_rev_id = engine.get_head_rev_id();
         Editor {
@@ -88,6 +92,8 @@ impl Editor {
             new_cursor: None,
             scroll_to: Some(0),
             col: 0,
+            tab_ref: tab_ref,
+            plugins: Vec::new(),
         }
     }
 
@@ -185,8 +191,9 @@ impl Editor {
     }
 
     // render if needed, sending to ui
-    pub fn render(&mut self, tab_ctx: &TabCtx) {
+    pub fn render(&mut self) {
         if self.dirty {
+            let tab_ctx = self.tab_ref.lock().unwrap();
             tab_ctx.update_tab(&self.view.render(&self.text, self.scroll_to));
             self.dirty = false;
             self.scroll_to = None;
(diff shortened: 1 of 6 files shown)


rust/src/editor.rs
```diff
@@ -179,17 +187,24 @@ impl Editor {
...
-        if self.dirty {
+    pub fn render(&mut self) {
+        if self.text_dirty || self.view_dirty {
+            let tab_ctx = self.tab_ref.lock().unwrap();
             tab_ctx.update_tab(&self.view.render(&self.text, self.scroll_to));
-            self.dirty = false;
             self.scroll_to = None;
+            if self.text_dirty {
+                for plugin in &self.plugins {
+                    plugin.update();
```
rkusa: Do you plan to send some more information to the plugin later on? Like sending the same notification the frontend gets (the result of `&self.view.render(&self.text, self.scroll_to)`)?

rust/src/editor.rs
```diff
@@ -552,19 +567,24 @@ impl Editor {
...
+        self.view_dirty = true;
     }
 
-    fn debug_run_plugin(&mut self, tab_ctx: &TabCtx) {
+    fn debug_run_plugin(&mut self) {
         print_err!("running plugin");
-        let plugin_ctx = tab_ctx.to_plugin_ctx();
-        start_plugin(plugin_ctx);
+        let tab_ctx = self.tab_ref.lock().unwrap();
+        start_plugin(tab_ctx.get_editor_ref());
```
rkusa: If I haven't missed anything, this is the only position where the `editor_ref` from the `TabCtx` is used. In this case, the editor could maybe manage its `Arc<Mutex<Editor>>` on its own and the `editor_ref` from `TabCtx` could be removed?

rust/src/tabs.rs
```diff
@@ -57,46 +51,49 @@ impl Tabs {
...
         self.id_counter += 1;
-        let editor = Editor::new();
-        self.tabs.insert(tabname.clone(), Arc::new(Mutex::new(editor)));
+        let tab_ctx = TabCtx {
+            tab: tabname.clone(),
+            kill_ring: self.kill_ring.clone(),
+            rpc_peer: rpc_peer,
+            editor_ref: Weak::new(),
+        };
+        let tab_ref = Arc::new(Mutex::new(tab_ctx));
```
rkusa: Since the Editor is the only one using it, couldn't the editor just get the ownership of the `TabCtx` (instead of the Arc Mutex)?

rkusa:
Looks good! 👍
I still added some code comments, but they are just some minor thoughts I had while reading the changes.

rust/src/editor.rs
raphlinus: Yes, but it won't be the same information as for rendering views. It will be a delta of the string (possibly with the actual string replaced by a byte count when it's too large), an edit type (keyboard insert, delete, paste, undo, redo), the id of the agent doing the editing (user, plugin), the undo group, and the new revision number.

The revision number will be used for CRDT. The RPCs to fetch data give a revision number, and query the snapshot at that revision. In addition, updates are given relative to that revision.

The undo group is mostly so that the plugin can do edits in the same group. Example, if the user types "{", the plugin inserts "}" after, to balance the parentheses, and sets the same undo group. Then when the user undoes the insert "{", both get undone.

rust/src/editor.rs
raphlinus: Yes, thanks!

rust/src/tabs.rs
raphlinus: Indeed. Uploading a new revision that has these simplifications.

rust/src/editor.rs
raphlinus: I should also add that I'm strongly considering making this a request rather than a notification. When the request returns, it basically unreferences all previous revisions. The plugin would be responsible for getting all its pending updates sent before returning from the request. Of course, this would be implemented as a non-blocking call from the editor, with callback.

rkusa: Sounds good, thanks for the explaination!

rkusa:
👍 for the Update! I think this is now a good and easy to understand solution! Can't wait to update my proof of concept plugins!

raphlinus:
I think I'll merge the latest, but wanted to give you another chance to comment - is f7c263b an improvement? I think so, but it's a judgment call.

rkusa:
Getting rid of the self_ref of the Editor struct sounds even better! I think it is again an improvement. I may have some comments later, once I am back on my laptop. If you like, you can of course already go ahead and merge it. In case I have a comment later, I can also present it as a PR.

raphlinus:
Thanks for the comments and suggestions. I feel it's in a "good enough" state where I can focus on the actual RPC's between editor and plugin, but I'm definitely open to iterating if there are further improvements.

DemiMarie:
How big is the RPC overhead?  I feel like the syntax highlighting should be in the editor.

chriskrycho:
@DemiMarie, you might find the discussion in #23 helpful on that note—there's a *lot* of discussion both about how heavy the RPC is relative to the other things going on, as well as how and why this plugin architecture works for something like syntax highlighting.

PR #111, opened 2016-07-30
Improve Editor reference handling

rkusa:
This PR is based on the discussion in #110. It does not make any functional changes, but rather suggests an improvement of the internal API.

The part I am trying to improve is the following method from the `Editor` struct, which gets both a reference to itself (`&mut self`), but also an additional external reference to itself (`self_ref: &Arc<Mutex<Editor>>`).

``` rust
fn do_rpc_with_self_ref(&mut self, cmd: EditCommand, self_ref: &Arc<Mutex<Editor>>) -> Option<Value>
```

This PR highly borrows from the`RpcPeer` implementation. It makes `Editor` a `pub struct Editor(Arc<Mutex<EditorState>>);`. The previous `Editor` is renamed to `EditorState`.

The question would now be, if this change makes the code easier to understand, or not? If not, I would say we simply keep the current implementation! If yes, before merging it, we may need to consider, for the sake of consistency, which methods go into `Editor` and which in `EditorState` (currently only the ones needed to make the code to compile are in `Editor`).

### DIFF
--- a/rust/src/editor.rs
+++ b/rust/src/editor.rs
@@ -16,7 +16,7 @@ use std::cmp::max;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::collections::BTreeSet;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, Weak};
 use serde_json::Value;
 
 use xi_rope::rope::{LinesMetric, Rope};
@@ -37,7 +37,9 @@ const MAX_UNDOS: usize = 20;
 
 const TAB_SIZE: usize = 4;
 
-pub struct Editor {
+pub struct Editor(Arc<Mutex<EditorState>>);
+
+pub struct EditorState {
     text: Rope,
     view: View,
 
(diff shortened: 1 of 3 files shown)


rust/src/editor.rs
```diff
@@ -97,9 +99,100 @@ impl Editor {
...
+    }
+
+    pub fn on_plugin_connect(&self, plugin_ref: PluginRef) {
+        let mut state = self.0.lock().unwrap();
+        plugin_ref.ping_from_editor(state.plugin_buf_size());
+        state.plugins.push(plugin_ref);
+    }
+
+    pub fn downgrade(&self) -> Weak<Mutex<EditorState>> {
+        Arc::downgrade(&self.0)
```
rkusa: With this part, I am currently least happy with, because it creates the requirement of making the `EditorState` `pub`lic

raphlinus:
@rkusa Thanks for sending this PR, and sorry for taking so long to review it.

I honestly can't decide whether this is an improvement or a sideways move. The signature of the `downgrade` method is not very appealing to me - people will have to deal with both `Editor` and `EditorState` and it's not 100% obvious why one or the other is being used.

Given that the goal of this patch is clarity, what do other people think? I'll happily merge it if there's consensus it's an improvement.

rkusa:
No problem! When reflecting on the PR now, I also think that *a sideways move* probably describes this PR well. That being said and since it is not an obvious improvement, I think it is maybe not worth the time to invest further in it. Which is why I am also totally fine with just closing this PR to move on with focusing on actual plugin functionality instead. I still appreciate the feedback, so thanks for having a look anyway! 🙂

PR #120, opened 2016-09-14
Support Unicode 9.0

makotokato:
UAX#14 is updated to revision 37 for Unicode 9.0 support.  It adds some rules such as emoji.  I would like to add Unicode 9.0 rules to xi-unicode.

http://www.unicode.org/reports/tr14/tr14-37.html


raphlinus:
Thanks!

PR #122, opened 2016-09-21
faster newline count via bit twiddling

llogiq:
The alignment splitting technique as well as the idea to partially unroll the loop to 2×wordsize was nicked from @burntsushi 's rust-memchr crate, and the bit twiddling is classic Stanford. :smile:

In my benchmarks, I see between 100% and 200% speedup.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -17,3 +17,7 @@ path = "unicode"
 
 [dependencies.xi-rpc]
 path = "rpc"
+
+[features]
+avx-accel = ["xi-rope/avx-accel"]
+simd-accel = ["xi-rope/simd-accel"]
--- a/rust/rope/Cargo.toml
+++ b/rust/rope/Cargo.toml
@@ -8,3 +8,7 @@ version = "0.1.0"
 
 [dependencies]
 bytecount = "0.1.2"
+
+[features]
+avx-accel = ["bytecount/avx-accel"]
+simd-accel = ["bytecount/simd-accel"]
--- a/rust/rope/Cargo.toml
+++ b/rust/rope/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
-name = "xi-rope"
-version = "0.1.0"
-license = "Apache-2.0"
 authors = ["Raph Levien <raph@google.com>"]
-repository = "https://github.com/google/xi-editor"
 description = "A generic rope data structure built on top of B-Trees."
+license = "Apache-2.0"
+name = "xi-rope"
+repository = "https://github.com/google/xi-editor"
+version = "0.1.0"
+
+[dependencies]
+bytecount = "0.1.2"
--- a/rust/rope/src/lib.rs
+++ b/rust/rope/src/lib.rs
@@ -130,9 +130,49 @@ impl Debug for Rope {
 }
 */
 
-// TODO: explore ways to make this faster - SIMD would be a big win
+#[cfg(target_pointer_width = "16")] const USIZE_BYTES: usize = 2;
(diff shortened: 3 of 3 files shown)


BurntSushi:
Nice! One note: @bluss is the one responsible for the initial `memchr` implementation. I just started the crate. :-)

I also have a [fast line counting routine](https://github.com/BurntSushi/ripgrep/blob/0263a401f628ff910858b303b52e169559cd4ca9/src/search_stream.rs#L549) that is similar to the one you have here. One difference is that `count_zero_bytes` is implemented this way:

```rust
(x.wrapping_sub(LO) & !x & HI).count_ones() as u64
```

On architectures that support it, this will emit a `POPCNT` instruction, which gives *another* huge perf boost. (With `rustc`, I believe you need to explicitly enable this with either `RUSTFLAGS="-C target-cpu=native"` or `RUSTFLAGS="-C target-feature=+popcnt"`.)

llogiq:
Cool :sunglasses:  – I'll be sure to benchmark that, but first I need to catch some sleep. :sleeping:

llogiq:
@BurntSushi see [benchmark](/llogiq/newlinebenchmark) – I widened the loop to up to 8 words per run to amortize the bit counting better.

zdanz:
@llogiq @BurntSushi sorry if I'm missing something obvious, but it seems to me that both of your `count_zero_bytes`  / `count_eol`  implementations are relying on an algorithm which *detects* zero bytes -- it doesn't *count* them correctly.  For example if you pass in 0x100, you will get back 8, even though there are only 7 zero bytes.

Is there higher-level code which corrects this?  I confess that I didn't go over it carefully.

BurntSushi:
@zdanz See this for an explanation: http://bits.stephan-brumme.com/null.html --- Note how it is called: https://github.com/BurntSushi/ripgrep/blob/0263a401f628ff910858b303b52e169559cd4ca9/src/search_stream.rs#L621

(There's no other higher level code that corrects this in `ripgrep`.)

zdanz:
@BurntSushi Thanks for the reply! I did see the explanation at that link: note that it "detects" zero bytes. `hasZeroByte` there returns a `bool`.

 I thought I might be missing the point when I looked at the search_stream code you pointed me to, so I cloned ripgrep and added this line to the top of main():

`println!("zdanz: {}", search_stream::count_lines(&[1,0,1,1,1,1,1,1, 1,1,1,1,1,1,0,1], 1));`

It prints out 16 when it should print out 14.  Shouldn't it?  (If you replace the 0's with 2's, then it does print out 14.)

BurntSushi:
@zdanz That's interesting. I've done quite a bit of out-of-band testing (by comparing the output of `ripgrep` with `grep`), and I haven't seen it get line counts wrong yet. But, you're example is pretty compelling. I'll put it on my list to check out. @llogiq This might be a good time to write a quickcheck property the checks the naive solution with the optimized version.

llogiq:
I have a quickcheck in place, and the current version no longer exhibits the incorrect overflow behavior.

llogiq:
I just added features (similar to what ripgrep uses) to allow for SSE/AVX usage where available.

llogiq:
@raphlinus What's the status on this? Is there something we need to do before merging?

llogiq:
Again, is there something we need to do?

raphlinus:
Sorry, I was delaying merging this because I was uneasy about unsafe code, but now I consider bytecount to be part of the "honorary part of rust's stdlib so it's ok if it has unsafe".

llogiq:
Ah,I see :smile:

It may relieve you to hear that bytecount has reduced its use of unsafe code to one line where we build the slice to aligned memory.

PR #132, opened 2016-11-07
Do not insert chars with unassigned command key

modelorganism:
Type a command that key that Xi doesn't implement, such as cmd-f. 
Xi enters an 'f' into the document. This PR makes Xi just Beep on unassigned command keys.
Although this current behavior appears to be deliberate, I belive it is surprising, and likely to unintentionally edit documents. If I wanted to type a letter, I wouldn't be holding down the command key.
I not sure what Sublime or BBEdit do when they encounter an unassigned key, since I couldn't find any keys that they didn't assigned.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -442,7 +442,7 @@ class EditView: NSView, NSTextInputClient {
         } else {
             let commandName = camelCaseToUnderscored(aSelector.description).stringByReplacingOccurrencesOfString(":", withString: "");
             if (commandName == "noop") {
-                sendRpcAsync("key", params: eventToJson(currentEvent!));
+                NSBeep()
             } else {
                 sendRpcAsync(commandName, params: []);
             }


bbuck:
I'm not a fan of beeping unless it can be turned off. Atom handles this by doing nothing. Still seems a bit odd, but much preferred to a beep and/or having to mute my system if I accidentally hit keys in the editor. I get I'm just one voice but I thought I'd throw in my opinion on this matter. 

I do agree with the general fix though, Command+f should not enter an f into the editor.

modelorganism:
@bbuck  You may be right. I suspect this whole issue will become less important as more keys are assigned. Then Xi will only beep, or do nothing, when you hit one of the few blank command keys.
For now while I am not married to the beeping,  I do I think it would be nice if Xi didn't insert
on normally non-editing command keys.

raphlinus:
I think beeping is fine for now. It's a reminder that there's an unimplemented command. As @modelorganism points out, it'll become much less of an issue in practice as more commands are implemented.

Thanks for the PR, and sorry to take so long to review.

Issue #57, opened 2016-05-04
Implement tabs in the UI


rtsuk:
I've started work on this. The plan is to use an [NSSegmentedControl](https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/SegmentedControl/Articles/SegmentedControlBasics.html) to display the tabs and an [NSTabView](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/ApplicationKit/Classes/NSTabView_Class/) to hold one EditView per tab.

wezm:
Maybe consider [MMTabView](http://mimo42.github.io/MMTabBarView/) as well.

rtsuk:
It looks good but I will have to check with @raphlinus on whether he wants to use something like that here.

raphlinus:
MMTabView looks _really_ nice. I especially like the tear-off tab functionality, as I use that all the time. The license is compatible, and it looks like it would be quite a bit of work to duplicate what it already provides.

I don't think there's any good way to declare a reference to it and have Xcode resolve it the way that gradle or cargo would (correct me if I'm wrong), so probably the best way to get this dependency added is to copy it in under a third_party directory.

Some very minor polish issues: the "dirty" close buttons are lacking `@2x` (retina) PNGs, and some of the tab-reorder animations could be a little smoother. I'm sure both can be readily patched.

rtsuk:
There's CocoaPods, but I don't think MMTabView is hosted there and I've never used CocoaPods.

I will experiment with MMTabView. Thanks @wezm!

raphlinus:
So Sierra is out and has tabs built in. Things work to some extent without any code changes, although windows open by default to new windows, so getting to tabs is not very discoverable (there's a "Merge All Windows" added to the Window method).

I think the app delegate needs to implement newWindowForTab to get the "+" button, but that's easy enough.

Should the app window controller have its tabbing mode set to preferred? That would more or less match the Sublime behavior, but I'm not sure it follows the macOS recommendations. Also, I'm not sure how easy it will be to get a side bar (see #78) with tabs.

PR #136, opened 2016-11-12
Use NSDocument for document management, storyboards for UI, statically typed RPC actions

bjtitus:
There are a few major changes in this Pull Request:

- An NSDocument subclass to better integrate with macOS
- A switch to storyboards from the previous xib based layout
- A switch to NSViewControllers to separate concerns (and because of the storyboard addition)
- Using AutoLayout from the storyboard, which reduces some of the constraint code

Some specifics about this implementation that may be of interest:
- The NSDocument subclass handles RPC communication instead of the EditView. Eventually, I think this should be further expanded as a protocol or something, but I didn't want to bite off too much in one go.
- I added statically typed RPC actions to collect all of the RPC strings that were floating around in EditView. Not sure if that's something that makes sense for the project.


bjtitus:
I signed it!

raphlinus:
Hi bjtitus!

I'm so sorry it's taken me so long to respond to your PR. I do want to move in this direction, specifically to make the Cocoa parts more Mac-native.

Regarding the static typing of RPC actions, I haven't decided exactly where I want to end up with this. I'll accept this part of your PR because making it more static and less "stringly type" (or "jsonly typed") is a good thing, though that can be handy for prototyping. If I were rewriting this, I'd probably organize RPC requests into a protocol (trait in rust) similar to the `Handler` trait of the plugin: https://github.com/google/xi-editor/blob/master/rust/syntect-plugin/src/caching_plugin.rs#L27 . But this is a (minor, imho) style issue. I will be changing the RPC actions somewhat, making them more structured (at the very least making movement / update selection a param rather than multiplying the number of argument-less actions by two). But that's orthogonal to this update.

Headers on new files much conform to the existing pattern (Copyright <year> Google and Apache banner). I don't want to shirk crediting you (or other contributors), though. If you want your name credited in the repo, the proper way to do that is to change the Google copyright to "The Xi Editor Authors" and add an AUTHORS file at the toplevel. I'm happy to do that if you request (I'll probably want to do that at some point).

I can make the changes to the header files if you like. I'm also happy to fix the merge conflicts.

cmyr:
@bjtitus @raphlinus What's the state of this PR? I started playing around yesterday adding bare-bones tab support (without adopting NSDocument) but this seems like a better approach.

If nobody has any outstanding problems I would be happy to clean this up and get it mergeable.

raphlinus:
@cmyr I'm happy for you (or anyone) to clean this up. I must admit I'm still fuzzy on the exact advantages of NSDocument. Also note that Sierra has tabs built-in (I've tested them and they seem to work), see #57 for a bit more discussion.

cmyr:
@raphlinus I'm not totally clear on the NSDocument model either, but after reading some docs it does strongly seem to be the preferred model for document-focused applications. It provides a bunch of things we (probably) aren't interested in, like using CoreData or iCloud as backing stores, but it also gives us things like out-of-the-box support for tabs, opening & creating documents, and integration with the system clipboard. 

I'm going to go ahead and build off of this PR with the goal of moving to the document/storyboard model, and holding off on the RFC changes for the time being. (unless @bjtitus or anyone else has any other thoughts)

cmyr:
I think we can close this?

raphlinus:
Yes, the NSDocument migration subsumed by the commits listed above. There's other stuff in the patch, but that should go forward in other patches.

PR #137, opened 2016-11-21
Support simple text search

modelorganism:
Add support for simple text search.  
Adds a new Find dialog. 
Adds the support for the following standard menu commands
 cmd-e - enter selection as find string
 cmd-f - show find dialog
 cmd-g - find next occurrence
These commands were already on the Xi Edit menu, which is not changed by this PR. I am hooking up the existing menu items. 
This PR introduces a single app-global find string and find dlg. This differs from SublimeText and Atom which have find controls in the editor window. Those editors however can open multiple files in one window.  Also the separate dialog seemed less likely to conflict with any other UI changes.


modelorganism:
With more thought we could probably have a more plausible implementation.

PR #139, opened 2016-12-05
Fix Autoresize/Constraint conflict that breaks scrolling for some Xcode

modelorganism:
Fix a conflict with constraints generated from the EditView’s
auto-resize mask. These extraneous constraints cause Xi edit windows
not to be scrollable or resizable on some versions of Xcode. The effect
of this conflict depends on which of the conflicting constraints are
discarded, which seems to somewhat unpredictable. This change marks the
EditView as translatesAutoresizingMaskIntoConstraints=NO. 
We know that this is a constraint conflict because the xi stderr says so:
```
Unable to simultaneously satisfy constraints:
(
    "<NSAutoresizingMaskLayoutConstraint:0x610000083f20 h=--& v=&-- V:[XiEditor.EditView:0x101d06bc0(270)]>",
    "<NSLayoutConstraint:0x608000082f80 V:[XiEditor.EditView:0x101d06bc0(>=4880.32)]>"
)

```
  and because this one change fixes the conflict message,  and make the xi edit window scrollable.
This is not an unknown issue see for example  [https://www.innoq.com/en/blog/ios-auto-layout-problem/](url)

### DIFF
--- a/XiEditor/AppWindowController.xib
+++ b/XiEditor/AppWindowController.xib
@@ -32,6 +32,9 @@
                             <subviews>
                                 <customView fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="hxn-ZY-xOb" customClass="EditView" customModule="XiEditor" customModuleProvider="target">
                                     <rect key="frame" x="0.0" y="0.0" width="480" height="270"/>
+                                    <userDefinedRuntimeAttributes>
+                                        <userDefinedRuntimeAttribute type="boolean" keyPath="translatesAutoresizingMaskIntoConstraints" value="NO"/>
+                                    </userDefinedRuntimeAttributes>
                                 </customView>
                             </subviews>
                         </clipView>


raphlinus:
This looks like it's probably a good fix, but I'm studying #136 and am likely to merge that (after at least fixing merge conflicts), and I think that reworks these constraints.

Thanks for the patch in any case!

adamhammes:
This bug is preventing me from using xi. What's the status on merging it?

raphlinus:
Sorry, taking longer to figure out #136 than I thought, and in the meantime this looks like a simple fix. Thanks for the contribution!

raphlinus:
I'm very confused still; now I get this warning:

[Layout] Detected missing constraints for <XiEditor.EditView: 0x100f0c2d0>.  It cannot be placed because there are not enough constraints to fully define the size and origin. Add the missing constraints, or set translatesAutoresizingMaskIntoConstraints=YES and constraints will be generated for you. If this view is laid out manually on macOS 10.12 and later, you may choose to not call [super layout] from your override. Set a breakpoint on DETECTED_MISSING_CONSTRAINTS to debug. This error will only be logged once.

I don't understand this stuff, hopefully someone who does can enlighten me. But it's possible that we'll go with #136 and this may go away.

Issue #141, opened 2016-12-08
Split into xi-core and OS X frontend


theduke:
Since there are multiple frontends popping up, and the goal is a independent structure,
I think it would make sense to split into a xi-core repo with the server and a separate xi-cocoa project for the frontend.

Eh2406:
Could there be a list of frontends somewhere?

raphlinus:
I have the repository created, just need to do the split.

theduke:
How's that coming along, @raphlinus ?

While there are relatively few issues at the moment, it's confusing to separate what's related to the core vs the Cocoa frontend.

PR #145, opened 2017-01-05
Fuchsia Boilerplate

jxson:
I wasn't able to verify the build instructions on my Mac today due to rust-lang#24887 but they "should work" on Linux machines.


adrientetar:
So Fuchsia runs flutter apps which are originally meant as iOS/Android portable apps?

Can we put each frontend into its own repo? I'm writing an Android frontend and this puts a lot of code that isn't needed in there (cc #141).

raphlinus:
@adrientetar Interesting that you're writing an Android front-end. And yes, at the moment all UI code for Fuchsia is written in Flutter. We will probably have our front-end running on at least Android, for easier/faster development and debugging.

I'm a bit torn on this (and also #141), as I feel that having the core and the officially developed front-end(s) together in the same repo makes it easier to atomically change the protocol. I also think that the amount of code here will be fairly small, so the burden of downloading additional code won't be too much. But I see a fair amount of support for splitting into smaller repos, so if people feel very strongly about this, we'll create a new fuchsia repo.

jxson:
@adrientetar FWIW I think it would be good to have a nice separation between core and any front-ends. I'll work on separating this PR's code into a separate repo.

raphlinus:
Yes, @jxson and I discussed this, and we'll do the Fuchsia front-end in a separate repo. I'll also work on preparing a patch that splits out most of the binary (other than the stdio mainloop) into a lib, so we can put the Fuchsia/fidl runtime logic into that new repo as well.

jxson:
Okay PR ported to https://fuchsia-review.googlesource.com/#/c/18132/

Thanks for the feedback @adrientetar 💃

szotp:
Hi, found this while searching for flutter and xi. Was this code merged anywhere? https://fuchsia.googlesource.com/xi appears to be empty.

Since flutter somewhat supports desktops, it could be possible to make a clone of VSCode/Atom in it. Hopefully with better startup time and overall performance.

raphlinus:
The good news is that the code has simply moved. It's now at:

https://fuchsia.googlesource.com/topaz/+/master/app/xi/

It works on Fuchsia. I haven't tested it on iOS/Android, so I wouldn't be surprised if it needs some small modifications to get running again there, but it's possible it just works.

Issue #144, opened 2017-01-04
Migrate Swift code to 3.0


raphlinus:
Xcode 8.2 gives a warning that Swift 2.3 is deprecated. Using the conversion tool still leaves 31 errors. Most are "Contextual type 'AnyObject' cannot be used with dictionary literal", which is easily fixable by adding `as AnyObject` after the literal. The NSData for the receive buffer must be replaced with Data.

Part of the reason I'm posting this as an issue is to get feedback on when this should be done. I'm happy chugging along with 2.3 for now, but at some point it needs to change. It also obviously can cause merge conflicts for PR's in flight.

adamhammes:
I think it would be appropriate to bump the version now. There are only 4 open PRs currently, two of which have no code changes.

AquaGeek:
I'd push to do this sooner rather than later. Xcode 8.2 is the [last version that will support Swift 2.3](https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Introduction.html).

rkusa:
The only affected PR seems to be #136, which needs to be touched (copyright header added, merge conflicts) anyway. So I'd agree that now would be a good time.

PR #148, opened 2017-01-08
Put back sometimes missing constraints.

modelorganism:
Sorry to have led you astray. I had thought that your version of Xcode 7.3 was just more lenient. 
Anyway Can you try this. I have put back the 2 missing constraints, but as constraints, not coverting from auto-resize. This allows me to set different priorities so that we can control which constraint is broken when the scroll view has content. 
Unfortunately I am not able to repro what you are seeing in my XCode Version 7.3.1 (7D1014). But it make sense to me that this could work, I think this version of the xib this is more upfront in that auto-resize is *off*, not overridden, and replaced with explicit constraints so xcode has a better chance of understanding  what we are doing, thus it hopefully means more that the xib compliles w/o warning.


raphlinus:
Thanks! This seems to work for me fine in Xcode 8.2. Maybe there are still issues in other versions, but I don't think we'll be supporting older Xcode versions long, as at some point we'll migrate to Swift 3 (see #144).

PR #149, opened 2017-01-11
Travis CI

jdub:
I couldn't find any evidence of previous work or discussions about Travis CI, so thought I'd whip up a quick proof of concept.

My key concern is that I don't know if you're allowed to use Travis CI with Google organisation projects, or whether you have any desire for CI at all.

* It builds on OS X (with Xcode 7.3, for the GUI) and Linux (mainly for turnaround time, as there are usually long queues for OS X builders)
* It builds on Rust stable and nightly, allowing nightly builds to fail
* It doesn't do anything fancy like uploading tagged releases to GitHub; I'll add that (at least for the OS X app) if you think it's worthwhile
* I have left the default email notifications settings; that can always be changed
* There's a minor change to your Xcode build script to work around a Travis bug, but it won't affect non-Travis builds

Getting much fancier than uploading builds of the OS X app to GitHub might prompt some repo splitting, but we don't have to worry about that for now. 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+language: rust
+
+rust:
+  - stable
+  - nightly
+
+os:
+  - linux
+  - osx
+
+matrix:
+  allow_failures:
+    - rust: nightly
+
+cache:
+  directories:
+    - $HOME/.cargo
+
+script:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then xcodebuild; fi
+  - cd rust
+  - export CARGO_TARGET_DIR=/tmp/target
+  - if [[ "$TRAVIS_OS_NAME" != "osx" ]]; then cargo build && cargo build --manifest-path syntect-plugin/Cargo.toml; fi
+  - for MANIFEST in Cargo.toml */Cargo.toml; do cargo test --manifest-path $MANIFEST; done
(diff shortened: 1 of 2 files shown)


jdub:
@googlebot I signed it!

raphlinus:
I'll look at this now. Yes, CI is a good thing, and as far as I know there are no problems with using Travis.

I do plan to split off the Cocoa front-end fairly soon (see #141), not sure how that's going to impact this work. I don't know much about tagged releases.

raphlinus:
Thanks! I've never really done CI outside the Google infrastructure, so happy to have help with it.

jdub:
Ah, sweet. Happy to help out with the Travis configuration when you do the split.

When you want to start making (pre-)releases, you can create an annotated tag and let Travis run the build, upload the release notes (tag annotation), and upload the resulting binary to GitHub. No clicky-clicky work, just tag and push. The same workflow can be applied to crate releases, too.

jdub:
btw, now you need to connect Travis and GitHub to each other, which may require the help of the admin of Google's GitHub organisation.

But give it a try anyway: Go to your Travis CI account page and see if the `google` organisation is listed on the left, and if the `xi-editor` repo is in it. You may need to hit "Sync account" at the top. Hit the switch!

raphlinus:
I hit it! As far as I can tell, everything is working.

jdub:
Looking good. Anything you'd like to change or add? Is there a test framework I can run for the OS X bits, perhaps?

Probably easiest to add binary uploads of the OS X app after the repo split. Then we can add a debug build for testers, too.

PR #151, opened 2017-01-16
Improve mouse cursor switching

manuel-rhdt:
This makes the cursor switch from standard pointer to I-Beam cursor work
regardless of the window size.


manuel-rhdt:
@googlebot I signed it!

raphlinus:
Good catch, thanks!

Issue #96, opened 2016-07-20
selections and color schemes


eyelash:
For the GTK+ front-end I'm developing I would like to invert the text color for the selection to achieve some more contrast but I noticed this would interfere with the "fg" line attribute.
So I propose the following incremental changes:
1. The core assumes text is rendered black on white and sends both a "sel" and a bright "fg" attribute for selections.
2. The "sel" attribute is replaced with a "bg" attribute that also contains a color.
3. The core somehow informs the front-end about a default background and foreground text color (and possibly a cursor color)

After these changes the core would have complete control over the colors and color schemes could be implemented and shared across all front-ends.

raphlinus:
I'm thinking about this, still haven't come a conclusion. There are some problems with background color spans in the Cocoa front-end, which is one reason I haven't done that yet. See [this thread](https://github.com/google/xi-editor/commit/a9dcdbd1846b15efd7355ed707301b064ae81c3b#commitcomment-18367250) for more discussion.

More deeply, who should be the final source of truth for the theme? It's easiest to let the plugin do it (this would be implemented by your (3) above, a command to the front-end to set the theme). But this could lead to flashing if the plugin takes more than a frame to load.

There's also more to the selection appearance than just the background color. If you look at Sublime, the selection has rounded corners, while background colors set by the theme are rectangular. Also, within a selection, spaces are rendered with a tiny dot, and tabs by a horizontal line.

In any case, I'll take PR's that evolve in the right direction. What you propose would be a good start.

sp3d:
I'd like to give my perspective on this, as someone who primarily uses GTK+-based GUI programs and wants to see Xi be usable as a powerful backend for text-editing widgets.

Editors for code obviously have more specialized needs than most of the built into toolkits, but the GtkSourceView library has done a really good job of handling both code-specific needs like paren matching, search result highlighting, code folding, and syntax highlighting while still fitting into the user's system theme, and I think we should take some hints from what it does.

GtkSourceView has its own style scheme and syntax highlighting system (and Xi will presumably have its own), but styles mutate colors and font attributes starting from the initial state established by the GTK+ theme's palette and settings.

@raphlinus, I'm not 100% sure what you mean by "final source of truth for the theme", but in order to follow GTK+ theming effectively, the GTK frontend needs to be able to present a base set of colors which are merged with recommendations from syntax highlighting, paren matching, etc.

[...]

PR #152, opened 2017-01-26
Add a doc for the proposed update protocol

raphlinus:
Relevant to the discussion in #95

### DIFF
--- a/doc/update.md
+++ b/doc/update.md
@@ -1,0 +1,94 @@
+# Xi view update protocol
+
+This document describes a proposal for a new protocol for sending view updates from the core to the front-end.
+
+## Background
+
+Goals: keep everything async. Keep network traffic minimal. Allow front-end to retain as much information as possible (including text if only cursors are updated). Allow front-end to use small amounts of memory even when document is large.
+
+Conceptually, the core maintains a full view of the document, which can be considered an array of lines. Each line consists of the *text* (a string), a set of *cursor* locations, and a structure representing *style* information. Many operations update this view, at which point the core sends a notification to the front-end.
+
+The front-end maintains a *cache* of this view. Some lines will be present, others will be missing. A cache is consistent with the true state when all present lines match.
+
+To optimize communication, the core keeps some state about the client. One bit of this state is the *scroll window;* in general, the core tries to proactively update all lines within this window (plus a certain amount of slop on top and bottom). In addition, the core maintains a set of lines in the client's cache. If a line changes, the update need only be communicated if it is in this set. This set is conservative; if a line is missing in the actual cache held by the front-end (evicted to save memory), no great harm is done updating it.
+
+## Requests from front-end to core
+
+```
+scroll: [number, number]  // first line, last line
+```
+
+```
+request: [number, number]  // first line, last line
+```
+
+## Requests from core to front-end
+
+
+```
+set_style
+  id: number
+  fg_color?: number // 32-bit RGBA value
+  bg_color?: number // 32-bit RGBA value, default 0
+  weight?: number // 100..900, default 400
+  italic?: boolean  // default false
+```
+
+It's not hard to imagine more style properties (typeface, size, OpenType features, etc).
+
+The guarantee on id is that it is not currently in use in any lines in the view. However, in practice, it will probably just count up. It can also be assumed to be small, so using it as an index into a dense array is reasonable.
+
+Discussion question: should the scope of set_style be to a tab, or to the global session?
+
+Style number 0 is reserved for selections. Discussion question: should other styles be reserved, like 1 for find results?
+
+```
+scroll_to: [number, number]  // line, column (in utf-8 code units)
+```
+
+```
+update
+  rev: number
+  ops: Op[]
+
+interface Op {
+  op: "copy" | "skip" | "invalidate" | "update" | "ins"
+  n: number  // number of lines affected
+  lines?: Line[]  // only present when op is "update" or "ins"  
+}
+```
+
+An update request can be seen as a function from the old client cache state to a new one. During evaluation, maintain an index (`old_ix`) into the old `lines` array, initially 0, and a new lines array, initially empty. [Note that this document specifies the semantics. The actual implementation will almost certainly represent at least initial and trailing sequences of invalid lines by their count; and the editing operations may be more efficiently done in-place than by copying from the old state to the new].
+
+The "copy" op appends the `n` lines `[old_ix: old_ix + n]` to the new lines array, and increments `old_ix` by `n`.
+
+The "skip" op increments `old_ix` by `n`.
+
+The "invalidate" op appends n invalid lines to the new lines array.
+
+The "ins" op appends new lines, specified by the "`lines`" parameter, specified in more detail below. For this op, `n` must equal `lines.length` (alternative: make n optional in this case). It does not update `old_ix`.
+
+The "update" op updates the cursor and/or style of n existing lines. As in "ins", n must equal lines.length. It also increments `old_ix` by `n`.
+
+In all cases, n is guaranteed positive and nonzero (as a consequence, any line present in the old state is copied at most once to the new state).
+
+```
+interface Line {
+  text?: string  // present when op is "update"
+  cursor?: number[]  // utf-8 code point offsets, in increasing order
+  styles?: number[]  // length is a multiple of 3, see below
+}
+```
+
+The interpretation of a line is different for "update" or "ins" ops. In an "ins" op, text is always present, and missing cursor or styles properties are interpreted as empty (no cursors on that line, no styles).
+
+In an "update" op, then the text property is absent from the line, and text is copied from the previous state (or left invalid if the previous state is invalid), and the cursor and styles are updated if present. To delete cursors from a line, the core sets the cursor property to the empty list.
+
+The styles property represents style spans, in an efficient encoding. It is conceptually an array of triples (though flattened, so triple at is `styles[i*3]`, `styles[i*3 + 1]`, `styles[i*3 + 2]`). The first element of the triple is the start index (in utf-8 code units), but encoded as a delta relative to the *end* of the last span (or relative to 0 for the first triple). It may be negative, if spans overlap. The second element is the length (in utf-8 code units). It is guaranteed nonzero and positive. The third element is a style id. The core guarantees that any style id sent in a styles property will have previously been set in a set_style request.
+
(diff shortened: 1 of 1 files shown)


raphlinus:
[Rendered](https://github.com/google/xi-editor/blob/protocol_doc/doc/update.md)

eyelash:
Are you planning to add more requests to this later on for things like default background color, cursor color, selection color, active line (line with a cursor) background color? These will definitely be needed for Xi to look great and I think it would be good to add them to this specification.
I just skipped through the document because I don't have much time but I will take a closer look at the protocol tomorrow or the day after tomorrow.

rkusa:
Thanks for writing this down @raphlinus! Just for clarification, one thing that we discussed an reddit some time ago is how to remove styles. This is something you wrote

> The set_line_*_spans method replaces all the spans on the line. More precisely, it deletes all spans that are fully enclosed in the region being replaced (ie the line). Thinking about it in response to your post, this is a problem for spans that cross multiple lines - and that will happen if the user inserts a newline in the middle of a span. The evolution of the protocol will allow for replacing a multiline region, but it feels ugly that the plugin has to keep track of that. I'll see if I can come up with a better approach.

If I understand this writeup correct, there is no multiline span possible, is that correct? If this is true, then "it deletes all spans that are fully enclosed in the region being replaced" is probably enough to remote styles.

doc/update.md
```diff
@@ -0,0 +1,94 @@
...
+interface Line {
+  text?: string  // present when op is "update"
+  cursor?: number[]  // utf-8 code point offsets, in increasing order
+  styles?: number[]  // length is a multiple of 3, see below
+}
+```
+
+The interpretation of a line is different for "update" or "ins" ops. In an "ins" op, text is always present, and missing cursor or styles properties are interpreted as empty (no cursors on that line, no styles).
+
+In an "update" op, then the text property is absent from the line, and text is copied from the previous state (or left invalid if the previous state is invalid), and the cursor and styles are updated if present. To delete cursors from a line, the core sets the cursor property to the empty list.
```
rkusa: I am not sure about the meaning of the first sentence. Is it maybe supposed to be:

> In an "update" op, **when** the text property is absent from the line, **then** text is copied from the previous state (or left invalid if the previous state is invalid), and the cursor and styles are updated if present.

doc/update.md
```diff
@@ -0,0 +1,94 @@
...
+  id: number
+  fg_color?: number // 32-bit RGBA value
+  bg_color?: number // 32-bit RGBA value, default 0
+  weight?: number // 100..900, default 400
+  italic?: boolean  // default false
+```
+
+It's not hard to imagine more style properties (typeface, size, OpenType features, etc).
+
+The guarantee on id is that it is not currently in use in any lines in the view. However, in practice, it will probably just count up. It can also be assumed to be small, so using it as an index into a dense array is reasonable.
```
adrientetar: What is the purpose of id here?

doc/update.md
```diff
@@ -0,0 +1,94 @@
...
+
+```
+request: [number, number]  // first line, last line
+```
+
+## Requests from core to front-end
+
+
+```
+set_style
```
adrientetar: nit: this could be named `def_style` instead of `set_style`

doc/update.md
adrientetar: nit: this section could say that id will be used to reference a style when updates are sent, right now it's only explicitly stated at the very bottom of the document.

raphlinus:
@eyelash All those things sound good. I'm still struggling with the question of whether the core gets full control over all theming, or whether the front-end shares some of it (as it does now for selection color). Thinking about it, the former is ultimately going to be simpler and more consistent. Also, not all of these are purely colors, in some cases you might also want to draw an outline. Finally, we also need styling for bracket matching (actually something I hope to implement soon).

@rkusa The view update is orthogonal to whether there are multiline spans. The latter is basically the domain of the plugin protocol. I would definitely like to support them, but as that thread indicated there are technical challenges, and they're not needed for syntax highlighting. But out of scope for this proposal.

@adrientetar The reason I chose "set" rather than "def" is that I wanted to make it clear that an id might be reused. But "def" is appealing, as it emphasizes the fact that it must be defined before use.

Thanks all for the editorial comments, I'll clean up the language a bit. Also, I left out underlines.

eyelash:
> The styles property represents style spans, in an efficient encoding. It is conceptually an array of triples (though flattened, so triple at is `styles[i*3]`, `styles[i*3 + 1]`, `styles[i*3 + 2]`). The first element of the triple is the start index (in utf-8 code units), but encoded as a delta relative to the *end* of the last span (or relative to 0 for the first triple). It may be negative, if spans overlap. The second element is the length (in utf-8 code units). It is guaranteed nonzero and positive. The third element is a style id. The core guarantees that any style id sent in a styles property will have previously been set in a set_style request.

Why is the first element relative and not absolute (just curious)?

I'm not sure I like this whole concept of the style ids. It feels like trading simplicity for a possible speed-up. Do we have any benchmarks to show that it actually makes a noticeable difference?

> Discussion question: should the scope of set_style be to a tab, or to the global session?

If the tabs all use the same theme I'd expect styles to be largely the same between tabs, so I vote for global scope.

> Should offsets be utf-8 or utf-16? The majority of front-ends use utf-16. It might be more efficient for xi to do the conversions (in Rust) than the front-end.

GTK/Pango uses UTF-8 offsets AFAIK (the documentation doesn't actually say so but they seemed to work so far).

Generally, I like the design of the protocol :+1:

cmyr:
I'm out of my depth here, but I've been following along for a little while. Anyway: there seems to be a real move going forward to treat UTF-8 as the de facto standard encoding in most use cases? E.G. Swift made a big deal out of having native UTF-8 strings, where Objective-C had used UTF-16. Similarly the newer APIs provided in Unicode's icu4c (such as the `UText` type) seem to be moving towards UTF-8 as well.

Anyway, enjoying your work. cheers! 👍

raphlinus:
@eyelash It's relative basically to keep the byte count smaller. In the common case where styles don't overlap, the triples are basically a run length encoding - number of bytes with no style, number of bytes of given style, style id.

There are a few arguments for style id's (though it's good to be skeptical). One is performance. I just checked the RPC for fetching 100 lines, and it's about 500µs for plain text, 3ms for syntax-highlighted Rust. I don't have numbers (yet, probably will soon) on the array encoding, but gut feeling I expect it to be about 2x faster.

The other major background is that I plan to take this "id" concept all the way through from plugins to the front-end (see #96, though it doesn't contain my recent thinking). More specifically, I plan to have plugins return a list of scopes, just like Atom, Sublime, and most editors (e.g. `["source.rust", "meta.function.rust", "entity.name.function.rust"]`) and have theme resolution done in the core. The theme resolution logic only needs to be invoked on a new unique list of scopes, of which there will probably only be a few dozen in a typical source file. Other than that, the id's can be passed through with minimal or no processing (likely a simple int->int map, which costs almost nothing). To me, this strikes a good balance between keeping the theme processing at a high level and maximal performance.

@cmyr Glad to have you following along, and please feel free to ask questions etc. One of the (many) goals of this project is teaching and learning. I agree it's logically simpler to keep everything utf-8 (and plan to do so), but at the same time it's tempting to do utf-8/utf-16 conversions in Rust where they're likely to be so much faster than in the language of the front-end :)

cmyr:
Okay, I'll bite: what are the target frontends you're thinking of, which use utf-16? Cocoa seems to. What else? I can definitely see the advantage of doing this in Rust vs. doing this in say electron or something, but it feels sort of blunt to just try and pick the encoding that is best for the majority of frontends, and the rest have degraded performance. Might it make sense to keep it in utf-8 and have `utf16-offsets` be a Cargo feature, which performs the conversion? (or the other way around). I'm not sure what the implementation will look like, mind, so the added complexity could be a good argument against.

adrientetar:
Android/Java does. Generally operating systems use UTF-16 internally.

raphlinus:
And Windows. And Flutter/Dart. So a huge majority of the front-ends will be UTF-16. That said (at least for now), they're all already getting their _strings_ in UTF-8 and having to do the conversion if necessary. Also... Swift has native UTF-8 strings? Last I checked, they were all UTF-16 as well, with a fast-path for ASCII.

cmyr:
I know that NSString was UTF-16, but I remember reading that Swift's native `String` was going to be encoded as utf-8[1]. This is from Swift 1.0. There are some changes planned for Swift 4.0, outlined [here](https://github.com/apple/swift/blob/master/docs/StringManifesto.md#the-unicode-protocol); that document teases talking about working with Cocoa NSString ranges, but doesn't seem to actually talk about it.


[1] https://github.com/apple/swift/blob/master/docs/StringDesign.rst#strings-are-encoded-as-utf-8

josephg:
Everyone uses UTF-16 because when they were written nobody thought we'd need >65536 unicode values, so most systems (java, obj-c, c#, win32, javascript, python2) encoded strings using UCS2. And UTF-16 was much easier to migrate to from UCS2 because its another 2 byte encoding.

... Which is to say, systems use UTF-16 for terrible legacy reasons. And UTF-16 has the property that *almost* all codepoints have a string length of 1, but a few characters have a string length of 2. And you won't run into those characters if you don't test for them explicitly.

Dart, Ruby, Python 3, Rust and Go all use UTF-8 and by default count string length correctly in unicode codepoints. (I'm not sure about swift). And everything accepts utf-8 on the boundary (its usually a more compact form, even encoding non-english languages).

The question is really about whether you count character positions in codepoints (which is the obvious answer) or count character positions in UTF-16 2-byte position offsets, which is a much more complicated system and it causes the reverse problem in forward looking languages of having to convert.

Its pragmatic to support UTF-16 2-byte position offsets for legacy languages / systems, but UTF-8 is the future and personally I think Xi should embrace that. You can always write efficient rope libraries which can convert efficiently - [I wrote one](https://github.com/josephg/librope#wide-character-string-compatibility) when dealing with this problem for collaborative text editing between C and JS. (My rope library can convert positions in <1μs for any string smaller than 1M characters, and it gets slower in logarithmic time. It requires memory linear with the string size though.)

rkusa:
> Style number 0 is reserved for selections. Discussion question: should other styles be reserved, like 1 for find results?

For soft and weak spans we may need two more reservations.

cmyr:
I was thinking yesterday that we might also want a reservation for find results?

raphlinus:
See note on #43 for why I'm merging this now. There's more work needed, and I captured that in #231.

PR #154, opened 2017-02-02
Add an example of a combine number parser

Marwes:
I cleaned up my example from https://www.reddit.com/r/rust/comments/5r6tib/parser_combinators_without_macros_my_take/ the time for combine is not quite as fast as nom or your peg parsers if one writes it in the normal way (first commit). I believe this is partly because of the drop glue from error types (no errors are allocated, but the drop glue still needs to run and check that `Vec::is_empty()` for the errors vector) and partly that LLVM is not quite able to optimize the `or` parsers as well as one might hope as modifying.

First commit
```
test combine_benches::bench_combine ... bench:          48 ns/iter (+/- 2)
test nom_benches::bench_nom         ... bench:          24 ns/iter (+/- 3)
test pom_benches::bench_pom         ... bench:       3,425 ns/iter (+/- 164)
test tests::bench_my_peg            ... bench:          20 ns/iter (+/- 2)
```

However, taking into account that the input string did not have a sign for the exponent I modified it slightly which seems to let LLVM optimize the parser to be more or less equal to nom.

```
test combine_benches::bench_combine ... bench:          25 ns/iter (+/- 4)
test nom_benches::bench_nom         ... bench:          24 ns/iter (+/- 1)
test pom_benches::bench_pom         ... bench:       3,240 ns/iter (+/- 660)
test tests::bench_my_peg            ... bench:          20 ns/iter (+/- 1)
```

### DIFF
--- a/rust/lang/Cargo.toml
+++ b/rust/lang/Cargo.toml
@@ -10,6 +10,7 @@ description = "A language syntax coloring and indentation plugin for xi-editor."
 pom = { version = "0.8", optional = true }
 regex = { version = "0.2", optional = true }
 nom = { version = "2.1", optional = true }
+combine = { version = "2.2.2", optional = true }
 
 [dependencies.xi-plugin-lib]
 path = "../plugin-lib"
--- a/rust/lang/benches/peg_tests.rs
+++ b/rust/lang/benches/peg_tests.rs
@@ -38,6 +38,9 @@ extern crate regex;
 #[macro_use]
 extern crate nom;
 
+#[cfg(feature = "combine")]
+extern crate combine;
+
 const TEST_STR: &'static str = "1.2345e56";
 
 #[cfg(all(test, feature = "pom"))]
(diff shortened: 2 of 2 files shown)


Marwes:
I signed it!

Marwes:
So this was pretty interesting to do since I actually discovered a performance problem with the optional parser which caused my first attempt to clock in at around 200ns since an error would be created. I fixed that in https://github.com/Marwes/combine/commit/a4bf28d2e490405ba362a3324a0b7115972d4044.

The second commit makes the parser a bit uglier but it generates straighter code which makes LLVM happy. May make some further tweaks or additions to combine to make this kind of code easier to write.

raphlinus:
Excellent, thanks! For the record, I get 19ns on my machine (compared with 11ns for my code).

Marwes:
Seems I accidentally rebased everything to a single commit but the second commit I mentioned was transforming `optional(token(b'+').or(token(b'-'))` into `token(b'-').map(Some).or(token(b'+').map(Some)).or(value(None))`.

What this change does is make the parser right associative which lets us avoid running a tiny bit of code after each failed parse.

```

match parse1() {
	Ok(x) = > Ok(x),
		Err(err) = > {
		match parse2() {
			Ok(x) = > Ok(x),
			Err(err2) => {
				match parse3() {
					Ok(x) = > Ok(x),
						Err(err3) = > merge(err, err2, err3),
				}
			}
		}
	}
}
```
versus
```
let result = match parse1() {
	Ok(x) = > Ok(x),
		Err(err) = > {
		match parse2() {
			Ok(x) = > Ok(x),
                        // We avoid this call to merge which is effectively a no-op in this case but unfortunately
                        // LLVM does not realise this
			Err(err2) => merge(err, err2),
		}
	}
};
match result {
	Ok(x) = > Ok(x),
	Err(err) = > {
		match parse3() {
			Ok(x) = Ok(x),
			Err(err3) => merge(err, err3)
		}
	}
}

```

PR #156, opened 2017-02-11
Incremental update protocol

raphlinus:
Starting to implement the new protocol described in #152.

As of this commit, the protocol should be in place, and functionality
should be basically unchanged, but the implementation is not very
optimized. In particular, updates just send the scroll area (with
slop), invalidating everything else, rather than actually sending a
small delta.

In addition, there are a whole bunch of optimizations not done yet,
for example style spans are stored as (old-style) view::Style objects,
and looked up in the style map one-at-a-time while rendering. A far
more efficient approach would be to use a similar style id system in
the plug-in protocol, and then basically plumb style id's through.

That said, all these optimizations should be possible without change
to the update protocol, and I'm reasonably satisfied that will be
stable now.

### DIFF
--- a/XiEditor.xcodeproj/project.pbxproj
+++ b/XiEditor.xcodeproj/project.pbxproj
@@ -8,7 +8,9 @@
 
 /* Begin PBXBuildFile section */
 		50370A2B1D4209F300EA1B18 /* Dispatcher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 50370A2A1D4209F300EA1B18 /* Dispatcher.swift */; };
+		AE0243F61E4BDF8A00641BDA /* StyleMap.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE0243F51E4BDF8A00641BDA /* StyleMap.swift */; };
 		AE041BA61D4327EB0069AB8B /* xi-syntect-plugin in Resources */ = {isa = PBXBuildFile; fileRef = AE041BA51D4327EB0069AB8B /* xi-syntect-plugin */; };
+		AE168C5C1E4AD87B008249AE /* LineCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE168C5B1E4AD87B008249AE /* LineCache.swift */; };
 		AE228F5D1C897CE7000E9B0F /* xi-core in Resources */ = {isa = PBXBuildFile; fileRef = AE228F5C1C897CE7000E9B0F /* xi-core */; };
 		AE499DD01C82BB2B002D68AF /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = AE499DCF1C82BB2B002D68AF /* AppDelegate.swift */; };
 		AE499DD21C82BB2B002D68AF /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = AE499DD11C82BB2B002D68AF /* Assets.xcassets */; };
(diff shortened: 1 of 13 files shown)


raphlinus:
@kulakowski is probably implemented in the Rust changes.

There are still some loose ends (including deleting a bunch of the old code), but I think it's in reasonably good shape. I think people working on front ends can implement this as well, without too much worry of things changing again.

eyelash:
Right now the LineCache creates a new array for each update. Are you planning to optimize this or is this the preferred approach for front-ends?

Also is there a reason the core does not automatically make sure the scroll area is fully present in the cache? (The front-end still needs to send `request_line`s when scrolling.)

raphlinus:
@eyelash I plan to do a lot more in the way of optimization. The goal is to get ordinary typing (including with syntax highlighting enabled) to actually invalidate a thin band on the screen. There are a lot of pieces to that. As well, I plan to do as you suggest and eagerly push the view area plus some slop on scroll.

Regarding the creation of the new array, that's an interesting question. I certainly thought about an in-place update, but in the general case it can be tricky. My guess is that the array copying accounts for only microseconds of time, which would make it not worth optimizing further. Of course, one piece of that is to trim the size of the array (I'm thinking a few hundred to a thousand lines beyond the scroll area). One possible compromise is to do the update in place only if it doesn't involve insertions or deletions.

Thanks for the comments!

eyelash:
While testing the new update protocol with xi-gtk I noticed a few things:

- When opening a large document and then slowly scrolling to the bottom the cache will finally contain all the lines of the document. I don't have access to a mac right now so I can't verify if it's a bug in xi-gtk or in xi-core.
- If the front-end does not respect the `scroll_to` events from the core (and displays an area that does not contain the cursor), the core will send `update` events that invalidate visible lines when inserting some text.

As I said before it would be nice if the core could send `update` events automatically when the front-end sends `scroll` events so the front-end does not need to send additional `request_lines` events when scrolling.

eyelash:
@raphlinus I still have a few questions:
### How is scrolling supposed to work?
The front-end sends `scroll` events to the core but right now the core does not automatically send any `update` events so the front-end has to send `request_lines` events as well when scrolling. Wouldn't it be easier if the core just made sure that the current scroll window is up to date?
### How is cache invalidation supposed to work?
Is the front-end supposed to just blindly follow the `update` requests from the core (so effectively the core decides and knows what the front-end cache contains) or is the front-end supposed to decide whether or not to keep things in its cache? Because right now if a front-end just follows the `update` requests from the core and you slowly scroll through a large document the front-end will send `request_lines` for each new line and the `update`s from the core will copy the lines that are currently in the cache and add the newly requested line so in the end the front-end cache contains the whole document. I don't think this is how it is supposed to work.

raphlinus:
@eyelash these are good questions. Again I want to emphasize that the current state is that I think the _protocol_ is ok, but what the core actually _sends_ is far from ideal.

Yes, the core should send updates back in response to scroll notifications. In fact, I think the core should predict where the scrolling is going to be, with the goal that the update in response to a scroll notification covers the range of the next one. I was thinking simple heuristics, but am now leaning towards something like a Kalman filter - you want to send a minimal update when not scrolling or when scrolling is slow, but it's fine to compute and send more data to avoid flickering when scrolling quickly. I think the goal should be no flickering except when scrolling frenetically. The other thing about doing a Kalman filter in the core is that it's amortized over all the front-ends.

And yes, the front-end should trim its cache. That has always been part of the design, but I didn't get around to implementing it yet. The tradeoffs are a little different here, as there is little computational cost to retaining more cache, and obvious benefit to increasing hit rates to retain more, but it shouldn't be unbounded. Incidentally, any change to the view (including just moving the cursor) will _currently_ blow away such a cache, but that will also change.

I definitely want to work on optimizing this stuff, but it's hard to prioritize because what's there now works, and there are plenty of other things that feel more urgent. Hopefully, when I do, everything will get smoother and faster without front-ends needing to do anything extra.

cmyr:
@raphlinus glad to see you're thinking about this. On macOS I've been looking into a bunch of scrollview performance stuff; there's a while API around 'overdraw', drawing beyond the view's current bounds to enable smooth scrolling. I think this (along with a few other tweaks such as using `needsDisplay(inRect:)`) and maybe moving to CALayer-backed drawing, should help us get much better scrolling performance.

raphlinus:
@cmyr Yes, I think the protocol can support extremely good scrolling performance, in a variety of different ways. From my understanding, invalidation rectangles (`needsDisplay(inRect:)`) made sense for software rendering. These days, hardware rendering tends to focus on retaining display lists, and having the GPU draw the scenes from those.

Either way, the client should be able to make use of it. In the former case, your invalidation rectangle covers lines that have been replaced, and if there's insertion or deletion, then you invalidate the half-plane below the edit.

In the latter case, caching of lines can go as far down the pipeline as you like. You could store attributed strings with layout, you could store a display list representation, you could even render the line into a texture and store that. One of the advantages of this approach is that inserting and deleting don't require a lot of re-rendering, you can just change the translation offset for the display list subtree, or where you draw the rendered line.

Drawing the cursors should be on a separate layer; in the invalidation case, the invalidation rectangle should just cover the old and new cursor(s).

I haven't looked into CALayer in great detail. My first experiments with it were marred by the fact that subpixel RGB rendering of text requires an opaque background. I do think it would work for this, though. In particular, it would be nice to blink the cursor by animating the opacity for the cursors layer, without the app having to actually redraw anything.

All that said, I'm not convinced how important these optimizations are on modern desktops. I am very interested in getting top performance on Flutter, though.

cmyr:
@raphlinus okay, interesting. I hadn't really thought very much about the historical situation that brought us a lot of those APIs.

I agree with cursors being layers, this was something I had mentally put down as an eventual to-do.

I'm back playing around here quite a bit again (I was down this week with the flu) and I'm developing a better picture of a front-end architecture that I think adequately respects & abstracts the relationship with the backend, but I'll open a PR explaining my thinking there before I get too carried away. 

In the meantime, I have two quick questions: 

1. _on "invalid lines" / invalidation_: . Does 'invalid' refer to content staleness or style staleness or something else? It seems like we often invalidate large swathes of lines during things like cursor movement. Also I don't understand the phrase "_invalidate the half-plane below the edit_", above. If this means invalidating all lines below that points, then I don't understand why this would always be necessary.

2. _clarification around who is responsible for initiating an update, and when_: It makes sense that after sending an edit or performing a selection, the backend automatically computes ands sends the frontend an update. However, the frontend is clearly responsible for initiating updates in other situations, such as on load, or when scrolling. Have you talked about this relationship in more detail anywhere?

raphlinus:
I'm also getting over a cold, sorry for taking a while to respond.

1. Yeah, this word "invalid" is overloaded. In the line cache, it means that the contents of the line are not known. In a UI drawing context, it means that the region needs redrawing. What happens for something like cursor update will depend on whether cursors have their own layer - if they do, then hopefully an "update" will cause redrawing of the cursor layer and leave the text layer untouched. If they don't have their own layer, then "update" of cursors will cause UI invalidation of the lines that had their cursors change (in the simple case of 1 cursor, it is the line that had the cursor and the line that has the cursor after the update).

2. I could write about it, but I don't think it's all that deep. An update happens when it's needed, hopefully no sooner and no later. What might be interesting is to describe heuristics by which the core tries to proactively keep the front-ends line cache fresh. The front-end having to explicitly request an update should be seen as a failure of the heuristic in normal operation.

Hope that clears things up a bit, lemme know if you have more questions.

I'll also try to get some roadmap written, but no promises, between getting over my cold and all the other things needing doing.

cmyr:
Great yea that's helpful. Hope you feel better soon! Don't worry too much about having an extensive roadmap, but some very loose thoughts on next steps could be helpful.

Issue #161, opened 2017-02-27
no cursor when creating new window


cmyr:
The code that draws the cursor doesn't execute if there are no lines in the in the LineCache, which there aren't until any text has been entered into the window. 

A possible fix would be to initialize with a single empty line?

raphlinus:
Yeah, this happened with the new update protocol.

There are (at least) three ways to fix it, I'm not sure which is better.

Choice 1: initial state is zero-length line list, no cursor is drawn, but core sends the diff to be a length 1 list with just a cursor immediately after creating a new tab (that's the new tab behavior; on loading a file, it might send the diff to show the file). If the core is very slow, there might be a gap where no cursor is shown.

Choice 2: if the line cache is empty, the front-end draws the cursor anyway.

Choice 3: the initial state of a line cache is a single empty line with a cursor. The first delta sent by the core will be against that rather than against the empty list.

I'm leaning toward 1 here because it's the least work for front-ends, they don't have to do anything special.

cmyr:
I think option 1 is also clear conceptually, since the cursor _is_ present even if the buffer is empty.

PR #162, opened 2017-02-27
NSDocument changes, take #2

cmyr:
This PR is a continuation of #136. It does the just about the _bare minimum_ required to migrate to the NSDocument model, and rolls back some of the broader changes in that PR. (I do, for the record, think those changes were a net positive, but I want to make this PR as single-purpose as possible.)

Some notes:

The more I read docs, the more I suspect that the NSDocument model is the correct model for this application; however there's an argument for making this change more incrementally and more thoughtfully. If this PR looks too messy (and it looks pretty messy to me) I'm happy to start again, and work more incrementally.

In any case, the codebase after this PR is asking for a refactor, and some clarity around separation of responsibilities: There is no clear separation of responsibilities between `Document`, `EditView`, and `EditViewController`. We also no longer have an `AppWindowController` class, but I suspect we will want one again at some point in the future. Anyway: I would like to have a clearer conceptual model of what code should go where. I think that the `EditView` should only be responsible for code related to drawing and rendering the view. Any non-drawing code (responder chain/event handling, communicating with the back end) should be handled by the `EditViewController`. I'm not totally sure right now where the `Document` fits in, but I haven't read too too deeply into the NSDocument model yet. 

I also have some ideas for clearing up all the `[String: AnyObject]`'s, that's halfway between going fully-static and staying stringy. But that will be a separate PR.

Some issues:
- saving / opening needs more work. Save, in particular, is currently always bringing up the 'save as' dialog box.

- I haven't implemented the New Tab menu item & associated hotkey, because my first look at this made it seem like I'll need to add back the NSWindowController, and if I do that I would like it to be a separate PR.

- the cursor does not blink when a new document opens. (https://github.com/google/xi-editor/issues/161)

- I couldn't for the life of me figure out where the initial window size used to be set, and all windows were being created with the minimum size; I added some placeholder code to `Document` that sets sane (to me) defaults for a new file, although this is pretty sloppy.

tl; dr: 

If you're okay with this PR, then I promise to do some cleanup shortly. If you aren't okay with this PR, I still think moving in this direction is probably the right decision (and now is the right time) but I'm happy to do that work more incrementally.


XiEditor/AppDelegate.swift
```diff
@@ -72,17 +51,28 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             if let obj = params as? [String : AnyObject], let update = obj["update"] as? [String : AnyObject] {
                 guard let tab = obj["tab"] as? String
                     else { print("tab missing from update event"); return }
-                guard let appWindowController = appWindowControllers[tab]
-                    else { print("tab " + tab + " not registered"); return }
-                appWindowController.editView.updateSafe(update: update)
+                
+                //FIXME: a better way to get a document
```
raphlinus: This is ok, but at least factor it out into a separate method to avoid duplication.

raphlinus:
Overall this looks like a good direction. Thanks for making it more focused, and also offering to break it up into even smaller chunks - that was one of the reasons why it took me so long to get to the previous one. But this one doesn't look _too_ overly scary.

Having a separate EditViewController to deal with responder events does seem like it fits into the Cocoa approach.

Let's chat a bit before making too much change on the `[String: AnyObject]` stuff. I'm certainly happy to hear your ideas on it.

Would you like to merge this PR as-is, and do further cleanup in followups, or refine before merging? As you point out, some functionality is regressing, but it doesn't seem too bad as a temporary situation.

XiEditor/AppDelegate.swift
cmyr: Okay, made the change. I think during cleanup I'd like to use some sort of dispatch table mapping "tab names" to documents.

I'm also wondering if "tab name" is the clearest way of describing what a tab name is doing? My understanding is that it's essentially a buffer identifier?

cmyr:
RE: cleanup: my thought was that if this seemed like a direction you were interested in, we could merge this as-is, so that subsequent changes can be in smaller & easier to read PRs, and to avoid big feature branches.

Re: `[String: AnyObject]`: I appreciate your interest in keeping the flexibility of stringly typing, and I don't have a totally coherent proposal right now, but I'm playing around with some ways to reduce & concentrate the boilerplate. Nothing drastic.


As it sounds like you're approximately okay with this direction, I'll spend a few hours this morning doing some clean-up, with the expectation that that will be a separate PR. Some goals for that:

1. clear up the relationship between the `EditView`, the `EditViewController`, and the `Document`.
2. add a menu item for new tab, and figure out a way to intelligently incorporate that into our architecture.
3. Ensure that whatever choices I make will make it fairly trivial to implement split views, feel like a logical extension of tabs, and something that might be a logical next step for me.
4. Fix saving. I assume we would like to opt-out of all the built-in autosave etc behaviour?

raphlinus:
Sounds great, thanks! Yes, I'd like to opt out of autosave for now, but I'm not opposed to the idea of keeping the state somewhere so that exiting without saving might give the option to restore. However, I feel the need to be extra careful about where potentially sensitive information should be stored. Future work in any case.

Issue #163, opened 2017-02-27
selection can clip neighbouring lines [macOS]


cmyr:
I saw this mentioned in #114 but didn't see an issue for it:

In some circumstances the background color of a selection can clip another line. As far as I can tell, I've only seen this when the text in the selected line is bold? 

![screen shot 2017-02-27 at 2 24 55 pm](https://cloud.githubusercontent.com/assets/3330916/23376038/a77bc688-fcf8-11e6-9402-ba34e2e44b66.png)

raphlinus:
The solution to this is to paint all the background color spans first, then paint the text over it. Probably the reason why bold is affected more is that metrics in the font indicate a large ascent, which would make the background rectangle taller.

PR #164, opened 2017-02-28
Fix selection clipping

cmyr:
This is a quick shot at #163.

Naive solution: when updating the view we now do a first pass through the lines and draw all of the selection backgrounds. 

To do this I needed more concrete access to the style information, so I introduced a `StyleSpan` struct that is just an NSRange and an identifier, and when we get a style array I convert it into an array of these.

I haven't profiled this, but it at least _looks_ fine.

Let me know if you have any thoughts!

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -202,8 +202,8 @@ class EditView: NSView, NSTextInputClient {
             let y = linespace * CGFloat(lineIx + 1)
             context.setFillColor(selcolor().cgColor)
             for selection in selections {
-                let selStart = CTLineGetOffsetForStringIndex(ctline, utf8_offset_to_utf16(line.text, selection.range.location), nil)
-                let selEnd = CTLineGetOffsetForStringIndex(ctline, utf8_offset_to_utf16(line.text, selection.range.location + selection.range.length), nil)
+                let selStart = CTLineGetOffsetForStringIndex(ctline, selection.range.location, nil)
+                let selEnd = CTLineGetOffsetForStringIndex(ctline, selection.range.location + selection.range.length, nil)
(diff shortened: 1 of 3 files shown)


XiEditor/EditView.swift
```diff
@@ -193,6 +193,22 @@ class EditView: NSView, NSTextInputClient {
...
+        // first pass, for drawing background selections
+        for lineIx in first..<last {
+            guard let line = getLine(lineIx), line.containsSelection == true else { continue }
+            let selections = line.styles.filter { $0.style == 0 }
+            let attrString = NSMutableAttributedString(string: line.text, attributes: self.attributes)
+            let ctline = CTLineCreateWithAttributedString(attrString)
+            let y = linespace * CGFloat(lineIx + 1)
+            context.setFillColor(selcolor().cgColor)
+            for selection in selections {
+                let selStart = CTLineGetOffsetForStringIndex(ctline, utf8_offset_to_utf16(line.text, selection.range.location), nil)
```
raphlinus: I'm pretty sure you don't want the utf8_offset_to_utf16 call here (or on the following line), as the selection will already be in utf-16 units. I get a crash when doing selections over non-ascii text.

cmyr:
@raphlinus pushed a commit to remove the range conversion. Also cleaned up the StyleMap a bit, while I was looking at how I would implement the drawing of background colors.

Happy to take a look at this at some point. I'm not very familiar with the low-level Core Text primitives but I'm sure I can figure it out. Not sure if it's a priority or not?

raphlinus:
Thanks, works nicely now!

None of the other work is a priority. Much of it is performance-polishing that likely wouldn't even be noticeable under normal conditions. Maybe we want to throw it into an issue?

porglezomp:
Yay! ✨🎆✨👍✨

PR #171, opened 2017-03-01
Reuse empty views when opening a new document

cmyr:
This fixes another small regression from the document model changes.

I'm not totally in love with this fix; it involves stashing state in the app delegate at one point, which feels kind of gross. If someone has a cleaner approach, I'd be happy to try it.

**N.B.** This doesn't work with files opened from the Open Recent > menu, which uses different API, and would be hackier. We can revisit at some point.

I'm starting to develop a better understanding of the NSDocument architecture, which is giving me a stronger sense of its limitations, and of the assumptions it makes which don't really hold true for us. This becomes clearest when I think about split windows.

### DIFF
--- a/XiEditor/LineCache.swift
+++ b/XiEditor/LineCache.swift
@@ -81,7 +81,7 @@ class LineCache {
     /// - Note: An empty line cache will still contain a single empty line, this
     /// is sent as an update from the core after a new document is created.
     var isEmpty: Bool {
-        return lines.count == 1 && lines[0]?.text == ""
+        return lines.count == 0 || (lines.count == 1 && lines[0]?.text  == "")
     }
 
     func get(_ ix: Int) -> Line? {
--- a/XiEditor/LineCache.swift
+++ b/XiEditor/LineCache.swift
@@ -76,6 +76,13 @@ class LineCache {
             return nInvalidBefore + lines.count + nInvalidAfter
         }
     }
+    
+    /// A boolean value indicating whether or not the linecache contains any text.
+    /// - Note: An empty line cache will still contain a single empty line, this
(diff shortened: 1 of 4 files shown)


XiEditor/LineCache.swift
```diff
@@ -76,6 +76,13 @@ class LineCache {
             return nInvalidBefore + lines.count + nInvalidAfter
         }
     }
+    
+    /// A boolean value indicating whether or not the linecache contains any text.
+    /// - Note: An empty line cache will still contain a single empty line, this
+    /// is sent as an update from the core after a new document is created.
+    var isEmpty: Bool {
+        return lines.count == 1 && lines[0]?.text == ""
```
raphlinus: A case can be made this should return true if lines.count == 0 as well. Unlikely to happen much in practice, but maybe could happen when startup / file opening is scripted.

raphlinus:
Thanks for fixing it. I agree it's not pretty, but am inclined to merge it sooner and do cleanup later.

Regarding the global questions around NSDocument: to me, whether we use NSDocument or not is still an open question. I'm not clear on exactly how much benefit we get from it that we can't get reasonably straightforward by calling the lower-level Cocoa methods directly, and it does seem we have to work around its inbuilt behavior. I'm open to both choices, just wish I had more concrete evidence either way.

Happy to merge this, but will give you a chance to change the `lines == 0` behavior. Thanks!

cmyr:
Pushed a little fix.

Yea, I'm coming around a little bit on the NSDocument stuff. For the time being I think it helps make a few things simpler, but I strongly suspect we would want to move away from it at some point. I'm happy to at least understand _why_ this is the case now, though.

PR #172, opened 2017-03-01
'New Tab' menu item

### LINKED ISSUE #57: Implement tabs in the UI

### LINKED ISSUE #168: Cmd-T should open a new tab
How to go about programatically adding a new tab through the NSDocument model isn't totally clear to me, but there are some clues in the [AppKit 10.12 release notes](https://developer.apple.com/library/content/releasenotes/AppKit/RN-AppKit/#10_12Window%20Tabbing); if that doesn't make this clear then there was a talk at WWDC this year that was probably relevant: [Crafting Modern Cocoa Apps](https://developer.apple.com/videos/play/wwdc2016/239/).

cmyr:
This builds on #171, and this diff includes a commit from that PR. I was assuming this would take longer. It'll be an easier read if that gets merged. ;) 

This is a slightly fudgy way of getting tabbing working again, but it isn't a lot of code and I'm not aware of a _better_ approach. I've tried to provide clear documentation inline.

This commit also bumps the SDK to 10.12, because it takes advantage of some tabbing-related APIs that become available then.

Assuming you're okay with this work, do you have any other frontend features you would like to have worked on? I'm enjoying hacking away, but would like to make sure I'm moving in a useful direction.

closes #168 
closes #57


raphlinus:
Generally seems fine.

Regarding 10.12, my _preference_ would be to allow one major version back. Would it be possible to use `if #available(OSX 10.12, *)`, or is there a lot of messy conditional code?

Happy to have you working on this stuff! Here's a partial wishlist of front-end things I'd like to see (arguably these should be issues, but I'll keep it lightweight for now):

* Show "dirty" state in UI, and pop up confirm dialog if closing on dirty (right now, it's a way to lose work).

* UI side of showing line numbers. I think this is actually pretty tricky because of scrolling interactions. The text can scroll horizontally and vertically, but the line numbers only vertically. Also, the ShadowView should cover only the text, not the line numbers. The actual plumbing of line numbers will take work in the core and update protocol (due to line wrapping), but you can just show the visual line for now, and I can fix that later.

* UI side of "find". Should pop up a little text box, and maybe have a status field ("1 of 3"). Will need to scale to replace as well, but can start with find. Obviously, implementation (in core) is highly nontrivial (just dealing with Unicode case folding is a fairly major effort), but I think I can get something ASCII-only and not very efficient done pretty quickly.

cmyr:
Ok, I should be able to ifdef no problem. cmd-T won't work on 10.11. (I may be able to disable the menu item.)

I'm also going to move the state storage for opening a new document (#171) into the Document type, to keep that stuff together.

And I think I've found a bug where we incorrectly open a window when we expect a tab, so i'll update with these fixes shortly.

happy to have some next steps, I'm enjoying myself.

Some thoughts, while I'm typing, and you can let me know if any of my ideas have glaring problems:

- I think I'll try and make the line-number stuff part of a more general gutter view (left and right, optional, displaying line numbers but in the future warnings etc).

- is 'dirty'ness a frontend or backend responsibility? The backend has a clearer sense of save status etcetera, it would make sense to me if there was an RPC call for checking document status.

- similarly with a more general bottom status bar. Can display file dirtiness there. 

- find is fun.

As well, I'd love to hack on plugins a bit more, but I don't really have a sense of how far a way we are from having multiple plugin support with method lookup &c. It feels like a non-trivial problem. But I'd love to do a racer plugin (and maybe rust language server?); if I had racer & I had the `move_word_left` & right actions I think xi would suddenly be a decent option for editing rust files. 

I'm also day-dreaming about a simple fuzzy finder. This is a strange one because it firmly feels like plugin territory, but it's also largely a UI question. I think there's a good argument for having it managed in core, though.

raphlinus:
Thanks for this!

I added some comments to the break-out issues, and we should continue those discussions there.

I'll do word movement soon. Getting it unicode-aware is hard (see https://github.com/unicode-rs/unicode-segmentation/issues/21 for thoughts on getting graphemes right), but it makes sense to do something that works for ASCII first.

I suppose find could be a plugin, but it feels like it should be in the core; among other things, it's likely to want to access the entire buffer, which could be a lot of excess RPC. Again, this feels like something where we should get basic functionality in place first, then refine.

I'm with you in feeling we're fairly close to having xi suitable for day-to-day programming. I've got to juggle a whole bunch of other responsibilities, but let's get there.

cmyr:
Sounds good raph! I wasn't thinking of find as a plugin for searching in files, but rather a fuzzy-match file navigator, like sublime text's "goto anything" (⌘P) command or Xcode's 'open quickly'. 

And yea, I'm happy to keep plugging away at this. I had a very slow work week this week, so was able to get properly immersed in the codebase, but I won't be able to _quite_ keep this pace up, though I should certainly have a few evenings free a week. 

I think I have some pretty clear next-steps, but it might be nice to put together a bare-bones roadmap sometime in the next few weeks, if only to have a sense of where to direct thinking?

Issue #175, opened 2017-03-02
line numbers / gutter


cmyr:
from the discussion here: https://github.com/google/xi-editor/pull/172#issuecomment-283775673. Line numbers are a good immediate goal, but it's useful to also think about how the UI would interact with plugins in the future.

PR #176, opened 2017-03-06
clarify scrolling behaviour in edit view?

cmyr:
Was working on something else and I came across some stuff in the mouse event handling of the `EditView` that i don't fully understand.

Is there a compelling reason for me to not make the included changes? I don't understand why we're trying to handle the scrolling manually, but I want to make sure I'm not missing anything.

This change is part of a larger PR I'm working on around #175.

**edit** More broadly, I guess I don't really understand the `markedText` stuff. Guess I should go read up on `NSTextInputContext`.

### DIFF
--- a/XiEditor/EditView.swift
+++ b/XiEditor/EditView.swift
@@ -14,13 +14,6 @@
 
 import Cocoa
 
-func eventToJson(_ event: NSEvent) -> Any {
-    let flags = event.modifierFlags.rawValue >> 16;
-    return ["keycode": Int(event.keyCode),
-        "chars": event.characters!,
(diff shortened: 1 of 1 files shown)


raphlinus:
Yeah, the purpose of the timer is to smoothly scroll when the mouse is held in a pressed state while the cursor is above or below the scroll region. Your change regresses that behavior - it will scroll when the mouse is actually moving, but pause when held still.

I'm not sure if there's a better way to get that behavior, but that's what I came up with.

cmyr:
Okay great, I'll take a closer look.

PR #178, opened 2017-03-17
moving more things out of the editView

cmyr:
This is a collection of changes that came about as I started thinking about how I would implement the gutter view. 

The majority of these changes are about continuing to remove code from `EditView.swift`. Most significantly (and maybe annoyingly) I've moved all of the font-specific drawing values into a struct, and moved it into the view controller; this (along with the line cache and the style map) is now exposed to the editView through a protocol member variable. This has the downside of making some of the drawing code more verbose.

I decided to take this route because the line-height metrics are going to be shared by the gutter view. If it offends your sensibilities I'm open to an alternative; one simple one would be to have the lineDrawingMetrics struct be passed to relevant children whenever it's mutated. 

Other things: 

I moved the code for requesting missing lines out of the drawRect. I actually spent quite a long time playing around with the drawing code, especially around getting the view's initial lines to be correctly populated. To this end I made one change on the core side, marking the view as `dirty` when we receive a `scroll` command. This lets me send `scroll` when the view loads, which correctly sets the viewport size and causes an update with the required lines. Is this a reasonable overloading of the scroll action? It feels like semantically this action is something more like "set_viewport". 

Anyway: with this change there are no longer any problems with missing lines, that I can tell.

This change also makes me suspect that `do_scroll` might want to send some overhang lines? In the future it could also keep track of scroll-speed / direction and send lines ahead of time. 

Let me know if there are any complaints here. I feel a bit self-conscious making changes like a lot of these, which are more style than content, but I think they're all more-or-less defensible. If you find this sort of PR annoying and would prefer I work more around the code as it exists, I'd accept that graciously.


raphlinus:
I'm uncomfortable merging this as-is, because it no longer correctly performs the protocol between front-end and back-end. I have no trouble with moving font metrics out of the editview so they can be used by a gutter view.

I should probably just buckle down and implement the minimal invalidation, as I can see it's going to cause frustrations for front ends (you and @eyelash, for sure) until it's done. But (a) it's complicated, (b) it interacts with multiple cursors, and (c) it seems to me to be not absolutely required for the editor to be usable.

That said, when it is done, scrolling is going to be smoother, and a lot of the flickering you might see now will go away.

I could do something stop-gappy, where any actual invalidation of the view sends the whole view, but the core tracks the state of the line cache just well enough to keep it up to date for scrolling.

Sorry for the negative review, and also for not getting back to you more quickly.

cmyr:
yea no worries this PR was a mess. I'll cherry-pick the stuff around drawing metrics and put them in a gutter view PR.

PR #179, opened 2017-03-24
Implement select_all

cmyr:
Was trying to use the editor just now and realized this wasn't implemented. 

Not sure if this fits with your general sense of how the backend should operate, I played around with a few ways of doing this and this one seemed like the cleanest. Let me know if you have a better approach!


raphlinus:
Yeah, this is fine :)

PR #180, opened 2017-03-24
Update travis config

cmyr:
we no longer use `xcodebuild` when building the core on macOS. I may not have done this right, I'm mainly opening the PR to see if travis likes my changes. 🤷‍♂️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,7 @@ cache:
     - $HOME/.cargo
 
 script:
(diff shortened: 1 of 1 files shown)


raphlinus:
Thanks, I was going to do this, but didn't get around to it. Having been experimenting in the core, I think we'll also want to go into at least core-lib/ and do a `cargo test`, but that's a separate improvement.

PR #183, opened 2017-04-01
Update to serde 0.9

### LINKED ISSUE #182: update to serde 0.9
serde 0.9 introduces the nice `json!` macro, which gets rid of the awkward `ObjectBuilder` stuff. It also provides `#[derive(Serialize, Deserialize)]`, which might be helpful at some point.

I'm opening this issue because I'm going to start this conversion, and wanted to give @raphlinus the opportunity to interrupt me if there's some good reason to hold off.

cmyr:
The majority of this PR is replacing the use of `ObjectBuilder` and `ArrayBuilder` with the new `json!({})` macro. The only place where this got a little tricky was in `view.rs`; the code around building updates heavily relied on passing around builder objects. I've now mostly modified it to just build `Vec<Value>`s in place, which hopefully has the side-benefit of making the code a bit easier to follow. I'm not sure if there were any subtle optimizations that I might've undone, but in general I was careful to keep things as streamlined as possible.

closes #182

### DIFF
--- a/rust/core-lib/src/styles.rs
+++ b/rust/core-lib/src/styles.rs
@@ -38,19 +38,20 @@ impl Style {
             "id": id,
             "fg_color": self.fg,
         });
-            if (self.bg >> 24) > 0 {
-                dict_add_value(&mut json, "bg_color", self.bg);
-            }
-            if self.weight != 400 {
-                dict_add_value(&mut json, "weight", self.weight);
-            }
-            if self.underline {
-                dict_add_value(&mut json, "underline", self.underline);
-            }
-            if self.italic {
-                dict_add_value(&mut json, "italic", self.italic);
-            }
-       json 
+
+        if (self.bg >> 24) > 0 {
+            dict_add_value(&mut json, "bg_color", self.bg);
+        }
+        if self.weight != 400 {
+            dict_add_value(&mut json, "weight", self.weight);
+        }
+        if self.underline {
+            dict_add_value(&mut json, "underline", self.underline);
+        }
+        if self.italic {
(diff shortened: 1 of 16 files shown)


rust/core-lib/src/styles.rs
```diff
@@ -32,22 +34,23 @@ pub struct Style {
...
-        }
-        if self.italic {
-            builder = builder.insert("italic", self.italic);
-        }
-        builder.build()
+        let mut json = json!({
+            "id": id,
+            "fg_color": self.fg,
+        });
+            if (self.bg >> 24) > 0 {
```
raphlinus: The indentation looks weird to me. If you want these indented, maybe add a {} block?

cmyr: Nice catch, this was in a block in an intermediate revision. Pushed a fix.

raphlinus:
Thanks!

cmyr:
💯

rust/core-lib/src/view.rs
```diff
@@ -104,130 +104,30 @@ impl View {
...
-            );
-        }
-        let mut sel = Vec::new();
+        let cursors = match line_num == cursor_line {
+            true => {
+                let mut c = Vec::new();
+                c.push(cursor_col);
+                Some(c) 
+            },
+            false => None,
```
eyelash: These changes make the core send a `"cursor":null` attribute for every line that does not contain a cursor whereas before the `cursor` attribute would only be either an empty array or an array with one element.

cmyr: Okay, that's a bug, I'll put up a fix

Issue #170, opened 2017-03-01
[RFC] Implementing split windows 


cmyr:
After tabs, split windows is a possible next major UI feature that we likely want to have, at some point, and figuring out how they are going to fit into the overall ownership model is something that we might want to hash out sooner rather than later.

Before getting too excited, I can see a few different ways splits could done;

1. we could go Xcode style, and simply allow for a single optional vertical split. Relatively simple to implement, but not super flexible.

2. we could go more in the style of a terminal multiplexer/vim, and basically allow any existing view to be split into two. This is trickier to implement in the front end, but not terribly so (at least in terms of programatically laying out the views)

3. we could not support splits.

Something I'm much less clear on is how this will work architecturally. Currently an editor view is owned by a view controller object, and is backed by a subclass of `NSDocument`. That document is responsible for creating the logical window into which that view is drawn. With tabs, documents maintain individual logical windows where are swapped in and out as tabs are swapped; each tab is considered a window. With splits, this breaks down a little bit, because a single window can contain multiple editors, but each editor's document expects to own its own window.... 

I haven't gone too deep down this path yet, and I suspect this won't be too tricky, but doing it well will involve some selective subclassing of `NSDocument` and `NSWindowController`.

I'm going to play around with this at some point, and if I'm making any progress I'll put up a WIP PR.

Does anyone have any thoughts on this? Should the backend be aware of split relationships?

chriskrycho:
Input as a would-be user: I make *very* heavy use of multiple split panes. One of my most common use cases is developing [Ember] and other such frameworks, in which contexts it's extremely common to want to be able to see the JS implementation, the Handlebars template, the corresponding SASS file, and the test JS file all at once—those certainly wouldn't all go side-by-side comfortably even on my 5k screen; the fact that Atom and Sublime both support basically arbitrary splitting is a big selling point of both in the GUI text editors space for me. It's something I got used to in Vim and Emacs land and have wanted ever since. It's therefore also one of the bigger annoyances for me in the current version of VS Code, which does *not* support arbitrary splits.

[Ember]: http://emberjs.com

AljoschaMeyer:
[...]

Issue #93, opened 2016-07-17
VIM for XI ?


Byron:
This issue is merely a question. If it's a bad place to put it here, please let me know where to direct it otherwise. That said: **Do you think it's feasible to implement VIM-like command support in a way that allows it to be re-used by all frontends ?**

To me, without nearly no knowledge about the XI architecture, it seems like it should be a command-handler running in the backend. Front-ends would need some sort of connector to send all key-presses to it, and it would decide on what to do. Effects of its operation could either manifest themselves in changes to the document itself, or the GUI. The latter would happen during mode-changes for example.

The second question coming to my mind: **Would it be feasible to start with the backend VIM plugin already ?**

_(My motivation to become more active with this is the promise of having a next-gen editor, which from the very beginning is build around high-performance data structures and programming languages. Right now, both Atom and IntelliJ seem to lag even with small files, and during ordinary interactions.)_

porglezomp:
I'm wondering if there should maybe be one more layer. I haven't really contributed to the project, but every time I come by I see some people discussing sticking common interaction code into the core. How well would it work to have

- Frontend: Platform specific, native interactions, sends events down the chain
- Middle: Interaction specific, one for modal editing and one for single-mode editing
- Backend: Performs the editing operations, just like now

The downsides I can immediately see here are the obvious complexity one, and the obvious performance one (which could maybe be reduced by having a channel between 1 and 3 for all the purely presentational issues). Has an architecture like this been suggested before?

emoon:
Bump. I would like to hear more thoughts on this.

cmyr:
This is something we would like to get around to eventually, but an earlier priority is fleshing out more of the basic editor functionality.

emoon:
Sounds good. Thanks for the reply.

dotancohen:
+1 support for a VIM mode. Xi is the first editor that I've been excited about in over a decade, but without a VIM mode it would be useless to me.

dayuoba:
how i wonder such cool editor with vim mode!

autoferrit:
One thing that would be awesome, is to have an easy keymap structure like space-neovim, spacemacs, and others. I would still prefer vim bindings, but this adds essentially a UI layer over the keybinding to make them easier to discover and remember as well. which is the big thing for me (using space-neovim).

dineshbhosale:
Spacemacs keybindings are <3

dineshbhosale:
For adding some substance , to above comment 
[...]

PR #189, opened 2017-04-13
Plugin "next steps" RFC

cmyr:
For the past week and a bit I've been playing around with building up plugin support. This RFC is a basic outline of where I'd like to take that work. It isn't very complete, but I hope it's a useful outline. :)

### DIFF
--- a/doc/plugins2.md
+++ b/doc/plugins2.md
@@ -1,0 +1,156 @@
+# Next steps for plugins
+
+This document builds on the [existing plugins rfc](https://github.com/google/xi-editor/pull/43), with a focus on fleshing out a preliminary API + implementation that allows people to more easily & practically experiment with writing plugins for Xi.
+
+## Goals:
+
+This document aims to present an alpha/preliminary/experimental implementation that supports the global, editor-local, and one-shot/invocation plugin scopes described in the previous RFC. 
+
+## Plugin Capabilities + UI support
+
+For this preliminary implementation, I would like to provide basic support for a small subset of plugin capabilities:
+
+- _commands_ are actions a plugin can define which the user can specifically invoke, for instance through a command palette. 
+- _hooks_ are editor events that a plugin can be notified of, such as saving or opening a file or inserting text.
+- _exports_, similar to atom's [services](http://flight-manual.atom.io/behind-atom/sections/interacting-with-other-packages-via-services/); functionality that a plugin can provide to other plugins, as a list of named RPC methods. 
+
+And I would like to add a few simple UI elements to the editor, to make experimenting a bit easier:
+- a simple _command palette_, which will be automatically populated with commands provided by plugins,
+- _status fields_, elements a plugin can add to the bottom status bar of a window, and send ongoing updates to. A good simple UI paradigm, suitable for testing plugin->client interactions; 
+
+and then, a little later:
+- the ability to annotate the gutter (warnings/errors/lints)
+- an autocompletion UI/API (probably as similar as possible to the language server protocol)
+- inline documentation popovers
+
+
+_Question_: Are there any other client features that are obviously important for experimenting with plugins?
+
+
+ 
+## Plugin Lifecycle
+
+Depending on its invocation scope, a plugin will receive certain notifications and requests from the editor automatically.
+
+### Invocation Lifecycle
+
+Invocation-scoped (one shot) plugins have the simplest lifecycle. In general they will be invoked in response to a specific user command (say, to render a markdown preview) or an editor event, such as the saving of a file. There are two possible approaches to the API, which are very similar: in the first, after the RPC channel is opened, the plugin receives a `run` request, which will contain basic information about the invoking buffer, as well as the command responsible for the invocation.
+
+The second major component of `run` would be an `invoking_method` or similar field, which is a method/params pair representing the request or notification that caused the plugin to fire. A single plugin executable should be able to define multiple triggers that it responds to. 
+
+The alternative is to have their be two commands: an `initialize` notification followed by some `request` method. This has the advantage of standardizing the `initialize` event across all plugin categories. I prefer the first approach: I like that to the application and to other plugins, there would be no distinction between calling a method on a running plugin and calling a method that invokes a single-shot plugin. 
+
+possible example `run` rpc:
+
+```json
+"method": "run",
+"params": {
+    "buf_info": {
+        "buf_len": 1337,
+        "revision": 161,
+        "syntax": "rust",
+    },
+    "invocation": {
+        "method": "xi-python-pep8.command.format", # or some other way of namespacing a command
+        "params": {},
+    },
+}
+```
+
+Once it has received the `run` request, the plugin is expected to perform its work and exit. 
+
+### Editor-local lifecycle
+
+Editor-local scoped plugins are associated with a given buffer, receive lifecycle events for that buffer, and can query that buffer through the plugin API.
+
+_Question_: How will we handle multiple views per buffer? Are plugins buffer scoped or view scoped? It seems silly to be running a linter in multiple views into the same buffer, but certain use-cases (such as acting on selections, or showing the cursor position in the status bar) want to be running per-view. 
+
+The lifecyle for editor-local plugins is more involved than for invocation plugins. They should receive some version of `initialize` and `deinit`, for doing setup and teardown. They should also receive all of the [`EditCommand`](https://github.com/google/xi-editor/blob/master/rust/core-lib/src/rpc.rs#L52) events from core (probably?) and they should receive the special `update` event which sends deltas representing changes to the buffer.
+
+_Q_: Are there any other significant lifecycle events that would be useful for this alpha API?
+
+_Q_: Should a plugin receive events by default, or should it have to register for notifications it is interested in? The first is simpler, but inefficient; message volume on the wire will grow linearly with the number of plugins.
+
+_Q_: should plugins be able to register/deregister for notifications while running? If we're doing a registration-based system then yes, probably, at some point?
+
+
+Other events to add, eventually:
+- hover reporting, mouse position reporting;
+- window resizing? window gain/lose focus? 
+
+### Global lifecycle
+
+Global plugins receive all of the events that editor-local plugins do, and for multiple open buffers (there should be some mechanism for indicating buffers of interest) plus they also receive all file_open/file_close events.
+
+**n.b.**: file_open/file_close/new_file are currently a bit messy; there are no editor-level new_file or file_close methods at the moment, but arguably there should be.
+
+_Q_: What else might be useful here?
+
+_Q_: How are buffers identified to global plugins? Do we have some internal identifier (such as the tab identifer currently in use) or do we have some new identifier that is buffer specific, as opposed to being linked to a view?
+
+## Plugin API
+
+Once running, plugins are able to communicate with `xi-core` through the standard rpc mechanism. I'd like to flesh out this API a bit more, to provide a baseline set of functionality for people experimenting with plugins.
+
+Much of the particulars of this API could in many cases will be abstracted behind a good client library: in python, for instance, a `Buffer` class might have a `lines` member which implements `__getitem__` and fetches raw bytes over rpc as necessary.
+
+_Q_: How should the underlying RPC methods be namespaced? (Should they be namespaced?) This may be helpful for message routing, but also for debugging and legability. For the purposes of this document I am going to namespace things according to the struct where the given request would currently be handled:
+
+### proposed bare bones API:
+
+
+#### methods available to all plugins:
+_global plugins will need to specify a buffer when calling methods, other plugins will not_.
+
+**n.b.**  there's some conflation here right now between the API provided by the plugin library and the actual rpc protocol.
+
+_Q_: by what interface should the buffer be presented by the plugin library?
+
+#### requests
+- `buffer/get_data(offset, max_size, rev)` (implemented): returns raw bytes.
+- `buffer/path()`: returns the filepath, if the file is saved, else `Null`
+- `buffer/syntax()`: returns the currently active syntax definition
+    
+    _Q_: should we just have a single `buf_info` call that returns a variety of basic information?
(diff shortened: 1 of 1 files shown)


cmyr:
[Rendered](https://github.com/cmyr/xi-editor/blob/7c125eba67dd912719b22636287592c84471b25f/doc/plugins2.md)

raphlinus:
Generally good stuff. Some thoughts.

A goal of this effort (merging in the doc and discussion from #43, which you linked) should be to get some document (not necessarily a detailed spec, some items can be treated at a high level) checked into master.

Whether a plugin is buffer-scoped or view-scoped depends on what it's doing. Syntax highlighting is the former. Finding matching brackets depends on cursor state, which is view-scoped. These are highly related; it makes sense to have the same plugin handle both. I think one of the more straightforward approaches is that plugins are basically buffer-scoped, but certain interactions (like with cursors) carry a view id.

For manifest and config (which are different but related), I'm thinking toml. Of course, that's just a low level syntax choice, the much more important question is what is represented.

Aren't view/selections() and view/cursors() the same? (a cursor is just a selection with start == end, a pattern common to most editing api's I've seen)

Profiling support early is good. There are some consequences, for example every update (whether it causes changes or not) should generate a response saying "all processing of this update is done". Since the design is async, we'll want to thread id's through these sequences.

I'm thinking of not including the actual EditCommand in the update to the plugin (this would require standardizing on the space of edit commands). However, you do need lots of fine-grained information about what caused the edit (kb typing, backspace, paste, undo, etc). In addition, you want some id threading (similar in mechanism but different in motivation from the performance stuff) so that edits made by the plugin can be associated with a user edit. If a user types "(" and the plugin inserts ")", then if the "(" is deleted or undone, the ")" should also be removed. As I discussed on IRC, the plugin->core insert method should also have a "priority" (basically an indication of whether to insert before or after the cursor) to deal with more complex cases such as "{|}" becoming "{\n |\n}" (here | represents the cursor).

Queries (not just buffer/get_data) will need a revision number to deal with concurrency races.

It's worth noting that the set of cursors/selections can become very large. Consider opening a million-line file, doing "select all", then "split into lines". Probably the best way to deal with this is to have an "overflow" value (perhaps with metadata such as the first and last selection and the total number of selections). Designing a protocol for partial-view access to a large selection object (as is done for text) is possible but probably not worth the complexity, considering that these cases are rare.

Inter-plugin communication is potentially very important but I think should be deferred a bit. It's reasonable to consider the core acting as a publish-subscribe mechanism for updates, with mechanisms for discovery, etc. That said, when designing manifests and formats for notifications, might be worth thinking about the fact that they may be consumed by other plugins as well as the core. It might be worth considering a more layered structure, where a lower layer deals with lifecycle and allows subscription to generic "objects" which can be many things, while a higher level layer instantiates those objects with (at least) buffers and views (likely also workspaces, and such an object could represent the state of, say, a "find" interactive query), and defines specific methods to be done on those objects. But I don't want to get too caught up in architecture astronautics.

Thanks for the input! Looking forward to hopefully some discussion, then we can synthesize the results and merge as a doc.

rkusa:
I'd like to add a some thoughts to the discussion. I am not sure about buffer-scoped and view-scoped plugins. They are indeed useful, however, it could be a good idea to force plugins "singletons" that are able to handle multiple views/buffers instead of having single plugin processes per view/buffer. Because the latter is easier, plugin developers probably prefer buffer-/view-scoped plugins over global ones and we end up with a lot of processes per buffer/view. Additionally, since files are not edited in parallel but more in a task-switching manner, plugins should be easily able to handle multiple buffers.

I think it could make sense to just have syntax-scoped plugins. E.g. when I am opening the first Rust file, all Rust-syntax-scoped plugins are started. If I open more Rust files, they are handled by the already started plugins. When all Rust files are closed, the plugin is terminated until I open a Rust file again.

There is one issue that came to my mind while writing this down. What if I have a general regex-driven syntax highlighting plugin that is able to handle all syntaxes, but also a Rust-specific syntax-highlighting that is more specialized. How is it possible to receive conflicting color spans from both plugins. This issue applies to all plugins scopes.
This issue may require to specify a usecases/features of plugins, so that only one "syntax-highlighting" plugin is responsible for certain buffer / for a certain syntax. Does anyone know how this is solved in Sublime/Atom/...?

The plugin definition could in this case look like

```toml
[[plugin]]
executable = "rls" # rust language server
features = ["syntax-highlighting", "auto-complete"]
scope = "text.rust"

[[plugin]]
executable = "syntect"
feature = ["syntax-highlighting"]
scope = "text.*"
```

"text.rust" wins over "text.*", because it is more specific.

raphlinus:
I'm in the process of merging PR's that are draft docs, but this still feels like an active discussion. My preferred resolution would be to capture the useful discussion in the comments (worst case, just pasting it in, but ideally integrating it some) then merge. Not a huge rush though; I know we're actively iterating on this now.

cmyr:
Okay I have some changes to this sitting around and am actively prototyping, will update in the next few days. 

@rkusa: the current plan is to support 'singletons', as you discuss. We've been calling them 'globally scoped plugins', and they receive events from all active views. I agree that in general they will be a better use of resources for a variety of use-cases, and we can hopefully make that clear through A) documentation, and B) good API design. (A plugin library could be designed in such a way that a buffer-scoped plugin could be implemented as a global-scoped plugin with very little/no change from the developer's POV, e.g. in my python examples the lib would just instantiate a new instance of the plugin class for each buffer). 

my current thinking about 'activation events' is very similar to [VSCode](https://code.visualstudio.com/docs/extensionAPI/activation-events)s. 'Active syntax definition' (e.g. filetype, in the general case) would be one of the activation events. (edit: you linked to this too, cool 😄)

In terms of the broader problem of plugin conflicts: A robust system for detecting conflict would be nice. If we move to doing syntax handling in core, which we've been talking about lately, plugin behaviour can just be prioritized; this doesn't really do much for handling conflicts between plugins. I haven't thought about this too much.

trishume:
### Buffer-scoped plugins

I mostly agree with @rkusa.

Here's my worry: Xi developers will spend a bunch of time implementing two separate APIs, one for globally scoped plugins and one for buffer-scoped plugins. Then Xi plugin developers will develop plugins, which they will start as buffer-scoped plugins because of limited functionality at the beginning. Then, they'll want to add a feature or make things more efficient, and that will necessitate them refactoring their entire plugin to be globally scoped instead, requiring a bunch of work switching to a separate API.

I think that if such a thing as buffer-scoped plugins is to exist, they should use the exact same API as globally-scoped plugins. That is, they should include a reference to the buffer in messages.

Then buffer-scoped plugins can just be a value in a config file that determines how Xi manages the lifetime of the process. And the only difference for the plugin author is that they don't have to implement buffer lifecycle management because they can rely on only receiving one "new buffer" message. 

It might be worth implementing global-scope plugins first because they're kind of a superset of buffer-scoped plugins.

Also because of the work involved in implementing buffer synchronization and caching from scratch, I imagine most plugins will be written using a library, and the libraries can offer a separate interface for buffer-scoped plugins that doesn't require them to refer to the buffer if they really want.

### Message routing

One nice architecture I've seen in other plugin-based multi-process software ([example](https://docs.pupil-labs.com/#a-full-example)), is having everything be on a message bus where plugins can subscribe based on various filters, including the core. That way sending messages to the editor and receiving notifications from it is conceptually the same as between-plugin communication. The message bus code would of course have a fast path for in-process communication with the core.

Another nice thing is that communicating with a specific plugin and receiving communication directed at your plugin can use the same mechanism as trying to only receive certain types of editor events.

Each message could have some kind of hierarchical name like `editor.notification.buffer.change`, `editor.request.view.move` and `plugin.gofmt.cmd.formatfile`. Then plugins and the core editor could subscribe to prefixes like `editor`, `plugin.gofmt` and `editor.notification.buffer`.

By coincidence, there even happens to exist [a Rust library](//github.com/trishume/syntect) that has code for [matching prefix filters efficiently on structures like these](https://github.com/trishume/syntect/blob/f03fa0f672ab0a25ec93cddba11a8a6b036605f9/src/parsing/scope.rs#L251). I like the conceptual uniformity of this.

### Non-sequitur agreement about a past point

Not so much about this, but more about the Reddit discussion on plugins that I missed earlier. As mentioned, there's a lot of plugins like linters and formatters that act on the whole file at once occasionally. It's definitely important that there be an RPC to just get the entire document contents, perhaps with a parameter for size limit, but that should be pretty easy.

### Musings about tradeoffs

All this discussion about all the different caching and size limiting mechanisms the plugin protocol will require are making me realize how nice in-process plugin architectures like in Emacs/Vim/Atom/VSCode/etc are. When API calls are super cheap it's no big deal to read the whole file or ask for a bunch of specific text ranges. I'm worried that because of all the synchronization and caching mechanisms required to make a plugin not slow, every plugin will be written using one or two libraries, which will not end up being any better than just using dynamic libraries in-process for plugins and having one or two binding libraries to the C API for different languages. And it will have the downside of each buffer being cached in N different plugins and synchronized on every edit being synchronized to a bunch of places instead of just stored in one place and having plugins request what they need when they need it.

I'm sure it's possible to make it all work well, and I don't think it would be that bad. I'm just realizing that the only gain over just having plugins be dynamically loaded libraries communicating over a C API that I can think of is that plugins will be unable to crash the core process. Although that can be mitigated somewhat by just not using plugins written in C/C++. The (small) advantage of the dynamic-library approach being that if someone implemented a Vim plugin, it would respond just as fast as the core in @raphlinus's million line file with a cursor on every line approach when you issued a movement command.

It's a tradeoff. I imagine there's more benefits of separate processes communicating over serialized RPC that I'm missing, and I'd love to hear them.

One possibility is that there will end up being a small number of implementations of a "lightweight plugin host" that implement all the caching, communication and synchronization in different languages. These plugins will then end up loading lightweight plugins, either as dynamic libraries or using dynamic languages like Python, which communicate over an API in the native language fetching data directly from the shared cache. That way people will be able to write fast plugins (maybe in nice scripting language) that are easy to get started with. This will end up being very much like Sublime Text's plugin host process which keeps plugins easy to write while stopping them from messing with the core.

cmyr:
@trishume good comments, I think we're all pretty much in agreement. Buffer-scoped and global plugins (as currently imagined) have the same API, with the difference that the global plugins receive events from multiple buffers. 

The main reason I like this approach is that it is conceptually simpler to the plugin author, but there might be an argument that this is just a failure to come up wtih a good client API.

My current thinking is that total memory footprint (xi + all running plugins) is something we want to be more worried about than we do about, say, RPC traffic volume. So that is a good argument in favour of global scopes.

I'm not sure if there is a good argument I'm missing for the original one-process-per-buffer argument? @raphlinus do you have any thoughts on this?

message bus: this could be an interesting approach? One issue is that in certain cases the editor isn't just posting a notification, but needs to be informed of how many plugins are handling the notification, so it can keep track of in-flight revisions. 

re: non-sequitar— Getting the whole buffer contents is basically a responsibility of the plugin library. In the current python impl, for instance, if you just run a loop like `for line in self.lines` the library will fetch the whole buffer for you. 

The idea of having a "plugin host" process is something that is included in the current version of this RPC, but also as a sort of afterthought. 

It is currently imagined that modal (vim) functionality will be a special case, and included in core.

Raph will be better position than I to explain his reasoning for the rpc + stdio architecture, so I'll leave that one alone. My general thinking here, though, is that the more complicated optimizations we can make should be reserved for when we actually encounter some problem where rpc + stdio is the bottleneck, and I think that horizon is pretty distant. It would definitely be nice to have some better benchmarks here, though.

theduke:
I have scanned the RFC and the other issues relating to plugin development, but it's still hard for me to grasp how plugins are supposed to work and how the RPC protocol is supposed to look.

One big question I have relates to ordering and dependencies between plugins.

I think there should be a lifecycle related to file/buffer changes, so plugins can declare that they are supposed to run before/after other plugins.

One example I've been thinking about:

I'd love to see/develop a `Vim` plugin.
Implementing full Vim modal editing support is quite a complex affair.
Being able to use a Vim plugin and not reinventing the wheel in each frontend would be a big deal.

That would require the plugin to receive all keys strokes (including modifiers), and also have some form control over layouting and editor behaviour.

I don't know if there are any plans yet regarding things like vim splits aka multiple visible buffers.

Just this out there.

theduke:
Also, I think an auto completion workflow and "framework" should be considered.

Plugins can supply auto completion suggestions. The frontend receives a list of suggestions and can display them to the user.

Of note here is that completions should carry an identifier and a description, which does not necessarily relate to the expanded completion.

theduke:
And a third thought:

A common **configuration and keybindings** system.

Maybe there could be a standardized configuration format (eg toml) with common config options that can be used between frontends. Frontends extend this and support what they can and want to.

Plugins are already planned to support commands.
They could also offer a list of keyed and named key bindings.
This would make it possible for the frontend to show a nice list with key bindings.
Maybe that's a bit redundant if a key binding can just invoke a plugin command, though.

cmyr:
@theduke I think we're mostly on the same page; most of the things you mention are on the current plugin roadmap. Vim mode/modal support generally is something we definitely have planned (#93 is the issue for this, no real details there but current thinking is that it _won't_ be a plugin in the current sense, but will be in xi-core, implemented as a state machine that consumes keyboard events and outputs editor actions. Splits are also on the roadmap (#170). Plugins will definitely get config files. How we will approach general user settings is something we haven't discussed too much, but I imagine something like `~/.config/xi/xi_config.toml is the most likely approach.

Autocompletion is definitely also part of the eventual plugin scope.

Inter-plugin communication and inter-plugin dependencies as well.

Plugin work is still very much early days, but if you're interested in playing around writing some sample plugins that is very possible right now; you can take a look at the python sample plugins for a sense of what is currently possible. These plugins are mostly limited to basic editing, but there's lots that can be done in there, and more functionality will be landing in the next few weeks.

If you're curious/interested jump on irc.mozilla.org/#xi and say hi.

cmyr:
This is out of date and is not going to get merged, and I see no reason in keeping it open as a historical document.

I hope to be revisiting plugin stuff more thoroughly in the coming months.

PR #192, opened 2017-04-13
Add goto rpc command

### LINKED ISSUE #190: goto line
A basic standard editor feature that we'll want at some point. Should be straightforward; scroll the view and set the cursor. Will require some simple new UI for text entry.

cbrewster:
Fixes #190

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -616,6 +616,14 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.view.send_update_for_scroll(&self.text, &self.tab_ctx, first, last);
     }
 
+    /// Sets the cursor and scrolls to the beginning of the given line.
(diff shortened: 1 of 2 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -616,6 +616,14 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.view.send_update_for_scroll(&self.text, &self.tab_ctx, first, last);
     }
 
+    /// Sets the cursor and scrolls to the beginning of the given line.
+    fn do_go_to_line(&mut self, line: u64) {
+        if let Some(line) = line.checked_sub(1) {
```
cmyr: If this is a case we expect I'd prefer to crash than to fail silently. I'd probably prefer that the frontend normalize line numbers to be zero based when making the request.

raphlinus:
Nice and simple, thanks. And thanks @cmyr for taking a look.

Issue #191, opened 2017-04-13
modify RPC protocol to support better lifecycles + multiple views


cmyr:
## Summary

The current way documents are created, opened, and closed does not cleanly support integration with plugins, or future support for multiple views (which should be at least considered while developing the plugin system). This is true in two particular ways: first that the RPC protocol isn't adequately descriptive to allow for cleanly determining buffer/view lifecycles, and second that there is currently a strong coupling between a buffer and a view (one buffer, one view) that does not reflect the ultimate goals of the protocol; and it would be helpful to reconsider this now, because it will allow for a better understanding of how plugins should work. I will address these two points in turn.

### Lifecycle

I discussed this briefly in #186, but I want to take a more holistic approach.

Currently, when the client creates a new window, it sends the `new_tab` RPC to the core. The core then picks a tab name (incrementing an internal counter) creates a new `Editor` instance, a new `TabCtx` instance (basically a collection of references to shared resources like the client, the kill_ring, the stylemap, etc) and then returns the tab name to the client. 

This occurs regardless of why the window was created; the same process occurs for opening a window as for creating a new one. (I outline this pretty well in #186, I won't repeat myself here.)

When the client closes a window, the `Editor` receives no notification; the core simply removes its reference to the editor, and the memory is cleaned up when it goes out of scope.

It is feasible that an eventual plugin API will want to include lifecycle events such as `will_open_file`, `did_open_file`, `new_file`, `will_close_file`, and `did_close_file`. It is feasible that a new buffer might be created, _with a path_. (E.g, in vim: `:e ~/i_dont_exist_yet.txt`).

Semantically, there is no such thing as a view which is not backed by a buffer. If the client is creating a view, it is for either a new buffer, to load a file, or to create a new view into an existing buffer. This should be reflected by the protocol. 

A possible solution: replace `new_tab` with `new_buffer`, and add `open_file`. `open_file` (or some other name) could eventually be overloaded to handle opening new views into an open buffer, or (probably better) there could be a `clone` or `new_view` RPC. `open_file` would be a command at the top level (a `TabCommand`) alongside `new_tab`/`new_buffer` and `delete_tab`. 

Alternatively, there could be a single `new_view` RPC that handled all three cases, through a few params. (After more thinking I'm leaning in this direction: `new_view` + a `path` param; `close_view`, which closes a view; a buffer is closed when it has no views).

#### An aside, on naming:
[...]

PR #194, opened 2017-04-15
RPC protocol update + multi-view plumbing [ready]

cmyr:
~edit: After a bit of playing around there may be a few performance regressions I want to look at, but this is good for review.~ (all good)

This is approximately where I would like it to be in terms of the API modifications discussed in #191. It also includes a variety of naming changes, which are intended to more clearly document the use of particular APIs & methods. I'm touchy about renaming other people's code: please consider all naming changes to be suggestions, and I'm happy to revert them. 😄

### RPC Changes:
Because this already involved breaking changes to the RPC protocol, I took the liberty of normalizing certain other RPC signatures.

- In *all* RPC methods, the word `tab` as an identifier of 'client viewports' has been replaced with `view_id`.
- in *all* RPC methods, the word `filename` has been replaced with the word `file_path`. (We were using both in different places, I just grepped the src directory and picked the most common).
- `TabCommand` is now `CoreCommand`. (not a protocol change)
-  `NewTab` and `DeleteTab` (`new_tab`, `delete_tab`) have been removed from `CoreCommand`.
- `NewView { file_path }`, `CloseView { view_id }` and `Save { view_id, file_path }` have been added to `CoreCommand`.
- `Open` and `Save` have been removed from `EditCommand`.

In error messages and other names, the word `tab` has been replaced with `core`, where appropriate.

### Other changes:

The other main change here is that `Editor` now behaves as it if has multiple views, although the actual multi-view support isn't implemented. To reduce the diff size, `Editor` still does everything through a `self.view` instance. There is code in place to dynamically swap this variable out when an event arrives associated with a new view, but these codepaths are not currently available.

Anyway: the real consequence of this is that events and responses are associated with a specific view_id, rather than a buffer+view (e.g. a 'tab'). This plus the clarified view and editor lifecycle stuff enabled by the RPC changes set me up to _hopefully_ start implementing provisional plugin support. Maybe? 🤞 

### To think about:

- What is the role of `TabCtx`, now? with message routing now based on `view_id`, which isn't owned by a particular `TabCtx`, this should be rethought.
- How will actual multi-view support happen, and what refactoring is required to make that possible? It still feels like there's a lot of state split between the editor and the view. I haven't looked into this very much.
- **Are  there any other protocol changes to make now**, while we're already breaking things?

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -28,7 +28,7 @@ use styles::{Style, StyleMap};
 use MainPeer;
 
 /// ViewIdentifiers are the primary means of routing messages between xi-core and a client view.
-type ViewIdentifier = String;
+pub type ViewIdentifier = String;
 
 /// BufferIdentifiers uniquely identify open buffers.
 type BufferIdentifier = String;
@@ -50,7 +50,6 @@ pub struct Tabs<W: Write> {
 
 #[derive(Clone)]
 pub struct TabCtx<W: Write> {
-    tab: BufferIdentifier,
     kill_ring: Arc<Mutex<Rope>>,
     rpc_peer: MainPeer<W>,
     style_map: Arc<Mutex<StyleMap>>,
@@ -69,9 +68,8 @@ impl<W: Write + Send + 'static> Tabs<W> {
         }
     }
 
-    fn new_tab_ctx(&self, buffer_id: &str, peer: &MainPeer<W>) -> TabCtx<W> {
+    fn new_tab_ctx(&self, peer: &MainPeer<W>) -> TabCtx<W> {
         TabCtx {
-            tab: buffer_id.to_owned(),
             kill_ring: self.kill_ring.clone(),
             rpc_peer: peer.clone(),
             style_map: self.style_map.clone(),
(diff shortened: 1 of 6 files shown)


cmyr:
@raphlinus 🚀

rust/core-lib/src/tabs.rs
```diff
@@ -15,112 +15,238 @@
...
+        self.id_counter += 1;
+        format!("buffer-id-{}", self.id_counter)
+    }
+
+    pub fn do_rpc(&mut self, cmd: CoreCommand, rpc_peer: &MainPeer<W>) -> Option<Value> {
+        use rpc::CoreCommand::*;
+
+        match cmd {
+            CloseView { view_id } => {
+                self.do_close_view(view_id); self.
```
raphlinus:
```
error: no field `None` on type `&mut tabs::Tabs<W>`
  --> core-lib/src/tabs.rs:94:51
   |
94 |                   self.do_close_view(view_id); self.
   |  ___________________________________________________^ starting here...
95 | |                 None
   | |____________________^ ...ending here

error: aborting due to previous error
```

cmyr:
I was testing out RLS vim integration. `self.` produces lots of useful completion suggestions! 😞

raphlinus:
Thanks!

PR #203, opened 2017-04-17
Keep track of saved/unsaved state

cbrewster:
This initial implementation keeps track of a saved revision. This
revision is initially set to the latest revision when creating a new
editor and is updated to the latest revision during a save.

The unsaved state is calculated by comparing the deletion subset of both
revisions. This creates the following behavior: If new text is inserted
and the deleted back to the original state, there are no unsaved
changes; however, if existing text is deleted and retyped, there are now
unsaved changes. If you undo back to the state of the save revision,
there are no unsaved changes.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -56,6 +56,7 @@ pub struct Editor<W: Write> {
     view: View,
     engine: Engine,
     last_rev_id: usize,
+    save_rev_id: usize,
     undo_group_id: usize,
     live_undos: Vec<usize>, //  undo groups that may still be toggled
     cur_undo: usize, // index to live_undos, ones after this are undone
@@ -119,6 +120,7 @@ impl<W: Write + Send + 'static> Editor<W> {
             view: View::new(initial_view_id),
             engine: engine,
             last_rev_id: last_rev_id,
+            save_rev_id: last_rev_id,
             undo_group_id: 0,
             live_undos: Vec::new(),
             cur_undo: 0,
@@ -137,7 +139,7 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
 
-    #[allow(unreachable_code, unused_variables)] 
+    #[allow(unreachable_code, unused_variables)]
     pub fn add_view(&mut self, view_id: &str) {
         panic!("multi-view support is not currently implemented");
         assert!(!self.views.contains_key(view_id), "view_id already exists");
@@ -145,10 +147,10 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
     /// Removes a view from this editor's stack, if this editor has multiple views.
-    /// 
+    ///
     /// If the editor only has a single view this is a no-op. After removing a view the caller must
     /// always call Editor::has_views() to determine whether or not the editor should be cleaned up.
-    #[allow(unreachable_code)] 
+    #[allow(unreachable_code)]
     pub fn remove_view(&mut self, view_id: &str) {
         if self.view.view_id == view_id {
             if self.views.len() > 0 {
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -663,13 +670,17 @@ impl<W: Write + Send + 'static> Editor<W> {
...
         self.path = Some(path.as_ref().to_owned());
+        self.save_rev_id = self.last_rev_id;
+        self.view.set_dirty();
+        self.render();
     }
 
     fn do_scroll(&mut self, first: i64, last: i64) {
         let first = max(first, 0) as usize;
         let last = last as usize;
+        let unsaved_changed = self.has_unsaved_changes();
```
cmyr: nit: unsaved_change**s**, probably?

cbrewster: woops!

rust/core-lib/src/editor.rs
```diff
@@ -302,9 +304,14 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
     }
 
+    fn has_unsaved_changes(&self) -> bool {
```
cmyr: I wonder if it wouldn't make more sense to store this as a member of the editor struct, or somewhere like in `tab_ctx` (which gets automatically passed to the view when we render). Then we could compute it only after an edit occurs, instead of on each render cycle.

I like the idea of this living in something _like_ `tab_ctx`, because it saves it having to be passed to each call? I wonder if this isn't also true of the `style_spans`? But I'm not sure if that makes sense.

rust/rope/src/engine.rs
```diff
@@ -222,6 +234,18 @@ impl Engine {
         self.rev_id_counter += 1;
     }
 
+    pub fn compare(&self, base_rev: usize, other_rev: usize) -> bool {
```
cmyr: nit: the name of this function doesn't convey what it _does_, e.g. that it isn't just comparing but is checking equality, and that it returns a bool. 

I'm not sure what I'd call it, but maybe something like `are_equal` or `are_equivalent`?

rust/core-lib/src/editor.rs
```diff
@@ -56,6 +56,7 @@ pub struct Editor<W: Write> {
     view: View,
     engine: Engine,
     last_rev_id: usize,
+    save_rev_id: usize,
```
raphlinus: I prefer "pristine_rev_id" for this, as "save" sounds hortatory.

rust/core-lib/src/editor.rs
raphlinus: The general style in xi is to update the state fully when things change, then have querying of the state as trivial as possible (which is what @cmyr is suggesting). But if it's cheap to compute and simpler, I don't have a big problem with doing more work on query.

rust/core-lib/src/editor.rs
```diff
@@ -663,13 +670,17 @@ impl<W: Write + Send + 'static> Editor<W> {
...
         self.path = Some(path.as_ref().to_owned());
+        self.save_rev_id = self.last_rev_id;
+        self.view.set_dirty();
+        self.render();
     }
 
     fn do_scroll(&mut self, first: i64, last: i64) {
         let first = max(first, 0) as usize;
         let last = last as usize;
+        let unsaved_changes = self.has_unsaved_changes();
```
raphlinus: Hmm, now I'm a little more uncomfortable and would prefer this to be stashed. do_scroll() can happen with very high frequency. Plus, stashing the state might decrease the amount of threading of has_unsaved_changes as a parameter into method calls. (thinking about it, that's more true if the boolean lives in View than if it lives in Editor; conceptually you can make a case for either, so i'd go with whichever is less code)

rust/rope/src/engine.rs
raphlinus: agreed, "compare" could be lots of things. I'd go for "is_equivalent_revision" to go with [Rust naming style](https://aturon.github.io/style/naming.html#predicates).

rust/rope/src/engine.rs
```diff
@@ -222,6 +234,18 @@ impl Engine {
...
+        let base_subset = self.find_rev(base_rev).map(|rev_index| self.get_subset_from_index(rev_index));
+        let other_subset = self.find_rev(other_rev).map(|rev_index| self.get_subset_from_index(rev_index));
+
+        if let Some(base_subset) = base_subset {
```
raphlinus: Sweet. You could write this as `base_subset.is_some() && base_subset == other_subset`.

rust/core-lib/src/editor.rs
cmyr: If it goes in view it has to go in _all_ views?

raphlinus: @cmyr Yes, but keep in mind it needs to get broadcast to all views anyway; the iteration over all views to set it doesn't scare me.

cmyr: @raphlinus what about it going in `TabCtx`?

cbrewster:
Should be ready for review

raphlinus:
Excellent work, this is exactly what I was hoping! Glad to see people start to dive into the revision stuff, it's definitely one of the trickier areas of the code.

Issue #188, opened 2017-04-12
Multiple Cursors


cbrewster:
Hi there!
I was wondering if there are any plans to support multiple cursors? I find it apart of my workflow with the text editors I use. 

If this is something you think would be useful for xi, I would be willing to help with the implementation.

raphlinus:
Hi! Multiple cursors are high on my list of things I'd like to see in xi. It also makes sense to work on them early, as selections are fundamental to many editing operations, so the more other stuff gets built, the harder it would be to retrofit them later.

Here are some thoughts about what the implementation would take. First, the existing selection state (`sel_start` and `sel_end` in `View` + `Edit::col`) obviously needs to be changed into a data structure representing multiple selections. Conceptually, this is just a `Vec<Selection>` where a selection contains start, end, col, and upstream/downstream (the latter specifies whether the cursor should be drawn at the end of the previous line or the beginning of the next when there's a line break not caused by a newline or space). I'm split on whether this should be a plain Vec, or a more sophisticated rope-based implementation. One use case to consider is a large document, select all, split selection into lines. But I'm ok with doing it simply first and optimizing it maybe later. (nb, in [Sublime API](https://www.sublimetext.com/docs/3/api_reference.html#sublime.Selection) terminology, the collection is called a "Selection" and each element is a "Range", maybe we should adopt that).

In any case, the "set of selections" struct needs methods for adding a selection (coalescing as needed), binary search, iteration, etc. It's a fairly simple stand-alone data structure.

Then, a large class of editing commands become an iteration over all selections. I'd like to see some reduction in code repetition in the movement methods (see #159) specifically, because at heart a movement is a function from a selection to a cursor position, so a generic move_selection method just iterates over that function, collecting the results.

A warning, some interactions are probably going to be very hard. In particular, input methods with composing regions interact in complex ways with multiple cursors. The way Sublime and Atom (at least) handle it is to display one copy of the composing region for each cursor. Duplicating that behavior will almost certainly require making the front-end <-> core protocol deal with composing regions explicitly. This is not simple, but it is something that makes sense for me to work on (in my day job, I'm also currently designing an IME protocol for Fuchsia, so this is relevant). But because of the difficulty, it might make sense to partially break this scenario temporarily (maybe downgrading multiple selections to one when a non-empty composing region is set, and then letting the front-end deal with it).

I hope these notes are interesting (they're useful to myself, for thinking about what I want to do, if nothing else).

cmyr:
[...]

PR #211, opened 2017-04-19
Add Selection and SelRange data structures

raphlinus:
These types will hold multiple-cursor selections (#188). So far,
nothing in place to actually use these types, but eventually they will
replace the single selection currently in View.

### DIFF
--- a/rust/core-lib/src/selection.rs
+++ b/rust/core-lib/src/selection.rs
@@ -1,0 +1,309 @@
+// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/selection.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+        match self.regions.binary_search_by(|r| r.max().cmp(&offset)) {
+            Ok(ix) => ix,
+            Err(ix) => ix,
+        }
+    }
+
+    /// Add a region to the selection. This method implements merging logic.
+    ///
+    /// Two non-caret regions merge if their interiors intersect; merely
```
cmyr: to clarify some terms:

A 'caret region' is a size-zero region, e.g. a cursor? A non-caret region is a size > 0 region, e.g. a selection?

rust/core-lib/src/selection.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+        if ix == end_ix {
+            self.regions.insert(ix, region);
+        } else {
+            self.regions[ix] = region;
+            remove_n_at(&mut self.regions, ix + 1, end_ix - ix - 1);
+        }
+    }
+
+    /// Gets a slice of regions that intersect the given range. Regions that
```
cmyr: Nit: maybe "Gets **the** slice of regions that intersect the given range."

PR #217, opened 2017-04-21
Fix re-opening file opens empty buffer

rkusa:
Fixes the following bug:

1. Open file X
2. Close file X
3. Open file X again

Results in empty document instead of having file X open again.

Reason is that filenames are not remove from `open_files` when closing all views to a filename.

This PR fixes this behaviour.

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -187,14 +187,18 @@ impl<W: Write + Send + 'static> Tabs<W> {
     
     fn close_view(&mut self, view_id: &str) {
         let buf_id = self.views.remove(view_id).expect("missing buffer id when closing view");
-        let has_views = {
+        let (has_views, path) = {
             let editor = self.buffers.get(&buf_id).expect("missing editor when closing view");
-            editor.lock().unwrap().remove_view(view_id);
-            editor.lock().unwrap().has_views()
+            let mut editor = editor.lock().unwrap();
+            editor.remove_view(view_id);
+            (editor.has_views(), editor.get_path().map(PathBuf::from))
         };
 
         if !has_views {
             self.buffers.remove(&buf_id);
+            if let Some(path) = path {
+                self.open_files.remove(&path);
+            }
         }
     }
 


rkusa:
Thanks for the review!

I am also not entirely happy with copying the value. The only other solutions I found were to move the `open_files.remove` into the `let has_views = {}` block

```rust
let buf_id = self.views.remove(view_id).expect("missing buffer id when closing view");
let has_views = {
    let editor = self.buffers.get(&buf_id).expect("missing editor when closing view");
    let mut editor = editor.lock().unwrap();
    editor.remove_view(view_id);
    let has_views = editor.has_views();
    if !has_views {
      if let Some(path) = editor.get_path() {
          self.open_files.remove(path);
      }
    }
    has_views
};

if !has_views {
    self.buffers.remove(&buf_id);
}
```

or into the `if !has_views {}` block. (basically your suggestion `if let Some(path) = editor.lock().unwrap().get_path()`, but we have to retrieve `editor` first)

```rust
let buf_id = self.views.remove(view_id).expect("missing buffer id when closing view");
let has_views = {
    let editor = self.buffers.get(&buf_id).expect("missing editor when closing view");
    let mut editor = editor.lock().unwrap();
    editor.remove_view(view_id);
    editor.has_views()
};

if !has_views {
    self.buffers.remove(&buf_id);
    if let Some(path) = self.buffers.get(&buf_id).unwrap().lock().unwrap().get_path() {
        self.open_files.remove(path);
    }
}
```

I kind of dislike both, but still happy to adjust the PR. Or maybe you have even a better idea?

cmyr:
yea I remember this being a bit of a headache, which is why we're using the block in the first place. Lemme fire up an editor and play around for a sec. :)

rkusa:
Is cloning the Editor better then copying the path? If so:

```rust
let editor = {
    let editor = self.buffers.get(&buf_id).expect("missing editor when closing view");
    editor.clone()
};
```

So now I am letting you play ;-)

cmyr:
Okay, I'm back to preferring your original solution. :)

We don't really need to worry about the copy as a performance issue (closing views happens rarely), it's just something I'm trying to train myself to treat as codesmell, in Rust.

raphlinus:
Thanks, appreciate the fix!

PR #218, opened 2017-04-22
Do view rendering using new multi-selection

raphlinus:
This is part of a migration to the new multi-selection approach. The
old cursor/selection state (which is still what's used in edit
operations) is copied over (when dirty) to the new format. Then, all
rendering is done with the new format.

Most of the logic upstream affinity is in place, but it will require
some small changes. I suggest we get back to that when we can generate
upstream positions from mouse clicks.

Another step towards #188

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -148,10 +148,9 @@ impl View {
         for region in self.selection.regions_in_range(start_pos, pos) {
             // cursor
             let c = region.end;
-            let is_upstream = region.affinity == Affinity::Upstream;
             if (c > start_pos && c < pos) ||
-                (!is_upstream && c == start_pos) ||
-                (is_upstream && c == pos) ||
+                (!region.is_upstream() && c == start_pos) ||
+                (region.is_upstream() && c == pos) ||
                 (c == pos && c == text.len() && self.line_of_offset(text, c) == line_num)
             {
                 cursors.push(c - start_pos);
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/view.rs
```diff
@@ -126,32 +143,34 @@ impl View {
...
-                };
-                let sel_end_ix = if line_num == sel_max_line {
-                    self.sel_max() - start_pos
-                } else {
-                    l_str.len()
-                };
+        for region in self.selection.regions_in_range(start_pos, pos) {
+            // cursor
+            let c = region.end;
+            let is_upstream = region.affinity == Affinity::Upstream;
```
cmyr: nit: I would just add a `is_upstream()` method to `SelRegion`.

rust/core-lib/src/view.rs
```diff
@@ -126,32 +143,34 @@ impl View {
...
+                (c == pos && pos == text.len())
+            {
+                if cursors.is_none() {
+                    cursors = Some(Vec::new());
+                }
+                // TODO: make this handle upstream case (should be prev line)
+                // TODO: we really need utf-8 byte offsets; if the function starts
+                // using different units to measure "columns" we need to adapt.
+                let (cursor_line, cursor_col) = self.offset_to_line_col(text, c);
+                // this test is needed for previous line when at end of text
```
cmyr: this comment is unclear.

rust/core-lib/src/view.rs
raphlinus: Done, agreed.

rust/core-lib/src/view.rs
raphlinus: "was unclear" :)

PR #219, opened 2017-04-22
Improve cursor position handling for plugin edits

rkusa:
I was playing around with #214 (@cmyr good job on that!) and found various cases where applying the plugin edit (autopair of braces) moves the cursor back to the edit position instead of keeping it where I have moved the cursor after the edit.

I have prepared a PR that tries to improve this behaviour and I hope I am not stealing @cmyr's work (because he already stated in code:`//FIXME: this needs to take account of potential changes between this rev and head.`) 😉

Behaviour with this PR is (example has artificial delay for the plugin communication):

In words:

- When not moving cursor, cursor is moved before the edit
- When moving the cursor to the left of the edit, the cursor stays where it is
- When moving the cursor to the right of the edit, the cursor stays where it is (and is not shifted to the left)
- When typing after the the plugin edit is triggered, the cursor properly stays before the edit once the edit arrives

(It is possible that I have missed some cases)

I actually also wanted to provide unit tests, but stopped at some point where I was still struggling to create a testable `Editor` instance. But making the `Editor` more test-friendly is maybe something for another issue/PR.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -256,12 +256,15 @@ impl<W: Write + Send + 'static> Editor<W> {
         let text = Rope::from(&edit.text);
         let rev_len = self.engine.get_rev(edit.rev as usize).unwrap().len();
         let delta = Delta::simple_edit(interval, text, rev_len);
+        let prev_head_rev_id = self.engine.get_head_rev_id();
         //self.engine.edit_rev(0x100000, undo_group, edit.rev as usize, delta);
         self.engine.edit_rev(edit.priority as usize, undo_group, edit.rev as usize, delta);
         self.text = self.engine.get_head();
-        if edit.after_cursor {
-            //FIXME: this needs to take account of potential changes between this rev and head.
-             self.new_cursor = Some((edit.start as usize, edit.end as usize));
+
+        // adjust cursor position so that the cursor is not moved by the plugin edit
+        let (changed_interval, _) = self.engine.delta_rev_head(prev_head_rev_id).summary();
+        if edit.after_cursor && (changed_interval.start() as usize) == self.view.sel_end {
+            self.new_cursor = Some((self.view.sel_start, self.view.sel_end));
         }
 
         self.commit_delta(self_ref, Some(&edit.author));


rust/core-lib/src/editor.rs
```diff
@@ -256,12 +256,15 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+        let prev_head_rev_id = self.engine.get_head_rev_id();
         //self.engine.edit_rev(0x100000, undo_group, edit.rev as usize, delta);
         self.engine.edit_rev(edit.priority as usize, undo_group, edit.rev as usize, delta);
         self.text = self.engine.get_head();
-        if edit.after_cursor {
-            //FIXME: this needs to take account of potential changes between this rev and head.
-             self.new_cursor = Some((edit.start as usize, edit.end as usize));
+
+        // adjust cursor position so that the cursor is not moved by the plugin edit
+        let (changed_interval, _) = self.engine.delta_rev_head(prev_head_rev_id).summary();
```
rkusa: At this line I'd like some feedback, because there is maybe a better way of retrieving the data I am looking for. I'll try to explain what I want.

Lets consider the following scenario (cursor is represented by | ) :

1. `(|` add opening parenthesis which will trigger the plugin
2. `(asdf|` continue writing
3. `(asdf|)` closing parenthesis arrives from the plugin

In this case `edit.start` will be `1`, but I am already at column `5`. `self.engine.edit_rev` properly integrates the changes, taking my intermediate edits into account. However, for the if statement below, I need the actual edit interval of the plugin edit for the current head revision. This is why I retrieve the delta between the previous and newly created revision head here.
In other words: The edit is at interval `(1, 1)`, but it is in the end applied at `(5, 5)`, and `(5, 5)` is what I need to know.

This is probably hard to follow, so if it is impossible to understand please tell me and I'll try to rephrase it.

raphlinus:
I'm merging this because it improves behavior, but I think it's still heuristic rather than generally correct.

Here's my medium-term plan. I believe the best way to do this is to have the OT interaction between edits and cursors automatically update the cursor, without this function having to explicitly set a cursor. That logic is nontrivial, as inserting text at the cursor can end up with the cursor either before or after the inserted text, depending on priority. If you look at the existing logic (in `View::after_edit`), there's no hope of getting it right, as too much information is lost - among other things, it's using a summary of the edit, rather than the fine-grained edit. The distinction between the two will become very important when there are multiple disjoint inserts.

That OT will need to be redone to handle #188. As part of that work, I'll (by necessity) audit all places in the code that are setting the cursor. My expectation and hope is that nearly all of them (including edits from plugins) will be handled with the OT; the ones that require explicit update will be a lot more painful in the multi sel case.

Thanks for digging into this! You've certainly touched on one of the trickier aspects of the code.

PR #223, opened 2017-04-23
spellcheck sample plugin

cmyr:
This is a little sample plugin (and some minor improvements to the python lib in support of it) which illustrates how to set spans.

sprelling


python/xi_plugin/style.py
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+"""Provides convenience methods for styling text."""
+
+BOLD = 1
+UNDERLINE = 2
+ITALIC = 4
+
+
+def color_for_rgb_float(red, green, blue):
```
raphlinus: If I were writing this, I'd add `, alpha = 1`, and map it.

cmyr: +1 didn't realize we were ever using alpha values.

Issue #224, opened 2017-04-24
simple find / replace


cmyr:
~This should be case insensitive & unicode aware. It should be reasonably performant when dealing with ASCII.~

Reusing this issue as a tracking issue for shipping actual functioning find & replace. This is very close, thanks to @raphlinus and @rkusa. Current status:

- as of #226, basic case insensitive search now exists in xi-core.
- with of #286 (WIP), we have rpc methods wired up for find next / find prev.

next steps:
- [ ] basic UI for find/replace in xi-mac. This should support find next / find prev, replace current, replace all.
- [ ] implement replace RPC. This should act on the current selection.

It may make sense to just implement find UI first, and treat replace as a separate work item.

scholtzan:
This can also be closed now :D

PR #226, opened 2017-04-24
Find substring in rope

raphlinus:
Add string-find functions for rope
    
This patch implements both case-insensitive and case-sensitive
versions of the "find" primitive.
    
There are a number of optimizations that can be done, some code
duplication, and the API for "step" should probably be improved, but
this could be used to implement a usable find feature.
    
Progress towards #224

### DIFF
--- a/rust/rope/src/find.rs
+++ b/rust/rope/src/find.rs
@@ -1,0 +1,335 @@
+// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Implementation of string finding in ropes.
+
+use rope::RopeInfo;
+use tree::Cursor;
+
+use memchr::{memchr, memchr2, memchr3};
+
+pub enum FindResult {
+    Found(usize),
+    NotFound,
+    TryAgain,
+}
+
+/// Case-sensitive find (just does byte comparison, no Unicode logic).
+///
+/// Can panic if `pat` is empty.
+pub fn find(cursor: &mut Cursor<RopeInfo>, pat: &str) -> Option<usize> {
+    loop {
+        match find_step(cursor, pat) {
+            FindResult::Found(pos) => return Some(pos),
+            FindResult::NotFound => return None,
+            _ => (),
+        }
+    }
+}
+
+/// Case-insensitive find, using Rust's `to_lowercase` as the basis
+/// for comparison.
+///
+/// Can panic if `pat` is empty.
+
+// Note: one limitation of this method is that 'k' as the first
+// character in `pat` will not match against U+212A (kelvin). Seems
+// low-frequency enough, but should be fixed eventually.
+pub fn find_casei(cursor: &mut Cursor<RopeInfo>, pat: &str) -> Option<usize> {
+    let pat_lower = pat.to_lowercase();
+    let b = pat_lower.as_bytes()[0];
+    if b == b'i' {
+        loop {
+            match find_ascii_i_step(cursor, &pat_lower) {
+                FindResult::Found(pos) => return Some(pos),
+                FindResult::NotFound => return None,
+                _ => (),
+            }
+        }
+    } else if b >= b'a' && b <= b'z' {
+        loop {
+            match find_ascii_alpha_step(cursor, &pat_lower) {
+                FindResult::Found(pos) => return Some(pos),
+                FindResult::NotFound => return None,
+                _ => (),
+            }
+        }
+    } else if b < 0x80 {
+        loop {
+            match find_ascii_step(cursor, &pat_lower) {
+                FindResult::Found(pos) => return Some(pos),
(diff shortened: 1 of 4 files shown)


rust/rope/src/find.rs
```diff
@@ -0,0 +1,335 @@
...
+
+/// Case-sensitive find (just does byte comparison, no Unicode logic).
+///
+/// Can panic if `pat` is empty.
+pub fn find(cursor: &mut Cursor<RopeInfo>, pat: &str) -> Option<usize> {
+    loop {
+        match find_step(cursor, pat) {
+            FindResult::Found(pos) => return Some(pos),
+            FindResult::NotFound => return None,
+            _ => (),
```
cmyr: given that this is a single case, handling it explicitly would make the control flow clearer.

rust/rope/src/find.rs
```diff
@@ -0,0 +1,335 @@
...
+// limitations under the License.
+
+//! Implementation of string finding in ropes.
+
+use rope::RopeInfo;
+use tree::Cursor;
+
+use memchr::{memchr, memchr2, memchr3};
+
+pub enum FindResult {
```
cmyr: Is this public API? I don't see it used in any of the public functions.

rust/rope/src/find.rs
```diff
@@ -0,0 +1,335 @@
...
+    Found(usize),
+    NotFound,
+    TryAgain,
+}
+
+/// Case-sensitive find (just does byte comparison, no Unicode logic).
+///
+/// Can panic if `pat` is empty.
+pub fn find(cursor: &mut Cursor<RopeInfo>, pat: &str) -> Option<usize> {
```
cmyr: To make sure I understand how the API would be used: the caller would pass in a cursor and a pattern, and this function will return either the _first_ match or None. 

To implement findall the client would just call repeatedly until the result is `None`. (the cursor can be reused, because it is kept up to date during each find step). 

If I'm not missing anything, then it makes sense then that we would just add a public `findall` call here, which could could update results asynchronously.


Also: maybe find should also accept an optional `upto_offset` argument, for things like "find in selection".

rust/rope/src/find.rs
```diff
@@ -0,0 +1,335 @@
...
+                FindResult::NotFound => return None,
+                _ => (),
+            }
+        }
+    }
+}
+
+/// Case-sensitive find (just does byte comparison, no Unicode logic).
+///
+/// This method is like [`find`] but does a bounded amount of work.
```
cmyr: To have rustdoc generate the links, this should be (I believe) ``[`find`][find]``.

rust/rope/src/find.rs
raphlinus: It is now.

rust/rope/src/find.rs
raphlinus: I think your understanding is basically correct, except I'm not sure what a public `findall` would do; as it is, repeated calls to `find` behave a lot like an iterator, which seems like a nice way to expose that functionality.

I _could_ add an `upto_offset` argument, but there are two other ways to achieve the same functionality. First, you could substring the rope before creating the cursor (this does require a little subtlety about not slicing a codepoint in half, but is doable and efficient). Second, you could call `find_progress` and then discard the result if it's past your range.

rust/rope/src/find.rs
raphlinus: +1

rust/rope/src/find.rs
raphlinus: You also have to supply a link ref, but I figured it out.

raphlinus:
Another thought: to upgrade the case folding, we should use Simon Sapin's [rust-caseless](https://github.com/SimonSapin/rust-caseless). This wouldn't actually be a massive change to the existing code.

* The case-folded pattern should probably be Vec<char> instead of a utf-8 encoded string, to avoid repeatedly decoding it.
* We can easily impl `Iterator<Item=char>` for `Cursor<RopeInfo>`, at which point the methods in the `Caseless` trait would Just Work™.
* The tricky part is no doubt the scanner. For default (non-normalizing) case folding, this is not too bad. Possibly the trickiest case is 's', for which there are 6 unique bytes such that it's the first byte of the utf-8 encoding of a codepoint that case-folds to 's' ("sSßſẞﬅﬆ", but the last two have the same initial utf-8 byte). Probably we want to use jetscii to handle this kind of thing.
* For normalizing case-folding, things are much trickier. It's probably possible to do the same sort of analysis, but the sets will be quite a bit bigger (pretty much all of 'a-z' will pull in most of the extended Latin ranges, and this will likely blow the 8-byte limit of jetscii). This will also slow things down, as the scanner will stop at ~every accented character. An initial pass would just have the scanner return `Some(0)` - this is valid but very slow. If we're doing this, it's probably better to work on codepoints rather than the initial byte.
* [Difficulty analysis](https://github.com/google/xi-editor/blob/master/doc/rope_science/rope_science_02.md#difficulty) will help a _lot._ Aside from ASCII of course, I think there are two especially relevant difficulty levels:
  * A set of codepoints s.t. any string formed from those code points is invariant under nfc
  * A set of codepoints, [A-Z] excluded, invariant under case folding

Ping @SimonSapin in case he has thoughts on this.

SimonSapin:
Sorry, I don’t have more to add unless you have a more specific question for me.

cmyr:
@raphlinus is this good to go? As per our current "merge-first, integrate second" approach, I'd be happy to see this in master. It'd also be a bit of extra motivation to wire up UI.

raphlinus:
@cmyr I'd say yes. Obviously there are lots of improvements, but I think it's at something of a local optimimum, where doing _real_ Unicode case folding (for example) would be a ton of extra work. I also chatted briefly with burntsushi about the approach on IRC for additional validation. Will merge.

PR #227, opened 2017-04-24
Add rope science 11 doc: practical syntax highlighting

raphlinus:
This document goes into a fair amount of detail about how a plugin can
efficiently increment transducers (useful in particular for syntax
highlighting) in an incremental way.

### DIFF
--- a/doc/rope_science/rope_science_11.md
+++ b/doc/rope_science/rope_science_11.md
@@ -1,0 +1,363 @@
+# Rope science, part 11: practical syntax highlighting
+
+23 Apr 2017
+
+In this post, we present an incremental algorithm for syntax
+highlighting. It has very good performance, measured primarily by
+latency but also memory usage and power consumption. It does not
+require a large amount of code, but the analysis is subtle and
+sophisticated. Your favorite code editor would almost certainly
+benefit from adopting it.
+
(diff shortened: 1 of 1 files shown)


raphlinus:
[rendered](https://github.com/google/xi-editor/blob/86081000b7521c772163e18cb9b4ae761b95a6ed/doc/rope_science/rope_science_11.md)

raphlinus:
Any improvements (such as graphs) can be added in a later commit.

raphlinus:
Next time this gets updated, add:

[Chuck Jazdzewski](http://removingalldoubt.com/) commented:

Interesting. This describes, almost exactly, how syntax highlighting was implemented in Turbo Pascal 7.0 in 1993. We called the "frontier" the high-water mark. We stored a scanner state using one byte per line which included such things as the comment form the scanner was in. We also kept the "invalidated" states as described.

I also demonstrated a technique that would allow recalculating the states without rescanning in certain cases as the end-line state can contain enough information to determine what the state of the line would be given a change in the previous state. This was too difficult to maintain over time so it was never implemented.

raphlinus:
Next time this gets updated, add link to https://gist.github.com/marijnh/f29b71c6f751122d53eddd2e13a87e7f

PR #229, opened 2017-04-25
Fixed out-of-bounds in Editor::cursor_start

### LINKED ISSUE #228: crash on `MoveToBeginningOfParagraph` event
This came up while testing https://github.com/google/xi-mac/pull/28. I assumed it was a result of either 388174e or 7614924 but I played around quite a bit and apparently it's older.

Easy to reproduce: open a new document, select-all, and fire `move_to_begnning_of_paragraph`. (ctrl+A in xi-mac). 

The crash is a subtraction overflow in `Editor::cursor_start`, where view.sel_min() is 0 and the cursor_col is > 0. 

A quick fix would be to just use `checked_sub`, but I'd want to look at underlying behaviour first. I'll get around to this if nobody beats me to it. 

@rkusa if you're up soon and looking for a bite-sized one this might be fun?

I'm starting to warm up to more aggressive testing.

cmyr:
The function was not accounting for the fact that the cursor could be at
the beginning or end of the selection region, and that `cursor_col` was
set based on the cursor position. If the cursor was at the end of the
selection, the subtraction would over(under)flow.

This fix first moves the cursor to the `sel_min`, and then calls the
existing function for moving to the start of the current line.

- fixes #228

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -188,6 +188,10 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.add_delta(sel_interval, Rope::from(s), new_cursor, new_cursor);
     }
 
+    /// Sets the position of the cursor to `offset`, as part of an edit operation.
+    /// If this cursor position's horizontal component was chosen implicitly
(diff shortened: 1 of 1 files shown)


raphlinus:
Good catch!

PR #230, opened 2017-04-25
Soft spans

rkusa:
This PR implements soft spans (an idea of @raphlinus). Soft spans are maybe best shown with an example of an actual use case:

In this example I am not using arrow keys to move the cursor over the closing parenthesis, I am inserting an actual closing parenthesis instead

Here are some characteristics I've implemented:

- if a character is inserted before a soft span and the character matches with the first character of the soft span, the cursor is shifted over this character (implementation-wise the existing value of the soft span is overridden)
- if more than one character is inserted (e.g. using paste), it works according the point above, but shifts the cursor for as long as the input string matches the soft span string
- if the cursor is moved (arrow keys, mouse click, ...), all soft spans before the cursor are removed
- if text is deleted, soft spans are shifted accordingly

The implementation of soft spans is similar to style spans, with the difference that the editor keeps the soft spans in place (instead of relying on the plugin to handle the lifetime of the soft span).

However, merging asynchronous inserts that come in late, is not implemented yet, e.g.:

While adding soft spans at the correct position is easy (since it is already implemented for style spans), as a user I would expect the additional plugin-inserted closing parenthesis to not show up at all.

This brings me to the main open question of this PR: How and where to implement the merging of soft spans? I kind of have the feeling that something like this should actual go into `engine.rs`?

A quick'n'dirty branch of xi-mac for testing purposes can be found [here](https://github.com/rkusa/xi-mac/tree/soft_spans) (I have added an entry into the Debug menu).

_(commits will be squashed once the PR is ready to go)_

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -73,6 +73,7 @@ pub struct Editor<W: Write> {
     scroll_to: Option<usize>,
 
     style_spans: Spans<Style>,
+    soft_spans: Spans<()>,
     tab_ctx: TabCtx<W>,
     plugins: Vec<PluginRef<W>>,
     revs_in_flight: usize,
@@ -131,6 +132,7 @@ impl<W: Write + Send + 'static> Editor<W> {
             new_cursor: None,
             scroll_to: Some(0),
             style_spans: Spans::default(),
+            soft_spans: Spans::default(),
             tab_ctx: tab_ctx,
             plugins: Vec::new(),
             revs_in_flight: 0,
@@ -183,8 +185,33 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
     fn insert(&mut self, s: &str) {
-        let sel_interval = Interval::new_closed_open(self.view.sel_min(), self.view.sel_max());
-        let new_cursor = self.view.sel_min() + s.len();
+        let start = self.view.sel_min();
+        let mut end = self.view.sel_max();
+        let new_cursor = start + s.len();
+
+        // process soft spans in front of the cursor
+        let iv_ahead = Interval::new_closed_open(start, self.soft_spans.len());
+        let soft_spans = self.soft_spans.subseq(iv_ahead);
+        if let Some((interval, _)) = soft_spans.iter().next() {
+            // only interested in soft spans that are directly in front of the cursor
+            if interval.start() == 0 {
+                let mut cursor = Cursor::new(&self.text, start);
+                let are_eql = compare_cursor_str(&mut cursor, s);
+
+                // are_eql is checked for 1 codepoint long inserts and cursor position is checked
+                // to also match longer inserts partially
+                if are_eql || cursor.pos() > start + 1 {
+                    self.soft_spans.clear_left(cursor.pos());
+
+                    // update selection interval to replace matching part of soft span
+                    end += cursor.pos() - start;
+                } else {
+                    self.soft_spans.shift_right(s.len());
+                }
+            }
+        }
+
+        let sel_interval = Interval::new_closed_open(start, end);
         self.add_delta(sel_interval, Rope::from(s), new_cursor, new_cursor);
     }
 
@@ -200,6 +227,9 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
         self.view.scroll_to_cursor(&self.text);
         self.view.set_old_sel_dirty();
+
+        // remove soft spans before cursor
+        self.soft_spans.clear_left(self.view.sel_start);
     }
 
     // Apply the delta to the buffer, and store the new cursor so that it gets
@@ -342,7 +372,7 @@ impl<W: Write + Send + 'static> Editor<W> {
 
     // render if needed, sending to ui
     pub fn render(&mut self) {
-        self.view.render_if_dirty(&self.text, &self.tab_ctx, &self.style_spans);
+        self.view.render_if_dirty(&self.text, &self.tab_ctx, &self.style_spans, &self.soft_spans);
         if let Some(scrollto) = self.scroll_to {
             let (line, col) = self.view.offset_to_line_col(&self.text, scrollto);
             self.tab_ctx.scroll_to(&self.view.view_id, line, col);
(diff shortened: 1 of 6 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -73,6 +73,7 @@ pub struct Editor<W: Write> {
     scroll_to: Option<usize>,
 
     style_spans: Spans<Style>,
+    soft_spans: Spans<()>,
```
rkusa: While currently empty, `()` could be replaced with a bool or enum once weak spans are implemented

rust/core-lib/src/editor.rs
```diff
@@ -183,8 +185,40 @@ impl<W: Write + Send + 'static> Editor<W> {
...
     fn insert(&mut self, s: &str) {
-        let sel_interval = Interval::new_closed_open(self.view.sel_min(), self.view.sel_max());
-        let new_cursor = self.view.sel_min() + s.len();
+        let start = self.view.sel_min();
+        let mut end = self.view.sel_max();
+        let new_cursor = start + s.len();
+
+        // process soft spans in front of the cursor
+        let iv_ahead = Interval::new_closed_open(start, self.soft_spans.len());
+        let soft_spans = self.soft_spans.subseq(iv_ahead);
```
rkusa: Would using a `Cursor` here better?

rust/core-lib/src/editor.rs
```diff
@@ -183,8 +185,40 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+            // only interested in soft spans that are directly in front of the cursor
+            if interval.start() != 0 {
+                break
+            }
+
+            let mut cursor = Cursor::new(&self.text, start);
+            let are_eql = compare_cursor_str(&mut cursor, s);
+
+            // are_eql is checked for 1 codepoint long inserts and cursor position is checked
+            // to also match longer inserts partially
```
rkusa: I have noticed that the codebase has only few inline comments. If this is preferred I am also fine with removing them

rust/core-lib/src/editor.rs
```diff
@@ -767,6 +806,22 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+        let text = Rope::from("soft spans!");
+        let new_cursor = self.view.sel_start;
+
+        let end = new_cursor + text.len();
+        let mut sb = SpansBuilder::new(end);
+        sb.add_span(Interval::new_closed_open(new_cursor, new_cursor + 3), ());
+        sb.add_span(Interval::new_closed_open(new_cursor + 3, end), ());
+        self.soft_spans = sb.build();
+
+        self.add_delta(sel_interval, text, new_cursor, new_cursor);
```
rkusa: I might do something wrong here, because undo does not work properly with the text inserted by this method

rust/core-lib/src/editor.rs
```diff
@@ -1030,3 +1086,16 @@ fn n_spaces(n: usize) -> &'static str {
     assert!(n <= spaces.len());
     &spaces[..n]
 }
+
+use xi_rope::rope::RopeInfo;
+fn compare_cursor_str(cursor: &mut Cursor<RopeInfo>, pat: &str) -> bool {
```
rkusa: This one ist copied from #226 and should be replace once #226 has landed

rust/core-lib/src/styles.rs
```diff
@@ -18,7 +18,7 @@ use std::collections::HashMap;
 
 use serde_json::value::Value;
 
-const N_RESERVED_STYLES: usize = 1;
+const N_RESERVED_STYLES: usize = 2;
```
rkusa: Reserved id `1` for soft spans

rust/rope/src/spans.rs
```diff
@@ -189,6 +189,26 @@ impl<T: Clone + Default> Spans<T> {
...
+        let iv = Interval::new_closed_open(len, self.interval().end());
+        *self = self.subseq(iv);
+    }
+
+    pub fn shift_right(&mut self, len: usize) {
+        let mut b = TreeBuilder::new();
+        b.push(SpansBuilder::new(len).build());
+        self.push_subseq(&mut b, self.interval());
+        *self = b.build();
+    }
```
rkusa: To be honest, I don't completely understand the tree API, yet. While these methods are doing what I am intending to do, I have the feeling that their implementation could somehow be improved a lot?

rust/core-lib/src/editor.rs
cmyr: I think you need to at least call `commit_delta`.

rkusa: Thanks for the tipp! I think `commit_delta` is called at the end of `do_rpc_with_self_ref()`. I did some more testing and was able to reconstruct the undo issue without using soft spans. I am going to create an issue for that

trishume:
This is neat. I imagine this is supposed to work like the equivalent feature in XCode, so I tested XCode's behaviour and found that any movement off of the line the soft close paren is on, the close paren becomes real (and XCode has a cool flash animation that calls attention to this).

I think this behaviour makes more sense than only removing soft spans before the cursor. Even if you move up a line, I think the spans on the line you are at should become real, at least for paren matching.

Another interesting thing I noticed is that XCode only creates the soft close paren after you type a character following the opening paren. I'm not sure the reason behind this, but one thing that often annoys me about bracket matching in other text editors is when I do some editing operation that messes up paren nesting, so I go to insert a missing open paren, but then the editor adds a close paren for it! The XCode behaviour avoids doing this, which is cool.

I also tested Sublime's behaviour and it seems that typing a `)` with your cursor before any `)` always just moves past it. Typing an opening paren in the middle of other characters doesn't insert a close paren, but typing one before whitespace in the middle of a line does.

raphlinus:
My feeling on this is that we're not quite yet at a state where this would be ready to merge. That said, this is great as a sample implementation, and there's good discussion in the comments. Do people agree with closing this particular PR and opening an issue for "soft spans" which points to this?

rkusa:
I'd like to update this PR to work in async scenarios, thus, moving the logic from the editor into the engine. However, before investing to much time into the engine, I'd like to defer it for some weeks an await the changes to the engien/OT that might come.

That being said, I am fine with closing it and moving the discussion into an issue.

PR #234, opened 2017-04-26
Add gxi to front-ends

bvinc:
Hello!  Please add and check out my pure rust GTK+ front-end that I've been working on.


bvinc:
I signed it!

raphlinus:
Very cool! We might consider breaking line_cache (in particular) out into its own crate, so that other front-ends written in Rust can share the logic. I might have some PR's for you in coming weeks.

PR #242, opened 2017-04-29
Add toggle_selection gesture; new movement

raphlinus:
Adding a new "gesture" RPC call, and wiring it up to
"toggle_selection" logic.

Also start new "movement" module, and start wiring existing movement
commands to use it. As of this patch, only "left" and "right" are
adapted to be multi-selection.

Progress toward #188

### DIFF
--- a/rust/core-lib/src/movement.rs
+++ b/rust/core-lib/src/movement.rs
@@ -1,0 +1,90 @@
+// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Representation and calculation of movement within a view.
+
+use selection::{Affinity, HorizPos, Selection, SelRegion};
+use view::View;
+use xi_rope::rope::Rope;
+
+/// The specification of a movement.
+#[derive(Clone, Copy)]
+pub enum Movement {
+    /// Move to the left by one grapheme cluster.
+    Left,
+    /// Move to the right by one grapheme cluster.
+    Right,
+}
+
+/// Calculate a horizontal position in the view, based on the offset. Return
+/// value has the same type as `region_movement` for convenience.
+fn calc_horiz(view: &View, text: &Rope, offset: usize) -> (usize, Option<HorizPos>) {
+    let (_line, col) = view.offset_to_line_col(text, offset);
+    (offset, Some(col))
+}
+
+/// Compute the result of movement on one selection region.
+fn region_movement(m: Movement, r: &SelRegion, view: &View, text: &Rope, modify: bool)
+    -> (usize, Option<HorizPos>)
+{
+    match m {
+        Movement::Left => {
+            if r.is_caret() || modify {
+                if let Some(offset) = text.prev_grapheme_offset(r.end) {
+                    calc_horiz(view, text, offset)
+                } else {
+                    (0, None)
+                }
+            } else {
+                calc_horiz(view, text, r.min())
+            }
+        }
+        Movement::Right => {
+            if r.is_caret() || modify {
+                if let Some(offset) = text.next_grapheme_offset(r.end) {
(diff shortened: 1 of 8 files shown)


rust/core-lib/src/view.rs
```diff
@@ -125,8 +126,48 @@ impl View {
...
+    /// If `modify` is `true`, the selections are modified, otherwise the results
+    /// of individual region movements become carets.
+    pub fn do_move(&mut self, text: &Rope, movement: Movement, modify: bool)
+        -> Option<usize>
+    {
+        let new_selection = selection_movement(movement, &self.selection, self, text,
+            modify);
+        self.selection = new_selection;
+        // We somewhat arbitrarily choose the last region for setting the old-style
+        // selection state, and for scrolling it into view if needed. This choice can
```
trishume: I tested and what Sublime does, in order:
- If any cursor is in view, it doesn't scroll. 
- If there is a cursor above the viewport, it scrolls up until a cursor is in view.
- If there is a cursor below the viewport, it scrolls down until a cursor is in view.

I'm not sure I like the arbitrary preference for cursors above to cursors below, even if the one above is further. I think the ideal behaviour might be to scroll to the closest cursor.

rust/core-lib/src/movement.rs
```diff
@@ -0,0 +1,90 @@
...
+    modify: bool) -> Selection
+{
+    let mut result = Selection::new();
+    for r in s.iter() {
+        let (offset, horiz) = region_movement(m, r, view, text, modify);
+        let new_region = SelRegion {
+            start: if modify { r.start } else { offset },
+            end: offset,
+            horiz: horiz,
+            affinity: Affinity::default(),
```
trishume: I tested this in Cocoa, and it seems that while left and right movements reset the affinity to Downstream, up and down movements preserve affinity. Because of the limited number of Cocoa movement shortcuts I know, I don't know if this is a special case for those, or some general rule like if it preserves the horizontal position then the affinity is preserved.

rust/core-lib/src/view.rs
```diff
@@ -125,8 +126,48 @@ impl View {
...
+            self.selection.delete_range(offset, offset);
+            if !self.selection.is_empty() {
+                return;
+            }
+        }
+        let region = SelRegion {
+            start: offset,
+            end: offset,
+            // TODO: might want to set horiz to some meaningful value.
+            horiz: None,
```
trishume: This is interesting. The previous way you calculated a `HorizPos` required access to the text, but the view doesn't seem to have that. Does that mean that the text needs to be passed as a parameter to `toggle_sel` somehow, should the `HorizPos` be passed in, or should calculating a `HorizPos` not rely on the text of the document?

raphlinus: I'm rethinking HorizPos compared with the old cursor_col now. The latter was basically always calculated eagerly. Most operations that set the cursor, other than vertical movement, would set it. My new thinking (which is associated with the fact that it's now an `Option`) is that these operations would reset it, and the first vertical motion would transition it from None->Some.

Computing HorizPos, if we were going to stick with that, does require view and text, but those are not too hard to plumb.

I have another patch on my disk that has most movements implemented. Not sure whether it's best to push it to this branch or do it in another PR. It is, however, still based on computing HorizPos eagerly. I think I want to think on this a little more.

rust/core-lib/src/view.rs
raphlinus: That matches my measurements. I'm not 100% sure it's the best behavior, but it does have being symmetrical going for it.

At this stage in the game, I have to choose a single point. When `sel_start` and `sel_end` are actually deleted out of View, it will be easier to think about the best behavior.

Very likely this method will want to return a range rather than a single point. That would make it fairly easy to replicate the sublime behavior, or do something else equally reasonable.

One thing I'll just throw out there because it's simple and would be easy to implement: keeping the midpoint in view.

rust/core-lib/src/movement.rs
raphlinus: Interesting, and this makes sense. It wouldn't be hard to implement, we just add affinity to the result of `region_movement`, and preserve it on vertical movement.

In fact, I think the logic for preserving affinity would be the same as for preserving HorizPos. Maybe we should make region_movement return `(usize, Option<(Affinity, HorizPos)>` to emphasize these share the same fate.

raphlinus: I've thought about this a little more, and think that "preserving affinity" is not really the right approach. When you go through all the cases, the result of vertical motion may have the same affinity or different. Rather, when setting a cursor, the `View::line_col_to_offset` method (the same one for mouse hit testing) should be augmented to produce an affinity. This will also affect the signature of `region_movement`.

I won't open an issue, I'll consider it part of the multi cursor development track. Thanks for bringing it up though.

PR #243, opened 2017-05-01
Wire up additional movements to new scheme

raphlinus:
Movements now multi-selection capable: word_left, word_right, up,
down, left_of_line, right_of_line, beginning_of_paragraph,
end_of_paragraph, beginning_of_document, end_of_document, page_up,
page_down.

Also fixed a subtle bug in right_of_line (would move down on next to
last line when last line was empty).

Fixed the bug that beginning_of_paragraph was only going to beginning
of visible line.

Also compute HorizPos lazily (see internal commit)

Progress towards #188

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -449,25 +449,11 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
     fn move_up(&mut self, flags: u64) {
-        if (flags & FLAG_SELECT) != 0 {
-            self.modify_selection();
-        }
-
-        let old_offset = self.view.sel_end;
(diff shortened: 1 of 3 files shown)


raphlinus:
This is on top of #242. I'm not sure how the workflow works here (I'm used to Gerrit in which this is straightforward); it might be easiest to examine individual commits within this PR.

chriskrycho:
@raphlinus, you should be able to (re)target the branch used as in #242 itself as the base for the PR, which will isolate the commits exposed in this PR to just those which differ from it. Then you can merge this into that, and that into master, as ready. (I use this flow sometimes for relatively long-running feature branches.)

raphlinus:
@chriskrycho Thanks, just did that.

PR #246, opened 2017-05-01
Merge multi-sel movements patches

raphlinus:
#243 was accidentally merged into the wrong branch. I think this is the way to merge it into master.

Progress towards #188

### DIFF
--- a/rust/core-lib/src/movement.rs
+++ b/rust/core-lib/src/movement.rs
@@ -14,9 +14,13 @@
 
 //! Representation and calculation of movement within a view.
 
+use std::cmp::max;
+
 use selection::{Affinity, HorizPos, Selection, SelRegion};
 use view::View;
-use xi_rope::rope::Rope;
+use word_boundaries::WordCursor;
+use xi_rope::rope::{LinesMetric, Rope};
+use xi_rope::tree::Cursor;
 
 /// The specification of a movement.
 #[derive(Clone, Copy)]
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/movement.rs
```diff
@@ -25,43 +29,158 @@ pub enum Movement {
...
 
-/// Calculate a horizontal position in the view, based on the offset. Return
-/// value has the same type as `region_movement` for convenience.
-fn calc_horiz(view: &View, text: &Rope, offset: usize) -> (usize, Option<HorizPos>) {
-    let (_line, col) = view.offset_to_line_col(text, offset);
-    (offset, Some(col))
+/// Computes the actual desired amount of scrolling (generally slightly
+/// less than the height of the viewport, to allow overlap).
+fn scroll_height(view: &View) -> isize {
+    max(view.scroll_height() as isize - 2, 1)
```
cmyr: is `2` the 'slop' constant?

rust/core-lib/src/movement.rs
```diff
@@ -25,43 +29,158 @@ pub enum Movement {
...
+            let mut cursor = Cursor::new(&text, r.end);
+            let offset = cursor.prev::<LinesMetric>().unwrap_or(0);
+            (offset, None)
+        }
+        Movement::EndOfParagraph => {
+            // Note: TextEdit would start at modify ? r.end : r.max()
+            let mut offset = r.end;
+            let mut cursor = Cursor::new(&text, offset);
+            if let Some(next_para_offset) = cursor.next::<LinesMetric>() {
+                if cursor.is_boundary::<LinesMetric>() {
```
cmyr: to clarify: if `cursor.is_boundary() == false` in this situation we're at the end of the file?

rust/core-lib/src/movement.rs
raphlinus: No, this is something else; it's the overlap (number of lines in common) on successive page scrolls. As slop is increased, this should stay the same. It should become a constant with a comment though.

rust/core-lib/src/movement.rs
raphlinus: Yes.

raphlinus:
I want `flags` to go away, everything should be captured in enum parameters to the movement command.

PR #254, opened 2017-05-05
[WIP] Support multiple views

cbrewster:
Here is the initial approach behind allowing multiple views.

I am not happy with the ways I have to avoid making borrowck angry, so that will need to be cleaned up. I think this should be a good general direction though.

This is also waiting on cursor OT.

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -17,6 +17,7 @@
 use std::collections::BTreeMap;
 use std::io::{self, Read, Write};
 use std::path::{PathBuf, Path};
+use std::fmt::{self, Display, Formatter};
 use std::fs::File;
 use std::sync::{Arc, Mutex};
 use serde_json::value::Value;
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -239,18 +263,18 @@ impl<W: Write + Send + 'static> Tabs<W> {
 }
 
 impl<W: Write> TabCtx<W> {
-    pub fn update_view(&self, view_id: &str, update: &Value) {
+    pub fn update_view(&self, view_id: &ViewIdentifier, update: &Value) {
         self.rpc_peer.send_rpc_notification("update",
             &json!({
-                "view_id": view_id,
+                "view_id": view_id.clone().into_value(),
```
cmyr: maybe we just implement serialize/deserialize for `ViewIdentifier`?

cbrewster: Yeah, will do. Was just slopping it together and didnt want to look at docs for implementing Serialize/Deserialize manually

cbrewster:
I think there is an argument to be made about storing the id vs passing the id to different places. I think the less state we store, the better off we will be in the future. I could see storing the view_id on Editor causing some hard to track down bugs.

cmyr:
yea, I do appreciate that general philosophy. I don't have really strong feelings either way; the main thing that influences me is how we have, so far, done some of this state-stashing (around cursor pos, scrolling, and applying edits) so this sort of seems to fall in line with that. 🤷‍♂️

cbrewster:
Not sure if I will have the time to continue working on this in the near future.

PR #255, opened 2017-05-05
multi-plugin support

cmyr:
Okay, here is a first sketch of how I've been thinking about multi-plugin stuff.

There are a bunch of rough edges and a bunch of stuff that isn't implemented yet, but this at least ready to be looked at.

There is a xi-mac branch [here](https://github.com/cmyr/xi-mac/tree/multi-plug) that adds a menu for activating plugins.

Using the included python sample plugins will require installing the python lib: from `xi-editor/python/xi_plugin/` run `pip3 install .`.

Thoughts:

- I'm okay with, but not in love with, the overall approach. If anyone has thoughts or concerns, now is the time to voice them. 😈 
- The way buffers are shared between the main `Tabs` instance and the `PluginManager` feels a bit gross. I'd like to add a type that manages the buffers, and provides convenience methods for getting a buffer from a view etc.
- This changes the RPC protocol around creating a view; instead of just getting a `view_id` you get `{view_id: str, available_plugins: [str]}`. This feels like a hack.
- **There are some crashes** I'm seeing, mostly when using the bracket-matching plugin. I haven't investigated too much.


rust/core-lib/src/plugins.rs
```diff
@@ -0,0 +1,388 @@
...
+            };
+            let mut plugin_ref = PluginRef(Arc::new(Mutex::new(plugin)));
+            completion(Ok(plugin_ref.clone()));
+            looper.mainloop(|| BufReader::new(child_stdout), &mut plugin_ref);
+        });
+    }
+}
+
+
+type PluginName = String;
```
cbrewster: I wonder if we should use a struct/tuple struct later. At one point it may be nicer to create a `usize` for a Plugin Id (or any other id for that matter) since they are smaller and easier to send around. If we encapsulate it into a struct, it's much easier to adjust the structure later if needed.

rust/core-lib/src/rpc.rs
```diff
@@ -43,6 +43,8 @@ pub enum Request<'a> {
 #[derive(Debug, PartialEq, Eq)]
 pub enum CoreCommand<'a> {
     Edit { view_id: &'a str, edit_command: EditCommand<'a> },
+    /// A command from the client to a plugin.
+    Plugin {  plugin_command: PluginCommand },
```
cbrewster: I wonder if we should have the `plugin_name` here rather than `PluginCommand`, so each variant doesn't need to have it.

cmyr: yea, the way this is structured is the result of me wanting to use serde to handle generating the enums, and it not being easy to do as you described. Ideally `view_id` would be threaded that way as well.

I'm not sure if this is doable and I just haven't done it, or if there needs to be another approach.

rust/core-lib/src/plugins.rs
cmyr: Yea, I like the approach you take in #254 and I'd be happy to go that route.

cmyr:
Agree with the Arc<Mutex stacking, another rough edge I would like to improve.

rust/core-lib/src/plugins.rs
```diff
@@ -0,0 +1,376 @@
...
+    version: String,
+    //scope: PluginScope,
+    // more metadata ...
+    /// path to plugin executable
+    pub exec_path: PathBuf,
+}
+
+impl PluginDescription {
+    fn new<S, P>(name: S, version: S, exec_path: P) -> Self
+        where S: Into<String>, P: Into<PathBuf> {
```
raphlinus: Generally the style I prefer is to put `{` on the next line by itself, same indent as `fn`

rust/core-lib/Cargo.toml
```diff
@@ -11,6 +11,7 @@ serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
 time = "0.1"
+parking_lot = "0.4"
```
raphlinus: I'm actually going to ask this be a separate change, when I look at the dependencies it pulls in I think it's going to cause difficulty for Fuchsia. We should fix that, of course, but it feels pretty orthogonal to the thrust of this PR.

rust/core-lib/src/editor.rs
```diff
@@ -762,13 +757,12 @@ impl<W: Write + Send + 'static> Editor<W> {
...
     pub fn do_rpc(self_ref: &Arc<Mutex<Editor<W>>>, view_id: &str, cmd: EditCommand) -> Option<Value> {
-        self_ref.lock().unwrap().do_rpc_with_self_ref(view_id, cmd, self_ref)
+        self_ref.lock().unwrap().do_rpc_impl(view_id, cmd)
     }
 
-    fn do_rpc_with_self_ref(&mut self, view_id: &str,
-                  cmd: EditCommand,
-                  self_ref: &Arc<Mutex<Editor<W>>>)
-                  -> Option<Value> {
+    //FIXME: what is the relationship between this and `do_rpc`, now?
```
cmyr: we've moved the arc/mutex up a level; instead of wrapping each editor, it wraps all editors. Editors no longer call plugins directly, so they don't need self_ref.

rust/core-lib/src/tabs.rs
```diff
@@ -18,53 +18,183 @@ use std::collections::BTreeMap;
...
+/// `WeakBufferContaienrRef` provides a more ergonomic way of storing a `Weak`
+/// reference to a [`BufferContainer`][BufferContainer].
+///
+/// [BufferContainer]: struct.BufferContainer.html
+pub struct WeakBufferContainerRef<W: Write>(Weak<Mutex<BufferContainer<W>>>);
+
+impl <W:Write>BufferContainer<W> {
+    /// Returns a reference to the `Editor` instance owning `view_id`'s view.
+    ///
+    /// Panics if no buffer is associated with `view_id`.
```
cmyr: Agreed, there's a TODO to this effect in here somewhere. Basically I would like to do proper error handling here, but unwraps felt okay for this as a proof of concept.

rust/core-lib/src/tabs.rs
```diff
@@ -18,53 +18,183 @@ use std::collections::BTreeMap;
...
+            if !editor.has_views() {
+                editor.get_path().map(PathBuf::from)
+            } else {
+            None
+            }
+        };
+
+        if let Some(path) = path_to_remove {
+            let buffer_id = self.lock().views.remove(view_id).unwrap();
+            self.lock().open_files.remove(&path);
```
cmyr: totally. Also not sure it's worth it. from "Differences from the standard library Mutex" section of the  docs for [`parking_lot::Mutex`](https://amanieu.github.io/parking_lot/parking_lot/struct.Mutex.html): 

> - Inline fast path for the uncontended case.

rust/core-lib/Cargo.toml
cmyr: fair.

rust/core-lib/src/tabs.rs
```diff
@@ -364,9 +360,12 @@ impl<W: Write + Send + 'static> Documents<W> {
         self.buffers.lock().editor_for_view_mut(view_id).unwrap().do_rpc(view_id, cmd)
     }
 
+    #[allow(unused_variables)]
```
cbrewster: You can prefix unused variables with `_` instead of using allow here

cmyr: Yea I had it that way first, but these are going to become used in imminent future work. I guess I've generally been using `#[allow()]` mostly to annotate sections of code that are unimplemented? This might not be very idiomatic / great practice, generally.

raphlinus:
A few observations from playing with it. First, it doesn't work with master xi-mac, it looks like the protocol has changed incompatibly.

Second, the syntect plugin looks like it needs an update to track the plugin protocol changes:

plugin response json err: ErrorImpl { code: Message("data did not match any variant of untagged enum UpdateResponse"), line: 0, column: 0 }
This may actually have been previous plugin changes (to add an edit response), not this PR. Also, it works, just prints these errors. In any case, I definitely want to track it.

Third, (and this may be an issue in the way xi-mac does installation), the toplevel .py files for the plugins are copied to Resources, but the xi_plugin subdir is not, so the import of xi_plugin fails.

doc/frontend.md
```diff
@@ -38,14 +38,14 @@ example, the actual interaction on the wire for `new_view` is:
...
-from core: {"id":0,"result":"view-id-1"}
+from core: {"id":0,"result": view-id-1"}
 ```
 
 ## Top-level methods served by back-end
 
 ### new_view
 
-`new_view { "file_path": "path.md"? }` -> `"view-id-1"`
+`new_view { "file_path": "path.md"? }` -> `view-id-1"`
```
raphlinus: I'm guessing you didn't mean to delete this `"`.

doc/frontend.md
```diff
@@ -38,14 +38,14 @@ example, the actual interaction on the wire for `new_view` is:
 
 ```
 to core: {"id":0,"method":"new_view","params":{}}
-from core: {"id":0,"result":"view-id-1"}
+from core: {"id":0,"result": view-id-1"}
```
raphlinus: Don't think this change is intentional.

rust/core-lib/src/plugins.rs
```diff
@@ -0,0 +1,401 @@
...
+/// `Editor` a tx end of an `mpsc::channel`. As plugin updates are generated,
+/// they are sent over this channel to a receiver running in another thread,
+/// which forwards them to interested plugins.
+pub fn start_update_thread<W: Write + Send + 'static>(
+    rx: mpsc::Receiver<(ViewIdentifier, PluginUpdate, usize)>,
+    plugins: Arc<Mutex<PluginManager<W>>>)
+{
+    thread::spawn(move ||{
+        loop {
+            let (view_id, update, undo_group) = rx.recv().unwrap();
```
raphlinus: This seems like it would panic on shutdown, I think breaking out of the loop is better. That said, I wasn't able to observe a panic.

rust/core-lib/src/plugins.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct PluginEdit {
+    pub start: u64,
+    pub end: u64,
+    pub rev: u64,
+    pub text: String,
+    /// the edit priority determines the resolution strategy when merging
+    /// concurrent edits. The highest priority edit will be applied last.
+    pub priority: u64,
+    /// whether the inserted text prefers to be to the right of the cursor.
+    pub after_cursor: bool,
```
raphlinus: This will probably be subsumed by "priority", but I think that can wait until we actually implement that; it will require a bunch of changes.

raphlinus:
Thanks for addressing the issues, and very excited to see this progress.

Merging now, then we'll go from here.

PR #258, opened 2017-05-06
[WIP] Experimenting with mpsc

cbrewster:
This is far from complete, but is just a sample proof of concepts of using mpsc rather than using `Arc<Mutex<...>>`.

This will require a bit of reworking with plugins as `self_ref` can no longer be used.

Using mpsc, each `Editor` will run in its own thread. This means we can fire off messages to an `Editor` without worrying about any operation blocking another `Editor`.

For cases where we need a synchronous response, we can create another mpsc channel that can be sent through the channel itself. The receiving end will send back the value we need.

I did have to make messages own Strings and Paths, I am unsure if we can make it just use references or not. This was just a quick and dirty experiment.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -43,6 +43,14 @@ version = "0.2.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "memchr"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "libc 0.2.22 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "num-traits"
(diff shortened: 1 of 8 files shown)


cbrewster:
This is almost back to original functionality, I need to think about how we handle sending copy/cut contents back synchronously with mpsc

cbrewster:
Here is a case where having each editor in its own thread helps. It is a bit of an extreme case.
Setup:
1. Document with a loooong line
2. Another normal document
3. Rewrap the lines on the long document and scroll down to the middle of the file(this is currently slow)
4. Scroll down in the normal document

Using `Mutex`/single-thread:
slowscroll

Using `mpsc`/multi-threaded:
fastscroll

cmyr:
We've decided not to go this route for now, although this was a very helpful illustration. 🙏

PR #259, opened 2017-05-06
Do transformation of multi-selection after delta

raphlinus:
Cursor movement after edits is now done automatically by
View::after_edit instead of explicitly as part of the edit. The
patch includes logic to transform a selection by a delta.

Work in progress, as a couple of things are still rough: first,
transpose leaves non-caret selections in a funny state; second,
edits by plugins still set the cursor manually. Also, this commit
doesn't change the behavior of any edit operations to be multi
selection aware, just changes the cursor updating logic.

Progress toward #188


rust/core-lib/src/view.rs
```diff
@@ -449,26 +454,24 @@ impl View {
...
-            } else {
-                self.sel_end = iv.start() + new_len;
-            }
-        }
-        self.sel_start = self.sel_end;
+    /// Updates the view after the text has been modified by the given `delta`.
+    /// This method is responsible for updating the cursors, and also for
+    /// recomputing line wraps.
+    ///
+    /// Return value is a locatio
```
cmyr: location? but feels like the line is truncated

rust/core-lib/src/editor.rs
```diff
@@ -207,13 +206,17 @@ impl<W: Write + Send + 'static> Editor<W> {
...
-    // the logic for merging edits into the same undo group. At call time,
-    // self.this_edit_type should be set appropriately.
-    fn add_delta(&mut self, iv: Interval, new: Rope, new_start: usize, new_end: usize) {
+    /// Applies the delta to the text, and updates undo state.
+    ///
+    /// Records the delta into the CRDT engine so that it can be undone. Also contains
+    /// the logic for merging edits into the same undo group. At call time,
+    /// self.this_edit_type should be set appropriately.
+    ///
+    /// This method can be called multiple times. Note that it does not update
```
cmyr: "may be called multiple times" is a little unclear. Multiple times within an update cycle / rpc call?

raphlinus:
If the only changes are the nits, this is a good candidate for squashing on merge.

rust/core-lib/src/view.rs
```diff
@@ -449,26 +454,24 @@ impl View {
...
-        self.sel_start = self.sel_end;
+    /// Updates the view after the text has been modified by the given `delta`.
+    /// This method is responsible for updating the cursors, and also for
+    /// recomputing line wraps.
+    ///
+    /// Return value is a location of a point that should be scrolled into view.
+    pub fn after_edit(&mut self, text: &Rope, delta: &Delta<RopeInfo>, pristine: bool)
+        -> Option<usize>
+    {
         if self.breaks.is_some() {
```
trishume: `if let Some(breaks) = self.breaks.as_mut()`

and then modify the re-wrap line to use `breaks` instead of an unwrap.

cmyr: will wait for a won't-fix from raph before merging

rust/core-lib/src/view.rs
```diff
@@ -487,3 +490,23 @@ fn clamp(x: usize, min: usize, max: usize) -> usize {
         max
     }
 }
+
+/// Computes a new selection based on applying a delta to the old selection.
+///
+/// When new text is inserted at a caret, the new caret can be either before
+/// or after the inserted text, depending on the `after` parameter.
+// Should this move to Selection itself?
```
rkusa: Moving this into `Selection` itself, would make `selection.rs` less self-contained. However, it kind of feels out of place here, so that I would personally prefer it being in Selection itself.

cmyr:
I'm not going to squash b/c i'd like to keep 2365ca2.

Issue #256, opened 2017-05-05
Refactor Engine


trishume:
Possible refactorings below. These were written down while reading the Xi code, I don't necessarily know if these are a good idea yet. I'm writing them down now more to document what confused me in the hopes that we can make it less confusing to new readers.

- [ ] Should `Subset` conceptually represent a set of the ranges (i.e the current `dels` field) or the complement of it. This effects method names and documentation. For example maybe `is_trivial` should be named `is_empty` and `add_deletion` should be `add_range`. This may make it make more sense when representing insertions and deletions in revisions.
- [ ] Have a separate tuple-struct wrapper of `Subset` called `Transform` that conceptually represents the corresponding function. Would use function terminology like `compose` and `apply`. Would make it hard to mix up which `Subset`s represent deletion/insertion ranges and which represent coordinate transforms.
- [ ] Refactor method names on `Subset` in terms of Git terminology. This one may be dumb but it's a commonly known way of thinking of revision graphs. For example `transform_expand` seems analogous to `rebase` or `cherry-pick`.
- [x] `DeltaElement::Copy` needs a doc comment explaining exactly what it is copying from, and how the whole `Delta` representation works. I now think I understand that the `Vec` represents the entire document, which some segments copied from the previous `Rope` and some new nodes inserted. Deletions are represented by the gaps between `Copy` elements.
- [ ] Add checking for coordinate space transforms. This might be getting a little too extreme: Do the previously-mentioned `Transform` wrapper for `Subset`, then add fields for the from and to revision of the coordinate transform. Add a `checked_apply` function or parameters to the normal apply function that checks that the apply function is transforming the right coordinates. If this was Idris I'd do this all with type level revision numbers and check it at compile time, but this is the next best thing. I'm not sure how helpful it would actually be though.
- [x] [Doc comment for `transform_expand`](https://github.com/google/xi-editor/blob/master/rust/rope/src/subset.rs#L148) needs parentheses. I interpreted "not in s1 or in s2" as "not (in s1 or in s2)", but it is actually "(not in s1) or in s2". One of those ambiguities in English that is resolved verbally through cadence and inflection, but is ambiguous in text. I read it in my head wrong and thus was confused.

[...]

PR #260, opened 2017-05-06
Multi-cursor insert and delete

raphlinus:
The basic insert and delete operations now apply to all regions in a
(multi) selection.

Progress towards #188


rust/rope/src/delta.rs
```diff
@@ -341,6 +336,60 @@ impl<'a, N: NodeInfo + 'a> Transformer<'a, N> {
...
+                els: Vec::new(),
+                base_len: base_len,
+            },
+            last_offset: 0,
+        }
+    }
+
+    /// Deletes the given interval. Panics if interval is not properly sorted.
+    pub fn delete(&mut self, interval: Interval) {
+        let (start, end) = interval.start_end();
```
cmyr: Just musing, but I wonder if some sort of 'Range' type would be useful for us.

rust/rope/src/delta.rs
```diff
@@ -341,6 +336,60 @@ impl<'a, N: NodeInfo + 'a> Transformer<'a, N> {
...
+        }
+        if start > self.last_offset {
+            self.delta.els.push(DeltaElement::Copy(self.last_offset, start));
+        }
+        self.last_offset = end;
+    }
+
+    /// Replaces the given interval with the new rope. Panics if interval
+    /// is not properly sorted.
+    pub fn replace(&mut self, interval: Interval, rope: Node<N>) {
```
cmyr: If I were designing this I might have the API accept an `AsRef<str>`, and do the conversion to rope internally.

rust/rope/src/delta.rs
```diff
@@ -341,6 +336,60 @@ impl<'a, N: NodeInfo + 'a> Transformer<'a, N> {
...
+        }
+    }
+
+    /// Deletes the given interval. Panics if interval is not properly sorted.
+    pub fn delete(&mut self, interval: Interval) {
+        let (start, end) = interval.start_end();
+        if start < self.last_offset {
+            panic!("Delta builder: intervals not properly sorted");
+        }
+        if start > self.last_offset {
```
cmyr: It's unclear to me why this isn't `>=`.

rust/rope/src/delta.rs
```diff
@@ -341,6 +336,60 @@ impl<'a, N: NodeInfo + 'a> Transformer<'a, N> {
...
+            last_offset: 0,
+        }
+    }
+
+    /// Deletes the given interval. Panics if interval is not properly sorted.
+    pub fn delete(&mut self, interval: Interval) {
+        let (start, end) = interval.start_end();
+        if start < self.last_offset {
+            panic!("Delta builder: intervals not properly sorted");
+        }
```
rkusa: To be consistent (with other parts of the Xi code) a `assert!` could be used instead.

rust/rope/src/delta.rs
```diff
@@ -341,6 +336,60 @@ impl<'a, N: NodeInfo + 'a> Transformer<'a, N> {
...
+
+    /// Replaces the given interval with the new rope. Panics if interval
+    /// is not properly sorted.
+    pub fn replace(&mut self, interval: Interval, rope: Node<N>) {
+        self.delete(interval);
+        self.delta.els.push(DeltaElement::Insert(rope));
+    }
+
+    /// Determines if delta would be trivial if built
+    pub fn is_trivial(&self) -> bool {
```
rkusa: As similarly suggested in #256, it may be more clear to name this one `is_empty`

rust/core-lib/src/editor.rs
```diff
@@ -386,19 +398,25 @@ impl<W: Write + Send + 'static> Editor<W> {
...
-        };
+        let mut builder = delta::Builder::new(self.text.len());
+        for region in self.view.sel_regions() {
+            let start = if !region.is_caret() {
+                region.min()
+            } else if let Some(bsp_pos) = self.text.prev_codepoint_offset(region.end) {
+                // TODO: implement complex emoji logic
+                bsp_pos
+            } else {
+                region.end
```
raphlinus: `unwrap_or` is cleaner, will do that. Yes, that handles beginning of buffer, so it is reachable.

rust/rope/src/delta.rs
raphlinus: What would Range buy you that Interval doesn't? Interval does have the weirdness of supporting both open and closed endpoints, but I'm not sure that justifies having a whole nother type.

rust/rope/src/delta.rs
raphlinus: As an optimization to avoid putting 0-length Copy elements into the delta.

rust/rope/src/delta.rs
raphlinus: So here's where things get interesting. The `Delta` struct (and thus this method) is generic on all kinds of ropes, not just strings. A delta of, say, line breaks is also meaningful, and it's not hard to see how that might be used to do minimal updates.

There are two things that could be done. There could be an `AsRef<str>`-taking method added as an impl for `Delta<RopeInfo>`. But I think a more general approach would be to change this to take `Into<Node<N>>` here, and then do appropriate `From` implementations to get strings into ropes.

I think that's outside the scope of this PR, but still worth doing as a cleanup.

rust/rope/src/delta.rs
raphlinus: Done

rust/rope/src/delta.rs
raphlinus: Done

PR #264, opened 2017-05-07
doc: rename render_lines into request_lines

little-dude:


raphlinus:
It's not correct. The old `render_lines` was a sync RPC request that returned the rendered lines. The new `request_lines` is an async notification that causes an async `update` to be sent later including the lines requested.

Thanks for the effort to clean up the docs, it's much appreciated.

PR #266, opened 2017-05-07
Fix valid lines cache for scroll updates (Fixes #241)

### LINKED ISSUE #241: scrolling near the bottom of a file constantly resends the entire file
Please confirm if this is a bug, or some sort of misunderstanding of how scroll should work.

For testing, I used a file named big.txt that's over 100,000 lines.  You can get it here: http://norvig.com/big.txt

Here's a simple test where I scroll around those lines.

```sh
$ cat test0
{"id":0,"method":"new_view","params":{ "file_path": "/home/brain/src/gxi/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100002,100012] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100003,100013] }}
$ cat test0 | xi-core | wc
      6     337    3685
```
It sent around 3k bytes.  This test seems normal.

```sh
$ cat test1
{"id":0,"method":"new_view","params":{ "file_path": "/home/brain/src/gxi/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[0,10] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}
$ cat test1 | xi-core | wc
      5  820735 7827012
```
Scrolling from the top to the bottom ended up sending over 7 MiB.  It sends all 100,000 lines.  This is unexpected.

```sh
$ cat test2
{"id":0,"method":"new_view","params":{ "file_path": "/home/brain/src/gxi/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[0,10] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100001,100011] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100002,100012] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100003,100013] }}
$ cat test2 | xi-core | wc
      8 3282076 31298626
```
31 MiB.  Every scroll afterwards is sending every one of the 100,000 lines over and over.

I would expect for scroll to invalidate far away lines and only send lines that I don't already have.

rkusa:
Fixes #241

The bug was in `send_update_for_scroll` in `view.rs`.

(I have ignored the scroll slop in the following examples)

When first requesting `[0, 10]`, the lines `[0, 10]` will go into `valid_lines`. When requesting `[100000,100010]`, all lines not in `valid_lines` will be send, i.e. `[10, 100010]`. However, only  `[100000,100010]` will make it into `valid_lines`. That is, the next request of `[100001,100011]` will send lines `[10, 100000]` (because not added to the line cache previously) and `[100010,100011]`.

(`valid_lines` ... are  Ranges of lines held by the line cache in the front-end that are considered valid)

Re-running @bvinc test (thanks for the excellent instructions to reproduce the issue)

```bash
% cat test1
{"id":0,"method":"new_view","params":{ "file_path": "/Users/markus/Downloads/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[0,10] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}

% cat test1 | cargo run | wc
       5  820735 7827012
```

```bash
% cat test1
{"id":0,"method":"new_view","params":{ "file_path": "/Users/markus/Downloads/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[0,10] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100001,100011] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100002,100012] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100003,100013] }}

% cat test1 | cargo run | wc
       8  820771 7827847
```

I cannot use `self.valid_lines.union_one_range` in the for loop directly (mutable borrow, but immutable borrow happens before), which is why I had to create a temporary Vec. If there is a better way then creating this temporary variable, I am happy to hear about it!

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -299,6 +299,7 @@ impl View {
 
         let mut ops = Vec::new();
         let mut line = 0;
+        let mut valid_ranges = Vec::new();
         for (start, end) in self.valid_lines.minus_one_range(first_line, last_line) {
             // TODO: this has some duplication with send_update in the non-dirty case.
             if start > line {
@@ -317,6 +318,7 @@ impl View {
             }
             ops.push(self.build_update_op("ins", Some(rendered_lines), end - start));
             ops.push(self.build_update_op("skip", None, end - start));
+            valid_ranges.push((start, end));
             line = end;
         }
         if line == 0 {
@@ -331,7 +333,9 @@ impl View {
             "pristine": self.pristine,
         });
         tab_ctx.update_view(&self.view_id, &params);
-        self.valid_lines.union_one_range(first_line, last_line);
+        for range in valid_ranges {
+            self.valid_lines.union_one_range(range.0, range.1);
+        }
     }
 
     fn build_update_op(&self, op: &str, lines: Option<Vec<Value>>, n: usize) -> Value {


trishume:
I tested this and it is much better, but isn't this still really bad?

When scrolling, shouldn't it only send the lines necessary to display the view? Shouldn't it send lines `[0,10]` and then send lines `[100000,100010]` without sending all the lines in between?

rkusa:
> When scrolling, shouldn't it only send the lines necessary to display the view? Shouldn't it send lines [0,10] and then send lines [100000,100010] without sending all the lines in between?

I agree and I am glad you bring this up (because I was thinking the same when reading the source but somehow lost the thought while fighting the borrow checker).

trishume:
And it is also somewhat realistic even for normal scrolling, because when scrolling very quickly in a large file (e.g with OSX scroll acceleration), the scrolling will skip over large portions of the file between frames.

raphlinus:
I think this is not the right fix, #268 is. That's not to say things are perfect, but should be much better. I believe that with #268 merged, this change should be a no-op, so should be closed.

Thanks for looking into it! Definitely helped me narrow down where to look in the IndexSet impl.

rkusa:
Now it requests `[100000,100010]`, instead of `[10, 100000]`. Nice to see it fixed, thanks!

Issue #262, opened 2017-05-06
xi-unicode line break unit test failures


cbreeden:
I wrote a test which parses the unit tests from UAX-45 found [here](http://www.unicode.org/Public/9.0.0/ucd/auxiliary/LineBreakTest.txt), and compared them with the results from `xi-unicode::LineBreakIterator`.  There were 122 failures.  I posted the failures in this [gist](https://gist.github.com/cbreeden/46e895cbf07b40072dc0e4d4cace761a).

It's possible that there could be a bug in my unit tests, though I have verified a few of the failures by hand.  I have included the source for the tests in the gist -- just be sure to include [LineBreakTest.txt](http://www.unicode.org/Public/9.0.0/ucd/auxiliary/LineBreakTest.txt) in the `tests/` directory.

I can open a PR for including the unit test if desired.

One possible source of false positives for these tests is the allowance of some [customizations](http://www.unicode.org/reports/tr14/#Customization).  These will need to be verified individually.

cbreeden:
Changing [this line](https://github.com/google/xi-editor/blob/master/rust/unicode/tools/mk_tables.py#L370) from

```python
            elif R == 'CM' and L not in ['BK', 'CR', 'LF', 'NL', 'SP', 'ZW']:
```
to
```python
            elif (R == 'CM' or R == 'ZWJ') and L not in ['BK', 'CR', 'LF', 'NL', 'SP', 'ZW']:
```
reduces the number of failures to 53.  The new line seems to more accurately reflect [LB9](http://www.unicode.org/reports/tr14/#BreakingRules).

I will need to verify that this is a strict subset of errors, and probably run it through the fuzzy tests as well.

raphlinus:
I suspect that most of these failures are due to Unicode 8->9 changes, and the followup about ZWJ is consistent. I hand-tested against the Unicode 8 corpus when I was developing it.

I don't have a lot of customizations or tailoring. In fact, I explicitly decided against adding more breaks in the punctuation / number combinations as in ICU, as it complicates the logic a lot and I believe the new breaks are not robust in all use cases.

I would definitely welcome a PR that added explicit testing.

Thanks for digging into this!

PR #272, opened 2017-05-08
Fix Unicode 9.0 discrepancies

cbreeden:
Fixes: #262 

This is any attempt to fix some test failures from the Unicode 9.0 line break unit tests.  A few things to note about this PR:

1. I have added `linebreaktest.txt` which can be used for unit tests.  I have commented out the unit tests that used custom handling of numeric-like sequences.  You can see which were commented out here e31524fcaf09b75deb7d6faae47d415eda848807.
2. I'm not entirely sure that [this line](https://github.com/google/xi-editor/commit/e31524fcaf09b75deb7d6faae47d415eda848807#diff-00d31ae7cb5cea5cdd929eef5f468d21R357) is handling all cases correctly.  It might be worth taking a look into this, as I _feel_ that I might be able to construct a test example which could fail along these lines.


cbreeden:
I signed it!

cbreeden:
Thanks, I'd wouldn't definitely appreciate a second pair of eyes.  In case you have some questions, (and since the patch is so small), I can briefly explain the reasoning here:

The following two modifications seem like the straight forward way to handling the [LB9](http://unicode.org/reports/tr14/#BreakingRules), basically where `ZWJ` is treated as you would treat `CM`.

```python
           if R in ['CM', 'ZWJ'] and L in ['BK', 'CR', 'LF', 'NL', 'SP', 'ZW']:
```
and
```python
          elif R in ['CM', 'ZWJ'] and L not in ['BK', 'CR', 'LF', 'NL', 'SP', 'ZW']:
```

One issue is when `ZWJ` is on the left.  We need to treat `ZWJ`, like we do with `CM`, as an `AL` except for the case when `[ID|EB|EM]` is on the right.  This is the purpose of the following conditional:

```python
            elif L == 'ZWJ' and R not in ['ID', 'EB', 'EM']:
                #handling for LB10
                l_with_cm = 2  # AL
                bk = t['AL' + '|' + R]
```

Ignoring the whole `l_with_cm` part is enough to fix all but 3 unit tests from LineBreakTest.txt.  And the `l_with_cm` part is the part I am least comfortable with.  The `l_with_cm` along with the `elif R in ['CM', 'ZWJ']` part which assigns `state = l_with_cm` is logically equivalent to having a special case for  `ZWJ [ CM | ZWJ ]` to be treated as an `AL`.

The test case to have in mind is `ZWJ x CM - [ID|EB|EM]`.  Provided that `ZWJ` doesn't take an identity from a previous character (LB9), which occurs when a `ZWJ` is after a hardbreak or sot, then the `CM` will try to take the identity from `ZWJ`.  Which would render the test as equivalent to `ZWJ x ZWJ x [ID|EB|EM]`.  What should be happening is that the `ZWJ x CM` should be treated as an `AL` and there should be a break between the `CM` and `[ID|EB|EM]`.

raphlinus:
I had a look at this and it seems fine. It's a little more ugliness, but symmetrical with the existing logic, and the underlying rules require a bunch of special-casing. It's nice that you were able to do this without expanding the size of the state machine itself. Also really happy that there are tests, this will definitely help the next time the spec revs.

Thanks again!

cbreeden:
Thanks! I learned a lot with this PR.  I found the Trie data structure for the codepoint -> break class mapping particularly interesting.

> It's nice that you were able to do this without expanding the size of the state machine itself.

The thought did cross my mind, noticing that the 2 MSBs are only used for representing 3 states: no break, soft break, and hard break.  It's quite possible to add a "needs special handling" state which include exactly one additional conditional for only the breaking cases.  But that should definitely be saved as a last resort of course :)

PR #274, opened 2017-05-08
Fix deadlock in tabs::do_save

cmyr:
- fixes 273

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -344,10 +344,11 @@ impl<W: Write + Send + 'static> Documents<W> {
     fn do_save(&mut self, view_id: &ViewIdentifier, file_path: &str) -> Option<Value> {
         let file_path = PathBuf::from(file_path);
         // if this is a new path for an existing file, we have a bit of housekeeping to do:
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -344,10 +344,11 @@ impl<W: Write + Send + 'static> Documents<W> {
     fn do_save(&mut self, view_id: &ViewIdentifier, file_path: &str) -> Option<Value> {
         let file_path = PathBuf::from(file_path);
         // if this is a new path for an existing file, we have a bit of housekeeping to do:
-        if let Some(prev_path) = self.buffers.lock()
-            .editor_for_view(view_id).unwrap().get_path() {
+        let prev_path = self.buffers.lock()
```
cbrewster: Can we just lock once and store that in a var?

cmyr: I tried that at first and was finding it hard to do expressively, mostly because of the last call to `add_file`, which does its own lock management. The ideal solution would probably be to move this whole piece of logic up into `BufferContainerRef`.

raphlinus:
Thanks!

PR #275, opened 2017-05-08
Fix end-of-line movement when on last line (Fixes #261)

### LINKED ISSUE #261: move_to_end_of_line does not work on last line
A small bug in movement handling. This is likely a small logic error in the recent movement refactor.

Reproduction:

- move the cursor to the last line in a document
- (xi-mac) hit `ctrl + e`
- the cursor stays in place.

Maybe we need a '_probably_ easy' label. 😏

ineol:
Xi used to go to one grapheme before the next discontinuity for LineMetrics.
This is not correct when on the last line.


ineol:
@googlebot I signed it

raphlinus:
Thanks for the fix, looks good!

PR #276, opened 2017-05-08
Rationalize file_name state, plus tests

cmyr:
This is progress towards the next step of plugins stuff, but can stand alone as a PR.

- it fixes an issue where, if we failed to open an existing file, we would create an empty buffer associated with that file's path, which could cause data loss.

- it cleans up lifecycle stuff a bunch, moving more responsibility into the `BufferContainer` types. Most importantly, there is now a single function responsible for updating a buffer's associated filepath. (This will enable syntax-specific plugin activations).

- It adds a test for some basic file saving behaviour.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -169,7 +169,8 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.views.len() > 0
     }
 
-    pub fn set_path<P: AsRef<Path>>(&mut self, path: P) {
+    /// should only ever be called from `BufferContainerRef::set_path`
+    pub fn _set_path<P: AsRef<Path>>(&mut self, path: P) {
         self.path = Some(path.as_ref().to_owned());
     }
 
@@ -603,7 +604,6 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
         // TODO: we should probably bubble up this error now. in the meantime always set path,
         // because the caller has updated the open_files list
-        self.path = Some(path.as_ref().to_owned());
         self.pristine_rev_id = self.last_rev_id;
         self.view.set_pristine();
         self.view.set_dirty();
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -170,6 +170,7 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
     /// should only ever be called from `BufferContainerRef::set_path`
+    #[doc(hidden)]
     pub fn _set_path<P: AsRef<Path>>(&mut self, path: P) {
         self.path = Some(path.as_ref().to_owned());
     }
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -112,11 +112,34 @@ impl <W: Write + Send + 'static>BufferContainerRef<W> {
         self.lock().open_files.contains_key(file_path.as_ref())
     }
 
-    /// Adds `file_path` as an open file, associated with `view_id`'s buffer.
-    pub fn add_file<P: AsRef<Path>>(&self, file_path: P, view_id: &ViewIdentifier) {
+    /// Adds a new editor, associating it with the provided identifiers.
+    pub fn add_editor(&self, view_id: &ViewIdentifier, buffer_id: &BufferIdentifier,
+                      editor: Editor<W>, path: Option<&Path>) {
+        {
+            let mut inner = self.lock();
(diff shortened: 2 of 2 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -169,7 +169,8 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.views.len() > 0
     }
 
-    pub fn set_path<P: AsRef<Path>>(&mut self, path: P) {
+    /// should only ever be called from `BufferContainerRef::set_path`
+    pub fn _set_path<P: AsRef<Path>>(&mut self, path: P) {
```
rkusa: What is the purpose of the prepended underscore - to communicate that the API should not be used, but you still need it `pub` to be able to call it from `BufferContainerRef`?

rust/core-lib/src/tabs.rs
```diff
@@ -289,38 +308,36 @@ impl<W: Write + Send + 'static> Documents<W> {
         json!(view_id)
     }
 
-    fn do_close_view(&mut self, view_id: &ViewIdentifier) {
-        self.close_view(view_id);
+    fn close_view(&mut self, view_id: &ViewIdentifier) {
```
rkusa: Do we have some kind of rule of thumb when to use `do_` and when not?

rust/core-lib/src/tabs.rs
```diff
@@ -289,38 +308,36 @@ impl<W: Write + Send + 'static> Documents<W> {
...
+                    // TODO: we should be reporting errors to the client
+                    print_err!("unable to read file: {}, error: {:?}", buffer_id, err);
+                    self.buffers.add_editor(view_id, &buffer_id, ed, None);
+                } else {
+                    // if a path that doesn't exist, create a new empty buffer + set path
+                    self.buffers.add_editor(view_id, &buffer_id, ed, Some(path));
+                }
             }
-        }
+        };
```
rkusa: (I don't want to be a nitpicker, but since I am adding comments anyway ... semicolon could be omitted for consistency)

rust/core-lib/src/editor.rs
cmyr: yes, exactly. It might be that the path just shouldn't be part of the editor, but I want to at least make it clear right now that setting path is not intended as real public API.

rust/core-lib/src/tabs.rs
cmyr: The use of `do_` predates my involvement, and I don't have strong opinions. In `editor` the `do_` methods are generally the actual implementations?

It is not a pattern I'm using in new code.

rust/core-lib/src/tabs.rs
cmyr: Nits are always welcome!

rust/core-lib/src/editor.rs
raphlinus: Another pattern to signal the same is `#[doc(hidden)]`, especially given that you haven't written actual doc. But the underscore is ok here.

rust/core-lib/src/tabs.rs
```diff
@@ -453,3 +459,67 @@ impl<W: Write> DocumentCtx<W> {
         self.update_channel.send((view_id, update, undo_group)).unwrap();
     }
 }
+
+#[cfg(test)]
```
raphlinus: 👍 for adding tests

rust/core-lib/src/tabs.rs
```diff
@@ -453,3 +459,67 @@ impl<W: Write> DocumentCtx<W> {
...
+        assert!(!container_ref.has_open_file("a fake file, for sure"));
+        let view_id_1 = "view-id-1".to_owned();
+        let buf_id_1 = "buf-id-1".to_owned();
+        let path_1 = PathBuf::from("a_path");
+        let path_2 = PathBuf::from("a_different_path");
+        let editor = Editor::new(mock_doc_ctx(&view_id_1), &view_id_1);
+        container_ref.add_editor(&view_id_1, &buf_id_1, editor, None);
+        assert_eq!(container_ref.lock().editors.len(), 1);
+
+        // save this somewhere:
```
raphlinus: This says "save" but I don't see any file writing, just setting paths. Am I missing something?

cmyr: right this is a bit confusing. this test is against the tracking of save location, which happens alongside the actual save operation, which is not tested.

cmyr:
Okay this is gtg.

raphlinus:
Great, thanks!

PR #277, opened 2017-05-09
Refactor Engine, fixes #256

### LINKED ISSUE #256: Refactor Engine

trishume:
This PR contains a number of related refactorings to the `Engine`, `Delta` and `Subset` classes. It is entirely re-namings, doc comment changes, and one removal of a duplicate method. Split into multiple commits according to which refactors must occur together, you may want to review each commit individually. Fixes #256, see discussion there for context.

cc @raphlinus 

I've pasted the commit message of the most significant refactoring below:

### Rename methods of Subset to be more consistent.

There was previously an inconsistency in the naming in Engine and the naming
in Subset. See #256

The naming of the methods of Subset was as if the Subset represented
the complement of the ranges it contained, which if the ranges were deletions
meant it represented the elements that were kept.

However, this was inconsistent with the naming in Engine which had subsets like
`inserts` and `deletes` that according to the concept of Subset actually represented
the non-inserted and non-deleted ranges.

This fixes the inconsistency by renaming methods as if Subset just represents the
ranges it contains, which eliminates having to reason about the complement when
thinking about the entire system.

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -259,7 +259,7 @@ impl<N: NodeInfo> InsertDelta<N> {
 
     /// Return a Subset containing the inserted ranges.
     ///
-    /// `d.inserted_subset().delete_in_string(d.apply_to_string(s)) == s`
+    /// `d.inserted_subset().delete_from_string(d.apply_to_string(s)) == s`
     pub fn inserted_subset(&self) -> Subset {
         let mut sb = SubsetBuilder::new();
         let mut x = 0;
@@ -456,7 +456,7 @@ mod tests {
         let d = Delta::simple_edit(Interval::new_closed_open(1, 9), Rope::from("era"), 11);
         let (d1, ss) = d.factor();
         assert_eq!("heraello world", d1.apply_to_string("hello world"));
-        assert_eq!("hld", ss.delete_in_string("hello world"));
+        assert_eq!("hld", ss.delete_from_string("hello world"));
     }
 
     #[test]
@@ -476,7 +476,7 @@ mod tests {
     fn inserted_subset() {
         let d = Delta::simple_edit(Interval::new_closed_open(1, 9), Rope::from("era"), 11);
         let (d1, _ss) = d.factor();
-        assert_eq!("hello world", d1.inserted_subset().delete_in_string("heraello world"));
+        assert_eq!("hello world", d1.inserted_subset().delete_from_string("heraello world"));
     }
 
     #[test]
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -22,10 +22,16 @@ use subset::{Subset, SubsetBuilder};
 use std::cmp::min;
 
 enum DeltaElement<N: NodeInfo> {
+    /// Represents the text in a range of the base document. Includes beginning, excludes end.
     Copy(usize, usize),  // note: for now, we lose open/closed info at interval endpoints
     Insert(Node<N>),
 }
 
+/// Represents an entire new document as a sequence of sections copied from
+/// the old document and of new inserted text.
+///
+/// For example, Editing "abcd" into "acde" could be represented as:
+/// `[Copy(0,1),Copy(2,4),Insert("e")]`
 pub struct Delta<N: NodeInfo> {
     els: Vec<DeltaElement<N>>,
     base_len: usize,
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -80,7 +80,7 @@ impl<N: NodeInfo> Delta<N> {
         for elem in self.els {
             match elem {
                 DeltaElement::Copy(b, e) => {
-                    sb.add_deletion(e1, b);
+                    sb.add_range(e1, b);
                     e1 = e;
                 }
                 DeltaElement::Insert(n) => {
@@ -95,7 +95,7 @@ impl<N: NodeInfo> Delta<N> {
         if b1 < self.base_len {
             ins.push(DeltaElement::Copy(b1, self.base_len));
         }
-        sb.add_deletion(e1, self.base_len);
+        sb.add_range(e1, self.base_len);
         (Delta { els: ins, base_len: self.base_len }, sb.build())
     }
 
@@ -103,12 +103,12 @@ impl<N: NodeInfo> Delta<N> {
     /// insertions and the other representing deletions. This is basically the inverse
     /// of `factor`.
     pub fn synthesize(s: &Node<N>, ins: &Subset, del: &Subset) -> Delta<N> {
-        let base_len = ins.len(s.len());
+        let base_len = ins.len_after_delete(s.len());
         let mut els = Vec::new();
         let mut x = 0;
-        let mut ins_ranges = ins.range_iter(s.len());
+        let mut ins_ranges = ins.complement_iter(s.len());
         let mut last_ins = ins_ranges.next();
-        for (b, e) in del.range_iter(s.len()) {
+        for (b, e) in del.complement_iter(s.len()) {
             let mut beg = b;
             while beg < e {
                 while let Some((ib, ie)) = last_ins {
@@ -163,7 +163,7 @@ impl<N: NodeInfo> Delta<N> {
         let mut y = 0;  // coordinate within xform
         let mut i = 0;  // index into self.els
         let mut b1 = 0;
-        let mut xform_ranges = xform.range_iter(l);
+        let mut xform_ranges = xform.complement_iter(l);
         let mut last_xform = xform_ranges.next();
         while y < l || i < self.els.len() {
             let next_iv_beg = if let Some((xb, _)) = last_xform { xb } else { l };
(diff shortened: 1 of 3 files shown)


rust/rope/src/delta.rs
```diff
@@ -22,10 +22,16 @@ use subset::{Subset, SubsetBuilder};
 use std::cmp::min;
 
 enum DeltaElement<N: NodeInfo> {
+    /// Represents the text in a range of the base document. Includes beginning, excludes end.
```
cmyr: maybe "a range in the base document", or "a range of text in the base document".

rust/rope/src/delta.rs
```diff
@@ -22,10 +22,16 @@ use subset::{Subset, SubsetBuilder};
...
     Copy(usize, usize),  // note: for now, we lose open/closed info at interval endpoints
     Insert(Node<N>),
 }
 
+/// Represents an entire new document as a sequence of sections copied from
```
cmyr: I find this a little confusing. I suppose it is a representation of a document, but I find it easier to think of it as a representation of a _transformation_ from one document (or document state) to another. 

This PR is a welcome opportunity for me to clarify my own understanding of a bunch of this code. 

How I would write this:

"A representation of a transformation from one document state to another. 

This transformation is composed of a series of `Copy` and `Insert` operations. `Copy` operations indicate unmodified ranges of the existing document, and `Insert` operations indicate additions. 

Note: The deletion of some part of the document, in this schema, is represented as a set of `Copy` operations over the ranges to be kept.

For example, the transformation from "hello" to "hippo" can be represented as:
`[Copy(0, 1), Insert("ipp"), Copy(4,5)]`"



(sorry, I got a little carried away)

rust/rope/src/delta.rs
```diff
@@ -205,10 +211,10 @@ impl<N: NodeInfo> Delta<N> {
         Delta { els: els, base_len: l }
     }
 
-    /// Return a subset that inverts the insert-only delta:
+    /// Return a Subset containing the inserted ranges of an insert-only delta.
```
cmyr: Something that isn't clear to me: is "insert-only" a guarantee made anywhere? Is the caller expected to ensure that a given delta is an "insert-only delta"?

rust/rope/src/engine.rs
```diff
@@ -85,28 +85,22 @@ impl Engine {
         None
     }
 
+    /// Get the contents of the document at a given revision number
     fn get_rev_from_index(&self, rev_index: usize) -> Rope {
```
cmyr: I think this function might be another candidate for a renaming. Maybe something like `get_revision`? or `document_at_revision`? I think the word `from` is confusing, given the way it is used in Rust to indicate conversion constructors. Maybe `for` would be more communicative? I also don't love `get` for something that isn't a getter. 😕

rust/rope/src/subset.rs
```diff
@@ -48,57 +48,54 @@ impl SubsetBuilder {
...
+        for (b, e) in self.complement_iter(s.len()) {
             result.push_str(&s[b..e]);
         }
         result
     }
 
     // Maybe Subset should be a pure data structure and this method should
     // be a method of Node.
-    pub fn apply<N: NodeInfo>(&self, s: &Node<N>) -> Node<N> {
+    pub fn delete_in<N: NodeInfo>(&self, s: &Node<N>) -> Node<N> {
```
cmyr: This would benefit from documentation.

I also prefer `delete_from` to `delete_in`.

rust/rope/src/delta.rs
```diff
@@ -145,27 +159,65 @@ impl<N: NodeInfo> Delta<N> {
...
+        if let Some((&DeltaElement::Copy(beg, end), init)) = els.split_last() {
+            if end == iv_end {
+                iv_end = beg;
+                els = init;
+            }
+        }
+        (Interval::new_closed_open(iv_start, iv_end), Delta::els_len(els))
+    }
+
+    fn els_len(els: &[DeltaElement<N>]) -> usize {
```
raphlinus: Doc (and possible rename): this method returns the length of the transformed string.

rust/rope/src/engine.rs
```diff
@@ -33,7 +33,7 @@ pub struct Engine {
 
 struct Revision {
     rev_id: usize,
-    from_union: Subset,
+    deletes_from_union: Subset,
```
raphlinus: yes, this is a good name. probably main reason i didn't go there is i tend to favor terseness

rust/rope/src/delta.rs
trishume: I agree to some extent, but the reason I changed it is that while the struct conceptually represents a change, something that confused me at first was that the `els` field actually represents the entire new document, just by copying things from the old one.

I've changed the comment to reflect both of these things better.

rust/rope/src/subset.rs
```diff
@@ -48,57 +48,55 @@ impl SubsetBuilder {
...
-    pub fn _deletions(&self) -> &[(usize, usize)] {
+    pub fn _ranges(&self) -> &[(usize, usize)] {
         &self.0
     }
 
-    /// Compute the intersection of two subsets. In other words, an element exists in the
-    /// resulting subset iff it exists in both inputs.
-    pub fn intersect(&self, other: &Subset) -> Subset {
+    /// Compute the union of two subsets. In other words, an element exists in the
+    /// resulting subset if it exists either input.
```
raphlinus: 'iff' is correct here, but maybe people aren't familiar with the usage?

trishume:
@cmyr I addressed all your comments, thanks, they were valuable and I agreed with them all.

This PR now has a lot of commits, but before merging I plan on doing some squashing, although I don't want to squash the entire thing since it's useful to pick out the individual changes from being washed away in the diffs of the larger ones.

rust/rope/src/subset.rs
trishume: Oh that was just an accidental omission when I edited it to describe union instead of intersection. I just improved it.

rust/rope/src/engine.rs
trishume: It was a totally fine name back when `Subset` actually was supposed to represent a transformation of a document to a subset of it, then the type told you what it was. Under the new model though, it needs a longer name to be unambiguous.

Also, Sublime has really good auto-complete for terms that are used frequently within a single file, so for me it isn't any more difficult to type. Everyone else should get a better editor, although I know they're working on it. 😉

PR #278, opened 2017-05-09
Refactor plugins into a directory module

cmyr:
This PR contains no substantive changes.

It is based off of #276, which should be merged first.


cmyr:
let me incorporate the changes to #276

raphlinus:
Ok, looks good, I'll merge on green. One possible idea, instead of having two level `use plugins::foo::Bar`, you might consider deciding what is really part of the `plugins` public interface and re-exporting (`pub use`) just those types.

cmyr:
okay, I'll consider that approach for a future patch.

PR #279, opened 2017-05-09
Improve the worst case performance of Cursor::next()

ineol:
As mentioned in a TODO, try to walk up the cached subtree and skip the 0-measure
nodes. I checked that the two implementations coincide by walking line by line some
files.

It degrades a little the performance when the lines are short, for example in
the case of the sqlite source:
```
after    ... bench:   4,530,490 ns/iter (+/- 733,470)
before   ... bench:   4,128,834 ns/iter (+/- 411,246)
```
However, for a ~30MB file with a single newline at the end, it is significantly faster:
```
after    ... bench:       9,794 ns/iter (+/- 596)
before   ... bench:   2,388,631 ns/iter (+/- 318,327)
```

### DIFF
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -630,6 +630,7 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
     }
 
     /// Tries to find the next boundary in the leaf the cursor is currently in.
+    #[inline(always)]
     fn next_inside_leaf<M: Metric<N>>(&mut self) -> Option<usize> {
         if let Some(l) = self.leaf {
             let offset_in_leaf = self.position - self.offset_of_leaf;
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -196,7 +196,7 @@ impl<N: NodeInfo> Node<N> {
         }
     }
 
-    /// Returns the fist child with a positive measure, starting from the `j`th.
+    /// Returns the first child with a positive measure, starting from the `j`th.
     /// Also, returns the offset we have skipped; note that if it returns `None`in the first component, we skip all the children.
     fn next_positive_measure_child<M: Metric<N>>(&self, j: usize) -> (Option<usize>, usize) {
         let children = self.get_children();
@@ -534,6 +534,9 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
         result
     }
 
+    /// Moves the cursor to the previous boundary, or to the beginning of the
+    /// rope. In the former case, returns the position of the first character
+    /// past this boundary. In the latter case, returns `0`.
     pub fn prev<M: Metric<N>>(&mut self) -> Option<(usize)> {
         if self.position == 0 || self.leaf.is_none() {
             self.leaf = None;
@@ -581,9 +584,9 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
         }
     }
 
-    /// Moves the cursor to the next discontinuity, or to the end
-    /// of the rope. In the former case, returns the offset of said
-    /// discontinuity.
+    /// Moves the cursor to the next boundary, or to the end of the rope. In the
+    /// former case, returns the position of the first character past this
+    /// boundary. In the latter case, returns the length of the rope.
     pub fn next<M: Metric<N>>(&mut self) -> Option<(usize)> {
         if self.position >= self.root.len() || self.leaf.is_none() {
             self.leaf = None;
@@ -851,7 +854,7 @@ mod test {
 
     #[test]
     fn cursor_next_triangle() {
-        let n = 10_000;
+        let n = 2_000;
         let text = Rope::from(build_triangle(n));
 
         let mut cursor = Cursor::new(&text, 0);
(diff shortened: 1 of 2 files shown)


rust/rope/src/tree.rs
```diff
@@ -566,33 +579,74 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
+                    self.leaf = Some(node_down.get_leaf());
+                    self.offset_of_leaf = self.position;
+                    return self.next_inside_leaf::<M>();
                 }
-                if self.offset_of_leaf + l.len() == self.root.len() {
-                    self.position = self.root.len();
-                    return Some(self.position);
+            }
+            // At this point, we know that (1) the next discontinuity is not not in the
+            // cached subtree, (2) self.position corresponds to the begining of the firt
```
raphlinus: first

rust/rope/src/tree.rs
```diff
@@ -566,33 +579,74 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
+                return Some(self.position);
             }
+            if self.offset_of_leaf + l.len() == self.root.len() {
+                self.position = self.root.len();
+                return Some(self.position);
+            }
+        } else {
+            panic!("inconsistent, shouldn't get here");
         }
+        return None;
```
raphlinus: `return` isn't necessary, can just be `None`

rust/rope/src/tree.rs
```diff
@@ -566,33 +579,74 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
+            // cached subtree, (2) self.position corresponds to the begining of the firt
+            // leaf after the cached subtree.
+            self.descend();
+            return self.next::<M>();
+        } else {
+            panic!("inconsistent, shouldn't get here");
+        }
+    }
+
+    fn next_inside_leaf<M: Metric<N>>(&mut self) -> Option<usize> {
```
raphlinus: Would be better to have a doc comment explaining what this function does.

rust/rope/src/tree.rs
```diff
@@ -566,33 +579,74 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
         }
     }
 
+    // Moves the cursor to the next discontinuity, or to the end
```
raphlinus: `///` because this is fine as doc string.

I think I've been using "boundary" instead of "discontinuity," but not sure exactly which is better. Should be consistent though.

rust/rope/src/tree.rs
```diff
@@ -196,6 +196,19 @@ impl<N: NodeInfo> Node<N> {
         }
     }
 
+    fn next_positive_measure_child<M: Metric<N>>(&self, j: usize) -> (Option<usize>, usize) {
```
raphlinus: Would prefer a doc comment. Doesn't need to be detailed, as it's fairly clear, but should probably explain the return type.

rust/rope/src/tree.rs
```diff
@@ -566,33 +579,74 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
         }
     }
 
+    // Moves the cursor to the next discontinuity, or to the end
+    // of the rope. In the former case, returns the offset of said
+    // discontinuity.
     pub fn next<M: Metric<N>>(&mut self) -> Option<(usize)> {
```
raphlinus: Might be a good place to brag that this is O(log n) worst case now :)

ineol: hmm... I think it's still worse than that, since in the worst case we have to restart from the top when the cache is too small.

ineol:
Thank you for the comments.

It appears that Cursor::prev is doesn't move when it is already is on a boundary, so I don't see an obvious way to test it; this is why I only wrote tests/benchmarks for next.

Now that I think of it, with Cursor::set I can write tests but no benchmarks...

ineol:
For information, the benchmark situation is as follows on my computer (the numbers seem stable accros runs):

before:
```
test benchmark_few_big_lines ... bench:     753,114 ns/iter (+/- 36,710)
test benchmark_short_lines   ... bench:  16,673,272 ns/iter (+/- 863,349)
test benchmark_triangle      ... bench: 110,738,718 ns/iter (+/- 15,690,231)
```
without inlining:
```
test benchmark_few_big_lines ... bench:       4,365 ns/iter (+/- 1,869)
test benchmark_short_lines   ... bench:  22,283,243 ns/iter (+/- 1,538,873)
test benchmark_triangle      ... bench:  22,053,478 ns/iter (+/- 1,272,111)
```
with inlining:
```
test benchmark_few_big_lines ... bench:       4,288 ns/iter (+/- 1,886)
test benchmark_short_lines   ... bench:  18,841,222 ns/iter (+/- 949,549)
test benchmark_triangle      ... bench:  22,529,569 ns/iter (+/- 1,679,518)
```

rust/rope/src/tree.rs
```diff
@@ -196,6 +196,21 @@ impl<N: NodeInfo> Node<N> {
         }
     }
 
+    /// Returns the fist child with a positive measure, starting from the `j`th.
```
raphlinus: first

rust/rope/src/tree.rs
```diff
@@ -566,33 +581,76 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
+            // of the first leaf after the cached subtree.
+            self.descend();
+            return self.next::<M>();
+        } else {
+            panic!("inconsistent, shouldn't get here");
+        }
+    }
+
+    /// Tries to find the next boundary in the leaf the cursor is currently in.
+    #[inline(always)]
```
raphlinus: Would just `#[inline]` work as well? Using `always` should be done with caution, see http://stackoverflow.com/questions/37639276/when-should-inline-be-used-in-rust for discussion.

rust/rope/src/tree.rs
```diff
@@ -774,3 +832,43 @@ impl Metric<BytesInfo> for BytesMetric {
...
+            s += &line;
+            s += "\n";
+            line += "a";
+        }
+        s
+    }
+
+    #[test]
+    fn cursor_next_triangle() {
+        let n = 10_000;
```
raphlinus: This test takes 3s, I think if this were 2_000 it would test it as well and be a lot faster.

rust/rope/src/tree.rs
ineol: I tried, and `#[inline]` does not seem to have an effect. I think it's because the function is too big for the LLVM heuristic, but here inlining can remove at least one branch and a few loads.

ineol:
About `prev`, it seems to stop just *after* the boundary, so that the next time it is called it does not need to move.

raphlinus:
Ok, looks good enough now to merge. There's definitely follow-up work on `prev` and possibly some more doc cleanup, but this is good forward progress. Thanks!

Issue #263, opened 2017-05-06
warn when the front-end sends a notification instead of a request


eyelash:
When the front-end sends a notification but the core expects a request, the request will be processed but the response will be silently dropped.

[Relevant code](https://github.com/google/xi-editor/blob/1f610e23e38fc1311eefb9c4e8d2498d4c111453/rust/core-lib/src/lib.rs#L90)

cmyr:
Anyone interested: this is only really a problem for people developing a frontend. Adding a print_err! here would be a good first step.

zombiefungus:
Is this noob-could-fix (like myself) issue?

cmyr:
@zombiefungus yes, this should be a very simple change. Let me know if you have any questions, or ask in the irc channel (irc.mozilla.org, #xi)

PR #280, opened 2017-05-09
Print error if notification returns a value (#263)

zombiefungus:


rust/core-lib/src/lib.rs
```diff
@@ -86,8 +86,11 @@ impl<W: Write + Send + 'static> Handler<W> for MainState<W> {
     fn handle_notification(&mut self, ctx: RpcCtx<W>, method: &str, params: &Value) {
         match Request::from_json(method, params) {
             Ok(req) => {
-                let _ = self.handle_req(req, ctx.get_peer());
-                // TODO: should check None
+                if let Some(value) = self.handle_req(req, ctx.get_peer()) {
+                    if !value.is_null() {
```
cmyr: We should print the error even if the actual value is `Null`, because it is still a violation of the protocol.

raphlinus:
Always nice to have greater transparency into error conditions, thanks!

PR #281, opened 2017-05-09
Print error if return value of a notification is not null

zombiefungus:
Changed `rust/core-lib/src/plugins/mod.rs` and `rust/plugin-lib/src/plugin_base.rs` checking if the result of handling a notification is Some, same as done [here](https://github.com/google/xi-editor/blob/fe7f47898c72e4fb476733eef5fd6f857ce2d369/rust/core-lib/src/lib.rs#L89-L91) acording to #280


zombiefungus:
Totally forgot to tell you, here it is @cmyr

raphlinus:
Thanks!

PR #285, opened 2017-05-12
Plugin activations + syntax-specific syntax highlighting

cmyr:
This is very rough around the edges but brings some tangible editor improvements:

- plugins now start when a new file is opened (syntect always starts, spellcheck starts if you're in plaintext or markdown, as a proof of concept)
- plugins can be _stopped_, from the dropdown menu (as well as when you close a window)
- If you open a file, syntect chooses a syntax definition based on the file path
- **saving a file does not change syntax**. Soon.

I haven't tested this super super extensively, but wanted to get it up for review sooner rather than later.

There's a compatible xi-mac PR here: https://github.com/google/xi-mac/pull/38

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -27,6 +27,7 @@ use editor::Editor;
 use rpc::{CoreCommand, EditCommand, PluginCommand};
 use styles::{Style, StyleMap};
 use MainPeer;
+
 use syntax::SyntaxDefinition;
 use plugins::{self, PluginManagerRef};
 use plugins::rpc_types::PluginUpdate;
@@ -408,12 +409,14 @@ impl<W: Write + Send + 'static> Documents<W> {
             Start { view_id, plugin_name } => {
                 // TODO: this is a hack, there are different ways a plugin might be launched
                 // and they would have different init params, this is just mimicing old api
-                let (buf_size, _, rev) = {
-                self.buffers.lock().editor_for_view(&view_id).unwrap()
-                    .plugin_init_params()
-                };
 
-                self.plugins.start_plugin(&view_id, &plugin_name, buf_size, rev);
+                // TODO: track syntax defs
+                let syntax = SyntaxDefinition::Plaintext;
+                let buffer_info = self.buffers.lock().editor_for_view(&view_id).unwrap()
+                    .plugin_init_info();
+
+                //TODO: stop passing buffer ids
+                self.plugins.start_plugin(&view_id, &plugin_name, &buffer_info);
                 None
             }
             Stop { view_id, plugin_name } => {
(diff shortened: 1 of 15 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -417,20 +452,33 @@ impl<W: Write> DocumentCtx<W> {
...
+    pub fn available_plugins(&self, view_id: &ViewIdentifier, plugins: Vec<&str>) {
         self.rpc_peer.send_rpc_notification("available_plugins",
                                             &json!({
                                                 "view_id": view_id,
                                                 "plugins": plugins }));
     }
 
-    pub fn get_kill_ring(&self) -> Rope {
-        self.kill_ring.lock().unwrap().clone()
+    /// Notify the client that a plugin ha started.
```
rkusa: s/ha/has/

rust/core-lib/src/tabs.rs
```diff
@@ -417,20 +452,33 @@ impl<W: Write> DocumentCtx<W> {
...
+                                            &json!({
+                                                "view_id": view_id,
+                                                "plugin": plugin,
+                                            }));
     }
 
-    pub fn set_kill_ring(&self, val: Rope) {
-        let mut kill_ring = self.kill_ring.lock().unwrap();
-        *kill_ring = val;
+    /// Notify the client that a plugin ha stopped.
```
rkusa: s/ha/has/

eyelash:
How is global installation (expecially on Linux) supposed to work with plugins? Right now xi-core expects plugins to be in the same directory as the main executable, so if I install xi-core to /usr/bin/xi-core I have to install plugins to /usr/bin/xi-syntect-plugin and /usr/bin/spellcheck.py which is not really that great. Also I would like to see some instructions on how to correctly install the plugins in the README because a simple `cargo install` won't be enough anymore.

cmyr:
@eyelash: the current practice of just throwing a bunch of plugins in the build directory is very much temporary. 'Plugin discovery' (checking some `$XI_HOME` env var, or else a default location, and reading plugin manifests from disk) is part of the current plugin roadmap,  but I'm holding off until more features are done.

If it would be helpful in the meantime that I check some `XI_PLUGIN_DIR` on startup and optionally look there, instead of in the directory containing the executable, that would be a possible quick solution?

eyelash:
> If it would be helpful in the meantime that I check some XI_PLUGIN_DIR on startup and optionally look there, instead of in the directory containing the executable, that would be a possible quick solution?

Yes, that would be helpful. Only starting xi-syntect-plugin and spellcheck.py if they are available would also be helpful in not breaking the current `cargo install` workflow that most other front-ends use.

cmyr:
okay, good to know. I'll add that to this PR.

eyelash:
Much better now, thanks a lot!

raphlinus:
Great to see the progress!

Issue #238, opened 2017-04-26
RPC protocol + methods for 'Find'


cmyr:
Preliminary find is now available (#226) but this isn't exposed to the client in any way. My first thought for a first version of this would be something that functions like `find_next`, taking a starting offset?

PR #286, opened 2017-05-12
Implement basic search RPC methods

rkusa:
This PR adds RPC methods for #226.
Refs #224 and #238.

~~Can be tested with [rkusa/xi-mac#feature/search](https://github.com/rkusa/xi-mac/tree/feature/search)~~

~~There is no UI, yet, the menu items can be used~~

UI: https://github.com/google/xi-mac/pull/47

Current RPC protocol (open for discussion):

Start a search. If `chars` is not provided, the string under the last selection is used for search.

```
find
  chars?: string
  case_sensitive?: bool
```

Find next search hit

```
find_next
  wrap_around?: bool
  allow_same?: bool
```

Find previous search hit

```
find_previous
  wrap_around?: bool
```

There are some open points that require some discussions and that might be suited for their own PR, e.g.:

- Use `find_progress` instead of `find` (requires some benchmark first to determine a good `num_steps` value.
- And something I'll comment on in the changes

... and some more points which are not strictly search related, but about how spans are handled:

~~- A way to merge spans (currently the syntect plugin would override the search highlight and vise versa)~~

Edit: Highlights are no Spans anymore, they are now handled like Selections.

~~- Update Spans when text changes (I am not yet sure how to solve this efficiently)~~

Edit: Implemented

- A way (hitting Esc?) for removing spans

Edit: refs #324

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -816,8 +816,46 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.insert(&*String::from(kill_ring_string));
     }
 
-    pub fn do_rpc(&mut self, view_id: &ViewIdentifier, cmd: EditCommand) -> Option<Value> {
+    pub fn do_find(&mut self, chars: Option<&str>, case_sensitive: bool) -> Option<Value> {
+        let mut from_sel = false;
+        let search_string = chars.map(String::from).or_else(|| {
+            // if the search string is not provided, use the string of the last selection for find
+            self.view.sel_regions().last().and_then(|region| {
+                if region.is_caret() {
+                    None
+                } else {
+                    from_sel = true;
+                    Some(self.text.slice_to_string(region.min(), region.max()))
+                }
+            })
+        });
 
+        if search_string.is_none() {
+            self.view.unset_find();
+            return Some(Value::Null);
+        }
+
+        let search_string = search_string.unwrap();
+        if search_string.len() == 0 {
+            self.view.unset_find();
+            return Some(Value::Null);
+        }
+
+        self.view.set_find(&search_string, case_sensitive);
+
+        Some(Value::String(search_string.to_string()))
+    }
+
+    fn do_find_next(&mut self, reverse: bool, wrap_around: bool) {
+        self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, false, true);
+
+        if self.scroll_to.is_none() && wrap_around {
+            // nothing found, search past end of file
+            self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, true, true);
+        }
+    }
+
+    pub fn do_rpc(&mut self, view_id: &ViewIdentifier, cmd: EditCommand) -> Option<Value> {
         use rpc::EditCommand::*;
 
         // if the rpc's originating view is different from current self.view, swap it in
@@ -883,6 +921,9 @@ impl<W: Write + Send + 'static> Editor<W> {
             Redo => async(self.do_redo()),
             Cut => Some(self.do_cut()),
             Copy => Some(self.do_copy()),
+            Find { chars, case_sensitive } => self.do_find(chars, case_sensitive),
+            FindNext { wrap_around } => async(self.do_find_next(false, wrap_around)),
+            FindPrevious { wrap_around } => async(self.do_find_next(true, wrap_around)),
             DebugRewrap => async(self.debug_rewrap()),
             DebugTestFgSpans => async(self.debug_test_fg_spans()),
         };
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -846,12 +846,12 @@ impl<W: Write + Send + 'static> Editor<W> {
         Some(Value::String(search_string.to_string()))
     }
 
-    fn do_find_next(&mut self, reverse: bool, wrap_around: bool) {
-        self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, false, true);
+    fn do_find_next(&mut self, reverse: bool, wrap_around: bool, allow_same: bool) {
+        self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, false, true, allow_same);
 
         if self.scroll_to.is_none() && wrap_around {
             // nothing found, search past end of file
-            self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, true, true);
+            self.scroll_to = self.view.select_next_occurrence(&self.text, reverse, true, true, allow_same);
         }
     }
 
@@ -922,8 +922,8 @@ impl<W: Write + Send + 'static> Editor<W> {
             Cut => Some(self.do_cut()),
             Copy => Some(self.do_copy()),
             Find { chars, case_sensitive } => self.do_find(chars, case_sensitive),
-            FindNext { wrap_around } => async(self.do_find_next(false, wrap_around)),
-            FindPrevious { wrap_around } => async(self.do_find_next(true, wrap_around)),
+            FindNext { wrap_around, allow_same } => async(self.do_find_next(false, wrap_around, allow_same)),
+            FindPrevious { wrap_around } => async(self.do_find_next(true, wrap_around, true)),
             DebugRewrap => async(self.debug_rewrap()),
             DebugTestFgSpans => async(self.debug_test_fg_spans()),
         };
--- a/rust/core-lib/src/rpc.rs
+++ b/rust/core-lib/src/rpc.rs
@@ -17,7 +17,7 @@
 use std::error;
 use std::fmt;
 use serde_json::{self, Value};
-use xi_rpc::{dict_get_u64, dict_get_string, arr_get_u64, arr_get_i64};
+use xi_rpc::{dict_get_u64, dict_get_string, dict_get_bool, arr_get_u64, arr_get_i64};
 use tabs::ViewIdentifier;
 
 // =============================================================================
@@ -119,6 +119,9 @@ pub enum EditCommand<'a> {
     Redo,
     Cut,
     Copy,
+    Find { chars: Option<&'a str>, case_sensitive: bool },
+    FindNext { wrap_around: bool },
+    FindPrevious { wrap_around: bool },
     DebugRewrap,
     DebugTestFgSpans,
 }
@@ -150,7 +153,7 @@ impl<'a> CoreCommand<'a> {
             "new_view" => params.as_object()
                 .map(|dict| NewView { file_path: dict_get_string(dict, "file_path") }) // optional
                 .ok_or_else(|| MalformedEditParams(method.to_string(), params.clone())),
-                
+
             "save" => params.as_object().and_then(|dict| {
                 dict_get_string(dict, "view_id").and_then(|view_id| {
                     dict_get_string(dict, "file_path").map(|file_path| {
@@ -286,6 +289,21 @@ impl<'a> EditCommand<'a> {
             "redo" => Ok(Redo),
             "cut" => Ok(Cut),
             "copy" => Ok(Copy),
+
+            "find" => params.as_object().map(|dict| {
+                let chars = dict_get_string(dict, "chars");
+                let case_sensitive = dict_get_bool(dict, "case_sensitive").unwrap_or(false);
+                Find { chars: chars, case_sensitive: case_sensitive}
+            }).ok_or_else(|| MalformedEditParams(method.to_string(), params.clone())),
+            "find_next" =>  params.as_object().map(|dict| {
+                let wrap_around = dict_get_bool(dict, "wrap_around").unwrap_or(false);
+                FindNext { wrap_around: wrap_around }
+            }).ok_or_else(|| MalformedEditParams(method.to_string(), params.clone())),
+            "find_previous" =>  params.as_object().map(|dict| {
+                let wrap_around = dict_get_bool(dict, "wrap_around").unwrap_or(false);
+                FindPrevious { wrap_around: wrap_around }
+            }).ok_or_else(|| MalformedEditParams(method.to_string(), params.clone())),
+
             "debug_rewrap" => Ok(DebugRewrap),
             "debug_test_fg_spans" => Ok(DebugTestFgSpans),
 
--- a/rust/core-lib/src/rpc.rs
+++ b/rust/core-lib/src/rpc.rs
@@ -120,7 +120,7 @@ pub enum EditCommand<'a> {
     Cut,
     Copy,
     Find { chars: Option<&'a str>, case_sensitive: bool },
-    FindNext { wrap_around: bool },
+    FindNext { wrap_around: bool, allow_same: bool },
     FindPrevious { wrap_around: bool },
     DebugRewrap,
     DebugTestFgSpans,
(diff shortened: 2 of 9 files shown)


rust/rope/src/find.rs
```diff
@@ -37,6 +37,7 @@ pub enum FindResult {
 /// A policy for case matching. There may be more choices in the future (for
 /// example, an even more forgiving mode that ignores accents, or possibly
 /// handling Unicode normalization).
+#[derive(Clone, Copy)]
```
rkusa: alternative: make `find` method expect `&CaseMatching` instead of `CaseMatching`

rust/rope/src/spans.rs
```diff
@@ -209,3 +224,53 @@ impl<'a, T: Clone + Default> Iterator for SpanIter<'a, T> {
...
+            return None;
+        }
+        if let Some((leaf, start_pos)) = self.cursor.get_leaf() {
+            if leaf.spans.is_empty() { return None; }
+            let leaf_start = self.cursor.pos() - start_pos;
+            let span = &leaf.spans[leaf.spans.len() - self.ix - 1];
+            self.ix += 1;
+            if self.ix == leaf.spans.len() {
+                if self.cursor.prev_leaf().is_none() {
+                    self.done = true;
```
rkusa: This is currently a workaround, because if there are leafs, `self.cursor.get_leaf()` seems to never return `None`. This is maybe a bug in `cursor.prev_leaf()`, or I am doing something wrong.

rust/rope/src/spans.rs
```diff
@@ -155,6 +155,12 @@ pub struct SpanIter<'a, T: 'a + Clone + Default> {
     ix: usize,
 }
 
+pub struct SpanBackwardsIter<'a, T: 'a + Clone + Default> {
+    cursor: Cursor<'a, SpansInfo<T>>,
+    ix: usize,
+    done: bool,
+}
+
```
rkusa: Instead of using an iterator, I was also considering implementing `Metric<SpansInfo<T>>` instead, however `Metric::prev` and `Metric::next` does not seem to be suitable for `Spans`, because this has to support multiple `Span`s at a given `offset`.

rust/core-lib/src/editor.rs
```diff
@@ -762,6 +762,83 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+        } else {
+            CaseMatching::CaseInsensitive
+        };
+
+        let mut sb = SpansBuilder::new(self.text.len());
+        let len = search_string.len();
+
+        let mut cursor = Cursor::new(&self.text, 0);
+        while let Some(pos) = find(&mut cursor, case_matching, &search_string) {
+            let style = Style { fg: 0xffc00000, font_style: 0 };
```
rkusa: The search highlight `Style` should not be hardcoded here, but where to put it?

rust/core-lib/src/editor.rs
```diff
@@ -762,6 +762,83 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+            let offset = iv.start();
+            let spans = self.style_spans.subseq(iv);
+            let span = if reverse {
+                spans.iter_backwards().find(|&(_, _style)| {
+                    let style_id = 1; // self.doc_ctx.get_style_id(style);
+                    return style_id == 1 // 1 ... reserved style id for search results
+                })
+            } else {
+                spans.iter().find(|&(_, _style)| {
+                    let style_id = 1; // self.doc_ctx.get_style_id(style);
```
rkusa: I cannot use `self.doc_ctx.get_style_id(style)` here, because `get_style_id` expects a `styles::Span` but we have `view::Style` here, and converting them each time does not make sense. I am not sure how to solves this elegantly ... 😕

rust/rope/src/find.rs
raphlinus: Clone+Copy is fine for things that fit in a word.

rust/rope/src/spans.rs
raphlinus: It's an inconsistency between `next_leaf()` and `prev_leaf()`. I'd be ok with adding `self.leaf = None;` to the first case in `prev_leaf()` instead.

rust/rope/src/spans.rs
```diff
@@ -189,6 +195,15 @@ impl<T: Clone + Default> Spans<T> {
             ix: 0,
         }
     }
+
+    /// iterates through all spans, but in backwards order
```
raphlinus: One interesting question about this: if you have nested spans, like `<a>...<b>...</b>...</a>`, are you returning `a` first or `b`, i.e. are you sorting by span begin or span end? Pretty sure it's the former, but probably worth documenting.

rust/rope/src/spans.rs
```diff
@@ -209,3 +224,53 @@ impl<'a, T: Clone + Default> Iterator for SpanIter<'a, T> {
...
+    use spans::SpansBuilder;
+    use interval::Interval;
+
+    #[test]
+    fn iter_backwards() {
+        let mut sb = SpansBuilder::new(8);
+        sb.add_span(Interval::new_open_open(1, 2), 1);
+        sb.add_span(Interval::new_open_open(4, 4), 3);
+        sb.add_span(Interval::new_open_open(4, 4), 2);
+        sb.add_span(Interval::new_open_open(3, 5), 4);
```
raphlinus: This test is dubious because it violates the precondition of `add_span` (sorted start order).

rust/core-lib/src/rpc.rs
```diff
@@ -281,6 +284,15 @@ impl<'a> EditCommand<'a> {
             "redo" => Ok(Redo),
             "cut" => Ok(Cut),
             "copy" => Ok(Copy),
+
+            "find" => params.as_object().map(|dict| {
+                let chars = dict_get_string(dict, "chars");
+                let case_sensitive = dict_get_bool(dict, "case_sensitive").unwrap_or(false);
```
raphlinus: You can probably use the newer serde-json features to make this more concise, something like `dict["case_sensitive"].as_bool().unwrap_or(false)`

cmyr: Or just `#[derive(Deserialize)]` on some appropriate struct.

rust/rope/src/spans.rs
rkusa: I've implemented `DoubleEndedIterator` instead in https://github.com/google/xi-editor/pull/286/commits/3588ad811650efad64f7a2d627aed08dc01ea26c
I like doing the it the Rust way, if there is one (and I wasn't really aware of `DoubleEndedIterator` previously).

rust/rope/src/spans.rs
rkusa: Do you think it is necessary to find a appropriate new position for such a comment, since I've implemented `DoubleEndedIterator`? Or is `DoubleEndedIterator` clear enough about what to expect?

rust/rope/src/spans.rs
rkusa: I've also added `self.leaf = None` for the first case of `next_leaf()` - for consistency (I hope I haven't missed something and that it wasn't there on purpose)

rust/core-lib/src/rpc.rs
rkusa: Unfortunately, this will panic with `panicked at 'no entry found for key'` if `case_sensitive` is not provided at all. Using `#[derive(Deserialize)]` would probably be a nice solution, but maybe better suited for a PR that converts all RPC methods to `#[derive(Deserialize)]`?

rust/core-lib/src/editor.rs
```diff
@@ -762,6 +762,83 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+                Interval::new_closed_open(region.max(), self.text.len())
+            }
+        });
+        if let Some(iv) = iv {
+            let offset = iv.start();
+            let spans = self.style_spans.subseq(iv);
+            let mut iter: Box<Iterator<Item=(Interval, &Style)>> = if reverse {
+                Box::new(spans.iter().rev())
+            } else {
+                Box::new(spans.iter())
```
rkusa: I disliked the repetition of `find` closure below, but the only way around it seemed to be using a `Box`.

rust/core-lib/src/editor.rs
```diff
@@ -762,6 +762,83 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+            };
+            let span = iter.find(|&(_, _style)| {
+                let style_id = SEARCH_STYLE_ID; // self.doc_ctx.get_style_id(style);
```
rkusa: Instead of using `self.doc_ctx.get_style_id(style);`, this still accepts all spans as a search highlight. Just wanted to highlight this issue again, even through I personally see this being fixed in later PRs (that e.g. focus on span-layers).

rust/rope/src/spans.rs
```diff
@@ -195,17 +200,122 @@ impl<'a, T: Clone + Default> Iterator for SpanIter<'a, T> {
...
+        if self.cursor.pos() != pos {
+            // switch from forward to backward iteration
+            self.cursor.set(pos);
+        }
+        if let Some((leaf, start_pos)) = self.cursor.get_leaf() {
+            if leaf.spans.is_empty() { return None; }
+            if let Some(last_leaf) = self.last_leaf {
+                // stop, if forward and backward iteration met
+                if ptr::eq(last_leaf, leaf) && leaf.spans.len() - ix == self.forward.1 {
+                    println!("NONE A!");
```
trishume: Should this and the other `println` still be here?

rkusa: Oops, good catch! I've removed them. Thanks

rust/core-lib/src/editor.rs
```diff
@@ -762,6 +762,83 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+                style_id == SEARCH_STYLE_ID // 1 ... reserved style id for search results
+            });
```
cmyr: so to clarify: you're using the spans you've defined to represent the state of the find operation, and then for find next you just look through the spans? This feels a bit iffy to me, I guess in the sense of model/view boundaries. My first thought is that maybe there could have a new `FindResult` or `FindState` type, which you move through in `find_prev` and `find_next`? This might also reuse the `Selection` type: search is a sort of selection, especially when replace is added to the mix.

rust/core-lib/src/editor.rs
cmyr: so I guess for selections this is handled by the front-end. This should eventually be handled by some future `Theme` or `ColorScheme`. This seems okay, for now...

rkusa:
@cmyr I think this is an important part and I am glad you bring this up (because I kind of forgot that this is something that should be discussed). For my understanding, `Span`s are not just about styling, but also for carrying metadata (more to come, like soft spans).

The main reasons I went for just using spans were:

- need them for styling anyway
- they are already transformed during an edit

That is, introducing a new type would require to also transform this type after edits (to make the find result move with the cursor).

So I would have implemented a `FindResult` type using a BTree of ranges (1 range = 1 find result). Additionally, I would have applied the transformation on each edit. So in the end, I would have kind of copied Spans (but yes, maybe also very similar to the `Selection` type). Once we have something like span layers, so that find results have their own layer, it is even more like having a dedicated `FindResult` type.

I also understand the reasoning behind having a separate type (maybe reusing the `Selection` type) for search results. And now that I am writing this... for "replace all" re-using `Selection` could indeed be a good fit. However, I do not know the `Selection` infrastructure to well, especially how tightly coupled they are with the engine. For now, my feeling is still that `Span`s are a appropriate place for search highlights, but when I think of Spans I might already have the state of spans in mind that is yet to be implemented. Anyway, I'll have a look at the `Selection` type and reconsider my initial reasoning :)

rkusa:
I was rethinking it today, and agree that search highlights should not live inside the `Span`s of the editor. Mainly because they should be view-scoped and not buffer-scoped. That is, they live in a data structure within a `View`. But I am not sure what data structure to re-use. I am leaning towards re-using the `Spans` type, because it uses a BTree. However using the `Selection` type could ease implementing replace/replace-all functionality that will come eventually.

cmyr:
paging @raphlinus

rkusa:
I've moved the search highlights into the view re-using the `Selection` type.

rkusa:
I've updated the PR.

Find is now only executed in the current view region (unless there is no occurrence in the current view region, then view port is scrolled to the first occurrence).
When scrolling, find is executed in the new regions to provide highlights. It thereby keeps track of already searched regions.
On edits, regions around the edits are re-searched.

rust/core-lib/src/view.rs
```diff
@@ -517,6 +590,175 @@ impl View {
...
+                        if !wrapped && ix < occurrences.len() && occurrences[ix].end == sel.1 {
+                            ix += 1;
+                        }
+                        if ix >= occurrences.len() {
+                            None
+                        } else {
+                            Some(occurrences[ix].clone())
+                        }
+                    }
+                }
```
rkusa: - [ ] I've to admit that this block (above) is really ugly (in terms of readability and understandability). I may try to come up with a simplified logic or add some meaningful code comments ...

rkusa:
I've pulled the incremental search out of this PR (can be found here for now https://github.com/rkusa/xi-editor/commit/3f7debf3b275062a8d72cc9342e85834630e1119).

This PR is rdy for review.

eyelash:
I wonder whether `find` and `find_next` should really be two separate commands. I assume a front-end would only show one button for both.

rkusa:
> I wonder whether find and find_next should really be two separate commands. I assume a front-end would only show one button for both.

Thanks for the feedback! Besides reducing the amount of params that are send over the wire, there is also an actual use case. If you open the find bar that already contains a search string (from previous searches), the search should be re-done but without moving the cursor to the next occurrence. So in this case only the `find` command is called and not both `find` and `find_next`. We could consider re-naming the `find` command to `set_find` (or similar), if this makes the behaviour more clear.

rust/core-lib/src/view.rs
```diff
@@ -665,9 +665,10 @@ impl View {
     /// next occurrence before (`true`) or after (`false`) the last cursor is selected. `wrapped`
     /// indicates a search for the next occurrence past the end of the file. `stop_on_found`
     /// determines whether the search should stop at the first found occurrence (does only apply
-    /// to forward search (i.e. reverse = false).
+    /// to forward searc, i.e. reverse = false). If `allow_same` is set to `true` the current
+    /// selection is considered a valid next occurrence.
     pub fn select_next_occurrence(&mut self, text: &Rope, reverse: bool, wrapped: bool,
-                                  stop_on_found: bool) -> Option<usize>
+                                  stop_on_found: bool, allow_same: bool) -> Option<usize>
```
rkusa: I am not particularly proud that I had to add another boolean option here 😕 and I am wondering whether using flags for `reverse`, `wrapped`, `stop_on_found` and `allow_same` would be a better approach?

rust/core-lib/src/view.rs
```diff
@@ -665,9 +665,10 @@ impl View {
     /// next occurrence before (`true`) or after (`false`) the last cursor is selected. `wrapped`
     /// indicates a search for the next occurrence past the end of the file. `stop_on_found`
     /// determines whether the search should stop at the first found occurrence (does only apply
-    /// to forward search (i.e. reverse = false).
+    /// to forward searc, i.e. reverse = false). If `allow_same` is set to `true` the current
```
cmyr: typo

raphlinus:
Thanks so much for this, it's really impressive work, and a critically important feature.

PR #291, opened 2017-05-15
Start vim keybindings

raphlinus:
This is very much work in progress, with many of the bindings not
doing quite what they're supposed to, and many others missing, but
it's a start.

Progress toward #93


jakalope:
If you're interested, I have a project [ViXi](https://github.com/jakalope/vixi) that manages vim style remapping. I was thinking about integrating it into xi-tui but it would make a lot more sense to make it available to all front-ends by putting it in the core lib.

jansol:
I believe this should just be closed at this point? Discussion about a vi mode has been going in a quite different direction since this PR was opened.

PR #292, opened 2017-05-15
Identifiers are now real types

cmyr:
This patch makes better use of the type system to distinguish between `BufferIdentifier` and `ViewIdentifier`. 

There were a few spots in the plugin code where this was poorly distinguished, and this change makes me a bit more considered about those places. It also will make it easier to use some value type as an identifier in the future.

### DIFF
--- a/rust/core-lib/src/plugins/manager.rs
+++ b/rust/core-lib/src/plugins/manager.rs
@@ -44,7 +44,8 @@ impl <W: Write + Send + 'static>PluginManager<W> {
     /// Returns plugins available to this view.
     pub fn available_plugins(&self, view_id: &ViewIdentifier) -> Vec<ClientPluginInfo> {
         self.catalog.iter().map(|p| {
-            let key = (view_id.to_owned(), p.name.clone());
+            let buffer_id = self.buffer_for_view(view_id);
+            let key = (buffer_id, p.name.clone());
             let running = self.running.contains_key(&key);
             let name = key.1;
             ClientPluginInfo { name, running }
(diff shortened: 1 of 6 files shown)


rust/core-lib/src/plugins/manager.rs
```diff
@@ -110,43 +112,67 @@ impl <W: Write + Send + 'static>PluginManager<W> {
...
+                    Ok(plugin_ref) => {
+                        plugin_ref.initialize(&init_info);
+                        let mut inner = me.lock();
+                        let mut running = false;
+
+                        if let Some(ed) = inner.buffers.lock()
+                            .editor_for_view(&view_id2) {
+                                ed.plugin_started(&view_id2, &key.1);
+                                running = true;
+                            }
```
trishume: This close paren is mis-indented.

But more importantly, why did you change this from how it was before to add the `running` variable? Since `running` only becomes true in the `if` block, why not remove the `running` variable and put `inner.running.insert(key, plugin_ref)` inside the `if` block.

cmyr: hmm, I was doing it this way because `inner.running.insert` is a mutable borrow, and we have an immutable borrow out when we take `editor_for_view`. It's possible I'm missing something. Just moving it into the `if` block doesn't compile.

My editor wants to align that closing brace with the second line of the opening expression. I don't really have an opinion here, will change, but it might get reverted next time I fix whitespace 😐.

trishume: Darn borrow checker. Having to do dumb things like that for no good reason when using mutable borrows is one of my least favourite things about Rust.

PR #296, opened 2017-05-16
[RFC] Plugin commands

cmyr:
This RFC outlines a possible protocol for describing plugin commands, and making them available to the user.

### DIFF
--- a/doc/plugin_commands.md
+++ b/doc/plugin_commands.md
@@ -1,0 +1,102 @@
+
+## Summary
+
+A major feature of plugins is the ability to make new commands available to the user. This RFC outlines a preliminary system for describing these commands.
+
+Thank you to @rkusa for the discussion and feedback that has been incorporated into this RFC.
+
+### Notes on UI and implementation
+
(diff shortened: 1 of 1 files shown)


cmyr:
[Rendered](https://github.com/cmyr/xi-editor/blob/dab48814a62cd26ec5c3a6c80e6fc47c2fb6917c/doc/plugin_commands.md)

raphlinus:
This seems like a good direction to me. I was struck by how much this resembles the original gimp plugin system. There's a similar feeling of trying to project a nontrivial ui from the plugin and have it displayed by the main app. That's useful for a lot of things, but at some point if your plugin is super-rich it might want to display its own UI. I don't want to build out those bridges yet, but it might be good to keep it in mind as a future direction, as a bound on the scope of this effort.

That said, any generic method to display UI makes things much easier for plug-in authors to work across all front-ends.

trishume:
I like it. But then again, this is basically exactly like Sublime and all I seem to do in discussions is talk about what Sublime does and how it is great.

cmyr:
This is stale; will revisit at some point in the future, but no need to keep it open for eternity.

PR #297, opened 2017-05-16
Prepare Engine code for switch to separate tombstones

trishume:
This PR contains most of the refactoring work I did in order to switch `Engine` to use separate tombstones.

It contains the following changes:
- Improves documentation comments on a number of functions, including using Rust's example compilation checking.
- Add a `Mapper` API for mapping coordinates in string to coordinates in a `Subset`.
- Add a `transform_shrink` method to InsertDelta.
- Add `Debug` implementations for a number of structures.
- Move test helper functions that are conceivably useful in multiple places into their own private module.
- Add a bunch of tests for `Engine`. I also wrote tests for `undo` and `gc` but unfortunately tied the commits to my separation of tombstones so I couldn't easily re-order them here via rebase.

There's also one more refactoring which I think should be done and I could either add to this PR or create a new PR for: I think `Subset::transform_shrink` has its arguments in the wrong order. For `InsertDelta::transform_shrink` I had to use a different order because of the type difference, but I think that's indicative that `Subset::transform_shrink` has the wrong order.

`Subset::transform_expand` returns the subset from `self` transformed (expanded) by the parameter, and so does `InsertDelta::transform_shrink`, but `Subset::transform_shrink` currently returns the subset from the parameter transformed (shrunk) by `self`. This seems weird and counterintuitive for me, in fact when I called it with the wrong order because of this I didn't even notice it when I looked for the bug in exactly that code.

After this PR is merged I'll submit another PR with the rest of the work from the [separate-tombstones branch](https://github.com/google/xi-editor/compare/separate-tombstones)

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -21,6 +21,7 @@ use tree::{Node, NodeInfo, TreeBuilder};
 use subset::{Subset, SubsetBuilder};
 use std::cmp::min;
 use std::ops::Deref;
+use std::fmt;
 
 #[derive(Clone)]
 enum DeltaElement<N: NodeInfo> {
@@ -230,6 +231,24 @@ impl<N: NodeInfo> Delta<N> {
                 DeltaElement::Insert(ref n) => n.len()
             }
         )
+    }
+}
+
+impl<N: NodeInfo> fmt::Debug for Delta<N> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        try!(write!(f, "Delta("));
+        for el in &self.els {
+            match *el {
+                DeltaElement::Copy(beg,end) => {
+                    try!(write!(f, "[{},{}) ", beg, end));
+                }
+                DeltaElement::Insert(ref node) => {
+                    try!(write!(f, "<ins:{}> ", node.len()));
+                }
+            }
+        }
+        try!(write!(f, ")"));
+        Ok(())
     }
 }
 
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -134,25 +134,30 @@ impl<N: NodeInfo> Delta<N> {
     ///     assert_eq!(String::from(d2.apply(r)), String::from(d.apply(r)));
     /// }
     /// ```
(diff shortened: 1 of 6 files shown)


rust/rope/src/delta.rs
```diff
@@ -107,28 +115,50 @@ impl<N: NodeInfo> Delta<N> {
...
-                    last_ins = ins_ranges.next();
+                    last_old = old_ranges.next();
                 }
-                if last_ins.is_some() && last_ins.unwrap().0 <= beg {
-                    let (ib, ie) = last_ins.unwrap();
+                // If we have a range in the old text with the character at beg, then we Copy
+                if last_old.is_some() && last_old.unwrap().0 <= beg {
+                    let (ib, ie) = last_old.unwrap();
                     let end = min(e, ie);
+                    // Try to merge contigous Copys in the output
```
raphlinus: contiguous

rust/rope/src/delta.rs
```diff
@@ -423,37 +499,18 @@ impl<N: NodeInfo> Builder<N> {
 
 #[cfg(test)]
 mod tests {
-    use rope::{Rope, RopeInfo};
+    use rope::Rope;
     use delta::{Delta};
     use interval::Interval;
-    use subset::{Subset, SubsetBuilder};
+    use test_helpers::{find_deletions};
```
raphlinus: don't need the braces here (i know delta above has them)

rust/rope/src/engine.rs
```diff
@@ -118,11 +118,18 @@ impl Engine {
         self.find_rev(rev).map(|rev_index| self.rev_content_for_index(rev_index))
     }
 
-    /// A delta that, when applied to previous head, results in the current head. Panics
+    // TODO: doesn't this totally break in the presence of Undo currently?
```
raphlinus: I don't _think_ so

raphlinus: I don't _think_ so. Certainly it works in the case where the new revision is the result of changing the undo state.

rust/rope/src/test_helpers.rs
```diff
@@ -1,7 +1,24 @@
+// Copyright 2016 Google Inc. All rights reserved.
```
raphlinus: Should in general be current year for new files, but not worth blocking the merge over. i'll push on green

PR #299, opened 2017-05-16
Store separate tombstones and text instead of a union string

trishume:
This PR refactors the representation of `Engine` to store the text and tombstones separately. At the moment this has no significant effect on behaviour or time complexity, other than making retrieving the head revision instant. It paves the way however for a tree/rope based representation of subsets to make many other operations much faster.

This PR does the refactoring using changes from #297.

It also adds some more tests that weren't added in #297 due to me being too lazy to extricate them with advanced Git.

This PR should not be merged before #297 and you should probably review that first if you are thinking of reading this. I've targeted this PR at the branch of #297 so that the diff on top of that is easier to look at.

@raphlinus

### DIFF
--- a/rust/rope/src/engine.rs
+++ b/rust/rope/src/engine.rs
@@ -25,12 +25,15 @@ use rope::{Rope, RopeInfo};
 use subset::Subset;
 use delta::Delta;
 
+#[derive(Debug)]
 pub struct Engine {
     rev_id_counter: usize,
-    union_str: Rope,
+    text: Rope,
+    tombstones: Rope,
     revs: Vec<Revision>,
 }
 
+#[derive(Debug)]
 struct Revision {
     rev_id: usize,
     deletes_from_union: Subset,
@@ -40,6 +43,7 @@ struct Revision {
 
 use self::Contents::*;
 
+#[derive(Debug)]
 enum Contents {
     Edit {
         priority: usize,
@@ -62,7 +66,8 @@ impl Engine {
         };
         Engine {
             rev_id_counter: 1,
-            union_str: initial_contents,
+            text: initial_contents,
+            tombstones: Rope::default(),
             revs: vec![rev],
         }
     }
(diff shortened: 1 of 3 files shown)


rust/rope/src/engine.rs
```diff
@@ -110,7 +119,7 @@ impl Engine {
 
     /// Get text of head revision.
     pub fn get_head(&self) -> Rope {
-        self.rev_content_for_index(self.revs.len() - 1)
+        self.text.clone()
```
raphlinus: Sweet! Make me wonder though if the signature of this should be changed to `&Rope`

rust/rope/src/engine.rs
```diff
@@ -140,11 +149,14 @@ impl Engine {
         }
 
         let head_rev = &self.revs.last().unwrap();
-        Delta::synthesize(&self.union_str, &prev_from_union, &head_rev.deletes_from_union)
+        // TODO: this does 3 calls to Delta::synthesize and 2 to apply, this should definitely be better.
+        let old_tombstones = Engine::shuffle_tombstones(&self.text, &self.tombstones, &head_rev.deletes_from_union, &prev_from_union);
+        Delta::synthesize(&old_tombstones, head_rev.union_str_len, &prev_from_union, &head_rev.deletes_from_union)
     }
 
+    // TODO: don't construct transform if subsets are empty
```
raphlinus: Could use doc explaining the meaning of the two `Ropes` in the return.

rust/rope/src/engine.rs
```diff
@@ -170,41 +182,67 @@ impl Engine {
...
+        // Taking the complement of deletes_from_union leads to an interleaving valid for swapped text and tombstones,
+        // allowing us to use the same method to insert the text into the tombstones.
+        let union_len = text.len() + tombstones.len();
+        let inverse_tombstones_map = old_deletes_from_union.complement(union_len);
+        let move_delta = Delta::synthesize(text, union_len,
+            &inverse_tombstones_map, &new_deletes_from_union.complement(union_len));
+        move_delta.apply(tombstones)
+    }
+
+    /// Move sections from text to tombstones and vice versa based on a new and old set of deletions
```
raphlinus: Again, explain the meaning of the two `Rope`s on return

rust/rope/src/engine.rs
trishume: Isn't `Rope::clone()` super cheap? I assumed it was basically just an `Arc::clone()` but I admit I didn't actually check. But even if that's the case it probably would make more sense to let the user do that if necessary.

raphlinus: Yes, it's cheap.

At the back of my head I'm probably thinking about the impact on the "mutate in place" optimization - if a clone of the text is retained it could defeat that optimization, while returning &Rope would force the user to drop the reference before doing any mutating change. But I'm not sure that's an issue in practice.

trishume: Okay done, at the moment every non-test reference to `get_head` now immediately clones it, but that could change in the future.

PR #300, opened 2017-05-17
[WIP] Fuzz testing for xi-rope

trishume:
This PR adds a fuzz testing framework for xi-rope with two starter fuzz tests. The first one just tests that single edits work, the second tests that linear sequences of edits involving undo work.

However, this is totally blocked on https://github.com/rust-lang/rust/issues/39882 / https://github.com/rust-fuzz/cargo-fuzz/issues/81 because undo uses `BTreeSet` and due to that bug any code involving BTreeSet will immediately turn up false positive address sanitizer errors.

I'm pretty sure the `op_seq` fuzz test is correct, but I can't test it until that issue is fixed. I don't think this PR should be merged before then.

cc @raphlinus

### DIFF
--- a/rust/rope/fuzz/.gitignore
+++ b/rust/rope/fuzz/.gitignore
@@ -1,0 +1,4 @@
+
+target
+corpus
+artifacts
--- a/rust/rope/fuzz/Cargo.lock
+++ b/rust/rope/fuzz/Cargo.lock
@@ -1,0 +1,53 @@
+[root]
+name = "xi-rope-fuzz"
+version = "0.0.1"
+dependencies = [
+ "libfuzzer-sys 0.1.0 (git+https://github.com/rust-fuzz/libfuzzer-sys.git)",
+ "xi-rope 0.2.0",
+]
+
+[[package]]
+name = "bytecount"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "gcc"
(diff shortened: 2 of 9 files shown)


mqzry:
The rust  issue (rust-lang/rust#39882 ) is closed by now. This might be fixed, needs triage.

trishume:
I confirm I've done the CLA and consent to this work being included.

cmyr:
🤖💥🔫

cmyr:
@trishume so I have this working again, and I'm not totally sure what's going on in the op_seq test. No pressure, but if you'd like to take a quick look I wouldn't complain?

trishume:
I unfortunately don't remember anything about this really, I took a quick look and nothing jogged my memory. I don't think I'd be any more effective than you. 😛

cmyr:
Yep fair enough. will poke around a bit and see if I can make sense of it.

cmyr:
Okay, thinking about this more:

1. fuzzing is definitely something we will want to explore at some point
2. It isn't totally clear where and how best to apply fuzzing, right now; we don't have an obvious entrance point (although actually fuzzing the RPC surface would be interesting)
3. When we _do_ do this, it probably makes the most sense to keep it out-of-band; that is, we can just have an external 'xi-fuzzing' repo we can work from, so that we aren't committing corpuses and other test data to the main repo


And with that in mind, I'm going to close this; we'll keep the branch around for future reference.

PR #304, opened 2017-05-19
Switch to multi-set representation of Subset

trishume:
This PR adds a new `multiset` module that mostly replicates the interface of the `subset` module but with a new representation.

It now is a multi-subset in that each element has a count that represents how many times it is in the `Subset` as segments with a length and a count. These segments cover the whole base document and there are asserts in the transforms that check these properties. So now you no longer need to store and pass `base_len` separately, and it catches more mistakes.

This PR just adds it alongside the old `subset` module and includes `quickcheck` tests that I've ran 100,000 times each that check its behaviour against the old module. I have [a branch](https://github.com/google/xi-editor/tree/remove-old-set) that removes these tests and the old representation that I'll PR after this is merged. The disadvantage is that you don't see `multiset.rs` as a diff from `subset.rs`, but that's not too bad since basically every single line of code was rewritten except the doc comments and the tests.

The transforms are completely re-done from scratch without looking at the previous implementations, don't bother digging for them to try and compare the changes. I implemented them using a different style that hopefully should be easier to understand and now they share more code. This also means the quickcheck comparisons are more credible since the two implementations are less likely to share bugs. Surprisingly, it at least appears like I got all the transforms right on the first try, except for a single copy-paste error with a one character fix. I'm not sure whether to credit myself, or to credit Rust for approaching "compiles iff correct".

This PR also doesn't refactor Engine to take advantage of the new functionality for reduced storage or increased efficiency. That's next on my todo list, perhaps after swapping the argument order of `transform_shrink` to actually make sense.

@raphlinus

### DIFF
--- a/rust/rope/src/multiset.rs
+++ b/rust/rope/src/multiset.rs
@@ -55,65 +55,46 @@ impl SubsetBuilder {
     /// `Subset` corresponds to the document length.
     pub fn pad_to_len(&mut self, total_len: usize) {
         if total_len > self.total_len {
-            // might be able to extend last segment
-            if let Some(last) = self.segments.last_mut() {
-                if last.count == 0 {
-                    last.len += total_len - self.total_len;
-                    self.total_len = total_len;
-                    return;
-                }
-            }
-
-            self.segments.push(Segment {len: total_len - self.total_len, count: 0});
-            self.total_len = total_len;
+            let cur_len = self.total_len;
+            self.push_segment(total_len - cur_len, 0);
         }
     }
 
     /// Sets the count for a given range. This method must be called with a
-    /// non-empty range with beg not before the largest range or segment added
+    /// non-empty range with `begin` not before the largest range or segment added
     /// so far. Gaps will be filled with a 0-count segment.
-    pub fn add_range(&mut self, beg: usize, end: usize, count: usize) {
-        assert!(beg >= self.total_len, "ranges must be added in non-decreasing order");
-        // assert!(beg < end, "ranges added must be non-empty: [{},{})", beg, end);
-        if beg >= end {
+    pub fn add_range(&mut self, begin: usize, end: usize, count: usize) {
+        assert!(begin >= self.total_len, "ranges must be added in non-decreasing order");
+        // assert!(begin < end, "ranges added must be non-empty: [{},{})", begin, end);
+        if begin >= end {
             return;
         }
-        let len = end-beg;
-
-        // merge into previous segment if possible
-        if let Some(last) = self.segments.last_mut() {
-            if last.count == count && self.total_len == beg {
-                last.len += len;
-                self.total_len += len;
-                return;
-            }
-        }
+        let len = end-begin;
+        let cur_total_len = self.total_len;
 
         // add 0-count segment to fill any gap
-        if beg > self.total_len {
-            self.segments.push(Segment {len: beg - self.total_len, count: 0});
+        if begin > self.total_len {
+            self.push_segment(begin - cur_total_len, 0);
         }
 
-        self.segments.push(Segment {len, count});
-        self.total_len = end;
+        self.push_segment(len, count);
     }
 
     /// Assign `count` to the next `len` elements in the string.
     /// Will panic if called with `len==0`.
     pub fn push_segment(&mut self, len: usize, count: usize) {
         assert!(len > 0, "can't push empty segment");
+        self.total_len += len;
 
         // merge into previous segment if possible
         if let Some(last) = self.segments.last_mut() {
             if last.count == count {
                 last.len += len;
-                self.total_len += len;
                 return;
             }
         }
 
         self.segments.push(Segment {len, count});
-        self.total_len += len;
     }
 
     pub fn build(self) -> Subset {
@@ -285,8 +266,7 @@ impl Subset {
     /// the count for both self and other in that range. The two `Subset`s
     /// must have the same total length.
     ///
-    /// Each returned `ZipSegment` will be have a differing count iff
-    /// both input `Subset`s change count at every `Segment` boundary.
+    /// Each returned `ZipSegment` will differ in at least one count.
     pub fn zip<'a>(&'a self, other: &'a Subset) -> ZipIter<'a> {
         ZipIter {
             a_segs: self.segments.as_slice(),
@@ -420,7 +400,8 @@ impl<'a> Mapper<'a> {
     /// the total cost to be O(n) where `n = max(calls,ranges)` over all times
     /// called on a single `Mapper`.
     pub fn doc_index_to_subset(&mut self, i: usize) -> usize {
-        assert!(i >= self.last_i, "method must be called with i in non-decreasing order. i={}<{}=last_i", i, self.last_i);
+        assert!(i >= self.last_i,
+            "method must be called with i in non-decreasing order. i={}<{}=last_i", i, self.last_i);
         self.last_i = i;
 
         while i >= self.cur_range.1 {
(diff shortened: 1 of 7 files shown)


rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+            "12346ABCGHIJOQRSTUVWXYZcefghijklmnopqrstuvxz");
+        transform_case("12456789ABCEFGJKLMNPQRSTUVXYadefghkrtwxz", "15ACEFGKLPRUVYdhrtx",
+            "0135ACDEFGHIKLOPRUVWYZbcdhijlmnopqrstuvxy");
+        transform_case("0128ABCDEFGIJMNOPQXYZabcfgijkloqruvy", "2CEFGMZabijloruvy",
+            "2345679CEFGHKLMRSTUVWZabdehijlmnoprstuvwxyz");
+        transform_case("01245689ABCDGJKLMPQSTWXYbcdfgjlmnosvy", "01245ABCDJLQSWXYgsv",
+            "0123457ABCDEFHIJLNOQRSUVWXYZaeghikpqrstuvwxz");
+    }
+
+    use subset;
```
trishume: New test code starts here, all previous tests are unchanged except for updating based on the API changes.

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+                }
+            }
+
+            self.segments.push(Segment {len: total_len - self.total_len, count: 0});
+            self.total_len = total_len;
+        }
+    }
+
+    /// Sets the count for a given range. This method must be called with a
+    /// non-empty range with beg not before the largest range or segment added
```
raphlinus: Say `` `beg` ``, especially given that it's an English word.

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// so far. Gaps will be filled with a 0-count segment.
+    pub fn add_range(&mut self, beg: usize, end: usize, count: usize) {
+        assert!(beg >= self.total_len, "ranges must be added in non-decreasing order");
+        // assert!(beg < end, "ranges added must be non-empty: [{},{})", beg, end);
+        if beg >= end {
+            return;
+        }
+        let len = end-beg;
+
+        // merge into previous segment if possible
```
raphlinus: I think you could just get rid of this, the following two calls to `push` take care of this case.

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+        if let Some(last) = self.segments.last_mut() {
+            if last.count == count {
+                last.len += len;
+                self.total_len += len;
+                return;
+            }
+        }
+
+        self.segments.push(Segment {len, count});
+        self.total_len += len;
```
raphlinus: Can move up, this happens in both branches.

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+}
+
+impl SubsetBuilder {
+    pub fn new() -> SubsetBuilder {
+        SubsetBuilder::default()
+    }
+
+    /// Intended for use with `add_range` to ensure the total length of the
+    /// `Subset` corresponds to the document length.
+    pub fn pad_to_len(&mut self, total_len: usize) {
```
raphlinus: Pretty sure this can just be replaced by a `push` of a 0-count segment (if len>0).

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+    pub fn is_empty(&self) -> bool {
+        (self.segments.is_empty()) || ((self.segments.len() == 1) && (self.segments[0].count == 0))
+    }
+
+    /// Compute the union of two subsets. The count of an element in the
+    /// result is the sum of the counts in the inputs.
+    pub fn union(&self, other: &Subset) -> Subset {
+        let mut sb = SubsetBuilder::new();
+        for zseg in self.zip(other) {
+            sb.push_segment(zseg.len, zseg.a_count + zseg.b_count);
```
raphlinus: Ok this is nice :)

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// Semantically iterates the ranges of the complement of this `Subset`.
+    pub fn complement_iter(&self) -> RangeIter {
+        self.range_iter(CountMatcher::Zero)
+    }
+
+    /// Return an iterator over `ZipSegment`s where each `ZipSegment` contains
+    /// the count for both self and other in that range. The two `Subset`s
+    /// must have the same total length.
+    ///
+    /// Each returned `ZipSegment` will be have a differing count iff
```
raphlinus: Isn't this an invariant of Subset? So you can state it as a guarantee here?

rust/rope/src/multiset.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// return one more than the largest index in the subset (i.e the length).
+    /// This behaviour is suitable for mapping closed-open intervals in a
+    /// string to intervals in a subset of the string.
+    ///
+    /// In order to guarantee good performance, this method must be called
+    /// with `i` values in non-decreasing order or it will panic. This allows
+    /// the total cost to be O(n) where `n = max(calls,ranges)` over all times
+    /// called on a single `Mapper`.
+    pub fn doc_index_to_subset(&mut self, i: usize) -> usize {
+        assert!(i >= self.last_i, "method must be called with i in non-decreasing order. i={}<{}=last_i", i, self.last_i);
```
raphlinus: Break long line

raphlinus:
Some general comments on this:

1. I'm not yet positive that we want to get rid of the old Subset; it's still valid for the `ins` and `del` sets in the log. The multi-set is primarily needed for the `from_union`. This data structure is strictly more general, but also takes 2x the space.

Keeping both is somewhat annoying, though. One approach to reducing code size is making a common SegmentIterator that just alternates 0 and 1 for the count in the (single) Subset case, and writing code in terms of that SegmentIterator (basically all the methods that work by doing a linear scan).

Going forward, I think the use in `ins` and `del` would retain a simple Vec-based implementation, but the multi-set would become a rope.

I don't have a super strong feeling about any of this; all other things equal, having fewer data structures is better than more. When we start talking optimization (esp things like RAM use) maybe it's better to focus on really fine-tuning the more general data structure, rather than having variants that are somewhat specialized. I mostly write this to illuminate the design choices.

2. I think you might want to store cumulative length rather than segment length. The main advantage of this is that query can be O(log n).

trishume:
@raphlinus 

### Re keeping old `Subset`:

That's some new information. In retrospect it makes sense that `ins` and `del` don't have to be ropes, but part of why I unified them was so that only one data structure would need to be rope-ified.

There's two major reasons why I unified them that are still valid though:

- This refactoring also eliminates the need for `base_len` and checks transforms against the size of the document. This benefits operations on simple `Subset`s as well.
- If both data structures are kept, I foresee having to write 1-2 new implementations for each transform that transform a subset by a multiset and vice versa.

Like all problems though, I think this could be solved by adding an additional level of abstraction.  `Subset` could take a type parameter conforming to a `Segment` trait with `len` and `count` methods. There would be two implementations, one which stored a length and count, and one which stored just a length, and then somehow stored a `first_segment_in_subset: bool` in the `Subset` (probably another type parameter) and computed the count from that and the index. Then all the various combination operations necessary would be implemented by the compiler through monomorphiziation. I'm not sure this is worth doing any time soon though, since most usages of the code would change straightforwardly it could be done at any time.

### Re storing cumulative length:

I can do that, it might even make some of the code simpler, but I'm not sure yet. Would this representation also work better for turning it into a Rope later?

rust/rope/src/multiset.rs
trishume: Done

raphlinus:
Thanks!

On old subset: I'm swayed by the argument of less coding work.

On cumulative length: I think it's basically orthogonal. In the Leaf of a rope, cumulative lengths would be relative to the start of the leaf. So it's actually easier to leave it as is. If we're switching to a rope, then everything becomes O(log n) so the optimization isn't needed. I'm happy deferring all that to later.

PR #305, opened 2017-05-21
Make more editing commands multi-select aware

raphlinus:
The following editing commands are now based on Selection rather than
the old single cursor:

* deletions
* cut/copy
* transpose
* select all
* insert tab
* do_click

In addition, the cursor fix-up logic in apply_plugin_edit has been
removed for the time being. The correct handling is to fully implement
priority, rather than do an explicit modification of the selection.

There are a number of TODOs remaining, and this patch does not delete
the old selection state, but at this point all editing operations
should be updated not to rely on it.

Progress towards #188

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -24,12 +24,12 @@ use serde_json::Value;
 use xi_rope::rope::{LinesMetric, Rope, RopeInfo};
 use xi_rope::interval::Interval;
 use xi_rope::delta::{self, Delta, Transformer};
-use xi_rope::tree::Cursor;
 use xi_rope::engine::Engine;
 use xi_rope::spans::{Spans, SpansBuilder};
 use view::{Style, View};
 use word_boundaries::WordCursor;
-use movement::Movement;
+use movement::{Movement, region_movement};
+use selection::{Affinity, Selection, SelRegion};
 
 use tabs::{ViewIdentifier, DocumentCtx};
 use rpc::{EditCommand, GestureType};
@@ -241,12 +241,12 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.view.set_old_sel_dirty();
     }
 
-    /// Applies a simple edit to the text, and updates undo state.
-    ///
-    /// See `add_delta` for more details.
-    fn add_simple_edit(&mut self, iv: Interval, new: Rope) {
-        let delta = Delta::simple_edit(iv, new, self.text.len());
-        self.add_delta(delta);
+    /// Sets the selection to a single region, and scrolls the end of that
+    /// region into view.
+    fn set_sel_single_region(&mut self, region: SelRegion) {
+        let mut sel = Selection::new();
+        sel.add_region(region);
+        self.scroll_to = self.view.set_selection(&self.text, sel);
     }
 
     /// Applies a delta to the text, and updates undo state.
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -693,20 +758,15 @@ impl<W: Write + Send + 'static> Editor<W> {
...
+        let deletions = self.view.sel_regions().to_vec();
+        self.delete_sel_regions(&deletions);
+        result
     }
 
-    fn do_copy(&mut self) -> Value {
-        if self.view.sel_start != self.view.sel_end {
-            let val = self.text.slice_to_string(self.view.sel_min(), self.view.sel_max());
+    fn do_copy(&self) -> Value {
+        if let Some(val) = self.extract_sel_regions(self.view.sel_regions()) {
```
cmyr: I don't love how this handles paste after multi-sel copy. It feels like, in that instance, the individual regions should be pasted, at least if the number of regions is equal.

It feels like there should be some intermediate representation, that only becomes a string when paste context is known?

rust/core-lib/src/editor.rs
```diff
@@ -732,44 +792,40 @@ impl<W: Write + Send + 'static> Editor<W> {
...
-        }
-
-        self.doc_ctx.set_kill_ring(Rope::from(val));
+        self.delete_by_movement(Movement::EndOfParagraphKill, true);
     }
 
     fn yank(&mut self) {
+        // TODO: if there are multiple cursors and the number of newlines
+        // is one less than the number of cursors, split and distribute one
+        // line per cursor.
```
cmyr: okay, addresses part of my earlier Q. 

🤔 what about newlines that are _part_ of the copied text?

rust/core-lib/src/movement.rs
```diff
@@ -49,6 +49,8 @@ pub enum Movement {
     StartOfParagraph,
     /// Move to the end of the text line.
     EndOfParagraph,
+    /// Move to the end of the text line, or next line if already at end.
+    EndOfParagraphKill,
```
cmyr: Is this about accommodating cocoa's weird two-selector behaviour? If so I have a PR to reopen..

rust/core-lib/src/editor.rs
```diff
@@ -431,18 +419,90 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
     }
 
+    /// Common logic for a number of delete methods. For each region in the selection,
+    /// if the selection is a caret, delete the region between the caret and the
+    /// movement applied to the caret, otherwise delete the region.
+    ///
+    /// If `save` is set, save the deleted text into the kill ring.
+    fn delete_by_movement(&mut self, movement: Movement, save: bool) {
+        // We compute deletions as a selection because the merge logic is convenient.
```
raphlinus: Yes, totally reasonable. The only hesitation is that it's _just_ the intervals, so horiz and affinity have no role. This is the familiar question of a small number of general types, or types specialized to their actual role. In this case, the value of specializing is marginal, and being able to reuse the merge logic is a win.

rust/core-lib/src/editor.rs
raphlinus: Yes, should probably be a TODO like below, or in any case captured in an issue.

rust/core-lib/src/editor.rs
raphlinus: So that's an interesting question. One compelling reason to use a string as the intermediate representation rather than something with more structure is that it can use the plain text type, and interoperate with other things. My sense is that the cases where you need the structure are going to be pretty marginal. It's worth considering for the long term though.

rust/core-lib/src/movement.rs
raphlinus: No, it's just that these two movement variations do two different things when the cursor is at the end of the line.

FWIW, we'll probably have a lot more movements to take care of vi.

rust/core-lib/src/editor.rs
trishume: I actually like the fact that the Sublime paste buffer is just a string, it makes it easier to reason about and in the past I've actually used the ability to copy N lines of text as one selection and then paste the lines in N different places/cursors.

PR #307, opened 2017-05-21
Update doc to show delete_forward

pforpallav:


raphlinus:
Thanks, always good to bring the docs up to date.

PR #309, opened 2017-05-22
Swap Subset::transform_shrink receiver and argument

trishume:
transform_expand, transform_union and Delta::transform_shrink all take the
transform as the parameter and return something with the changes of self.
Before this commit, transform_shrink was different for no apparent reason.

This weirdness has caused me lots of extra thinking and a couple bugs,
this commit fixes it.

I don't know why the argument order was what I see as reversed to begin with, perhaps @raphlinus might know why.


raphlinus:
Yes, this was simply a case of me being confused. I've had it both ways at various times in the past, but somehow convinced myself that the transform as receiver was better. I have no confidence in that reasoning though, and agree that this is more consistent.

PR #310, opened 2017-05-23
Make travis builds fail reliably

raphlinus:
The build was failing to fail because bash error handling is not
intuitive.


trishume:
This PR made the tests fail, which means the fix worked. LGTM 😛

PR #312, opened 2017-05-23
Only store deletes_from_union in Engine

trishume:
This PR refactors `Engine` to store `deletes_from_union` only for the head revision in `Engine` rather than in every revision. It makes `Undo` reversible and then uses the reversibility of operations to compute the `deletes_from_union` for any revision by inverting changes from the head revision backwards.

I made `Undo` reversible by representing it as two symmetric differences / xors that can be applied in exactly the same way either direction and computed using the same operation. For undo groups this is `BTreeSet::symmetric_difference` and it conceptually represents the set of undo groups toggled between done and undone. 

Since `deletes_from_union` is a multiset with counts, it's slightly weirder but still simple. You could think of it like a set of toggled deletes if it was a normal set, but instead it is the bitwise xor of the counts for elements. This has the same algebraic xor properties that let it be used as a symmetric difference, and it has counts of `0` where there is no difference, but for elements that differ the counts don't have a nice human meaning.

I haven't yet re-written `compute_undo` to be more efficient, I'll do that in a different PR, but for now I preserved the correctness of `gc` so everything should still work.

In an intermediate step, I instrumented all the tests to check the old `deletes_from_union` stored on the `Revision` against the new computation, and it found no differences for any revision in any test, but I reverted that commit and then removed the old field. The commits are un-squashed so we can look in the git history if we ever want to check that or similar things again.

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -270,7 +270,7 @@ impl<N: NodeInfo> InsertDelta<N> {
     //
     // TODO: write accurate equations
     // TODO: can we infer l from the other inputs?
-    pub fn transform_expand(&self, xform: &Subset, l: usize, after: bool) -> InsertDelta<N> {
+    pub fn transform_expand(&self, xform: &Subset, after: bool) -> InsertDelta<N> {
         let cur_els = &self.0.els;
         let mut els = Vec::new();
         let mut x = 0;  // coordinate within self
@@ -279,6 +279,7 @@ impl<N: NodeInfo> InsertDelta<N> {
         let mut b1 = 0;
         let mut xform_ranges = xform.complement_iter();
         let mut last_xform = xform_ranges.next();
+        let l = xform.count(CountMatcher::All);
         while y < l || i < cur_els.len() {
             let next_iv_beg = if let Some((xb, _)) = last_xform { xb } else { l };
             if after && y < next_iv_beg {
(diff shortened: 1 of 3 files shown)


rust/rope/src/delta.rs
```diff
@@ -270,7 +270,7 @@ impl<N: NodeInfo> InsertDelta<N> {
     //
     // TODO: write accurate equations
     // TODO: can we infer l from the other inputs?
```
raphlinus: This TODO has been resolved, can be deleted

rust/rope/src/engine.rs
```diff
@@ -147,31 +168,30 @@ impl Engine {
...
-        let old_tombstones = Engine::shuffle_tombstones(&self.text, &self.tombstones, &head_rev.deletes_from_union, &prev_from_union);
-        Delta::synthesize(&old_tombstones, &prev_from_union, &head_rev.deletes_from_union)
+        // TODO: this does 2 calls to Delta::synthesize and 1 to apply, this probably could be better.
+        let old_tombstones = Engine::shuffle_tombstones(&self.text, &self.tombstones, &self.deletes_from_union, &prev_from_union);
+        Delta::synthesize(&old_tombstones, &prev_from_union, &self.deletes_from_union)
     }
 
     // TODO: don't construct transform if subsets are empty
     /// Retuns a tuple of a new `Revision` representing the edit based on the
     /// current head, a new text `Rope`, and a new tombstones `Rope`.
```
raphlinus: Return value needs to be updated.

rust/rope/src/multiset.rs
```diff
@@ -180,6 +180,28 @@ impl Subset {
...
+    pub fn subtract(&self, other: &Subset) -> Subset {
+        let mut sb = SubsetBuilder::new();
+        for zseg in self.zip(other) {
+            assert!(zseg.a_count >= zseg.b_count, "can't subtract {} from {}", zseg.a_count, zseg.b_count);
+            sb.push_segment(zseg.len, zseg.a_count - zseg.b_count);
+        }
+        sb.build()
+    }
+
+    /// Compute the bitwise xor of two subsets, useful as a symmetric
```
raphlinus: This comment confused me a bit. In the case where the inputs are not multi (ie guaranteed to be 0 or 1), then this is a traditional symmetric difference. But afaict it's not primarily used in this context, deletes_from_union doesn't in general have this property. The explanation above (that it's a reversible difference) is good. I'm not sure it needs a full explanation about the difference between the multi and non-multi cases, but I think including the magic word "reversible" would help.

PR #313, opened 2017-05-24
Better Undo

trishume:
`compute_delete` no longer takes time proportional to the length of the history.
Now it only takes time proportional to how far in the past the earliest
toggled undo group is.

`Revision` now stores the largest undo group number seen up to that point, and works backwards from the head until it finds a revision that can't have any of the toggled undo groups before it, then back-computes the `deletes_from_union` at that point, and re-computes the `deletes_from_union` forward from that point with the new undos.

This is now `O(n)` in how far back the earliest undo group is, rather than `O(n)` in the length of the history. That still means spamming `cmd+z` `n` times is still `O(n^2)`, but I tested it in-editor and it isn't noticeable at all even when I crank up MAX_UNDOS and spam 200+ undos.

In the future, this could be made `O(log n)` in how far back in history it is by using binary search to find the base revision, and then using a skip list or tree data structure to store coordinate transforms and undo groups to skip over large swathes.

This PR is currently targeted at #312 and should probably be reviewed after it, and definitely should be merged after it.

### DIFF
--- a/rust/rope/src/engine.rs
+++ b/rust/rope/src/engine.rs
@@ -39,6 +39,9 @@ pub struct Engine {
 #[derive(Debug)]
 struct Revision {
     rev_id: usize,
+    /// The largest undo group number of any edit in the history up to this
+    /// point. Used to optimize undo to not look further back.
+    max_undo_so_far: usize,
     edit: Contents,
 }
 
@@ -68,6 +71,7 @@ impl Engine {
         let rev = Revision {
             rev_id: 0,
             edit: Undo { toggled_groups: BTreeSet::new(), deletes_bitxor: deletes_from_union.clone() },
+            max_undo_so_far: 0,
         };
         Engine {
             rev_id_counter: 1,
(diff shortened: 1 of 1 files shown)


trishume:
I didn't bump `MAX_UNDOS` in this PR, in case we want to keep it low just to passively test garbage collection, but theoretically GC could now be disabled without using too much memory or time. Since revs no longer store `deletes_from_union`, tens of thousands of edits still won't use much memory.

Bumping up `MAX_UNDOS` to a number in the hundreds could work, but `gc` is still linear in the length of the history, or possibly worse in some cases I dunno.

Also, I discovered some bugs in xi-mac while testing: https://github.com/google/xi-mac/issues/44  https://github.com/google/xi-mac/issues/43

rust/rope/src/engine.rs
```diff
@@ -39,6 +39,9 @@ pub struct Engine {
 #[derive(Debug)]
 struct Revision {
     rev_id: usize,
+    /// The largest undo group number of any edit in the history up to this
```
raphlinus: This works but my idea is to use the earliest revision number in an undo group as the (otherwise opaque) undo group identifier. Not 100% clear to me now which approach is better.

trishume: For now that approach to allocating numbers is compatible with this approach, and if anyone ever wants to optimize memory and we can guarantee that revisions are allocated in that way, removing this field is an easy optimization.

Issue #270, opened 2017-05-08
Scrolling causes a large copy ops, never invalidating, causing line cache to grow unbounded


bvinc:
If you would like a copy of this text file.  I downloaded it at http://norvig.com/big.txt .

```sh
$ cat test2
{"id":0,"method":"new_view","params":{ "file_path": "/home/brain/src/gxi/big.txt"}}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[0,10] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100000,100010] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100001,100011] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100002,100012] }}
{"method":"edit","params": {"view_id": "view-id-1", "method": "scroll", "params":[100003,100013] }}

$ cat test2 | xi-core
{"id":0,"result":"view-id-1"}
{"method":"update","params":{"update":{"ops":[{"lines":[{"styles":[],"text":"The Project Gutenberg EBook of The Adventures of Sherlock Holmes\n"},{"styles":[],"text":"by Sir Arthur Conan Doyle\n"},{"styles":[],"text":"(#15 in our series by Sir Arthur Conan Doyle)\n"},{"styles":[],"text":"\n"},{"styles":[],"text":"Copyright laws are changing all over the world. Be sure to check the\n"},{"styles":[],"text":"copyright laws for your country before downloading or redistributing\n"},{"styles":[],"text":"this or any other Project Gutenberg eBook.\n"},{"styles":[],"text":"\n"},{"styles":[],"text":"This header should be the first thing seen when viewing this Project\n"},{"styles":[],"text":"Gutenberg file.  Please do not remove it.  Do not change or edit the\n"},{"styles":[],"text":"header without written permission.\n"},{"styles":[],"text":"\n"}],"n":12,"op":"ins"},{"n":12,"op":"skip"},{"n":128446,"op":"copy"}],"pristine":true},"view_id":"view-id-1"}}
[...]

PR #314, opened 2017-05-24
Invalidate both for selection dirty and text changes

raphlinus:
We need to send a full invalidation even when only the selection
changes (although this will change with minimal invalidation).
Previously, these changes set `View::dirty` but with multi selection
it's being tracked separately

Fixes google/xi-mac#43

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -293,6 +293,9 @@ impl View {
         first_line: usize, last_line: usize)
     {
         let dirty = self.dirty || self.sel_dirty;
+        if dirty {
+            self.valid_lines.clear();
+        }
         let height = self.offset_to_line_col(text, text.len()).0 + 1;
         let last_line = min(last_line, height);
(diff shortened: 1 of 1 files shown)


eyelash:
It looks like these changes regress the behaviour that was introduced in 3af52ce609a97ddf2f5341ff73ce99548ce89faf for clients that rely on the fact that the core automatically keeps the scroll region up to date in the client's cache.

eyelash:
Your latest commit fixes the issue I mentioned above for xi-gtk :tada:

raphlinus:
I think #270 captures the work of (2) above and can be used as a tracking issue.

For (1), yes, there should be a tracking issue. It's complicated (I wonder if we should have a "hard" tag to go along with "easy") but at least has the nice property of being isolated to a fairly small section of the code (dirty tracking and rendering) and having no observable changes other than performance.

The issues are largely separate, but minimal invalidation will have to respect whatever policy we have in place for the line cache.

PR #320, opened 2017-05-26
implement add_selection_above and add_selection_below

eyelash:
The behaviour approximately matches the behaviour of Sublime Text (with the exception that it adds a cursor at the beginning of the document if you call `add_selection_above` when there is a cursor on the first line). I decided for this approach because it was simple to implement, but generally I'd prefer Atom's behaviour (if you have a selection that is not a caret and call `add_selection_below`, Atom will duplicate that selection and Sublime Text will only add a caret on the line below).

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -431,13 +431,7 @@ impl<W: Write + Send + 'static> Editor<W> {
         let mut deletions = Selection::new();
         for r in self.view.sel_regions() {
             if r.is_caret() {
-                let (offset, horiz) = region_movement(movement, r, &self.view, &self.text, true);
-                let new_region = SelRegion {
-                    start: r.start,
-                    end: offset,
-                    horiz: horiz,
-                    affinity: Affinity::default(),
-                };
+                let new_region = region_movement(movement, r, &self.view, &self.text, true);
                 deletions.add_region(new_region);
             } else {
                 deletions.add_region(r.clone());
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -587,6 +587,16 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.view.select_all(self.text.len());
     }
 
+    fn add_selection_by_movement(&mut self, movement: Movement) {
+        let mut sel = Selection::new();
+        for region in self.view.sel_regions() {
+            sel.add_region(region.clone());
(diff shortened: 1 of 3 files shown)


raphlinus:
This looks pretty good. @cmyr wanna have a look, and/or play with it? Given the choice between the Atom and Sublime behavior, from what you've written I tend to agree that Atom sounds better, but I don't personally have a strong feeling.

eyelash:
In order to support Atom's behaviour I would want to refactor the movement code to offer an API that is based on offsets instead of `SelRegion`s so one could move the start and the end of a `SelRegion` separately. I didn't want to do that just yet because it would touch a lot of code.

cmyr:
I don't think I'll have time this weekend, but will give it a close look on monday. 😎

cmyr:
Okay, had a chance to play around with this (there's a branch for xi-mac [here](https://github.com/cmyr/xi-mac/tree/feature/add-sel-move) that I'll PR if we go with this) and I have some non-blocking thoughts:

I don't especially like the behaviour around adding a selection at the top of the document when there's a cursor on the first line; I can't see a real use-case, and it feels buggy (if you add a cursor in this way and then move down, it merges with the other first-line cursor). Is there a good argument for this behaviour that I'm overlooking?

The Atom behaviour also sounds better to me, but I'm okay with merging this and adding an issue to investigate making that change later.

Now that we have multi-sel, we should implement `cancel_operation` (esc) and use that to exit multi-sel?

eyelash:
> I don't especially like the behaviour around adding a selection at the top of the document when there's a cursor on the first line; I can't see a real use-case, and it feels buggy (if you add a cursor in this way and then move down, it merges with the other first-line cursor). Is there a good argument for this behaviour that I'm overlooking?

I just went for the simplest possible solution, to get a pull request up and start the discussion. The same problem happens with a cursor on the last line and `add_selection_below` and both Atom and Sublime Text add a selection at the very end of the document so I thought these edge cases would be unimportant enough to be dealt with later.

Generally I feel a lot of things are in flux right now and I'm not sure whether it is worth it adding a lot of code now to deal with all the edge cases when there is a good chance that code will be need to be rewritten anyway. What are your thoughts here?

> Now that we have multi-sel, we should implement `cancel_operation` (esc) and use that to exit multi-sel?

Yes, definitely.

raphlinus:
Thanks. I agree good next steps are to discuss desired behavior in issues before committing to code.

The selection and movement code _has_ been changing a lot in support of multi-select, but I think quieting down now.

Issue #319, opened 2017-05-25
move themes + style resolution into core


cmyr:
Currently, plugins send raw style information. As per the discussion in #284, we would like to move to a system where plugins send scope information, which is resolved into styles in core. This has the added benefit of making core ‘theme aware’, which means we can start exposing themes to the client for UI styling, and we can (soon) start allowing themes to be set by the client.

This Involves bringing in a subset of `syntect` as a core dependency. 

Rough outline:

- [x] refactor syntect to avoid compiling parser
- [x] update plugin protocol:
  ~- [ ] add `def_styles` method~
    - [x] add `add_scopes` method
    - [x] replace `set_fg_spans` with `add_spans` and `update_spans`.
- [x] migrate plugins towards sending scope spans instead of style spans
     - [x] first with the spellcheck plugin
     - [x] then syntect plugin
- [x] store raw scopes, associated with their sources (e.g. the creating plugin) in editor
- [x] remove scopes on source exit (#318)
- [x] resolve scopes into styles in editor; handle merges & prioritization
- [ ] add protocol methods around managing themes

If I’m on a roll:

 - [ ] [rope science 11](https://github.com/google/xi-editor/blob/master/doc/rope_science/rope_science_11.md) 🤷‍♂️

Issue #284, opened 2017-05-10
[RFC] Style Identifiers


rkusa:
## Current Situation

We have a `set_style` RPC call (which is actually called `def_style` in the source):

```
set_style
  id: number
  fg_color?: number // 32-bit RGBA value
  bg_color?: number // 32-bit RGBA value, default 0
  weight?: number // 100..900, default 400
  italic?: boolean  // default false
```

This is intended to be used by syntax highlighting, linting and similar plugins, which set color spans for parts of the text. Instead of having to define and therefore repeat the style for each span separately, `set_style` allows to reuse them.

When a `Line` is send from the core to the frontend, it could contain a `number[]` array which defines the style spans of this line.

To quote the current [`update.md`](https://github.com/google/xi-editor/blob/master/doc/update.md):

> It is conceptually an array of triples (though flattened, so triple at is styles[i*3], styles[i*3 + 1], styles[i*3 + 2]). [...] **The third element is a style id.**

## Shortcomings

When having different plugins that set style spans on their own (1 syntax highlighting, 1 linting, ...), we probably end up with colors that do not match well (i.e. how does the lint plugin know to use a red which matches my dark or light theme).

## Possible Solution

As in other editors, styles could be represented by a string (often called a "scope" in other editors). For example Sublime Text defines the following for a minimal scope coverage:

```
entity.name
entity.other.inherited-class
entity.name.section
entity.name.tag
entity.other.attribute-name
variable
variable.language
variable.parameter
variable.function
constant
constant.numeric
constant.language
constant.character.escape
storage.type
storage.modifier
support
keyword
keyword.control
keyword.operator
string
comment
invalid
entity.name
entity.other.inherited-class
entity.name.section
entity.name.tag
entity.other.attribute-name
variable
variable.language
variable.parameter
variable.function
constant
constant.numeric
constant.language
constant.character.escape
storage.type
storage.modifier
support
keyword
keyword.control
keyword.operator
string
comment
invalid
```

However, to not restrict plugins in their power, we might not predefine/restrict these scopes and let plugins define them theirselves.
So the Syntax Highlighting plugin might define a style for `invalid`, while a Lint plugin re-uses this `invalid` style definition or defines its own, if a style for `invalid` does not already exist.

That is, the `set_style` RPC call would have two different semantics, one that overrides a style, and one which only sets the style if it does not already exist:

```
set_style
  id: string
  redefine: bool // whether the style should be redefined if it already exists
  fg_color? ... italic?
```

[...]

PR #329, opened 2017-05-30
[ready] Send scopes not styles

cmyr:
This is the first big bit of resolution for #284  and #319. 

Plugins now send scopes, which are resolved into styles in core. Syntect is now a dependency, but it's a stripped down Syntect.

~There's some weird rendering bugs I wouldn't mind someone (e.g. @trishume) taking a look at, see if the behaviour is familiar from past debugging. This [gist](https://gist.github.com/cmyr/2774217adcaf5fd36714cccd586c8f73) contains a good example of this. Removing the comments on line 40 make things weird.~


cmyr:
okay I broke xi-lang, will investigate tomorrow. 😒

rust/core-lib/src/editor.rs
```diff
@@ -340,8 +343,14 @@ impl<W: Write + Send + 'static> Editor<W> {
         // Currently it breaks any such span in half and applies no spans to the inserted
         // text. That's ok for syntax highlighting but not ideal for rich text.
         let empty_spans = SpansBuilder::new(new_len).build();
+        self.style_scopes.update_all(iv.clone(), empty_spans);
+
+        //TODO: style_spans should just live in style_scopesa,
```
raphlinus: typo

rust/core-lib/Cargo.toml
```diff
@@ -16,6 +16,13 @@ xi-rope = { path = "../rope", version = "0.2" }
 xi-unicode = { path = "../unicode", version = "0.1.0" }
 xi-rpc = { path = "../rpc", version = "0.2.0" }
 
+[dependencies.syntect]
```
raphlinus: This will have to be a release, as we can't publish core-lib to crates.io if it has git dependencies.

cmyr: yea, just waiting on the upstream PR

rust/core-lib/src/layers.rs
```diff
@@ -0,0 +1,171 @@
...
+                .filter(|result| match *result {
+                   Err(ref err) => {
+                       print_err!("failed to resolve scope {}\nErr: {:?}",
+                                  &stack.join(" "),
+                                  err);
+                       false
+                       }
+                   _ => true
+                })
+            .map(|s| s.unwrap())
```
trishume: this indentation seems weird. I think this and the other iterator call below should be indented just like `filter` above.

rust/core-lib/src/layers.rs
```diff
@@ -0,0 +1,171 @@
...
+            .collect::<Vec<_>>();
+            stacks.push(ScopeStack::from_vec(scopes));
```
trishume: ScopeStack contains another `Vec` that isn't really useful in this application. Here it is just a useless wrapper and it should really just be a `Vec<Scope>`.

I did warn you not to use `ScopeStack`, but in a conversation about something else so you might have missed it.

rust/core-lib/src/layers.rs
```diff
@@ -0,0 +1,171 @@
...
+
+        // compute styles for each new stack
+        let themes = doc_ctx.theme_set.lock().unwrap();
+        let theme = themes.themes.get("InspiredGitHub").expect("missing theme");
+        //let theme = themes.themes.get("base16-ocean.dark").expect("missing theme");
+        let highlighter = Highlighter::new(theme);
+
+        let mut new_styles = Vec::new();
+        for stack in &stacks {
+            let style = highlighter.style_for_stack(&stack);
```
trishume: Similarly you'll want to change this method in your syntect PR to take a `&[Scope]` like the other `Highlighter` methods do.

rust/core-lib/src/editor.rs
```diff
@@ -913,27 +930,35 @@ impl<W: Write + Send + 'static> Editor<W> {
         self.text.measure::<LinesMetric>() + 1
     }
 
-    pub fn plugin_set_fg_spans(&mut self, start: usize, len: usize, spans: Vec<Span>, rev: usize) {
+    pub fn plugin_add_scopes(&mut self, plugin: &str, scopes: Vec<Vec<String>>) {
+        self.style_scopes.add_scopes(plugin, scopes, &self.doc_ctx);
```
trishume: In the future it would be nice if each plugin could have more than one layer, but in practice one layer per plugin will cover 99% of use cases and everything except large multi-functional plugins that maybe should be split into separate plugins. And the only functionality they lose is being able to separately dispose of and add spans for different things from the same plugin. So this is totally fine for now, and as a default.

rust/core-lib/src/plugins/rpc_types.rs
```diff
@@ -127,10 +127,35 @@ pub struct Span {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+pub struct ScopeSpan {
+    pub start: usize,
+    pub end: usize,
```
trishume: At least for syntax highlighting, every single character will be covered by a span, so it would save space to make this just a `length`. But on the other hand the JSON format right now has like 5x the characters per span it could have. So this is fine for now but if serialization volume is ever an issue the format should be changed to something like `[len,id],[len,id],[len,id],...`.

By my estimation this format pushes the total size for all of jquery to 6mb (including the stacks), which depending on serialization overhead may not be bad.

rust/core-lib/src/plugins/rpc_types.rs
```diff
@@ -127,10 +127,35 @@ pub struct Span {
...
+    pub scope_id: u32,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Style {
+    pub fg: u32,
+    #[serde(rename = "font")]
+    pub font_style: u8,
```
trishume: Why isn't this just named `font` and then not renamed? And if it is to clarify that it doesn't really hold a font, wouldn't `font_id` be better (without even a serde rename)?

rust/core-lib/src/plugins/rpc_types.rs
```diff
@@ -127,10 +127,35 @@ pub struct Span {
...
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct StyleDef {
+    pub style: Style,
+    pub scope: String,
```
trishume: This isn't used yet right?

But when it is used, this should be more like the struct in syntect in that `scope` should be named `selector` since it can be a fancier pattern than a single scope. Also `Style` should probably be a `StyleModifier` like syntect where each of foreground, background, font modifier is optional so they can be merged properly.

You could fix this in this PR but if it is actually not used, I would remove it from this PR and stash it in a branch to work off of later when it actually is used.

rust/syntect-plugin/src/main.rs
```diff
@@ -82,35 +71,56 @@ impl<'a> PluginState<'a> {
...
+        let mut prev_cursor = 0;
+        for (cursor, batch) in ops {
+            if self.scope_state.len() > 0 {
+                let scope_ident = self.stack_idents.get_value(self.scope_state.as_slice());
+                //print_err!("scope ident: {:?}", scope_ident);
+                let scope_ident = match scope_ident {
+                    LookupResult::Existing(id) => id,
+                    LookupResult::New(id) => {
+                        let stack_strings = self.scope_state.as_slice().iter()
+                            .map(|slice| slice.build_string())
```
trishume: So, this will probably be a bottleneck, because this operation isn't meant to be efficient when called like this.

All scopes in `syntect` store their actual string contents in a shared atom repository called `SCOPE_REPO` that this method locks each time it is called, which is terribly inefficient.

Instead what you can do is lock `SCOPE_REPO` (it's public) once outside this loop and use its `to_string` method here. See [here](https://github.com/trishume/syntect/compare/unique-scopes#diff-35cf218b39de628b31722bb466ecb837R84) for an example of how to do this.

rust/syntect-plugin/src/stackmap.rs
```diff
@@ -0,0 +1,120 @@
...
+        self.children.get_mut(first).unwrap().get_value(&stack[1..], next_id)
+    }
+}
+
+impl StackMap {
+    /// Returns the identifier for this stack, creating it if needed.
+    pub fn get_value(&mut self, stack: &[Scope]) -> LookupResult {
+        assert!(!stack.is_empty());
+        let result = self.scopes.get_value(stack, self.next_id);
+        if result.is_new() {
```
trishume: This could become `if let LookupResult::New(_) = result {` and then the method below would be unnecessary.

But I do sympathize in that I wish Rust had an easier way to get "is this thing this type of the enum" as a boolean by default.

rust/syntect-plugin/src/stackmap.rs
```diff
@@ -0,0 +1,120 @@
...
+    Existing(u32),
+    New(u32),
+}
+
+impl Node {
+    pub fn new(value: u32) -> Self {
+        Node { value: Some(value), children: HashMap::new() }
+    }
+
+    fn get_value(&mut self, stack: &[Scope], next_id: u32) -> LookupResult {
```
trishume: As is I bet this lookup is slower than a `HashMap<Vec<Scope>,Node>` would be because of repeated pointer chasing costs. However, this tree structure is still good because if this is slow it's possible to use `ScopeStack::apply_with_hook` to find the basic pushes and pops of an operation, and use that to move a cursor around the tree with each operation, making the lookup time tiny and no longer proportional to stack depth.

The trie might also use less memory but I'm not certain.

rust/core-lib/src/layers.rs
cmyr: yea, mentioned this in the pr in syntect, but I make use of a convenience method that's available on `ScopeStack`. If you think it's worth it I'm happy to change both, no clear winner to me. `ScopeStack` is slightly less efficient, but semantically clearer.

rust/core-lib/src/editor.rs
cmyr: Interesting, I hadn't really thought about a plugin wanting more than one layer, but given the limitation on spans overlapping, I do think it makes sense as a future feature.

It's worth noting that using the plugin name as the layer identifier is temporary. Eventually we will assign plugins PIDs; and layers might come from core, as well.

rust/core-lib/src/plugins/rpc_types.rs
cmyr: Yea, syntax highlighting is definitely a special case: in most others we wouldn't expect spans to be contiguous. If we run into performance issues, we can add a new protocol method to handle this case.

rust/core-lib/src/plugins/rpc_types.rs
cmyr: This is compat with some previously defined type somewhere. We have a bunch of redundant Style defs in various places, which will be ripped out at the end of this work.

rust/core-lib/src/layers.rs
trishume: I'd prefer you change it. For people who don't know what `ScopeStack` does its easier to figure out what it represents when it is just a `Vec<Scope>` directly.

rust/syntect-plugin/src/stackmap.rs
cmyr: I prefer implementing small helpers like this when they improve readability.

PR #332, opened 2017-05-31
Implement span merging logic

cmyr:
This is the span merging algo.

It includes a test, and I feel okay about it; but feel invited to give it a close look.


rust/rope/src/spans.rs
```diff
@@ -181,6 +181,108 @@ impl<T: Clone + Default> Spans<T> {
...
+            assert!(!red_iv.intersect(blue_iv).is_empty());
+
+            // if these two spans do not share a start point, create a new span from
+            // the prefix of the leading span.
+            if red_iv.start() < blue_iv.start() {
+                let iv = red_iv.prefix(blue_iv);
+                sb.add_span(iv, f(red_val, None));
+                red_iv = red_iv.suffix(iv);
+            }
+            else if blue_iv.start() < red_iv.start() {
```
trishume: This being on a new line is both inconsistent with the `else if` above, and also not correct Rust style AFAIK.

rust/rope/src/spans.rs
```diff
@@ -181,6 +181,108 @@ impl<T: Clone + Default> Spans<T> {
...
+            // create a new span by merging the overlapping regions.
+            let iv = red_iv.intersect(blue_iv);
+            assert!(!iv.is_empty());
+            sb.add_span(iv, f(red_val, Some(blue_val)));
+
+            // if an old span was consumed by this new span, advance
+            // else manually set next_, for the next loop iteration
+            red_iv = red_iv.suffix(iv);
+            blue_iv = blue_iv.suffix(iv);
+            assert!(red_iv.is_empty() || blue_iv.is_empty());
```
trishume: I like all the asserts 👍

cmyr:
Yea, I also feel that these algorithms always feel kind of clunky, and I've started to move my focus towards doing very prosaic implementations that focus on being coherent and readable. This is maybe the extreme end of that spectrum with the red/blue cuteness, but I'm happy with the result.

rust/rope/src/spans.rs
cmyr: yea I think I originally broke this line because it looked visually dense without, but that doesn't seem super necessary in hindsight, and I can always use comments or something if I feel the need to break it up.

PR #335, opened 2017-06-06
Incremental Find

rkusa:
This PR adds incremental search.

**Current behaviour:**

When changing a buffer, the search for the whole file is invalidated.

**PR behaviour:**

When changing a buffer, all ranges of deletions and inserts of the edit delta, extended by twice the search string length (because case matching might influence the occurrence length) to both left and right, are invalidated. That is, only the area around changes need to be re-searched.

This could, however, lead to ambiguous occurrences. E.g. when searching for “aba” in “ababababa”.

**Ambiguous search result:**

In case of an ambiguous search result, the leftmost occurrence wins (i.e. the occurrence closer to the file beginning). There are two cases: 1) the new occurrence wins or 2) the existing occurrence wins. For (2), the new occurrence can simply be ignored. However, for (1), the search occurrences cache needs to be invalidated starting from the end of the new occurrence. When searching for a multi-line search string, the occurrences cache is invalidated until the end of the file. Otherwise, for single-line search string, the occurrences cache is invalidated until the end of the line. An example where this occurs:

This approach has one caveat, when starting to search for a multi-line search query in the middle of the file and then scrolling backwards. If there are ambiguous occurrences in this scenario, the highlighted occurrences might jump around while scrolling backwards, as shown in this example:

However, this is fine IMHO, since I would consider this scenario to be very rare and I would prefer the editor to only search what needs to be searched, instead of trying to deal with every possible scenario. Since this is my somehow subjective opinion, this is of course open for discussion.

### DIFF
--- a/rust/core-lib/src/index_set.rs
+++ b/rust/core-lib/src/index_set.rs
@@ -124,7 +124,8 @@ impl IndexSet {
         }
     }
 
-    /// Computes a new set based on applying a delta to the old set.
+    /// Computes a new set based on applying a delta to the old set. Collapsed regions are removed
+    /// and contiguous regions are combined.
     pub fn apply_delta(&self, delta: &Delta<RopeInfo>) -> IndexSet {
         let mut ranges: Vec<(usize, usize)> = Vec::new();
         let mut transformer = Transformer::new(delta);
@@ -330,7 +331,7 @@ mod tests {
         let mut e = IndexSet::new();
         e.union_one_range(1, 3);
         e.union_one_range(5, 9);
-        
+
         let d = Delta::simple_edit(Interval::new_closed_open(2, 2), Rope::from("..."), 10);
         let s = e.apply_delta(&d);
         assert_eq!(s.get_ranges(), &[(1, 6), (8, 12)]);
--- a/rust/core-lib/src/index_set.rs
+++ b/rust/core-lib/src/index_set.rs
@@ -19,6 +19,8 @@
 // crate. Maybe we don't need both.
 
 use std::cmp::{min, max};
+use xi_rope::delta::{Delta, Transformer};
+use xi_rope::rope::RopeInfo;
 
 pub struct IndexSet {
     ranges: Vec<(usize, usize)>,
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/selection.rs
```diff
@@ -109,16 +109,63 @@ impl Selection {
...
+           || (!delete_adjacent && self.regions[last].min() < end)) {
             last += 1;
         }
         remove_n_at(&mut self.regions, first, last - first);
     }
 
+    /// Add a region to the selection. This method does not merge regions and does not allow
+    /// ambiguous regions (regions that overlap).
+    ///
+    /// On ambiguous regions, the regions closer to the left (to the file beginning) wins. That is,
```
cmyr: language nit: although this is currently used primarily to deal with ranges of lines in a file, the idea of a 'file' is external to this data structure. I'm also unsure about the word 'left'. Where possible I would prefer something like 'lower', e.g. something that references a property of the value, not a property of a particular conception of the number line.

rust/core-lib/src/index_set.rs
```diff
@@ -85,6 +124,30 @@ impl IndexSet {
         }
     }
 
+    /// Computes a new set based on applying a delta to the old set.
```
cmyr: Maybe a bit more detail of how the delta is applied, e.g. that collapsed regions are removed and contiguous regions are combined?

rust/rope/src/delta.rs
```diff
@@ -243,6 +244,31 @@ impl<N: NodeInfo> Delta<N> {
             }
         )
     }
+
+    /// Iterates over all the inserts of the delta. Each insert is represented by a tuple of
+    /// three elements, where the first element represents the position of the insertion relative
+    /// to the old document, the second element the position relative to the new document and the
+    /// third element the length of the insertion.
```
cmyr: nit: this seems like a sort of weird/esoteric representation. I might in any case be inclined towards defining some `DeltaOpSummary` struct that gives these values useful names? `old_offset, new_offset, len` or something. Not sure if that improves ergonomics in any practical way, or not...

Anyway I'm just thinking out loud, you should ignore unless you personally think this would be an improvement.

rkusa:
> Looks good to me, with a few little comments around naming and docs.

@cmyr They are very welcome!

> I don't love the behaviour in the pathological case you demonstrate with the multiple lines: [...] to do this right we would need to search from the beginning of the document.

Thanks for the feedback! I also don't like the behaviour, however, I dislike the approach of always searching from the beginning of the document even more. This is really a constructed example, and I cannot think of an example of an actual search I did in the past where this would have happened.

Anyway, there is maybe a solution in-between:

Single-line search → only search visible area
Multi-line search → always search from the beginning of the file

> This is something we should be thinking about anyway: there are lots of cases where we will want full document (and multi-document) results. Do you have a sense of how that will fit into the current incremental approach?

It is only a matter of parameters to search the whole document instead of just the visible area. The incremental-search part would also keep the result list updated when edits happen.

cmyr:
Okay, good info!

I'm okay with the weird multi-line behaviour for now, and I like that there's a fix available if we decide it's actually a problem. 👍

raphlinus:
I skimmed this and didn't see any problems. Thanks @rkusa for the feature, and @cmyr for reviewing.

PR #336, opened 2017-06-06
PluginManager Tweaks

cmyr:
Checkpoint PR.

These are mostly cherry-picks from a branch where I was working on plugin commands; this is mostly cleanup and better error handling.

### DIFF
--- a/rust/core-lib/src/plugins/manager.rs
+++ b/rust/core-lib/src/plugins/manager.rs
@@ -44,8 +44,12 @@ impl <W: Write + Send + 'static>PluginManager<W> {
 
     /// Returns plugins available to this view.
     pub fn available_plugins(&self, view_id: &ViewIdentifier) -> Vec<ClientPluginInfo> {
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/plugins/manager.rs
```diff
@@ -116,93 +135,129 @@ impl <W: Write + Send + 'static>PluginManager<W> {
...
+                Ok(plugin_ref) => {
+                    plugin_ref.initialize(&init_info);
+                    let mut inner = me.lock();
+                    let mut running = false;
+
+                    if let Some(ed) = inner.buffers.lock()
+                        .editor_for_view(&view_id) {
+                            ed.plugin_started(&view_id, &plugin_name);
+                            running = true;
                         }
```
trishume: That thing where your text editor keeps wanting to indent close brackets like these incorrectly. This could also be solved by making the if condition a single line, it looks like it fits, it wouldn't even be the longest line in the file.

PR #338, opened 2017-06-13
Tweaks to be able to build Xi inside the Fuchsia tree.

trishume:
The debug plugins are disabled (with #cfg) on Fuchsia. Also, add
BUILD.gn build rules, so the core can be built using GN rather than
pulled in as a third-party dependency.

- [x] Should Xi be in `third_party/` instead of `rust/`?

### DIFF
--- a/rust/core-lib/src/plugins/manifest.rs
+++ b/rust/core-lib/src/plugins/manifest.rs
@@ -23,6 +23,7 @@ use syntax::SyntaxDefinition;
 static PLUGIN_DIR: &'static str = "XI_PLUGIN_DIR";
 
 // example plugins. Eventually these should be loaded from disk.
+#[cfg(not(target_os = "fuchsia"))]
 pub fn debug_plugins() -> Vec<PluginDescription> {
     use self::PluginActivation::*;
     use self::PluginScope::*;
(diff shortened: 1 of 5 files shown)


trishume:
@cmyr yah it's not great, but we should be able to remove them eventually when a real Fuchsia SDK is released. Gradle is a good guess, but the build files are for [GN](https://chromium.googlesource.com/chromium/src/tools/gn/+/48062805e19b4697c5fbd926dc649c78b6aaa138/README.md), a build system used AFAIK only by Chromium and Fuchsia so almost nobody has heard of it.

raphlinus:
After some discussion we decided on `//third_party/xi-editor`

rust/core-lib/src/plugins/manifest.rs
```diff
@@ -96,6 +96,19 @@ pub enum PluginActivation {
 }
 
 #[cfg(not(target_os = "fuchsia"))]
```
cmyr: i'm curious about this flag, `PluginScope` is referenced in `PluginDescription`, which _isn't_ behind a gate. Am I missing something?

trishume: My mistake, sloppy rebasing in the Github web UI. I just fixed it and did it as an actual rebase without a merge commit.

PR #339, opened 2017-06-13
Synchronize Editor state to the Ledger on Fuchsia

trishume:
Adds the ability to synchronize the state of an Editor to and from the Ledger on Fuchsia. First a Ledger handle has to be given to MainState and then any buffer with a file name will be synchronized with any other buffer with the same file name.

I isolated as much of the Fuchsia-specific code as possible to the new `fuchsia` subdirectory/module, but I couldn't isolate some of the plumbing code since it needed access to private fields. Also some of the code on `Editor` isn't really Fuchsia/ledger specific but it is for now because we don't have any other sync backends and the Fuchsia build system doesn't support features so I can't scope it behind a feature instead of an OS.

A substantial amount of the code in the `fuchsia` folder is copy-pasted from [this fuchsia/rust_ledger_example CL](https://fuchsia-review.googlesource.com/#/c/33483/) but then modified. I couldn't preserve all the generic-ness because of the fact that the `SyncUpdater` needs to know about buffers because that is the level where the `Mutex` is, and that contaminates the rest of `sync.rs` to be Xi-specific.

Also includes a refactor of `Engine` to make its initial contents a `Revision`. In order for CRDT to eventually work even on documents that had initial contents loaded from a file, the two `Engine`'s being merged need a common ancestor, which they won't have if they have different initial Revisions. This mostly involved changing the tests to avoid assuming the initial state is revision 0, which is a refactor that would be necessary anyway when I change how revision IDs are generated in order to support merge.

**Edit:** forgot to note that this currently doesn't do a CRDT merge since that isn't implemented yet. It just does last-write-wins. This means that for now having the garbage collector enabled works but eventually the GC will have to be disabled in `Editor` once a real merge function is hooked up.

cc @jimbeveridge

**Not to be merged before #338**

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -22,7 +22,7 @@ default-features = false
 features = ["assets"]
 
 [target."cfg(target_os = \"fuchsia\")".dependencies]
-sha2 = "^0.5"
+sha2 = "0.5"
 
 [features]
 avx-accel = ["xi-rope/avx-accel"]
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -21,6 +21,9 @@ version = "1.5"
 default-features = false
 features = ["assets"]
 
+[target."cfg(target_os = \"fuchsia\")".dependencies]
+sha2 = "^0.5"
+
 [features]
 avx-accel = ["xi-rope/avx-accel"]
 simd-accel = ["xi-rope/simd-accel"]
--- a/rust/core-lib/src/fuchsia/sync.rs
+++ b/rust/core-lib/src/fuchsia/sync.rs
@@ -1,0 +1,186 @@
+// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Architecture for synchronizing a CRDT with the ledger. Separated into a
+//! module so that it is easier to add other sync stores later.
+
+use magenta::{Channel, ChannelOpts};
+use fuchsia::read_entire_vmo;
+use super::ledger::{ledger_crash_callback, self};
+use apps_ledger_services_public::*;
+use std::sync::mpsc::{Sender, Receiver, RecvError};
+use std::io::Write;
+use fidl::{Promise, Future, self};
+use magenta::HandleBase;
+
+use tabs::{BufferIdentifier, BufferContainerRef};
+use xi_rope::engine::Engine;
+use serde_json;
+
+// TODO switch these to bincode
+fn state_to_buf(state: &Engine) -> Vec<u8> {
+    serde_json::to_vec(state).unwrap()
+}
+
+fn buf_to_state(buf: &[u8]) -> Option<Engine> {
+    serde_json::from_slice(buf).ok()
+}
+
+/// Stores state needed by the container to perform synchronization.
+pub struct SyncStore {
+    page: Page_Proxy,
+    key: Vec<u8>,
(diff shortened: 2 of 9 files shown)


rust/core-lib/src/fuchsia/sync.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// Run this in a thread, it will return when it encounters an error
+    /// reading the channel or when the `Stop` message is recieved.
+    pub fn work(&self) -> Result<(),RecvError> {
+        loop {
+            let msg = self.chan.recv()?;
+            match msg {
+                SyncMsg::Stop => return Ok(()),
+                SyncMsg::TransactionReady { buffer }=> {
+                    let mut container = self.container_ref.lock();
+                    let mut editor = container.editor_for_buffer_mut(&buffer).unwrap();
```
cmyr: In some of the plugin stuff I've had crashes in places like this if an editor goes out of scope while the msg is pending.

rust/core-lib/src/tabs.rs
```diff
@@ -462,6 +479,69 @@ impl<W: Write + Send + 'static> Documents<W> {
     }
 }
 
+// =============== Fuchsia-specific synchronization plumbing
```
cmyr: I like breaking this stuff out (including the `use`s) but maybe go a step further and put all of the fuschia specific stuff at the bottom (before tests) of the file, where possible? Although I get keeping related impls together too.

rust/core-lib/src/tabs.rs
```diff
@@ -43,7 +46,7 @@ pub struct BufferIdentifier(usize);
 
 impl fmt::Display for ViewIdentifier {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "{}", self.0) 
+        write!(f, "{}", self.0)
```
cmyr: 🙋‍♂️

rust/core-lib/Cargo.toml
```diff
@@ -16,6 +16,9 @@ xi-rope = { path = "../rope", version = "0.2" }
 xi-unicode = { path = "../unicode", version = "0.1.0" }
 xi-rpc = { path = "../rpc", version = "0.2.0" }
 
+[target."cfg(target_os = \"fuchsia\")".dependencies]
```
raphlinus: There's a story here. Idiomatic would be single quotes, ie `[target.'cfg(target_os = "fuchsia")'.dependencies]`, but our Python-based Cargo.toml rewriting logic breaks on avakar/pytoml#28.

Any special reason there's an explicit caret here? It doesn't change the semantics (caret is the default if no explicit operator is given).

rust/core-lib/src/fuchsia/sync.rs
```diff
@@ -0,0 +1,186 @@
...
+        let watcher = PageWatcherServer { updates: updates.clone(), buffer: buffer.clone() };
+        let _ = fidl::Server::new(watcher, s2).spawn();
+
+        let (mut snap, snap_request) = PageSnapshot_new_pair();
+        page.get_snapshot(snap_request, Some(key.clone()), Some(watcher_client_ptr)).with(ledger_crash_callback);
+
+        let initial_state_chan = updates.clone();
+        let initial_buffer = buffer.clone();
+        snap.get(key.clone()).with(move |raw_res| {
+            let res = raw_res.expect("fidl failed on initial state response");
```
raphlinus: The core should probably be more robust and not panic on fidl RPC errors.

rust/core-lib/src/fuchsia/sync.rs
raphlinus: Yeah, agree this looks like it can happen in normal use and shouldn't be a panic.

rust/core-lib/src/fuchsia/sync.rs
```diff
@@ -0,0 +1,186 @@
...
+struct PageWatcherServer {
+    updates: Sender<SyncMsg>,
+    buffer: BufferIdentifier,
+}
+
+impl PageWatcher for PageWatcherServer {
+    fn on_change(&mut self, page_change: PageChange, result_state: ResultState) -> Future<Option<PageSnapshot_Server>, fidl::Error> {
+        let (future, done) = Future::make_promise();
+
+        assert_eq!(ResultState_Completed, result_state, "example is for single-key pages");
```
raphlinus: Messages seem to be written for "example" rather than xi.

trishume:
This should be ready to merge after #338, I've addressed all comments.

trishume:
@raphlinus can I get a final review?

PR #345, opened 2017-06-20
Style progress (plus basic themes)

cmyr:
This is the last little bit of work towards #319 and #284, with the exception of clarifying custom style _declaration_; that will be revisited at some point down the line.

A consequence of this work is that the theme now lives in `xi-core`. It seemed like a shame not to expose that to the user. Some of the implementation details around setting themes will change, but it works well enough for now.

Theme selection is testable from the xi-mac branch PR'd here: https://github.com/google/xi-mac/pull/51

### DIFF
--- a/rust/core-lib/src/layers.rs
+++ b/rust/core-lib/src/layers.rs
@@ -24,17 +24,19 @@
 use std::collections::BTreeMap;
 use std::io::Write;
 use syntect::parsing::Scope;
-use syntect::highlighting::Highlighter;
 
 use xi_rope::interval::Interval;
 use xi_rope::spans::{Spans, SpansBuilder};
+
 use tabs::DocumentCtx;
 use styles::Style;
 use plugins::PluginPid;
 
 /// A collection of layers containing scope information.
+#[derive(Default)]
 pub struct Scopes {
     layers: BTreeMap<PluginPid, ScopeLayer>,
+    merged: Spans<Style>,
 }
 
 /// A collection of scope spans from a single source.
@@ -43,13 +45,17 @@ pub struct ScopeLayer {
     style_lookup: Vec<Style>,
     /// Human readable scope names, for debugging
     name_lookup: Vec<Vec<String>>,
-    pub spans: Spans<u32>,
+    pub spans: Spans<Style>,
 }
 
 impl Scopes {
 
     pub fn define_styles() {
         unimplemented!();
+    }
+
+    pub fn get_merged(&self) -> &Spans<Style> {
+        &self.merged
     }
 
     /// Adds the provided scopes to the layer's lookup table.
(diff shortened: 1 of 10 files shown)


rust/core-lib/src/layers.rs
```diff
@@ -59,38 +62,72 @@ impl Scopes {
...
-            let style = layer.style_lookup[*val as usize];
-            sb.add_span(iv, style);
+        let mut layer_iter = self.layers.values();
+        let mut resolved = layer_iter.next().unwrap().subseq_styles(iv);
+
+        for other in layer_iter {
+            //FIXME: this creates a whole lot of unnecessary styles, because
+            // subseq_styles creates a new spans object and a new Style for each
+            // id. Better would be to rewrite Spans::merge to be a bit more like
+            // a reduce/fold operation, and then we could just use &Styles.
```
trishume: I'd prefer that you do this before this PR is merged, but it's not a strong preference and I'm okay with it being merged before then.

cmyr: My thinking: I'm not sure the exact approach I want to take yet, and it doesn't feel like a particularly important optimization at this stage. It's also a very self-contained piece of work that would be a reasonable follow-on sometime down the road. Maybe this is really a `TODO` more than a `FIXME`? Creating `Style`s isn't catastrophically expensive or anything.

cmyr: @trishume I did go ahead and rework this, but with an aim less towards performance than towards ease of debugging. If It ends up being slow in a profiler I'll happily revisit.

raphlinus:
I did a skim and didn't see any problems, I'm sure tris's review was solid. Thanks!

eyelash:
Why are colors here sent differently than for `def_style`? Once as JSON object and once as integer.

cmyr:
@eyelash I'm not totally sure what you're referring to, but the style defs between the plugins & core are different than between core & client, but using a similar mechanism where we define a style first and then refer to it in the future by its definition order.

eyelash:
I forgot to mention that I meant from core to front-end. In `def_style` colors are sent like: `"fg_color": 4294967295` and in `theme_changed` they are sent like `"foreground": {"r": 255, "g": 255, "b": 255, "a": 255}`. I don't like the inconsistency here.

cmyr:
@eyelash ah gotcha.

The theme definitions are a data type we're borrowing from syntect ([theme::ThemeSettings](https://github.com/trishume/syntect/blob/master/src/highlighting/theme.rs)) and I'm putting them on the wire unchanged. Adding the theme support was sort of a side-effect of this work, and I didn't put a _ton_ of thought into it. Your complaint is fair. My inclination would be to leave it as-is for now (maybe opening an issue) and then we can address this at some point down the line, when we start modifying the theme format?

But I'm also maybe just feeling lazy. If you push back on this at all I'm happy to send the theme colours as u32 ARGBs.

PR #348, opened 2017-06-26
Implement CRDT merge for inserts only

trishume:
Based on converting Revisions on the right into Deltas and carrying them through the new commits on the left. Doesn't support undo or deletes yet.

Also adds a couple more tests and some `fmt::Debug` impls that use the alternative flag to display prettier but non-scalable representations.

In a follow-up PR I'm going to add a DSL for sequences of edits and merges that will allow me to test this much more thoroughly and eventually randomly generate test scripts. Then I'll move on to scaling this up to deletes and Undo, as well as making it faster.

cc @jimbeveridge

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -252,26 +252,26 @@ impl<N: NodeInfo> fmt::Debug for Delta<N> where Node<N>: fmt::Debug {
             for el in &self.els {
                 match *el {
                     DeltaElement::Copy(beg,end) => {
-                        try!(write!(f, "{}", "-".repeat(end-beg)));
+                        write!(f, "{}", "-".repeat(end-beg))?;
                     }
                     DeltaElement::Insert(ref node) => {
-                        try!(node.fmt(f));
+                        node.fmt(f)?;
                     }
                 }
             }
         } else {
-            try!(write!(f, "Delta("));
+            write!(f, "Delta(")?;
             for el in &self.els {
                 match *el {
                     DeltaElement::Copy(beg,end) => {
-                        try!(write!(f, "[{},{}) ", beg, end));
+                        write!(f, "[{},{}) ", beg, end)?;
                     }
                     DeltaElement::Insert(ref node) => {
-                        try!(write!(f, "<ins:{}> ", node.len()));
+                        write!(f, "<ins:{}> ", node.len())?;
                     }
                 }
             }
-            try!(write!(f, ")"));
+            write!(f, ")")?;
         }
         Ok(())
     }
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -46,6 +46,7 @@ pub struct Delta<N: NodeInfo> {
 /// A struct marking that a Delta contains only insertions. That is, it copies
 /// all of the old document in the same order. It has a `Deref` impl so all
 /// normal `Delta` methods can also be used on it.
+#[derive(Clone)]
 pub struct InsertDelta<N: NodeInfo>(Delta<N>);
 
 impl<N: NodeInfo> Delta<N> {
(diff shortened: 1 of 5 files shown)


rust/rope/src/multiset.rs
```diff
@@ -327,6 +333,54 @@ impl Subset {
             subset_amount_consumed: 0,
         }
     }
+
+
+    /// Print a debug representation where `#` for count 1, `-` for 0, and
```
cmyr: I  think this is missing a word somewhere between 'where' and '1'.

rust/rope/src/multiset.rs
```diff
@@ -327,6 +333,54 @@ impl Subset {
...
+    /// digits otherwise. Mainly useful for testing with small lengths.
+    pub fn debug_print(&self) {
```
cmyr: With the debug impl below, couldn't this just be `println!("{:#?}", self)`?

rust/rope/src/engine.rs
```diff
@@ -482,15 +460,221 @@ impl Engine {
...
+    // TODO: will the common revs always occur in the same order in both sets,
+    // can we take advantage of that to make this faster?
+    let a_ids: BTreeSet<usize> = a.iter().map(|r| r.rev_id).collect();
+    let b_ids: BTreeSet<usize> = b.iter().map(|r| r.rev_id).collect();
+    a_ids.intersection(&b_ids).cloned().collect()
+}
+
+/// Returns the operations in `revs` that don't have their `rev_id` in
+/// `base_revs`, but modified so that they are in the same order but based on
+/// the `base_revs`. This allows the rest of the merge to operate on
```
cmyr: to operate on... (the suspense is killing me)

rust/rope/src/engine.rs
```diff
@@ -482,15 +460,221 @@ impl Engine {
...
+struct DeltaOp {
+    rev_id: usize,
+    priority: usize,
+    undo_group: usize,
+    inserts: InsertDelta<RopeInfo>,
+    deletes: Subset,
+}
+
+/// Transform `revs`, which doesn't include information on the actual content of the operations,
+/// into a `InsertDelta`-based representation that does by working backward from the text and tombstones.
```
cmyr: biggest nit in the world: "...into a(n) `InsertDelta`-based...".

rust/rope/src/multiset.rs
trishume: Ooops, I deleted this but it was reintroduced by a faulty rebase and I didn't notice it.

rust/rope/src/engine.rs
```diff
@@ -482,15 +471,222 @@ impl Engine {
...
+    assert!(a.len() > 0 && b.len() > 0);
+    assert!(a[0].rev_id == b[0].rev_id);
+    // TODO find the maximum base revision.
+    // this should have the same behavior, but worse performance
+    return 1;
+}
+
+/// Find a set of revisions common to both lists
+fn find_common(a: &[Revision], b: &[Revision]) -> BTreeSet<usize> {
+    // TODO: will the common revs always occur in the same order in both sets,
```
raphlinus: I think the answer is probably no. If you have concurrent edits A and B, then it's pretty easy to imagine that one oplog will have (A, B'), while another has (B, A'), and that this merge will see both.

rust/rope/src/engine.rs
```diff
@@ -482,15 +471,222 @@ impl Engine {
         }
         self.revs.reverse();
     }
+
+    /// Merge the new content from another Engine into this one with a CRDT merge
+    pub fn merge(&mut self, other: &Engine) {
+        let (mut new_revs, text, tombstones, deletes_from_union) = {
+            let base_index = find_base_index(&self.revs[..], &other.revs[..]);
```
trishume: I wasn't sure if it would work without them, and I guess in the interest of avoiding long Rust compile loops, I just went with them. Fixed.

rust/rope/src/delta.rs
```diff
@@ -245,24 +246,43 @@ impl<N: NodeInfo> Delta<N> {
...
-        try!(write!(f, "Delta("));
-        for el in &self.els {
-            match *el {
-                DeltaElement::Copy(beg,end) => {
-                    try!(write!(f, "[{},{}) ", beg, end));
+        if f.alternate() {
+            for el in &self.els {
+                match *el {
+                    DeltaElement::Copy(beg,end) => {
+                        try!(write!(f, "{}", "-".repeat(end-beg)));
```
trishume: I think this is copy-pasted from code I wrote a while ago, but that was also after `?` was introduced. The real reason is I'm still getting used to remembering `?`, since most of my Rust experience before Google was before `?`.

PR #349, opened 2017-06-26
CRDT Merge test scripts

trishume:
Adds some test helpers for running "merge scripts" that make it easier to test sequences of edits and merges between multiple engines. Ports the two existing tests in #348 to use them and adds two more tests.

These also open the door to randomized or even fuzz testing of merge later by generating these merge scripts and adding tests at the end for properties like eventual consistency.

**Not to be merged before #348**

cc @jimbeveridge

### DIFF
--- a/rust/rope/src/engine.rs
+++ b/rust/rope/src/engine.rs
@@ -1217,7 +1217,7 @@ mod tests {
         MergeTestState::new(3).run_script(&script[..]);
     }
 
-    /// Tests that merging again when there is no new revisions does nothing
+    /// Tests that merging again when there are no new revisions does nothing
     #[test]
     fn merge_idempotent() {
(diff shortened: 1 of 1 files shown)


rust/rope/src/engine.rs
```diff
@@ -1098,126 +1098,174 @@ mod tests {
...
+            Assert(0, "acrbdz".to_owned()),
+            Assert(1, "arpbzj".to_owned()),
+            Merge(0,1),
+            Assert(0, "acrpbdzj".to_owned()),
+        ];
+        MergeTestState::new(3).run_script(&script[..]);
+    }
 
-        e1.merge(&e2);
+    /// Tests that merging again when there is no new revisions does nothing
```
raphlinus: "there are no new revisions" I think

PR #351, opened 2017-06-28
Combine sequential edits with same priority in merge

trishume:
Applies an optimization where it combines sequential revisions with the same priority into one transform to decrease the number of transforms that have to be considered in `rebase` substantially for normal editing patterns. Any large runs of typing in the same place by the same user (e.g typing a paragraph) will be combined into a single segment in a transform as opposed to thousands of revisions.

**Not to be merged before #350**

cc @jimbeveridge

### DIFF
--- a/rust/rope/src/engine.rs
+++ b/rust/rope/src/engine.rs
@@ -485,9 +485,10 @@ impl Engine {
             let b_new = rearrange(b_to_merge, &common, other.deletes_from_union.len());
 
             let b_deltas = compute_deltas(&b_new, &other.text, &other.tombstones, &other.deletes_from_union);
+            let expand_by = compute_transforms(a_new);
 
             let max_undo = self.max_undo_group_id();
-            rebase(a_new, b_deltas, self.text.clone(), self.tombstones.clone(), self.deletes_from_union.clone(), max_undo)
+            rebase(expand_by, b_deltas, self.text.clone(), self.tombstones.clone(), self.deletes_from_union.clone(), max_undo)
         };
 
         self.text = text;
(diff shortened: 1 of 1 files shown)


rust/rope/src/engine.rs
```diff
@@ -632,16 +633,38 @@ fn compute_deltas(revs: &[Revision], text: &Rope, tombstones: &Rope, deletes_fro
...
+            } else {
+                last_priority = Some(priority);
+                out.push((priority, inserts));
+            }
+        }
+    }
+    out
+}
+
 /// Rebase b_new on top of a_new and return revision contents that can be appended as new
```
raphlinus: Comment seems to be stale as params have been renamed.

rust/rope/src/engine.rs
```diff
@@ -632,16 +633,38 @@ fn compute_deltas(revs: &[Revision], text: &Rope, tombstones: &Rope, deletes_fro
...
+/// same priority into one transform to decrease the number of transforms that
+/// have to be considered in `rebase` substantially for normal editing
+/// patterns. Any large runs of typing in the same place by the same user (e.g
+/// typing a paragraph) will be combined into a single segment in a transform
+/// as opposed to thousands of revisions.
+fn compute_transforms(revs: Vec<Revision>) -> Vec<(usize, Subset)> {
+    let mut out = Vec::new();
+    let mut last_priority: Option<usize> = None;
+    for r in revs.into_iter() {
+        if let Contents::Edit {priority, inserts, .. } = r.edit {
```
raphlinus: A question: since this seems to only be aggregating the inserts, would it be possible to skip over revs with empty insert sets? That would reduce the number of `transform_union` calls, and also (potentially) aggregate more if you had deletes with one priority interleaved with inserts of another.

trishume: Good idea, that totally makes sense. Done.

PR #357, opened 2017-07-05
Also GC initial loading revision in Editor, fixing panic

trishume:
Fixes google/xi-mac#52 by including the revision that contains the initial contents in the GC call the Editor makes. The GC only supports GCing prefixes of the visible edit history but when the initial revision was introduced, the Editor no longer abided by that restriction, causing invalid states and thus crashes.
 
This PR also updates the Cargo.lock, because it should've been updated in another PR and now every build dirties it.

cc @cmyr @dsego

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -37,6 +37,11 @@ version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "byte-tools"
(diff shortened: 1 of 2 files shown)


trishume:
@raphlinus yes see the PR description, every time Xi is built it dirties the Cargo.lock file because it wasn't committed back when syntect was bumped. I'm just fixing that, it's a separate commit.

PR #363, opened 2017-07-10
implement delete_word_forward and delete_word_backward

### LINKED ISSUE #361: Implement delete_word_forward and delete_word_backward
In many editors, alt+backspace/delete deletes a word starting from the cursor position. Seeing as moving the cursor left/right and selecting by word is supported, I'd expect deletion to work the same way.

callodacity:
* fixes #361 

Let me know if there's anything else I need to do 👍 

---

As a side note: 
Is there a suite of tests for the editor - a place where I can add some tests for these extra methods?

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -456,6 +456,14 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
     }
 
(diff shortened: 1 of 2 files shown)


cmyr:
@callodacity And re: your side note, no, there really aren't any tests for the actual editor itself, but I agree that there should be eventually.

PR #365, opened 2017-07-10
update README image

callodacity:
Just a quick aesthetic update, nothing major to see here - just an update to the README's demo image.

---

I also have this image which demos multiple selections if you're interested.

xi-mac-screenshot2


callodacity:
@cmyr I was thinking the same thing!

eyelash:
I just realized the xi-mac repository still has the old screenshot. Maybe you could put the actual image file only in one repository and then link to the same image from both READMEs.

Issue #253, opened 2017-05-05
Preliminary plugin support


cmyr:
This is a tracking issue for my ongoing work towards preliminary plugin support, discussed in #189. Sections here will correspond approximately to individual PRs.

## Goal

The goal of this work is to make it easy for people to begin developing a wide variety of plugins for xi. When this is done, I'd like to write some examples/documentation and start trying to encourage people to play around with authoring plugins.

## Work
_Individual subheadings should represent a self-contained piece of functionality_
### multi-plugin support
This will be the extension of the current plugin system to support simultaneously running multiple plugins.
- [ ] centralize plugin dispatch (editor doesn't own plugins)
- [ ] add plugin RPC namespace (plugin commands move outside of the `Edit` namespace)
- [ ] allow starting/stopping plugins by name (inc simple UI, probably a menu item)

### plugin command support
- [ ] RPC support for arbitrary commands / requests
- [ ] add 'set syntax' command to syntect plugin 
- [ ] add 'set theme' command to syntect plugin

### plugin activation support
- [ ] activations on syntax change
- [ ] autorun

### global plugin support
- [ ] plugin lifecycle improvements (notif on start/stop/etc)
- [ ] multi-editor, single-plugin dispatch

### single-invocation plugin support
- [ ] RPC protocol for single invocation
- [ ] additional activations + manifest support

### UI support 
This will involve adding RPC protocol commands, and reference implementations on xi-mac
- [ ] status bar updates
- [ ] show available plugin commands (command palette?)
- [ ] line annotations ?

### manifest format + loading
- [ ] describe preliminary `TOML` format for plugin manifests
- [ ] loading manifests from disk
- [ ] standardize plugin discovery
- [ ] 'dev mode' (loading and reloading an in-development plugin)

### documentation + examples
- [ ] rustfmt example (single invocation, non-undo-group editing)
- [ ] braces example (editing using priorities, undo groups)
- [ ] spellcheck (setting styles)

### inter-plugin communication 
(this may get pushed)
- [ ] plugin manifests describe exported functions
- [ ] plugin manifests describe dependencies
- [ ] routing messages between plugins

PR #366, opened 2017-07-11
plugin commands

cmyr:
With this PR, plugins can declare custom commands.

This isn't super useful right now because plugins can't do a whole lot in _response_ to commands, but that will be up next.

There's an accompanying xi-mac PR coming (https://github.com/google/xi-mac/pull/55).

progress towards #253

### DIFF
--- a/doc/frontend.md
+++ b/doc/frontend.md
@@ -287,29 +287,36 @@ this writing, the following is valid json for a `Command` object:
 
 ```json
     {
-    "title": "Test Command",
-    "description": "Passes the current test",
-    "rpc_cmd": {
-        "type": "notification",
-        "method": "test.cmd",
-        "params": {
-            "this_works": "some value",
-            "a_choice": "some value"
-        }
-    },
-    "args": [
-        {"title": "This Works", "description": "Indicates something", "key": "this_works", "arg_type": "Bool"},
-        {
-            "title": "Favourite Number",
-            "description": "A number used in a test.",
-            "key": "a_choice",
-            "arg_type": "Choice",
-            "options": [
-                {"title": "Five", "value": 5},
-                {"title": "Ten", "value": 10}
-            ]
-        }
-    ]
+        "title": "Test Command",
+        "description": "Passes the current test",
+        "rpc_cmd": {
+            "rpc_type": "notification",
+            "method": "test.cmd",
+            "params": {
+                "view": "",
+                "non_arg": "plugin supplied value",
+                "arg_one": "",
+                "arg_two": ""
+            }
+        },
+        "args": [
+            {
+                "title": "First argument",
+                "description": "Indicates something",
+                "key": "arg_one",
+                "arg_type": "Bool"
+            },
+            {
+                "title": "Favourite Number",
+                "description": "A number used in a test.",
+                "key": "arg_two",
+                "arg_type": "Choice",
+                "options": [
+                    {"title": "Five", "value": 5},
+                    {"title": "Ten", "value": 10}
+                ]
+            }
+        ]
     }
 ```
 
--- a/doc/frontend.md
+++ b/doc/frontend.md
@@ -99,6 +99,19 @@ Starts the named given for the given view.
 
 Stops the named plugin for the given view.
 
+#### plugin_rpc
+
+```
+plugin_rpc {"view_id": "view-id-1", "receiver": "syntect",
+            "notification": {
+                "method": "custom_method",
+                "params": {"foo": "bar"},
+            }}
+ ```
+
+Sends a custom rpc command to the named receiver. This may be a notification
+or a request.
+
 ### edit
 `edit {"method": "insert", "params": {"chars": "A"}, "view_id":
 "view-id-4"}`
@@ -231,9 +244,8 @@ follow the existing pattern.
 `theme_changed {"name": "InspiredGitHub", "theme": Theme}`
 
 Notifies the client that the theme has been changed. The client should
-use the new theme to set colors as appropriate.
-The `Theme` object is directly serialized from a 
-[`syntect::highlighting::ThemeSettings`](https://github.com/trishume/syntect/blob/master/src/highlighting/theme.rs#L27) 
+use the new theme to set colors as appropriate. The `Theme` object is
+directly serialized from a [`syntect::highlighting::ThemeSettings`](https://github.com/trishume/syntect/blob/master/src/highlighting/theme.rs#L27)
 instance.
 
 ### plugins
--- a/doc/frontend.md
+++ b/doc/frontend.md
@@ -86,12 +86,6 @@ Dispatches the inner method to the plugin manager.
 
 ### Plugin methods
 
-#### initial_plugins
-
-`initial_plugins {"view_id": "view-id-1"}` -> `[plugin_name]`
-
-Returns an initial list of plugins available for this view.
-
 #### start
 
 `start {"view_id": "view-id-1", "plugin_name": "syntect"}`
@@ -241,6 +235,71 @@ use the new theme to set colors as appropriate.
 The `Theme` object is directly serialized from a 
 [`syntect::highlighting::ThemeSettings`](https://github.com/trishume/syntect/blob/master/src/highlighting/theme.rs#L27) 
 instance.
+
+### plugins
+
+#### available_plugins
+
+`available_plugins {"view_id": "view-id-1", "plugins": [{"name": "syntect",
+"running": true]}`
+
+Notifies the client of the plugins available to the given view.
+
+#### plugin_started
+
+`plugin_started {"view_id": "view-id-1", "plugin": "syntect"}`
+
+Notifies the client that the named plugin is running.
+
+#### plugin_stopped
+
+`plugin_stopped {"view_id": "view-id-1", "plugin": "syntect", "code" 101}`
+
+Notifies the client that the named plugin has stopped. The `code` field is an
+integer exit code; currently 0 indicates a user-initiated exit and 1 indicates
+an abnormal exit, i.e. a plugin crash.
+
+#### update_cmds
+
+`update_cmds {"view_id": "view-id-1", "plugin", "syntect", "cmds": [Command]}`
+
+Notifies the client of a change in the available commands for a given plugin.
+The `cmds` field is a list of all commands currently available to this plugin.
+Clients should store commands on a per-plugin basis; when the `cmds` argument is
+an empty list it means that this plugin is providing no commands; any previously
+available commands should be disabled.
+
+The format for describing a `Command` is in flux. The best place to look for
+a working example is in the tests in core-lib/src/plugins/manifest.rs. As of
+this writing, the following is valid json for a `Command` object:
+
+```json
+    {
+    "title": "Test Command",
+    "description": "Passes the current test",
+    "rpc_cmd": {
+        "type": "notification",
+        "method": "test.cmd",
+        "params": {
+            "this_works": "some value",
+            "a_choice": "some value"
+        }
+    },
+    "args": [
+        {"title": "This Works", "description": "Indicates something", "key": "this_works", "arg_type": "Bool"},
+        {
+            "title": "Favourite Number",
+            "description": "A number used in a test.",
+            "key": "a_choice",
+            "arg_type": "Choice",
+            "options": [
+                {"title": "Five", "value": 5},
+                {"title": "Ten", "value": 10}
+            ]
+        }
+    ]
+    }
+```
 
 ### RPCs from front-end to back-end
 
--- a/python/xi_plugin/host.py
+++ b/python/xi_plugin/host.py
@@ -64,7 +64,6 @@ class PluginHost(object):
 
         else:
             assert len(buffer_info) == 1
-            print(buffer_info, file=sys.stderr)
             first_view = buffer_info[0]["views"][0]
             self.plugin.initialize(self.views[first_view])
 
@@ -101,6 +100,17 @@ class PluginHost(object):
     def shutdown(self, peer, **params):
         peer.done = True
         self.plugin.shutdown()
+
+    def other_cmd(self, peer, method, params):
+        '''Unknown RPC, presumably a custom command provided by the plugin.'''
+        view_id = params.pop("view_id")
+        if view_id:
+            view = self.views.get(view_id)
+            if view is None:
+                print("plugin {} missing view for id {}",
+                      self.plugin.identifier, view_id)
+            params["view"] = view
+        return getattr(self.plugin, method)(**params)
 
     def _initialize_buffers(self, peer, buffer_info):
         for buf in buffer_info:
--- a/python/xi_plugin/host.py
+++ b/python/xi_plugin/host.py
@@ -101,15 +101,11 @@ class PluginHost(object):
         peer.done = True
         self.plugin.shutdown()
 
-    def other_cmd(self, peer, method, params):
-        '''Unknown RPC, presumably a custom command provided by the plugin.'''
-        view_id = params.pop("view_id")
-        if view_id:
-            view = self.views.get(view_id)
-            if view is None:
-                print("plugin {} missing view for id {}",
-                      self.plugin.identifier, view_id)
-            params["view"] = view
+    def custom_command(self, peer, method, params):
+        '''Custom command provided by the plugin.'''
+        if params.get('view'):
+            view = self.views.get(params['view'])
+            params['view'] = view
         return getattr(self.plugin, method)(**params)
 
     def _initialize_buffers(self, peer, buffer_info):
--- a/rust/core-lib/src/plugins/manager.rs
+++ b/rust/core-lib/src/plugins/manager.rs
@@ -27,7 +27,7 @@ use tabs::{BufferIdentifier, ViewIdentifier, BufferContainerRef};
 
 use super::{PluginCatalog, PluginRef, start_plugin_process, PluginPid};
 use super::rpc_types::{PluginCommand, PluginUpdate, UpdateResponse, PluginBufferInfo, ClientPluginInfo};
-use super::manifest::PluginActivation;
+use super::manifest::{PluginActivation, Command};
 
 pub type PluginName = String;
 type PluginGroup<W> = BTreeMap<PluginName, PluginRef<W>>;
@@ -219,7 +219,9 @@ impl <W: Write + Send + 'static>PluginManager<W> {
         let plugin_desc = self.catalog.get_named(plugin_name)
             .ok_or(Error::Other(format!("no plugin found with name {}", plugin_name)))?;
 
-        let init_info = if plugin_desc.is_global() {
+        let is_global = plugin_desc.is_global();
+        let commands = plugin_desc.commands.clone();
+        let init_info = if is_global {
             let buffers = self.buffers.lock();
             let info = buffers.iter_editors()
                 .map(|ed| ed.plugin_init_info().to_owned())
@@ -237,7 +239,13 @@ impl <W: Write + Send + 'static>PluginManager<W> {
             match result {
                 Ok(plugin_ref) => {
                     plugin_ref.initialize(&init_info);
-                    me.lock().on_plugin_launch(&view_id, &plugin_name, plugin_ref);
+                    if is_global {
+                        me.lock().on_plugin_connect_global(&plugin_name, plugin_ref,
+                                                           commands);
+                    } else {
+                        me.lock().on_plugin_connect_local(&view_id, &plugin_name,
+                                                          plugin_ref, commands);
+                    }
                 }
                 Err(_) => print_err!("failed to start plugin {}", plugin_name),
             }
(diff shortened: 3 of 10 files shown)


rust/core-lib/src/plugins/manager.rs
```diff
@@ -346,41 +382,58 @@ impl <W: Write + Send + 'static>PluginManager<W> {
...
-        PluginManagerRef(self.0.clone())
+        PluginManagerRef(self.0.clone(), self.1.clone(), self.2.clone())
     }
 }
 
 /// Wrapper around a `Weak<Mutex<PluginManager<W>>>`
-pub struct WeakPluginManagerRef<W: Write>(Weak<Mutex<PluginManager<W>>>);
+pub struct WeakPluginManagerRef<W: Write>(
+    Weak<Mutex<PluginManager<W>>>,
+    Weak<Mutex<Vec<Box<Fn(PluginManagerRef<W>) + 'static + Send>>>>,
```
trishume: I don't understand why these aren't just fields on `PluginManager`, or at least some struct that's all wrapped in one `Arc<Mutex<>>`.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -475,6 +531,46 @@ impl<W: Write + Send + 'static> PluginManagerRef<W> {
...
+        //TODO: our RPC error story is sort of non-existent. This should
+        //probably return a Result<T> with some error type that can go on
+        //the wire?
+        self.lock().dispatch_common(view_id, receiver, method, params, true)
+            .unwrap_or_else(|| json!({}))
+    }
+
+
+    pub fn handle_idle(&self) {
+        let to_run = match self.2.load(Ordering::SeqCst) {
```
trishume: This should probably just be an if-else expression.

rust/core-lib/src/plugins/manifest.rs
```diff
@@ -107,10 +111,132 @@ pub enum PluginScope {
...
+    pub key: String,
+    pub arg_type: ArgumentType,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    /// If `arg_type` is `Choice`, `options` must contain a list of options.
+    pub options: Option<Vec<ArgumentOption>>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum ArgumentType {
+    Number, Int, PosInt, Bool, String, Choice,
```
trishume: Why not make this an actual enum with variant-specific data. Serde has a bunch of good options for how to serialize these.

For example the `Choice` variant could store the choices. That way you get way more type safety.

rust/core-lib/src/plugins/manifest.rs
```diff
@@ -107,10 +111,132 @@ pub enum PluginScope {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "type")]
+#[serde(rename_all = "snake_case")]
+/// A placeholder type which can represent a generic RPC.
+///
+/// This is the type used for custom plugin commands, which may have arbitrary
+/// method names and parameters.
+pub enum PlaceholderRpc {
```
trishume: Given that both variants have the same fields, this looks like it would make more sense as a struct like

```rust
struct PlaceholderRpc {
    pub method: String,
    pub params: Value,
    pub type: RpcType,
}
enum RpcType {
    Notification, Request
}
```

That would eliminate the need for all the methods that match on the variant and extract the right field further down.

rust/core-lib/src/tabs.rs
```diff
@@ -371,9 +372,13 @@ impl<W: Write + Send + 'static> Documents<W> {
             let buffer_id = self.next_buffer_id();
             self.new_empty_view(rpc_peer, &view_id, buffer_id);
         }
+
+        // schedule plugin notification for after the view has received an identifer
+        let view_id2 = view_id.to_owned();
         let init_info = self.buffers.lock().editor_for_view(&view_id)
             .unwrap().plugin_init_info();
-        self.plugins.document_new(&view_id, init_info);
+        self.plugins.run_on_idle(move |p| p.document_new(&view_id2, &init_info));
```
trishume: Why is this (and by extension all the accompanying infrastructure) necessary?

rust/core-lib/src/tabs.rs
```diff
@@ -491,6 +496,27 @@ impl<W: Write + Send + 'static> Documents<W> {
                 self.plugins.stop_plugin(&view_id, &plugin_name);
                 None
             }
+            PluginRpc  { view_id, receiver, mut rpc } => {
+                // inserting view_id into params here because the plugin needs
+                // to know where this command originates. This feels a bit
+                // clumbsy though. At the very least we should use some obscure
+                // key so we don't collide with any possible user-supplied keys?
+                // Or we could validate keys on plugin load or something.
```
trishume: I think it would make more sense to have the view the command is run on be an argument to the custom RPC function with a `View` argument type. The plugin would have to explicitly add a `view` argument to the command if it needed the view, like how Rust methods require an explicit `self`. Then that argument would be implicitly filled in by the front end without it presenting a UI for it, although it could for example in a parameter filling dialog show that argument with the view file name the command is being run on.

This allows a distinction between for example global, buffer and view commands in a very functional-programming-esque way. It conveys if the command actually needs a view, allowing it to be potentially be called programmatically in contexts that don't have a view, and paving the way for other implicit argument types later.

rust/core-lib/src/tabs.rs
```diff
@@ -491,6 +496,27 @@ impl<W: Write + Send + 'static> Documents<W> {
...
+                assert!(rpc.params_ref().is_object(), "params must be an object");
+                rpc.params_ref_mut().as_object_mut()
+                    .map(|p| p.insert("view_id".to_owned(), json!(view_id)));
+                match rpc {
+                    PlaceholderRpc::Notification { ref method, ref params } => {
+                        self.plugins.dispatch_notification(&view_id, &receiver,
+                                                           method, params);
+                        None
+                    }
+                    PlaceholderRpc::Request { ref method, ref params } => {
```
trishume: Do you have any examples of a situation where a frontend would want a response from a plugin?

In Sublime and Emacs at least there is no concept of a return value for a command, they're asynchronous things that you just fire and forget, and it doesn't seem to hamstring them.

python/xi_plugin/host.py
```diff
@@ -99,6 +98,17 @@ def shutdown(self, peer, **params):
         peer.done = True
         self.plugin.shutdown()
 
+    def other_cmd(self, peer, method, params):
+        '''Unknown RPC, presumably a custom command provided by the plugin.'''
```
cmyr: when we receive the rpc (https://github.com/google/xi-editor/pull/366/files#diff-761f7b0ecb719bd1378e85f32c16fabbR77) we check to see if the method is one the host already implements; only if we don't know the command we call this. It is an expectation of the current architecture that any method that is explicitly part of the plugin protocol should be handled by the host.

It is currently possible to declare a custom command that shares a method name with a plugin protocol method; in the future we can check for this case when we parse plugin manifests.

rust/core-lib/src/plugins/manager.rs
cmyr: The reason these live outside of the `PluginManager` is because the public interfaces exist on `PluginManagerRef`.

The reason it isn't just one `<Arc<Mutex<_>>` is that we want to check for work on each idle call, and I wanted to avoid unnecessary locking, hence using the `AtomicBool` to store whether or not there's work.

But I do agree that these types look hairier than they need to. At the very least I should be able to combine the second two fields into some sort of `AtomicVec` type or something.

rust/core-lib/src/tabs.rs
cmyr: Because we need the client to have a `view_id` before we send it notifications about running plugins and available commands, etc. I'm very open to a better strategy if you have any ideas (although I think in general the pattern of doing plugin maintenance etc work in the idle call will become more useful)

rust/core-lib/src/tabs.rs
cmyr: I agree that in client->plugin RPC requests probably aren't necessary (at least as custom commands; they might be useful for things like requesting completions?) but I wanted to build in the machinery now because I think they _will_ be used for inter-plugin communication.

This does raise the point though that there are some places in this PR where I'm expecting client->plugin RPC requests; I'll clean those up.

rust/core-lib/src/tabs.rs
cmyr: Okay, I like this.

"view" will become a special key in a command's argument list, and it will be auto-filled by the client if present.

rust/core-lib/src/plugins/manifest.rs
cmyr: Some of my thinking here is that not all json parsing tools are as expressive / flexible as serde is. In particular, what I'm designing here is less a set of Rust data structures and more a JSON schema. Having this field be consistently a string, and having options serialized alongside if required, feels like the clearest representation I can get?

Otherwise I'd have to do something like allow this field to be either a string or an array, or have it be an object that is empty in all but the `Choice` case. If there's a better option, I'm all ears. ☺️

rust/core-lib/src/tabs.rs
cmyr: @trishume I pushed some changes addressing a bunch of your comments; the last commit (cd0b001) _slightly_ cleans up the `PluginManagerRef` stuff. However, if there's a way to avoid that mess altogether I'm happy to get rid of it.

One option I considered initially was a much less general solution, where we just stash the `view_id` and  the `init_info`, and send `document_new` directly if they've been stashed. 

Also: a bunch of the ugliness of this stuff is because going through the lock unnecessarily on every pass of the run loop felt bad to me, but maybe that's a grossly premature consideration? My intuition on some of these systems performance topics isn't very well calibrated.

python/xi_plugin/host.py
trishume: I think I'd prefer a design where the RPC to the plugin was a wrapper method like "custom_command" so that they are distinct from defined parts of the plugin API, and plugins should ignore top-level commands they don't know about. That way it would be possible to add new plugin notification RPCs in a backwards compatible way. I also think the separation more clearly distinguishes two different things: RPCs defined by the plugin, and RPCs defined by the host.

cmyr: 👌 sounds good to me.

rust/core-lib/src/plugins/manifest.rs
trishume: What I mean is that serde has attributes that let you get the JSON representation you want while also using the correct Rust representation:

https://serde.rs/enum-representations.html

If you use the "internally tagged" representation you should be able to get a representation like

```js
{"title": "My Argument", /* ... */, "type": {"name": "choice", "choices": ["Blah", "Foo", "Bar"]}
```

Which I think is better in that it puts the options as part of the type, which they are, and working with it in statically typed languages will be way nicer.

cmyr: But in the other cases, wouldn't the representation then be

```js 
{"title": "My Argument", /* ... */, "type": {"name": "bool"}
```
i.e, the `type` field would have to always be an object, which in <80% of cases will have just the single field?

rust/core-lib/src/tabs.rs
trishume: It does look better now but here's a different idea:

Accidentally sending notifications while responding to a RPC call with a reply sounds like it will be a recurring problem. Instead of having to know which things will send notifications and carefully defer them (and what if you need the result of the notifying thing to craft your reply?!), you should be able to send notifications whenever and have them be delivered correctly. What I'm thinking is that when you tell the event loop to send a notification, it knows if it is already in the middle of a procedure call and if so pushes the notifications onto a queue, which it then sends off when the call is done.

I don't like the current `run_on_idle` method for a number of reasons:

1. Queueing `run_on_idle` tasks sounds like the event loops job, not the plugin manager's.
1. "idle" doesn't really mean idle, there might be a thousand events queued up waiting to execute and you'll still run it on the next spin right? If not, does that mean document creation has unbounded latency? I feel that an idle mechanism should only be used for things you actually want to run on idle, like garbage collection, not things you just want to run "later because otherwise it'll mess things up, but not too late please!".
1. I can't tell if there's a race condition here where the frontend immediately sends a command to a plugin after creating a new view, but the plugin manager doesn't know the document has been created yet because the new event arrived before the idle call.
1. It feels like it might end up like Cocoa/IOS where if something doesn't work you shrug and wrap it in a timer that runs after 0.1ms and it fixes it without you even knowing why.

rust/core-lib/src/plugins/manifest.rs
trishume: Correct, but I don't think it's a big deal, especially since it conveys that the type can be more complex, and I care about languages with good type systems like Rust and Swift having a nice type-safe way of representing our protocol.

rust/core-lib/src/tabs.rs
cmyr: View creation feels like a special case, because it is responsible for generating and assigning `ViewIdentifier`s, which in turn are used for all message routing.

On 1/2 I agree. 3. isn't a problem, because the frontend can't send a command until it knows what commands are available, which it can't know until it has a view id. On 4., yes, I think a good rule of thumb is that if we ever add a `run_with_delay` function we've probably screwed up somewhere. 😏

How about this solution: when a client view has received an identifier, it sends a `view_loaded` notification, and we use _that_ to do the setup?

trishume: That indeed would be the simplest solution, but I don't like how it adds unnecessary complexity to the frontend API to fix a core plumbing problem.

Also re your atomics locking on every iteration point: I just thought about it and a lock is basically implemented as an operation on an atomic, which in the case of no contention (which will never happen with our single-threaded event loop, I don't even quite understand why a Mutex is needed) is very very cheap. So I wouldn't expect the atomic would improve performance much.

So at the very least I think it should be a `run_next_iteration` method on some kind of handle to the event loop (maybe using an unbounded `mpsc` queue, which gives you your queue in the same method as synchronization).

Even if it doesn't crop up that often, or only once, I still think the correct approach though is to make the event loop behave [what I think is] correctly with respect to notifications and RPCs with returns. At a stretch I can even imagine other resource creation cases like creating breakpoints. It might even be simpler than the `run_next_iteration` code, depending on how notifications actually work (I haven't looked).

cmyr: We need the mutex because each plugin runloop runs in a separate thread.

I'm not sure about enforcing sequential ordering on notifications; I could imagine situations where we explicitly _do_ want to send out notifications while holding on to a response, for instance if we want to do some `will_save` notification. 

I think `run_next_iteration` on `RpcPeer` could be a reasonable solution, though.

Also I was slightly confused: I just went digging through some code and there's actually _already_ an `available_plugins` request that the client sends to core when it gets its initial view identifier; and core doesn't send any notifications as part of the `document_new`; instead that function launches plugins, which when they connect attempt to notify the client that they've launched.

So at the very least that existing `available_plugins` request could be an `init_view` request, although we could also remove it and consolidate behind some new solution that offers correct message ordering.

Anyway I've got to run, but I'll pick this up tomorrow.

rust/core-lib/src/plugins/manifest.rs
cmyr: I'll make this change, but I don't think it's a clear win (which I've felt about the majority of the feedback on this PR). In particular, I think that the type safety guarantees are overstated: these types should only ever be parsed at runtime (from manifest files or from json-rpc), and so the compile-time safety of a type system isn't particularly relevant, and I think that the ease of actually reading and writing the manifest files _is_ important, and this change doesn't help in either case.

rust/core-lib/src/tabs.rs
cmyr: The more I think about this, the more I think that no plugin work should happen during handling of the `new_view` rpc, and that adding a notification for when a view is ready to be interacted with is a useful and reasonable part of view lifecycle.

This simplifies code in a number of places, and makes the API more explicit; a client should not receive any updates for a given view until it has sent a 'view_loaded` (or whatever name) notification.

Because this notification already exists (by another name) I'm going to remove all of the `run_on_idle` stuff currently here, and implement this in terms of that notification, which will be trivial; we can reconsider from there.

cmyr: Okay, I've flip flopped again, sort of:

Digging through the current RPC system, I noticed that @raphlinus has encountered a similar race situation around sending the cursor for new empty buffers; the solution there was also to schedule the idle handler.

So it is clearer to me, now, that this makes sense as a real feature of the rpc system. I'm not sure if we need a new method, that is higher priority than `schedule_idle`; in any situation where there is a backlog of active events it isn't clear that our new event should be prioritized.

I'm not going to dig into this in much more detail this weekend, but I'll explore a bit more and see if I can come up with a cleaner solution when I get a chance.

cmyr:
@trishume reworked a bunch of this.

I found an old TODO from @raphlinus around another situation where was using `handle_idle` to schedule work to occur after a view has been created; in this case it was sending the initial cursor.

I've stuck with that approach, but it's quite a bit cleaner. the main `Documents` instance now schedules the idle handler when it creates a new view, and stores a closure to do the post-creation work. It has to store and then call this closure itself, manually, when the idle handler is called.

The architecture I would most like is if various entities can register a closure with the event loop itself, but there are some issues with this: 

1. In general, it would be nice if the instance that is registering the closure was available in the closure. The obvious way to do this would be to have it be an argument of the closure, but I'm not sure how easy this will be in a generic way; I suspect it's possible (maybe with  an associated type) but it will take a bit of work in `xi-rpc`.

2. To allow for this registration process, we'd have to start passing the `RpcCtx` much further down the call tree than we do, currently. 

Anyway: this is starting to feel like it might be outside of the scope of this PR. Some of the changes in here already sort of feel like it should be a new PR.

We haven't talked about this, but for a while I've been playing around with a strongly-typed API for interacting with the runloop/rpc system; there's a sketch of that [here](https://gist.github.com/cmyr/c5d28931826fb6ac410d9bda9e389d79). It feels like this ability to schedule closures to execute on idle might be a reasonable component of that work.

trishume:
(Previous comment location was default-collapsed so continuing discussion of argument type schema here)

I didn't know you planned on using this JSON format for manifests as well as for communication with the frontend. If humans will be writing it regularly, that changes things and I agree it should be nicer. I stand by what I've said as far as formats for frontend-core communication go, what I meant by type safety is that serde does the validation for you and you don't have to write as much of your own checking code to ensure that the options are correct and present, and code that has to deal with types and options has an easier job dealing with one thing instead of two that have to match up.

If you plan on sharing the schema with the manifest it does make sense to focus more on niceness for humans. I'm okay with you reverting the change to the schema that I asked you to make in that case. 

However, I think I'd prefer an alternative hybrid format that I think makes sense from both a human point of view and a types point of view. Basically use the enum approach you have now, except add the option to deserialize a String as any of the `enum` options that have no data. You can do this using https://serde.rs/string-or-struct.html, you don't have to worry about the serializing, it's okay for that to use the more verbose form.

Also I previously suggested that types look like `"type": {"name": "choice", ...}`, but I think `"type": {"id": "choice", ...}` might be better, but I'm not sure. I think it's less ambiguous with a user-provided name, and as a bonus it's shorter.

rust/core-lib/src/plugins/manifest.rs
```diff
@@ -107,10 +111,131 @@ pub enum PluginScope {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+/// A user provided argument to a plugin command.
+pub struct CommandArgument {
+    /// A human readable name for this argument, for use as placeholder
+    /// text or equivelant.
+    pub title: String,
+    /// A short (single sentence) description of this argument's use.
+    pub description: String,
+    pub key: String,
+    pub arg_type: ArgumentType,
```
trishume: I would use `serde` to rename this to `type`, I know you can't name it `type` directly since it's a keyword. This is already an argument structure and none of the other fields are prefixed with `arg_`.

rust/core-lib/src/plugins/manifest.rs
```diff
@@ -107,10 +111,131 @@ pub enum PluginScope {
...
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(tag = "type_name")]
```
trishume: Similarly, I would tag this `id`. See my previous review comment, and my top level comment on the subject.

cmyr:
@trishume yea, the whole "this is parsed from a manifest file" is an important piece of missing information. Happy to see our thinking converge more in light of that.


Any strong feelings about the idle handling stuff? (8b160d8). 

I'm going to hold off on this all in general until @raphlinus has a chance to take a look, since he may have had some vision for where this should go, ultimately.

doc/frontend.md
```diff
@@ -237,11 +244,75 @@ follow the existing pattern.
...
+
+```json
+    {
+    "title": "Test Command",
+    "description": "Passes the current test",
+    "rpc_cmd": {
+        "type": "notification",
+        "method": "test.cmd",
+        "params": {
+            "this_works": "some value",
```
cmyr: My original thinking is that `params` is provided separately from `args`; in this case params could contain arguments that are not user supplied, which would be sent unchanged. So, for instance, a plugin could provide multiple commands that share a common rpc method. `args` are things we prompt the user for, and are used to update the `params` object.

I appreciate this definitely isn't clear, right now; and the examples don't help.

My goal with this plugins path work has been to get to a point where there's enough functionality and API in place for me to start actually writing _useful_ plugins; at that point I expect to iterate on a lot of these nitty things as I develop a better sense for the ergonomics, and at that point I'd also like to write up good documentation / examples (this is mentioned in #253). So fwiw I don't expect much of this code to be intact come september, and that belief has been reflected in the level of polish I've put into certain areas of it.

rust/core-lib/src/plugins/manifest.rs
cmyr: my thinking here is that 'type' is a reserved word in so many languages that I prefer not to use it as an identifier anywhere?

This is actually also my attitude towards `id`, although that is more a hangover from years of writing objc, and I'm more flexible there.

rust/core-lib/src/plugins/manifest.rs
cmyr: I gave this some thought, and `type_name` was the most semantically communicative key I could come up with. In my experience, `id` is mostly used as a unique identifier (as in the json-rpc protocol, or scads of REST APIs).

My inclination is to revert this change and go back to having `arg_type`/`type` be a string, and then have the optional `options` field. Using serde's 'either string or struct' doesn't offer any more safety than an assert, and it potentially complicates parsing for clients.

Keep in mind that these types are never actually used in in rust; they're deserialized in one format and then serialized in another.

doc/frontend.md
trishume: Okay that makes sense, thanks.

cmyr:
@trishume addressed the last few holdout issues and did some cleanup.

PR #367, opened 2017-07-11
Async plugin edits

cmyr:
This patch gives plugins the ability to initiate edit events; previously plugin edits could only occur in response to buffer updates.

This (in combination with #366) will make useful command-based plugins (code formatting tools, removing trailing whitespace) possible.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -318,25 +318,23 @@ impl<W: Write + Send + 'static> Editor<W> {
         }
     }
 
-    // generates a delta from a plugin's response and applies it to the buffer.
-    pub fn apply_plugin_edit(&mut self, edit: PluginEdit, undo_group: usize) {
+    //TODO: plugin edits should be represented by real Deltas.
+    /// generates a delta from a plugin's response and applies it to the buffer.
+    pub fn apply_plugin_edit(&mut self, edit: &PluginEdit, undo_group: Option<usize>) {
         let interval = Interval::new_closed_open(edit.start as usize, edit.end as usize);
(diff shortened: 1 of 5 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -998,6 +998,17 @@ impl<W: Write + Send + 'static> Editor<W> {
     // Note: the following are placeholders for prototyping, and are not intended to
     // deal with asynchrony or be efficient.
 
+    /// Applies an async edit from a plugin.
+    pub fn plugin_edit(&mut self, edit: &PluginEdit) {
+        self.this_edit_type = EditType::Other;
+        let interval = Interval::new_closed_open(edit.start as usize, edit.end as usize);
+        let text = Rope::from(&edit.text);
+        let rev_len = self.engine.get_rev(edit.rev).unwrap().len();
+        let delta = Delta::simple_edit(interval, text, rev_len);
```
trishume: We definitely eventually want to support full multi-edit Deltas from plugins because that is our story for how to cleanly handle asynchronous formatting and file reloading.

I would say add a TODO here and soon before building up too much stuff based on edits from plugins I would appreciate a refactor of the plugin API to send a full Delta representation. Perhaps even just based on the serde implementations for `Delta`.

cmyr: @trishume totally. The plan for a while has been to just put Deltas on the wire; I have the serde implementations for delta done in another working branch, but I've been slightly putting off finishing that patch because it requires changing both plugin libs, and i want to figure out what other changes are coming to the update protocol (we've been talking about including line break info/deltas) to avoid redundant work. But yea, I can get that work in soon.

PR #368, opened 2017-07-12
Add get_selections plugin RPC

cmyr:
Another small additional plugin capability: plugins can now query a view for the current selections. In conjunction with #367, this means that plugins can modify selected regions of a buffer.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -1058,6 +1058,17 @@ impl<W: Write + Send + 'static> Editor<W> {
         Some(text.slice_to_string(offset, end_off))
     }
 
+    pub fn plugin_get_selections(&self, view_id: &ViewIdentifier) -> Value {
+        //TODO: multiview support
+        assert_eq!(view_id, &self.view.view_id);
+        let sels: Vec<(usize, usize)> = self.view.sel_regions()
+            .iter()
+            .map(|s| { (s.start, s.end) })
+            .collect();
+
+        json!({"selections": sels})
+    }
+
     // Note: currently we route up through Editor to DocumentCtx, but perhaps the plugin
     // should have its own reference.
     pub fn plugin_alert(&self, msg: &str) {
--- a/python/xi_plugin/host.py
+++ b/python/xi_plugin/host.py
@@ -43,6 +43,9 @@ class PluginPeer(RpcPeer):
                                                'max_size': max_size,
                                                'rev': rev})
 
+    def get_selections(self, view_id):
+        return self.send_rpc_sync('get_selections', {'view_id': view_id})
+
 
 class PluginHost(object):
     """Handles raw RPC calls, updating state and calling plugin methods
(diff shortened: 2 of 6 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -1058,6 +1058,17 @@ impl<W: Write + Send + 'static> Editor<W> {
         Some(text.slice_to_string(offset, end_off))
     }
 
+    pub fn plugin_get_selections(&self, view_id: &ViewIdentifier) -> Value {
+        //TODO: multiview support
```
trishume: I've always been uneasy about the whole separation between the active view and the non-active views into two separate containers, and further the fact that RPCs for the views have to dispatch through the `Editor`, and this just makes me more uneasy.

I don't know how to fix this other than a large refactor so for now this is just a general comment about a sense of foreboding.

cmyr: I share your disquiet. I think the first problem is very tractable; I think we should be able to remove that inconsistency when we actually add multiview support. There was a PR a while ago from @cbrewster that included this change.

The second issue is one I struggle with, more. Part of this feels like a relic of the fact that historically there's been a 1-1 relationship between views and editors, and so they have been used interchangeably. A possible architectural change might be to actually have the main `Documents` object store views, instead of editors?

I'm very open to anybody's thoughts on this.

trishume: Yah I think that would make a lot of sense, instead of things running on "the main view of this editor", they would run on "this view". 

Sublime's entire API is on the view and it just happens that internally that views sometimes share buffers, and they give you a way to get the buffer ID of a view so you can tell which views share buffers.

Alternatively, there could be some things that run on views and some that run on buffers and there is a pool of each that you index with a buffer or view ID. There's APIs to get the buffer ID of a view or the views displaying a buffer. This I think is how Emacs works. This API more explicitly represents the underlying concepts and also enables buffers that aren't attached to views, on the other hand it is more complex, and in most cases you don't care and the Sublime-style API just does what you want. You also need some sort of garbage collector, or you can do what Emacs does and have your number of open buffers just gradually grow forever unless you intervene or install a garbage collector plugin 😛 .

Interestingly, the internal representation can be different than what the plugin API makes things look like. In fact I think I might be in favor of exposing a Sublime-style API to plugins but using two pools internally.

I have a feeling that there exists an architecture with two pools that is nice and clean, but I don't actually know what it would look like or how to go about doing that. On the other hand I think making buffers shared reference counted children of views is more straightforward.

cmyr: I definitely agree that it's important to maintain a distinction between the internal architecture and the API as exposed via plugin libraries; I've been trying to keep the plugin API (especially in the python lib) very view focused.

When I think about architecting this I come back to the asymmetry of the relationship between a `ViewIdentifier` and a `BufferIdentifier`; that is, that a `ViewIdentifier` implicitly also identifies a buffer; every view has a single constant buffer for the duration of its existence. 

Is there any advantage to thinking about some clever packed single identifier value?

cmyr:
I'd like to bring this in so I can add an example plugin to #366, so I've opened #371 for this discussion.

PR #369, opened 2017-07-12
Use CRDT merge to resolve conflicts with Fuchsia sync

trishume:
Use the CRDT merge both when incorporating new changes via the page watcher, and add the ledger conflict resolver machinery.

Tested with two Fuchsia devices syncing through the cloud with Ledger and it successfully handles concurrent editing and reaches a consistent state in a short time. 🎊 🎉 

cc @jimbeveridge

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -237,7 +237,7 @@ impl<W: Write + Send + 'static> Editor<W> {
         for v in self.views.keys() {
             views.push(v.to_owned());
         }
-        
+
         PluginBufferInfo::new(self.buffer_id, &views,
                               self.engine.get_head_rev_id().token(), self.text.len(),
                               nb_lines, self.path.clone(), self.syntax.clone())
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -419,8 +419,8 @@ impl<W: Write + Send + 'static> Editor<W> {
     }
 
     pub fn merge_new_state(&mut self, new_engine: Engine) {
-        // TODO: CRDT merge
-        self.engine = new_engine;
+        print_err!("Xi: merging new state from ledger");
```
raphlinus: I think this is fine for now but it's leaving me wishing for a more systematic way to introspect into the core's behavior.

rust/core-lib/src/tabs.rs
```diff
@@ -648,7 +648,8 @@ pub struct SyncRepo {
 
 #[cfg(target_os = "fuchsia")]
 impl<W: Write + Send + 'static> Documents<W> {
-    pub fn setup_ledger(&mut self, ledger: Ledger_Proxy, session_id: (u64,u32)) {
+    pub fn setup_ledger(&mut self, mut ledger: Ledger_Proxy, session_id: (u64,u32)) {
+        start_conflict_resolver_factory(&mut ledger, vec![0]);
```
raphlinus: The "0" doesn't have context. I'd personally `let key = 0` or something.

Issue #249, opened 2017-05-04
refactor tests in python plugin lib, and run with travis


cmyr:
There are currently a number of tests in the python plugin lib that I was running manually with [py.test](https://docs.pytest.org/en/latest/) during development. These should be pulled out into their own file(s), and travis should be configured to run them automatically.

little-dude:
I could help with. Where are the exisiting tests?
And do you have a kind of test plan in mind already?

cmyr:
all tests are currently in cache.py. Not sure I know what you mean by test plan? For additional tests, going forward?

little-dude:
> For additional tests, going forward?

Yes exactly

cmyr:
yea we haven't gone into that in too much depth. We would definitely like to do more testing. @cbrewster was talking about this a bit a while ago, in #193. 

One thing that I would like is regression tests against the RPC protocol itself, and eventually for the plugin protocol. Then there's lots of room for testing movement and edits, but we'd need some sort of harness such as discussed above. There might be simpler approaches, but I haven't looked into it too closely.

Melvie:
@cmyr  Does this still need to be worked on? I'd like to help out!

cmyr:
@Melvie sure! The python plugin lib is going to be getting a bunch of attention in july, so setting up proper testing for it would be welcome.

Melvie:
@cmyr  Awesome. I've never used travis before so I'll have some learning to do there. But you mentioned all the tests being in cache.py and braking them up into their own files. Any advice as to how you'd break them up at all?

cmyr:
@Melvie iirc there is currently exactly _one_ test. I'm not sure what best practices are for pytest, to be honest? but I suspect there should be a top level 'tests' directory? And I'd probably just add a single new file called like `test_cache.py` or something. But I'd personally just do some googling. I'm also not super familiar with travis, but there's lots of info available. :)

Melvie:
@cmyr So, I have the travis yml set up properly and the pytests will run locally. But when rust builds on travis and  locally I get the following error after it runs **cargo build || exit**:
```
error[E0599]: no function or associated item named `load_defaults` found for type `syntect::highlighting::ThemeSet` in the current scope
   --> core-lib/src/styles.rs:156:22
    |
156 |         let themes = ThemeSet::load_defaults();
    |                      ^^^^^^^^^^^^^^^^^^^^^^^
error: aborting due to previous error(s)
[...]

PR #372, opened 2017-07-15
Pytests refactor

Melvie:
Refactored test in python plugin lib and set up to run with travis as per Issue #249


python/xi_plugin.egg-info/PKG-INFO
```diff
@@ -0,0 +1,10 @@
+Metadata-Version: 1.0
```
cmyr: Is including the .egg necessary?

rust/Cargo.lock
```diff
@@ -37,6 +37,11 @@ version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
cmyr: can we exclude the Cargo.lock from this PR?

Melvie: I'm super new to rust, but from the reading I've done it seems like we can exclude the cargo.lock file for this PR.

python/xi_plugin.egg-info/PKG-INFO
Melvie: I branched and tested pushing with travis without the egg and it fails. So, I created a setup.py file which will build the egg when the tests are run. So, no  including the .egg isn't necessary.

I did find though that python wasn't building on the osx worker though so I might close this request and open a new one when I figure out that

rust/Cargo.lock
cmyr: Yes, my phrasing was maybe a bit misleading. I meant more like, "can we _please_ exclude the Cargo.lock from this PR?" ;)

PR #376, opened 2017-07-25
Add pronunciation to readme

shaunlebron:
Easier to spread the word on something if people immediately know how to pronounce it. [source]( https://www.youtube.com/watch?v=SKtQgFBRUvQ#t=18s)

screen shot 2017-07-25 at 3 14 16 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
   <a href="https://github.com/google/xi-editor">Xi Editor</a>
 </h1>
 
+<p align="center"><em>(pronounced "Zigh")</em></p>
+
 <h4 align="center">A modern editor with a backend written in Rust.</h4>
 
 The xi editor project is an attempt to build a high quality text editor,


shaunlebron:
I signed the CLA.

Agreed about the strange spelling "igh".  Luckily it can only be pronounced one way in English!

jeanm:
Some alternative ways of explaining it:
* Like *seye* in *bullseye*
* `/zaɪ/`
* Like *si* in *quasi* (won't work for everyone, since some people say `/kwɑzi/`)
* Like *sly*, but without the *l* (not quite)

eyelash:
Or maybe just link to https://en.wikipedia.org/wiki/Xi_%28letter%29 which includes pronunciation?

raphlinus:
This is fine as is, thanks! Linking to Wikipedia is not really going to work because the pronunciation of the Greek letter depends on context. My choice of "zigh" is somewhat arbitrary.

PR #378, opened 2017-08-02
CRDT Documentation

trishume:
This PR adds a big document with a bunch of diagrams that explains in detail the text CRDT representation and operations implemented in the `xi-rope` crate. Since a large part of the design of the Xi CRDT is the efficient representations and optimization of the operations, the document describes the CRDT as implemented rather than conceptually. 

It includes snippets of code that are mostly not mandatory to understand, but helpful to clarify the precise meaning of the description and to trace along with the examples. It's kind of like a [literate program](https://en.wikipedia.org/wiki/Literate_programming).

It also cleans up the old `crdt.md` document a little bit and cleans up the doc comments and code where I noticed things were suboptimal so that the code was easier to explain.

### 📝  👁  👁‍🗨  [See the rendered document here](https://github.com/google/xi-editor/blob/crdt-doc-1/doc/crdt-details.md) 👀  ✨ 

Related-ish document about using the Fuchsia Ledger with CRDTS: #380

I've described everything I want to describe and done all the major diagrams. However, there's a few things I have to do before this is ready for review:

- [x] Proofread and edit the document. I've currently done no proofreading other than minimal revision during the process of writing it.
- [x] Add some links between sections of the document.
- [x] Ensure that the image links work when Github renders the document.
- [ ] Maybe squash some commits and run the diagrams through ImageOptim to keep the repo size down.

Link to the source document for the diagrams: https://www.figma.com/file/UGOAcpKR5WIP81t3DGPIP2dR/CRDT-Merge-Diagrams

cc @raphlinus @jimbeveridge


doc/crdt-details.md
```diff
@@ -0,0 +1,1007 @@
...
+
+### Transform Property 2 (TP2) and Operational Transforms
+
+Operational Transformation (OT) is a common way to implement asynchronous text editing. It works by sending *operations* like inserts and deletes between peers and transforming them to apply to the current text. Unfortunately many implementations of OT have a problem where they don't always preserve ordering when text is deleted. For example see the following diagram showing 3 peers sending edits between each other ending up in an inconsistent state:
+
+![TP2 Problem](img/tp2.png)
+
+Acting consistently in cases like this is called having "Transform Property 2". The difficulty of solving this with Operational Transformation can be solved by having a central server do all the transformation so that although the order over deletes is not preserved at least the result can be made consistent. This is what Google Docs and many other collaborative editing systems do.
```
raphlinus: I'd cite [Operational Transformation in Real-Time Group Editors:
Issues, Algorithms, and Achievements](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.53.933&rep=rep1&type=pdf) by Sun & Ellis, 1998, here. I'm ok either with references being simple inline links, or having a references section at the end.

The discussion of TP2 vs what most systems do seems correct, but feels like it might be confusing. What's in my mind is, "there are at least two approaches to the TP2 problem. One is to use a central server that serializes all edits. With this approach, the ordering relative to deleted text may not be preserved, but at least all clients will converge to the same state."

doc/crdt-details.md
```diff
@@ -0,0 +1,1007 @@
...
+
+Xi avoids this problem by using "tombstones", which leave deleted characters in the representation so that ordering can be preserved. This will be described in detail later.
```
raphlinus: Xi takes another approach by using "tombstones," which...

Either here or below, a good chance to cite the [Tombstone Transformation Functions for Ensuring Consistency in Collaborative Editing Systems](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.103.2679&rep=rep1&type=pdf) paper. (http://ieeexplore.ieee.org/document/4207539/ may be a more stable but more paywally link)

doc/crdt-details.md
```diff
@@ -0,0 +1,1007 @@
...
+
+## Representation
+
+The conceptual representation described in `crdt.md` would be very inefficient to use directly. If we had to store an ID and ordering edges for each character and reconstruct the current text via topological sort every time we wanted to know what the current text, Xi would be incredibly slow and use tons of memory.
+
+Instead, we use a representation that allows all the operations we care about to be fast. We also take advantaged of the typical patterns of text document usage to make the representation more memory efficient for common cases.
```
raphlinus: "take advantage"

doc/crdt-details.md
```diff
@@ -0,0 +1,1007 @@
...
+    let old_deletes_from_union = self.deletes_from_cur_union_for_index(rev_index);
+    let delta = Delta::synthesize(&self.tombstones,
+        &self.deletes_from_union, &old_deletes_from_union);
+    delta.apply(&self.text)
+}
+```
+
+#### Engine::deletes_from_cur_union_for_index
+
+If you look back at the [example history scenario](TODO), you'll see `back_computed_deletions_from_6_union`, which shows that for any past `Revision` we can find a set of deletions from the current union string that result in the past text. This helper is what computes deletion sets like `back_computed_deletions_from_6_union`.
```
raphlinus: Probably want to fix up this link.

doc/crdt.md
```diff
@@ -56,9 +48,9 @@ The CRDT state is:
 
   - An undo group id.
 
-  - A set of delete intervals.
```
trishume: I added that text and a little more further up in the `Subset` section when I first mention characters.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+#### Subset::union
+
+Takes two `Subset`s and produces a new `Subset` of the same string where each character has the sum of the counts it has in each input. When treating a `Subset` as a normal set, this is just the union.
+
+#### Subset::transform_expand
+
+`Subset::transform_expand` takes a `Subset` and another "transform" `Subset` and transforms the first `Subset` through the coordinate transform represented by the "transform". Now what does this mean:
+
+`Revision`s are never modified even when the union string expands with newly inserted characters. We can deal with this by treating `Subset`s of inserted characters as coordinate transforms. The only difference to the union string is the inserted characters, so if we can modify the coordinates of a `Subset` from being based on one union string to another, we can work with edits from multiple `Revision`s together.
```
cmyr: >`Revision`s are never modified even when the union string expands with newly inserted characters.

This could be clearer. Like: "`Revision`s are never modified, and their `Edit`s always operate on the union string that exists when they are created. For operate on multiple `Revision`s, have to be able to map coordinates in one's union string to coordinates in another's. We do this by treating [...]`


And a question:

When we use a subset as a transform is it really a subset, or is this just an incidental homomorphism, where the `Subset` type just happens to be share a representation with some `SubsetTransform` type, and so we just overload it and get on with things? This operation felt like a strange use of a `Subset`, and it took me a few read-throughs to get a sense of what is going on.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+The actual procedure works by iterating over the segments of the transform:
+
+- When it encounters a 0-count segment it keeps putting (potentially partial) segments from `self` into the output until it fills its size.
+- When it encounters a non-0-count transform segment it outputs a 0-count segment of the same size.
+
+The `SubsetBuilder` used for the output automatically merges consecutive segments that have the same count. See the diagram below for how this plays out:
+
+![transform_expand workings](img/trans-expand-1.png)
+
+One example of how this can be used is to find the characters that were inserted by a past `Revision` in the coordinates of the current union string instead of the past one:
```
cmyr: One example of how this can be used is to find the characters that were inserted by a past `Revision` in the coordinates of ***another `Revision`'s union string:*** ~the current union string instead of the past one:~

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+///
+/// s2 = self.delete_from_string(s1)
+///
+/// element in self.transform_expand(other).delete_from_string(s0) if (not in s1) or in s2
+pub fn transform_expand(&self, other: &Subset) -> Subset {
+    self.transform(other, false)
+}
+```
+
+#### Subset::transform_union
```
cmyr: what is this used for?

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+    let old_deletes_from_union = self.deletes_from_cur_union_for_index(rev_index);
+    let delta = Delta::synthesize(&self.tombstones,
+        &self.deletes_from_union, &old_deletes_from_union);
+    delta.apply(&self.text)
+}
+```
+
+#### Engine::deletes_from_cur_union_for_index
+
+If you look back at the [example history scenario](#example-history), you'll see `back_computed_deletions_from_6_union`, which shows that for any past `Revision` we can find a set of deletions from the current union string that result in the past text. This helper is what computes deletion sets like `back_computed_deletions_from_6_union`.
```
cmyr: This helper is what computes **these** deletion sets~like `back_computed_deletions_from_6_union`~.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+We can find an `old_deletes_from_cur_union` by taking our current `deletes_from_union` and walking backwards through our list of `Revision`s undoing the changes they have made since the old revision.
+
+There's a problem with this though, the `inserts` and `deletes` subsets use indices in the coordinate space of the union string at the time the `Revision` was created, which may be smaller than our current union string.
```
cmyr: There's a problem with this though, the `inserts` and `deletes` subsets **at a given `Revision`** use indices in the coordinate

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+We could keep track of the transform and account for it, but it's easier to use a helper we need anyway elsewhere that computes what the `deletes_from_union` actually would have been at that previous point in time, that is, relative to the old union string not the current one like we want. This helper is called `Engine::deletes_from_union_for_index` and it performs the work of un-deleting and un-undoing everything after the old revision.
```
cmyr: > [...] that is, relative to the old union string not the current one like we want.

This phrasing is really chunky

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+            }
+        }
+    }
+    deletes_from_union
+}
+```
+
+#### Engine::deletes_from_union_for_index
+
+This function uses the property that each `Revision` contains the information necessary to reverse it to work backwards from the current state of `deletes_from_union` to the past state. For every `Edit` revision it `subtract`s the `deletes` (meaning if something was deleted twice, this will only reverse one), but only if they weren't undone, and then uses `transform_shrink` to reverse the coordinate transform of the `inserts` so that the indices in the intermediate `old_deletes_from_union` refer to the previous union string. `Undo` edits store the symmetric differences of the `deletes_from_union` and the currently undone groups, so those are just reversed.
```
cmyr: > This function uses the property that each `Revision` contains the information necessary to reverse it to work backwards from the current state of `deletes_from_union` to the past state.

"reverse it to": maybe needs a comma after "it"? or make "it" "itself"? or "in order to"?

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+pub fn synthesize(tombstones: &Node<N>, from_dels: &Subset, to_dels: &Subset) -> Delta<N> {
+  // ...
+}
+```
+
+### Engine::delta_rev_head
+
+This operation is similar to `Engine::get_head` except it returns a `Delta` from the text at a specified revision to the current head text. This is useful for things like updating the position of cursors and rich text spans when edits are made.
+
+Like `Engine::get_head` it starts by calling `Engine::deletes_from_cur_union_for_index` to get a `Subset` describing the state of the text at the old revision relative to the current union string. Now we can just use `Delta::synthesize` to create a `Delta` from the old to the new `deletes_from_union`. The problem is, `Delta::synthesize` expects the tombstones `Rope` you give it to correspond to `from_dels`, but we have one for `to_dels`. To fix this, we can use a helper called `shuffle_tombstones` to shuffle things to get an `old_tombstones` corresponding to `from_dels`.
```
cmyr: "to shuffle things to get" needs an edit

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+#### shuffle_tombstones
```
cmyr: General naming nit: I do not find it especially helpful that there are a bunch of functions with "shuffle" in their name that are not only _not_ doing any shuffling, but are actually quite opinionated about orderings. 😁

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+        old_deletes_from_union: &Subset, new_deletes_from_union: &Subset) -> Rope {
+    // Taking the complement of deletes_from_union leads to an interleaving valid for swapped text and tombstones,
+    // allowing us to use the same method to insert the text into the tombstones.
+    let inverse_tombstones_map = old_deletes_from_union.complement();
+    let move_delta = Delta::synthesize(text, &inverse_tombstones_map, &new_deletes_from_union.complement());
+    move_delta.apply(tombstones)
+}
+```
+
+### Engine::edit_rev
```
cmyr: Another name nit: does `edit_rev` actually 'edit' a revision? or is it more just like `new_rev_based_on_another_previous_rev`?

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+This operation has a number of stages:
+
+1. Use the `Delta::factor` helper to split the given `delta` into `ins_delta` and `deletes`. This helper returns an `InsertDelta` containing only the inserts and a `Subset` based on the text (before the inserts) containing the deleted characters. When you see "delta" from now on, it refers to both pieces.
+1. Use `Engine::deletes_from_union_for_index` to work backwards from the present to find the `deletes_from_union` at the time of the base revision: `deletes_at_rev`.
+1. Transform the delta to be based on the union string at the time of `base_rev` rather than the `text` at the time of `base_rev`. This makes later transformations easier.
+    - The `deletes` are transformed using `Subset::transform_expand` to expand the indices to include the characters that weren't in the text the `deletes` was based on.
+    - `ins_delta` is transformed using the similar `InsertDelta::transform_expand` helper. The thing is, with inserts it is ambigous if deleted characters between the same two characters in `text` should be placed before or after characters inserted in the same place. In this case we put the inserts after the deletes.
```
cmyr: why do we put them after? does it matter?

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+    - Each concurrent plugin has a different `priority` and they are useful for expressing what we expect concurrent edits to do. For example inserted auto-indentation should come before new user edits, but matched brackets should come after concurrent user edits, we can set the `priority` of the plugins to get this behavior. In the case of concurrent edits by the same plugin on different synced devices we break ties by session ID.
+1. `Subset::transform_expand` the `deletes` to apply to the head union string after `ins_delta` is applied instead of before. This matches the meaning of `inserts` and `deletes` in `Revision`, whereas `Delta::factor` gives them to use based on the same string. How can we `transform_expand` a `Subset` by an `InsertDelta`, by using the `InsertDelta::inserted_subset` helper to get a `Subset` of the post-insert string designating which characters were inserted.
```
cmyr: down at 5: s/use/us; and a sentence starts with 'How' that is not a question

doc/crdt-details.md
```diff
@@ -303,7 +332,8 @@ Takes two `Subset`s and produces a new `Subset` of the same string where each ch
 
 `Subset::transform_expand` takes a `Subset` and another "transform" `Subset` and transforms the first `Subset` through the coordinate transform represented by the "transform". Now what does this mean:
 
-`Revision`s are never modified even when the union string expands with newly inserted characters. We can deal with this by treating `Subset`s of inserted characters as coordinate transforms. The only difference to the union string is the inserted characters, so if we can modify the coordinates of a `Subset` from being based on one union string to another, we can work with edits from multiple `Revision`s together.
+
+`Revision`s are never modified, and their `Edit`s always refer to the union string that existed when they were created. For operations on multiple `Revision`s, have to be able to map coordinates in one's union string to coordinates in another's. We can deal with this by treating `Subset`s of inserted characters as coordinate transforms. The only difference to the union string is the inserted characters, so if we can modify the coordinates of a `Subset` from being based on one union string to another, we can work with edits from multiple `Revision`s together.
```
cmyr: missing a word, second sentence, after the first comma

cmyr: second last sentence: I might try "**As** the only difference [...] characters, ~so~ if [...]"

and later on, maybe "if we can **map** ~modify~ the coordinates of a `Subset` from ~being based on~ one union string to another,"

doc/crdt-details.md
```diff
@@ -121,26 +131,45 @@ pub struct Delta {
...
 
-If we wanted to go from the "union string" to the current text of the document, you'd delete the characters marked in `deletes_from_union` from the union string. The problem is you want to get the current text really often, so this would be an inefficient way to actually store the text.
+If we wanted to go from the "union string" to the current text of the document, you'd delete the characters marked in `deletes_from_union` from the union string. The problem is, Xi access the current text **very often**, so this would be an inefficient way to actually store the text for large documents.
+
+Instead, the "union string" isn't really stored in Xi, we store it as two separate parts called the `text` and `tombstones`. The `text` stores the current visible document contents, so it is really fast to access, and `tombstones` stores all the characters that are currently in the `deletes_from_union` set. We still store the `deletes_from_union` set to mark deleted characters. Note that the union string is still used for most indices as a coordinate space for `Subset`s like `deletes_from_union`.
```
cmyr: A different take on the first sentence: 

"As an optimization, then, we store the union string in two parts: `text` and `tombstones`."

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+`Engine` is the top-level container of text state for the CRDT. It stores the current state of the document and all the `Revision`s that lead up to it. This allows operations that require knowledge of history to apply `Revision`s in reverse from the current state to find the state at a point in the past, without having to store the state at every point in history. Be sure to read the code in this case, all the fields are described by doc comments.
+
+**SUPER IMPORTANT INSIGHT:** Because the union string preserves the textual ordering of inserted characters, indices in the union string only depend on the set of inserted characters and not what order they were added in the history. This means that the correct representation of a `Revision` for a given edit depends only on the *set* of `Revision`s before it in the history and not their order.
+
+### Example History
+
+Bringing it all together, here's a sketch of how a simple editing scenario would be represented this way.
+
+![Representation example](img/representation-overview.png)
```
trishume: I did most of this.

And I thought about making all the diagrams interactive widgets, but I realized I didn't have remotely enough time.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+
+## Table of Contents
+
+- [Motivation](#motivation): Why Xi's CRDT is the way it is.
+- [Representation](#representation): Describes the representation Xi uses to implement the CRDT in a memory and time efficient way.
+- [Operations](#operations): Describes all the operations implemented on the representation to allow it to support undo, asynchronous edits, distributed synchronization and more.
+    - [Engine::merge](#enginemerge): Description of the CRDT merge operation used for multi-device syncing.
+
+## Motivation
+
```
trishume: I put this in the introduction.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+- Inserting one piece of text in the middle of another producing a new piece of text
+- Deleting an interval from a piece of text, producing a new piece of text
+
+Behind the scenes, `Rope` is an immutable balanced tree structure using Rust's atomic reference counting smart pointer (`Arc`) to share data, so "copying" any sub-tree is a very fast `O(1)` operation. The leaves of the tree are chunks of text with a maximum size. For example, deleting an interval of a `Rope` only requires creating a few new nodes that reference the sub-tree to the right of the deleted interval and to the left, and creating up to two new leaves if the deleted interval doesn't lie on a chunk boundary.
+
+Obviously `Rope`s will be slower and take more memory than small `Strings` but they have an asymptotic advantage when working with large documents.
+
+For a deeper look at `Rope`s see the [Rope Science](rope_science/intro.md) series.
+
+### Subset
```
trishume: I did both of these things.

doc/crdt-details.md
```diff
@@ -0,0 +1,1054 @@
...
+Bringing it all together, here's a sketch of how a simple editing scenario would be represented this way.
+
+![Representation example](img/representation-overview.png)
+
+Shown below is a longer history of `Edit` `Revision`s. At each point, the `inserts` and `deletes` `Subset`s of the `Revision` that was just added are shown. Below that is what the contents of `Engine` would be immediately after that `Revision` was added to it.
+
+Also included is a "Concepts" section that includes things that aren't actually stored but are useful for understanding the representation. This includes the "union" string that all indices are based on. It also includes `back_computed_deletions_from_6_union`, which is like `deletes_from_union` except instead of being based on the union at the time of the edit, it is based on the union of `Revision` 6. This shows that since we never throw away information, we can represent the `text` at any past `Revision` as a set of deletions from the union string of any later revision.
+
+```
+()  Revision: before any revisions are made
```
trishume: That's why I added the diagram above it, but I kept this in since it's larger and more explicit/complete.

doc/crdt-details.md
trishume: This is talking about an example, specifically the diagram below, where it is actually talking about taking coordinates into the current union string.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+1. Apply the `ins_delta` to `text` using `Delta::apply` and also `transform_expand` `deletes_from_union` to include the newly inserted characters.
+1. Now that we've applied the inserts, we just need to apply the deletions. But wait! What if the undo group of this edit was undone between `base_rev` and now? The edit might already be undone, in which case the `deletes` shouldn't apply and the inserted characters should be deleted.
+    1. We solve this by checking if our undo group is currently undone and if it is we set `to_delete` to the inserted characters, and otherwise set it to `deletes`.
+    1. Then we compute a new `deletes_from_union` by using `Subset::union` to delete the characters in `to_delete`.
+1. We use a helper called `shuffle` which is like (and uses) `shuffle_tombstones` but it updates both the `text` and `tombstones` from corresponding to an old `deletes_from_union` to a new one. It "shuffles" characters between `text` and `tombstones` as necessary so that their new state corresponds to the new `deletes_from_union`.
```
cmyr: ~from corresponding to an old `deletes_from_union` to a new one~ to refer to a new `deletes_from_union`.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+/// controls whether the insertions in `self` come after those specific in the
+/// coordinate transform.
+pub fn transform_expand(&self, xform: &Subset, after: bool) -> InsertDelta<N> {
+    // ...
+}
+```
+
+#### shuffle
+
+This helper is used in many different operations that use the pattern of computing how `deletes_from_union` should change and then using that to fix up the `text` and `tombstones` to correspond to that.
```
cmyr: should change and then ~using that to fix up the `text` and `tombstones` to correspond to that.~ **updating `text` and `tombstones` to reflect that new state.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+    // Delta that deletes the right bits from the text
+    let del_delta = Delta::synthesize(tombstones, old_deletes_from_union, new_deletes_from_union);
+    let new_text = del_delta.apply(text);
+    (new_text, shuffle_tombstones(text,tombstones,old_deletes_from_union,new_deletes_from_union))
+}
+```
+
+### Engine::undo
+
+Undo works conceptually by rewinding to the earliest point the toggled undo group appears, and replaying history from there but with the correct undone groups not applied.
```
cmyr: 'correct' is ambiguous here. Maybe "[...] replaying history, skipping toggled undo groups."?

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+
+### Engine::merge
+
+This is the operation you were (maybe) waiting for! The CRDT merge operation that allows peer-to-peer syncing of edits in a conflict-free eventually-consistent way. It takes `self` and another instance of `Engine` and incorporates any changes which that `Engine` has into `self`.
+
+It does this by finding changes which the other `Engine` has but it doesn't and doing a whole bunch of transformation so that those edits can be appended directly on to the end of `self`'s list of `revs`. The append-only nature of merge preserves the ability for operations like `Engine::delta_rev_head` to work, and allows future optimizations of how things are persisted.
```
cmyr: i'd replace the second 'it' on this line with `self`. 

also missing an "s" on transformation.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+
+In practice the order of the `Revision` history will tend to be very similar between peers. Any edit that occurs while another edit is visible on screen (present in this peer's `Engine`) will never be re-ordered before that visible edit. Thus if there are no concurrent edits made on devices that sync with each other by merging, the devices will end up with the same `Revision` history.
+
+**Note:** As of the time this was written, Xi's `merge` implementation does not support undo operations, and will panic if one is encountered. We plan to fix this and expect that the structure of the merge operation will stay the same, just with more cases. So for the rest of this description, presume `Edit`s are the only type of `Revision`.
+
+#### Base and common revisions
+
+An important part of merging is figuring out which revisions the two sides have in common. These common revisions are found as two parts:
+
+1. The "base": A prefix of the same length of both histories such that the set of revisions in both prefixes is the same.
```
cmyr: unclear.

Maybe: "Starting from the first revision in each history, the longest sequence of revisions such that the set of revisions in the sequence is equal in both histories; in other words the maximal common prefix set."

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+    - Note that they aren't necessarily in the same order, just all shared. That's because all indices are relative to the union string, which is the same regardless of the order of the revisions that created it.
+    - We ignore everything in this prefix, so the longer it is the faster the merge runs, but a length of 0 is perfectly fine and in fact at the time this was written, that's what Xi uses. However, we plan on finding the maximal such prefix as an important optimization.
+1. The "common" revisions: After the base, some revisions on each side will be shared by both sides. The common revisions are the intersection of the two revision sets after the base.
+    - Note that the common revisions aren't necessarily in the same positions or order on each side.
+
+![CRDT Merge Flow Example](img/merge-intro.png)
+
+If you're wondering how you can end up with common revisions not in the base, and in different positions and ordering, that's because it's rare. But these cases can occur even with only two peers. See the example below.
```
cmyr: strictly speaking, it may not be the case that the reason I'm wondering how this is the case is "because it's rare". You could say "well: it's rare." (or something else conversational) or you could just jump to the explanation.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+
+![CRDT merge common in different order](img/merge-common-diff-3.png)
```
cmyr: This diagram could use some labels or a legend. It isn't clear what makes the red arrow special, or why edit 4 doesn't get synced etc.

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+    let b_ids: BTreeSet<RevId> = b.iter().map(|r| r.rev_id).collect();
+    a_ids.intersection(&b_ids).cloned().collect()
+}
+```
+
+#### Rearranging
+
+To make the rest of the algorithm easier, it would be nice if we didn't have to worry about any of the common revisions and could just work with the new revisions on each side. We can realize this by transforming and reordering all the new revisions on each side after all of the common revisions.
+
+We can safely use transforms to reorder two revisions if neither of the revisions depends on the other. A revision depends on all the revisions that were in the `Engine` at the time `Engine::edit_rev` was called. We know that a new revision can't depend on common revision after it, because the append-only nature means that ordering couldn't occur. Furthermore, the common revision can't depend on the new revision, because by definitions the other peer has the common revision but not the new revision, and merges never merge a revision but not its dependencies. Thus, even without explicit dependency information, we know we can reorder all the new revisions on each side after all the common revisions.
```
cmyr: third sentence, missing a word: "on **a** common revision after it"

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+
+Note that though it's nice to think about it as reordering the histories, really we just need to compute the transformed new revisions, transforming each new revision to be based on all common revisions after it. See the "computed" and "taken to next step" annotations of the diagram below, note that transformed operations are marked with a prime tick but they maintain their ID.
+
+![CRDT merge rearrange](img/merge-rearrange.png)
+
+We do the rearranging by working from the end of the list of revisions to the beginning. We build up a `Subset` representing all the characters that were added to the union string by the common revisions after a point. Starting with the identity transformation (an entirely 0-count `Subset`), whenever we encounter a common revision we add the characters it inserts to the transform. When we encounter a new revision we `transform_expand` its inserts and deletes by the transform `Subset` and add it to the output `Vec`. Then we `transform_shrink` the transform `Subset` by the characters inserted in the new revision, because we only want to fast-forward new revisions over common revisions after them, not new revisions after them.
```
cmyr: "after a point.": Which point? an arbitrary point? This is unclear.

Also I might rephrase the first sentence, and just say "we do the rearranging by working backwards from the end of the list of revisions."

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+
+We can safely use transforms to reorder two revisions if neither of the revisions depends on the other. A revision depends on all the revisions that were in the `Engine` at the time `Engine::edit_rev` was called. We know that a new revision can't depend on common revision after it, because the append-only nature means that ordering couldn't occur. Furthermore, the common revision can't depend on the new revision, because by definitions the other peer has the common revision but not the new revision, and merges never merge a revision but not its dependencies. Thus, even without explicit dependency information, we know we can reorder all the new revisions on each side after all the common revisions.
+
+Note that though it's nice to think about it as reordering the histories, really we just need to compute the transformed new revisions, transforming each new revision to be based on all common revisions after it. See the "computed" and "taken to next step" annotations of the diagram below, note that transformed operations are marked with a prime tick but they maintain their ID.
+
+![CRDT merge rearrange](img/merge-rearrange.png)
```
cmyr: in this graph: not clear what "computed" means. Does it mean "modified"?

doc/crdt-details.md
```diff
@@ -0,0 +1,1086 @@
...
+                        inserts: transformed_inserts,
+                        deletes: transformed_deletes,
+                        priority, undo_group,
+                    })
+                }
+            },
+            Contents::Undo { .. } => panic!("can't merge undo yet"),
+        };
+        if let Some(edit) = contents {
+            out.push(Revision { edit, re
```
cmyr: shouldn't `edit` have a field name?

cmyr: "has some desirable properties": maybe mention an example or two?

cmyr: end of this section: s/new/newly

cmyr: "they include" twice

doc/crdt-details.md
trishume: I think my existing phrasing better matches what `shuffle` does. Which characters are in `text` and which are in `tombstones` is based on the `deletes_from_union`. They don't refer to the `deletes_from_union`, it's more like their correct arrangement is determined by it.

doc/crdt-details.md
trishume: I added a clarifying paragraph above

doc/crdt-details.md
trishume: Clarified better in the text above

doc/crdt-details.md
trishume: Field punning

doc/crdt-details.md
trishume: Clarified below, we don't actually find the maximal prefix yet, we just plan on it.

trishume:
I did some more edits to clarify things based on comments. Thanks @cmyr!

I think this is now in a state good enough to merge.

PR #382, opened 2017-08-09
Replace print_err! with eprintln!

cmyr:
@raphlinus this is ready whenever fuschia is.


cmyr:
Updated this patch, it should be gtg.

PR #383, opened 2017-08-11
Pytests refactor

Melvie:
Refactored tests in python plugin lib and set up to run with travis as per Issue #249


.travis.yml
```diff
@@ -1,25 +1,55 @@
...
+  else
+    sudo curl -L https://raw.githubusercontent.com/pyenv/pyenv-installer/master/bin/pyenv-installer | bash
+    sudo apt-get update
+    export PATH="~/.pyenv/bin:$PATH"
+    eval "$(pyenv init -)"
+    eval "$(pyenv virtualenv-init -)"
+    sudo pyenv install 3.6.0
+    sudo pyenv virtualenv 3.6.0 general
+    sudo pyenv global general
+    sudo apt-get -y install python3-pip python-dev
```
eyelash: Why is this line needed here while on macOS it seems to be working without it?

Melvie: I ran a test with out it and it seems it is not required.

eyelash: @Melvie I guess you could use `push --force`

eyelash: @Melvie BTW my question was about the line `sudo apt-get -y install python3-pip python-dev` on Linux and not about pyenv and setuptools on macOS.

eyelash: I just saw that on line 561 of the travis output it says `sudo: pyenv: command not found` so it looks like the installation through pyenv did not work and `sudo apt-get -y install python3-pip python-dev` then installs python through the official Ubuntu repositories.

Melvie: @eyelash  Thanks for reviewing this and helping me out, really appreciate it. I added the pyenv and setup tools back to macOS and removed the sudo as you suggested from the Linux portion.

cmyr:
Going to close this for now; the python lib is out of date and these tests won't currently be passing anyway. Thanks for the PR though!

PR #387, opened 2017-08-25
Use trait object for RpcPeer

cmyr:
This change allows us to ignore the peer's concrete type, which has the
practical effect of removing all the places where we previously had to
pass around a generic <W: Write> parameter.

It will hopefully have the secondary effect of allowing for a cleaner
modification of the rpc system to support typed handlers and peers.


trishume:
I haven't reviewed the actual diff, but I'd like to express my support for the idea, I doubt the trait dispatch will be a significant performance bottleneck and all those generics were annoying.

cmyr:
@trishume I wrote some microbenchmarks (`cargo bench`) for static vs dynamic dispatch, and the difference was lost in the SD. Although it's possible the compiler is doing something tricky, since the trait is only implemented in one place...

PR #388, opened 2017-08-27
Fix build break, removal of W trait was incomplete

raphlinus:
Looks like one of the remaining <W> parameters didn't get removed
from #387. This should fix the build.


cmyr:
Ah yea that was an addition in the other outstanding PR that just merged.

PR #392, opened 2017-09-01
Update xi-unicode to Unicode 10.0

cbreeden:
This pull request updates the test suite and tables for Unicode 10.0.  There isn't any significant changes from Unicode 9.0 to 10.0 except for the excerpt:

```
- Updated the list of ranges and blocks that default to class ID.
```

This range update is what resulted in the modifications of `tables.rs`.  Unfortunately, the test suite provided from the Unicode Consortium didn't provide any tests to ensure this transition went smoothly.  That's too bad.

cf: <http://unicode.org/reports/tr14/#Modifications>


raphlinus:
Thanks for doing this work!

PR #393, opened 2017-09-01
Typed RPC handling

cmyr:
## Typed `Handler`s

This PR updates `xi-rpc`, chiefly to add associated types to the `Handler` trait.

Types that implement `Handler` now specify the types of RPCs they handle. This is chiefly intended to be used with serde; the general idea is that the set of accepted RPCs can be represented as a pair of enums, which can be automatically parsed using serde's `derive` macros.

This PR doesn't include the large refactors required to take advantage of this change, although I have them largely done in other working branches.

### Rationale

The major advantage of this change is that it means adding a new RPC call is as simple as adding a new member to an enum, and lets `xi-rpc` do more input validation and error handling.

#### `Parser` trait?

The `Parser` trait: I wanted to avoid requiring the associated types from having a `: Deserialize` trait bound; the approach I took was to split `Handler` into two traits, `Parser` and `MethodHandler`. We provide an implementation of `Parser` for types which _do_ implement `Deserialize`. (Technically `DeserializeOwned`, right now, although this may change if something like https://github.com/arcnmx/serde-value/issues/15 is resolved). However, it may not be unreasonable to just require the associated types to be `Deserialize`; another option would be to create a new trait, like `RpcCall`, and use that as a constraint on the associated types, and provide a free implementation for types that are `Deserialize`. Which of these approaches makes the most sense?

#### Performance
Performance: The current hand-rolled parsing is quite efficient, because after the initial parsing to a `Value` everything else is a borrow. The new system currently involves first parsing to a `Value`, and then parsing _again_ to one of the associated types. In some small benchmarks, the new approach is about 25% slower (the 1st and 3rd items here):

```
test borrow       ... bench:      14,427 ns/iter (+/- 48) // current impl in master
test own          ... bench:      18,127 ns/iter (+/- 409) // shim impl in this PR
test serde        ... bench:      19,348 ns/iter (+/- 499) // impl after refactor
test future_serde ... bench:       9,776 ns/iter (+/- 153) // (optimistic) future impl with serde borrowing
```

(the benchmark code is available here: https://github.com/cmyr/xi_der_bench)

There are some big potential future gains, however, because we do not currently take advantage of serde's borrowing functionality; the last benchmark suggests we should be able to at least equal current performance as work continues on serde.

However: if we had to live with this slowdown indefinitely, would that be an acceptable tradeoff?

### Conclusion

If this approach seems sound, I will prepare some follow-on PRs:
1. Improve the error handling story in `xi-rpc`, to formalize an error system based on the [json-rpc spec](http://www.jsonrpc.org/specification#error_object);
1. Refactor `xi-core` to use this for the core protocol
1. Ditto the rust plugin-lib

And then look into a few other possible changes in `xi-rpc`:

1. Investigate a similar change for the _peer_ side of the RPC channel
1. Investigate some changes to support more flexible idle scheduling

If this approach _doesn't_ seem sound, I will stop talking about this and move on to something more concretely useful. 🤷‍♂️

### DIFF
--- a/rust/core-lib/src/lib.rs
+++ b/rust/core-lib/src/lib.rs
@@ -91,7 +91,7 @@ extern crate xi_rope;
 extern crate xi_unicode;
 extern crate xi_rpc;
 
-use xi_rpc::{RpcPeer, RpcCtx, Handler};
+use xi_rpc::{RpcPeer, RpcCtx, RpcCall, MethodHandler};
 
 pub type MainPeer = RpcPeer;
 
@@ -112,27 +112,29 @@ impl MainState {
     }
 }
 
-impl Handler for MainState {
-    fn handle_notification(&mut self, mut ctx: RpcCtx, method: &str, params: &Value) {
-        match Request::from_json(method, params) {
+impl MethodHandler for MainState {
+    type Notification = RpcCall;
+    type Request = RpcCall;
+    fn handle_notification(&mut self, mut ctx: RpcCtx, rpc: Self::Notification) {
+        match Request::from_json(&rpc.method, &rpc.params) {
             Ok(req) => {
                 if let Some(_) = self.handle_req(req, &mut ctx) {
-                    print_err!("Unexpected return value for notification {}", method)
+                    print_err!("Unexpected return value for notification {}", &rpc.method)
                 }
             }
-            Err(e) => print_err!("Error {} decoding RPC request {}", e, method)
+            Err(e) => print_err!("Error {} decoding RPC request {}", e, &rpc.method)
         }
     }
 
-    fn handle_request(&mut self, mut ctx: RpcCtx, method: &str, params: &Value) ->
+    fn handle_request(&mut self, mut ctx: RpcCtx, rpc: Self::Request) ->
         Result<Value, Value> {
-        match Request::from_json(method, params) {
+        match Request::from_json(&rpc.method, &rpc.params) {
             Ok(req) => {
                 let result = self.handle_req(req, &mut ctx);
                 result.ok_or_else(|| json!("return value missing"))
             }
             Err(e) => {
-                print_err!("Error {} decoding RPC request {}", e, method);
+                print_err!("Error {} decoding RPC request {}", e, &rpc.method);
                 Err(json!("error decoding request"))
             }
         }
--- a/rust/core-lib/src/lib.rs
+++ b/rust/core-lib/src/lib.rs
@@ -91,7 +91,7 @@ extern crate xi_rope;
 extern crate xi_unicode;
 extern crate xi_rpc;
 
-use xi_rpc::{RpcPeer, RpcCtx, RpcCall, MethodHandler};
+use xi_rpc::{RpcPeer, RpcCtx, RpcCall, Handler, RemoteError};
 
 pub type MainPeer = RpcPeer;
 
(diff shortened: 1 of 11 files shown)


cmyr:
I've pushed a commit to my fork that includes the proposed changes to `xi-rpc`, which are usefully illustrative: https://github.com/cmyr/xi-editor/pull/1/commits/90ac5e29767681fa9c7282b2368a5e4dd8bcc530

rust/rpc/src/error.rs
```diff
@@ -0,0 +1,175 @@
...
+    Disconnect,
+}
+
+impl fmt::Display for ReadError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            ReadError::Io(ref err) => write!(f, "I/O Error: {:?}", err),
+            ReadError::Json(ref err) => write!(f, "JSON Error: {:?}", err),
+            ReadError::NotObject => write!(f, "JSON message was not an object."),
+            ReadError::Disconnect => write!(f, "Peer closd the connection."),
```
raphlinus: typo

rust/rpc/src/error.rs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        let (code, message, data) = match *self {
+             RemoteError::Parse(ref d) => (-32700, "Parse error", d),
+             RemoteError::InvalidRequest(ref d) => (-32600, "Invalid request", d),
+             RemoteError::MethodNotFound(ref d) => (-32601, "Method not found", d),
+             RemoteError::InvalidParams(ref d) => (-32602, "Invalid params", d),
+             RemoteError::Custom { code, ref message, ref data } => {
+                 (code, message.as_ref(), data)
+             }
+             RemoteError::Unknown(_) => panic!("The 'Unknown' error variant is\
```
raphlinus: need space: `is \`

rust/rpc/src/lib.rs
```diff
@@ -184,58 +185,80 @@ impl<W: Write + Send> RpcLoop<W> {
...
+                        self.peer.put_rx(Some(json));
                     }
                 }
-                self.peer.put_rx(Value::Null);
-                // TODO: send disconnect error to all pending
             });
+
             let mut idle = VecDeque::<usize>::new();
             loop {
+                // only do idle work if no pending messages
```
raphlinus: This comment is restating the code. I might write "pending messages take priority over idle work"

rust/rpc/src/error.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+impl From<io::Error> for ReadError {
+    fn from(err: io::Error) -> ReadError {
+        ReadError::Io(err)
+    }
+}
+
+//TODO: code review discussion: do we want, in general, to support these
```
raphlinus: Result of discussion: for request, we should return an error and then close the connection (without disrupting core). For notification, we should log an error and close the connection.

rust/core-lib/src/lib.rs
```diff
@@ -113,27 +113,29 @@ impl MainState {
...
-            Err(e) => print_err!("Error {} decoding RPC request {}", e, method)
+            Err(e) => print_err!("Error {} decoding RPC request {}", e, &rpc.method)
         }
     }
 
-    fn handle_request(&mut self, mut ctx: RpcCtx, method: &str, params: &Value) ->
-        Result<Value, Value> {
-        match Request::from_json(method, params) {
+    fn handle_request(&mut self, mut ctx: RpcCtx, rpc: Self::Request) ->
+       Result<Value, RemoteError> {
```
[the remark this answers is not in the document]
cmyr: This code will all be removed next PR. I fixed the indentation where the handler is declared in xi-rpc, though.

raphlinus:
Thanks for this!

PR #395, opened 2017-09-13
xi-rpc testing tools

cmyr:
This PR adds some tools for setting up a mock runloop, and adds some tests to xi-rpc that uses these tools.

In follow-on work, I want to use these tools to write a suite of integration tests for xi-core.


cmyr:
I'm rethinking some of my approach here as I start writing tests in xi-core, will reopen after a bit of revision.

PR #397, opened 2017-09-13
Clarify minimal supported rust version.

dsp:
Another simple PR to clarify we are relying on Rust 1.18 or later.


raphlinus:
Thanks! We'll need to be vigilant to keep this up to date, for example we're considering `eprintln!` which I think would bump this to 1.19.

PR #398, opened 2017-09-13
Two needless `mut`s removed

llogiq:
Just a trivial improvement


raphlinus:
Thanks for the cleanup, and for fixing the spelling :)

PR #399, opened 2017-09-14
Feature/runloop returns err

cmyr:
This is primarily to facilitate the writing of certain types of tests,
specifically those where we want to start a run loop with a specific
handler, and then verify that JSON is handled as expected.

This also has the benefit of providing more error information to the
call site, where more specific decisions can be made about error
handling and/or logging.


cmyr:
@raphlinus As in, create a serde_json value and then go to a string from there?

Part of my thinking with these tests is that I want the inputs to be completely clear, and so avoid any steps that might modify them. Also our expected input are just sequential newline-delimited JSON objects, which I can't create with `json!()` without manually concatenation.

raphlinus:
@cmyr Makes sense, I understand the reasoning, and the newline definitely does add more code in the test, which is generally something to avoid.

PR #402, opened 2017-09-15
refactor core-lib to use typed RPC system

cmyr:
This builds on the work done in #393.

A benefit of this work is that (potentially) the generated documentation for `core_lib::rpc` can eventually serve as the canonical documentation of the RPC protocol, using doctests to ensure that the documentation is up to date with protocol changes.

### DIFF
--- a/rust/core-lib/src/rpc.rs
+++ b/rust/core-lib/src/rpc.rs
@@ -12,7 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! RPC handling for communications with front-end.
+//! The main RPC protocol, for communication between `xi-core` and the client.
+//!
+//! We rely on [Serde] for serialization and deserialization between
+//! the JSON-RPC protocol and the types here.
+//!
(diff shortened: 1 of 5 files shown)


cmyr:
rebased to run the tests added in #399

rust/core-lib/src/rpc.rs
```diff
@@ -111,253 +340,141 @@ pub enum EditCommand<'a> {
...
+#[serde(rename_all = "snake_case")]
+#[serde(tag = "method", content = "params")]
+/// The edit related requests.
+pub enum EditRequest {
+    /// Cuts the active selection, returning their contents,
+    /// or `Null` if the selection was empty.
+    Cut,
+    /// Copies the active selection, returning their contents or
+    /// or `Null` if the selection was empty.
+    Copy,
```
little-dude: Nice! I'm surprised this works. I assumed `Copy` was a reserved keyword, so I went for an ugly `Copy_` in my own code.

PR #403, opened 2017-09-22
Add cache simulator to docs

raphlinus:
Add an interactive cache simulator to explore different cache eviction
policies for syntax highlighting.

It's an "interactive doc", can be viewed by pointing the browser at
the HTML. This version is _extremely_ spare wrt visual design and
supporting text, but the visualization is pretty compelling.


cmyr:
Thanks for addressing the nit! This is a very useful illustration, glad you took the time.

PR #404, opened 2017-09-26
Revert unintended "col" to "column" rename

raphlinus:
The recent RPC changes renamed the "col" parameter in the "gesture"
command to "column", which broke gestures. This reverts that change.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -985,7 +985,7 @@ impl Editor {
             Drag (MouseAction {line, column, flags, ..}) => {
                 self.do_drag(line, column, flags);
             }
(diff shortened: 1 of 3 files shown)


cmyr:
Heh, somehow managed to not include this one RPC in the integration tests. :| will add a commit.

cmyr:
@raphlinus I might've gotten a bit cute here: If you'd prefer this as a separate PR force your original commit and ping me

raphlinus:
Adding a test is good, thanks. Dear googlebot, I confirm everybody is ok with this commit.

raphlinus:
Nah, having a test with the change is good, I would have added it had it been on my radar. I'll just wait for travis before merging.

PR #406, opened 2017-10-05
use typed rpc system for plugins

cmyr:
Mostly just that; removing the duplicate definitions of various plugin RPC types from the plugin lib, which now imports those types from `xi_core::plugins::rpc`. 

I also made the tweak of no longer having `PluginRef` be the handler for the plugin process, since it was only ever forwarding everything to `PluginManager` anyway. `PluginManager` now implements `Handler` for those types, and receives events directly.

This means that plugins need to know their PID when communicating with core, but I think that makes sense anyway for logging and diagnostics.


cmyr:
okay @raphlinus this is ready to go. Unfortunately it's sort of two PRs in one: the first just moves away from manual parsing of plugin RPCs (and shares types between the core lib and the plugin lib) and the second makes the plugin RPC handler be the plugin manager directly, instead of being handled by the PluginRef.

ofc I feel strongly about burning most of this down at some point, but this is still an improvement on the current code.

There's a case to be made that core-lib should now have an 'rpc-types' feature; it seems like overkill to have the plugin lib bringing in the whole crate. When we go to futures however all the RPC types will get redefined, and at that point maybe we can move them to their own crate which can also be shared with various clients.

Issue #381, opened 2017-08-06
Can't Input Hard Tabs


OxyDeadbeef:
Pressing tab always gives 4 spaces.

cmyr:
Yea this is a current limitation, and can be addressed when we add user preferences. We should also probably add per-buffer tab preferences at some point.

linde12:
I solved this temporarily in the client by sending `\t`. I then render it like four spaces but treat it like a single character so that i can, e.g., still remove it by pressing backspace just once.

I imagine the core will do something similar to this. Looking forward to it!

raphlinus:
Agree this is important. It does need to be per-buffer because different languages have different requirements (I believe they're more or less mandatory for go and make, while for example Python's PEP8 recommends spaces). It's also related to a setting for the tab indent; 4 spaces is common but not universal.

I can't say exactly when we'll get to this. As @cmyr points out, it's not quite self-contained but depends on some mechanism for preferences.

OxyDeadbeef:
I'd like to point out that, as far as I know, no programming language requires a specific type of indentation.  Some have "preferred" indentation, but that doesn't really mean anything.

Changing tab input settings wouldn't necessarily require a preferences menu, but could do something similar to gedit:

![screenshot from 2017-08-17 22-24-12](https://user-images.githubusercontent.com/8846235/29443027-e311db6e-839a-11e7-89ab-725f349a563b.png)

joerixaop:
That depends on whether you consider Makefiles a language

NatElkins:
@OxyDeadbeef F# requires tabs.
![tabs](https://user-images.githubusercontent.com/1795462/29576909-92abd9ae-8737-11e7-8550-40f539cc2542.gif)

Like the gif says, it can be turned off with the `#indent "off"` directive, but code with tabs won't work out of the box.

maxnordlund:
[Gofmt](https://golang.org/cmd/gofmt/) also uses tabs for indentation.

eyelash:
Since #409 landed you can create a file `~/.config/xi/preferences.xiconfig` with the content
```
translate_tabs_to_spaces = false
```
which will cause Xi to input tabs when you press tab.

cmyr:
As @eyelash mentions, while the ergonomics aren't great, there is now a user setting for tabs vs spaces, which I think is enough to close this issue. On xi-mac, as of https://github.com/google/xi-mac/pull/71, tabs and spaces can now be set (crudely) on a per-file basis.

PR #409, opened 2017-10-11
Barebones user config, plus plugin loading

cmyr:
This PR implements very basic user config loading, and piggybacks on that to load plugins from disk.

There is a bunch of stuff this doesn't do: settings are global and only load once, there are no syntax-specific or platform-specific settings. (I guess a tracking/discussion issue for this would be nice; I'll do that shortly)

## Config

We ship with default config settings, in `core-lib/assets/default.toml`. Additionally, we look for a user config directory, first by checking the `XI_CONFIG_DIR` env var, and then falling back to `$XDG_CONFIG_HOME/xi`, or `$HOME/.config/xi`. We look for a TOML file in this directory named `preferences.xiconfig`; if present, any valid fields in this file override the defaults.

This config file now lets the user specify a global preference for tabs and spaces (#381), as well as spaces-per-tab, in addition to specifying the `plugin_search_path`, about which:

## Plugin Manifests

Plugins now require manifest files. These tell the editor where to find the plugin executable and when to run the plugin, and generally describe plugin capabilities.

Plugins are loaded by looking for files named `manifest.toml` in any directory that is a child of a directory in the `plugin_search_path`.

### Client installed plugins

If a client ships with some plugins (like we do on xi-mac, with the syntect plugin) they can install them in a directory outside of the user's `XI_CONFIG_DIR`, to conflicts with any of the user's plugins. In this case, the client should set the `XI_SYS_PLUGIN_PATH` environment variable to point to the folder containing the client's bundled plugins; this directory will be appended to the user's `plugin_search_path`, meaning these plugins won't load if an identically named plugin is in the user's `plugin_search_path`.

### xi-mac example
(A companion PR is available here: https://github.com/google/xi-mac/pull/68)

On xi-mac, we now include in our app bundle a folder called `plugins`, with the following structure:
```
plugins/
  syntect/
    manifest.toml
    bin/
      xi-syntect-plugin
```

and manifest.toml looks like this:

```toml
name = "syntect"
version = "0.1"
exec_path = "./bin/xi-syntect-plugin"
activations = ["autorun"]
```

At runtime, we set `$XI_SYS_PLUGIN_PATH` to the location of this folder within our app bundle, and syntect loads and runs as expected.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -61,6 +61,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "config"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "lazy_static 0.2.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "nom 3.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.4.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "crossbeam"
 version = "0.2.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -147,6 +158,14 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "odds 0.2.25 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "nom"
+version = "3.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


eyelash:
Why not pass the path as an argument to the `client_started` RPC instead of using the `XI_SYS_PLUGIN_PATH` environment variable?

jansol:
On macOS and Windows the default path should be ~/Library/{Application Support, Preferences, Caches}/Xi and %APPDATA%/Xi respectively. (Many applications skip the Preferences dir on mac and just use Application Support, which is fine too)

cmyr:
@eyelash I hadn't thought of that. The only big difference (to me) is that it means plugins would be slightly slower starting, don't think that would be a big problem though.

cmyr:
@jansol I _loosely_ agree, in that I think we should be respecting platform idioms as much as possible. We should probably be moving the _bundled_ plugins into `~/Application Support`. For the user config stuff, however, my inclination is to be more unix-y? I guess I sort of see Xi as a unix tool, and imagine a xi config being shared across multiple machines the way my `.vimrc` is currently.

In any case, a frontend author can enforce whatever policy they like by setting the `$XI_CONFIG_DIR` variable. There's maybe a good case that `xi-core` shouldn't make any assumptions at all about user config stuff, and just uses defaults of `$XI_CONFIG_DIR` is not set?

jansol:
Good point wrt frontends setting `$XI_CONFIG_DIR`. Maybe not setting it could be outright considered an error?

cmyr:
@jansol If we ship with reasonable defaults, I'm not sure it would make sense to crash if we can't find settings. I definitely think it's something we should log, though, and maybe we can consider it down the road when we see what approach frontends take?

cmyr:
Okay, I'm going to put this in with a few notes:

1. we should perhaps be using vargs instead of env vars;
2. we should perhaps be using the RPC system for setting the system plugin dir (although this makes sense as a varg as well)
3. I should probably rename `prefs.rs` to `config.rs` and import the `config` crate as `config_rs`.

Issue #410, opened 2017-10-12
User config tracking issue


cmyr:
I've started working on user config stuff, and it's a big enough piece of work I think it would benefit from a tracking issue.

## Goals

The goal of this bit of work is to implement a simple file-based system for storing core user settings, supporting syntax and platform-specific overrides, hot reloading, and maybe a simple client API.

## Non-goals

I'm not sure I want to tackle the problem of having xi-core store frontend specific settings, or plugin settings. My focus right now is on settings relevant to `xi-core`.

### Basic default and user settings (#409)

- [x] include reasonable default settings
- [x] search for user settings file on launch
- [x] override defaults with user settings
- [x] use settings for things like tab/space in xi-core

### Syntax specific settings (#411)

- [x] Include default syntax-specific settings for certain languages (make, TOML, YAML, F#)
- [x] search for user syntax-specific setting files
- [x] use syntax-specific overrides on a per-buffer basis
- [x] update settings when buffer syntax changes (on save e.g.)

### Platform specific settings (#411)

It seems reasonable to think about this now, but realistically this will be a noop until we have settings that need platform-specific overrides?

- [x] use platform-correct newline characters

### Hot reloading (#427)

- [x] monitor user settings directory for changes in settings files
~- [ ] when settings files change, load the changes and apply them intelligently (if, say, `Python.xiconfig` changes, only update buffers editing python files)~

### Validation (#428)
- [x] define valid keys for a given config set
- [x] validate configs before setting
- [x] make this abstract enough that using JSON-schema will be possible at some point?

### Client notifications (#430)
- [x] come up with some API for notifying clients of config changes
- [x] notify clients of config changes
- [x] xi-mac support (https://github.com/google/xi-mac/pull/75)

### Non file-based settings API (#431)
- [x] add RPC protocol support for client managed settings
- [x] add 'get_config' RPC

### simple client API (~#418~ #431)
- [x] clients should be able to set transient per-editor overrides, for instance changing tab settings for a given buffer. 
- [x] migrate this to use the eventual settings API
- [x] move to using `ConfigDomain` for buffer overrides

### documentation (#432)
- [ ] add something to `docs/` or to `frontend.md` on how clients should interact with config.

PR #412, opened 2017-10-16
Feature/syntax platform config

cmyr:
This PR adds default platform and syntax-specific overrides to settings, and allows for a user to specify their own syntax-specific overrides.

Progress on #410

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -565,7 +565,8 @@ impl Editor {
 
     fn insert_newline(&mut self) {
         self.this_edit_type = EditType::InsertChars;
-        self.insert("\n");
+        let text = self.config.newline.clone();
+        self.insert(&text);
     }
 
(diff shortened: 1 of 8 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -565,7 +569,8 @@ impl Editor {
 
     fn insert_newline(&mut self) {
         self.this_edit_type = EditType::InsertChars;
-        self.insert("\n");
+        let text = self.config.newline.clone();
```
raphlinus: Don't think you need the clone here, can just take a reference.

cmyr: @raphlinus we can't take a reference because we mutably borrow `self` in `self.insert`. I'm not sure if there's a simple workaround for this?

raphlinus: Oh, right. Well, the cost of the clone should be trivial in practice. Please proceed.

PR #413, opened 2017-10-16
add a simple makefile to install the syntect plugin

eyelash:


rust/syntect-plugin/makefile
```diff
@@ -0,0 +1,10 @@
+PLUGIN_DIR = ~/.config/xi/plugins
```
cmyr: can we check for `XI_PLUGIN_DIR` and use that if set?

eyelash: you mean the env variable `XI_PLUGIN_DIR` or just renaming the variable to `XI_PLUGIN_DIR`?

cmyr: @eyelash I mean check for the env var. We should also do some thinking about a longer term strategy with make (a set of conventions for verifying the environment, building, and installing, that we can standardize for plugins in general) but I think this is good for now!

eyelash: @cmyr I pushed another commit but for whatever reason I can't see it here in the pull request. It works for me on Linux but I don't know what version of make macOS uses so you need to test it there as well.

rust/syntect-plugin/makefile
```diff
@@ -1,10 +1,12 @@
...
+XI_CONFIG_DIR ?= $(XDG_CONFIG_HOME)/xi
+XI_PLUGIN_DIR ?= $(XI_CONFIG_DIR)/plugins
 
 target/release/xi-syntect-plugin:
 	cargo build --release
 
 install: manifest.toml target/release/xi-syntect-plugin
-	install -D manifest.toml $(PLUGIN_DIR)/syntect/manifest.toml
-	install -D target/release/xi-syntect-plugin $(PLUGIN_DIR)/syntect/bin/xi-syntect-plugin
+	install -D manifest.toml $(XI_PLUGIN_DIR)/syntect/manifest.toml
```
cmyr: good shout on having me check. Apparently `install -D` isn't available on BSD/macOS: https://superuser.com/a/548975

raphlinus:
Great, thanks!

PR #417, opened 2017-10-20
Fix #352

### LINKED ISSUE #352: syntect plugin is missing spans
With syntect enabled, some text regions are missing spans. I’m not sure yet if this is a regression: to be visible it requires using a theme where the normal text color is distinct from the system default text color.

In xi-mac, text with no associated spans is drawn using the default foreground color by default, so this isn’t especially visual. When I disable that, and draw unstyled text in another color, it looks like this:

(to reproduce these screenshots, comment out the line `attributes[NSForegroundColorAttributeName] = textColor` in `TextDrawingMetrics` in `EditView.swift`, line 55 as of this writing.)

screen shot 2017-06-30 at 9 56 08 am
screen shot 2017-06-30 at 9 56 17 am
screen shot 2017-06-30 at 10 01 12 am

cmyr:
While we were correctly calculating the stack state at EOL, we were not
generating a style span for the last state in each line.

### DIFF
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -20,9 +20,10 @@ extern crate xi_plugin_lib;
 
 mod stackmap;
 
+use std::sync::MutexGuard;
 use xi_plugin_lib::state_cache::{self, PluginCtx};
 use xi_plugin_lib::plugin_base::ScopeSpan;
-use syntect::parsing::{ParseState, ScopeStack, SyntaxSet, SCOPE_REPO};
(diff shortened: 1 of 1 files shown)


rust/syntect-plugin/src/stackmap.rs
```diff
@@ -10,7 +10,6 @@
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
-// limitations under the License.
```
raphlinus: Pretty sure this wasn't intentional.

cmyr: 😁 rups

PR #419, opened 2017-10-23
fix a warning

eyelash:


bkchr:
Because you initialize the value and not modify.

cmyr:
@bkchr well we call `table.insert`, which requires table to be `mut`? Doesn't that count?

raphlinus:
Because the type of `table` is `&mut HashMap<...>`.  I think this is valid.

eyelash:
@cmyr I guess `table` is a mutable reference and does not need to be mutable itself. You only want to mutate the thing `table` points to and not the variable `table` itself. The following code compiles:
```rust
let mut x = HashMap::new();
let y = &mut x;
y.insert(1, 2);
```

cmyr:
okay cool, thanks for walking me through this. 🙇

PR #420, opened 2017-10-24
Remove outdated caching_plugin

cmyr:
Unless there is a good reason to keep this around? Afaict it's now redundant.

This will have the advantage of making #406 a much cleaner PR.


cmyr:
It does seem like an auspicious milestone, but I can't quite remember why.

Are those TODOs documented anywhere? Are there any cases where `caching_plugin` would be a better choice than `plugin_cache`?

raphlinus:
The TODOs are inline in the code (see `get_line` in state_cache.rs). The main one I'm thinking of is writing the offset back into the cache. For the syntax highlighting use case, skipping this is not a big deal because the plugin calls `set` to update the state, and that refreshes the cache. A workaround is to call `set` with `()` state, but that seems hacky. Best to just fix it.

I don't think that should block deleting this code, this code never worked properly in the random access case.

cmyr:
@raphlinus Okay great then none of those TODOs are in the to-be deleted file anyway. Will merge whenever the ✅ shows up, so I can clean up #406. (On the train this evening, doing some little tasks)

Issue #415, opened 2017-10-17
Managing and persisting user preferences


cmyr:
# Preferences and persistence
This is a brief sketch of how I've been thinking about the preferences / config system, how the various levels of preferences fit together, and some thoughts on how we might manage client settings.

## Config levels

There are a number of configuration 'tiers', which override one another in a set order:

1. **Defaults**: the base level settings that ship with xi; these are included in `assets/defaults.toml`. These are always loaded first, and provide sane general-case defaults.
2. **Platform specific**: these are platform-specific overrides that ship with xi, in (for instance) `assets/windows.toml`. These override the default settings on a particular platform.
3. **General user preferences** (optional): these are persistent user preferences, which override defaults. When a user modifies preferences, they are modifying these. 
1. **Syntax specific (system)**: these are syntax-specific settings which are bundled with xi. These override even user preferences; if a user's preferences declare a desire for tabs over spaces, the yaml syntax (`assets/yaml.toml`) preferences (which want spaces over tabs) will be preferred.
1. **Syntax specific (user)** (optional): The user can optionally set their own syntax-specific settings, which will override all previous settings, and persist between application launches.
1. **Session overrides (system)**: finally, on a per-buffer basis xi-core may override some settings, for instance if we have a preference for spaces over tabs but open a file that is indented with tabs, we might override the user preference for that given file.
1. **Session overrides (user)**: through the RPC protocol, the client may override a given setting for the current buffer. These overrides are not persistent; they would be used to manually set a line-ending or encoding on a per-document, per-session basis.

Of these, only the third and fifth are user-defined and persistent; they're the focus of this discussion.

## Getting and persisting user preferences
Two of these tiers (general and syntax-specific) require us to persist user preferences. There are a few possible approaches to this:
1. **File based, core managed**: the classic unix text editor approach: all preferences are saved in preference files; files are watched and preferences reloaded on change. This is the easiest approach, and has pedigree, but it can confuse and frustrate newer users, and doesn't really respect platform conventions.

2. **RPC based, client managed** The opposite pole: the client manages all user settings, sending them to core at launch and then keeping core up to date about changes. This would involve RPC calls to the effect of:

    ```javascript
    {
        "method": "set_preferences",
        "params": {
            "user": {
                "font_size" 14
            }
            "python": {
                "translate_tabs_to_spaces": true,
                "tab_size": 2
            }
        }
    }
    ```
This approach gives the client total control over how preferences are managed, but also requires each client to implement a settings system.

[...]

PR #427, opened 2017-11-07
hot reloading configs

cmyr:
This PR adds support for updating a configuration _while the editor is running_.

This required a minor patch in `xi-rpc`, to support idle scheduling outside of the main thread.

It also moves us towards using the `client_init` RPC to do configuration setup; for the time being we will continue to fall back on using environment variables to set a configuration directory, but that will be deprecated.

This is progress towards #410. (I'm going to update that issue in light of #415)

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -14,7 +14,7 @@
 
 use std::env;
 use std::path::{PathBuf, Path};
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 
 use config_rs::{self, Source, Value, FileFormat};
 use syntax::SyntaxDefinition;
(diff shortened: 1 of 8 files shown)


rust/core-lib/src/config.rs
```diff
@@ -199,13 +259,14 @@ impl ConfigManager {
         let mut settings: Config = settings.try_into().unwrap();
         // relative entries in plugin search path should be relative to
         // the config directory.
+        if let Some(ref config_dir) = self.config_dir {
```
raphlinus: I think this needs an indent.

cmyr: 👍. For Fuchsia I suspect we will opt out of file-based config stuff altogether, and just use the (yet to be implemented) RPC protocol methods for managing configs.

PR #429, opened 2017-11-15
config notification

cmyr:
This is based on/blocked by #428, and shouldn't be reviewed until that is in.

This adds a new RPC method, `config_changed`, which notifies the client of the config values for each view. 

Some of these may be of particular interest to the client (such as font settings) and some of these may be useful for updating other state, such as indicating the current indentation setting in a status bar.

It means you can do this (on xi-mac, see https://github.com/google/xi-mac/pull/75)

font_config

- progress towards #410


cmyr:
Moved to #430

PR #430, opened 2017-11-17
config notification

cmyr:
(duplicated from #429)

This adds a new RPC method, `config_changed`, which notifies the client of the config values for each view. 

Some of these may be of particular interest to the client (such as font settings) and some of these may be useful for updating other state, such as indicating the current indentation setting in a status bar.

It means you can do this (on xi-mac, see https://github.com/google/xi-mac/pull/75)

font_config

- progress towards #410

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -21,6 +21,7 @@ use std::fmt;
 use std::rc::Rc;
 use std::path::{PathBuf, Path};
 use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
 use config_rs::{self, Source, Value, FileFormat};
 
@@ -134,7 +135,7 @@ pub struct ConfigPair {
     /// precedence over items in `base`.
     user: Option<Table>,
     /// A snapshot of base + user.
-    cache: Table,
+    cache: Arc<Table>,
     validator: Rc<Validator>,
 }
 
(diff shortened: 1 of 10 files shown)


rust/core-lib/src/config.rs
```diff
@@ -156,15 +161,34 @@ pub struct ConfigManager {
     extras_dir: Option<PathBuf>,
 }
 
+/// A collection of config tables representing a heirarchy, with each
```
raphlinus: typo nit: hierarchy

rust/core-lib/src/config.rs
```diff
@@ -156,15 +161,34 @@ pub struct ConfigManager {
...
+/// table's keys superceding keys in preceding tables.
```
raphlinus: typo nit: superseding

rust/core-lib/src/config.rs
```diff
@@ -413,11 +528,19 @@ impl Error for ConfigError {
         match *self {
             IllegalKey( .. ) => "illegal key",
             UnknownDomain( .. ) => "unknown domain",
-            FileParse( .. ) => "failed to parse file",
+            Parse( _, ref e ) => e.description(),
+            Io( ref e) => e.description(),
```
raphlinus: tiny nit: spacing asymmetrical

rust/core-lib/src/tabs.rs
```diff
@@ -47,6 +47,9 @@ use plugins::rpc_types::{PluginUpdate, ClientPluginInfo};
 #[cfg(target_os = "fuchsia")]
 use apps_ledger_services_public::{Ledger_Proxy};
 
+/// A client can use this to pass a path to bundled plugins
```
raphlinus: Should probably say "this environment variable" as it's not otherwise obvious what "this" is.

cmyr: this line is removed in #433.

PR #432, opened 2017-11-17
Document the new config system

cmyr:
([Rendered](https://github.com/cmyr/xi-editor/blob/f8b69be9d1c4bc81a41dbcd79364d87d6c83361d/doc/config.md))

This should not be merged before the remaining code PRs, #430 and #431. This represents the end of the config work as outlined in #410. 

Feedback on this is appreciated, I think it's all pretty straightforward but I've also been thinking about it for the past month.

This also includes a few formatting tweaks in `frontend.md`.


eyelash:
Does the file based approach also handle front-end specific settings? You mention `font_face` and `font_size` which look like front-end specific settings (they don't make sense for a TUI front-end for example). If yes then how does validation work in that case? How does the core know which front-end specific settings are valid?

cmyr:
@eyelash good Q. My thinking is that as a convenience we will offer a basic set of front-end settings, and a client that isn't interested in these can just ignore them. 

If in the future we want to support custom client settings, we  could add a new 'client' domain, and the client can provide a template which we'd use for validation?

cmyr:
I've updated this minorly to remove mention of xdg and to mention that xi-core handles watching/reloading.

PR #433, opened 2017-11-18
Feature/config cleanup

cmyr:
Final cleanup of env-var related config stuff.

The client must now specify a config directory and a plugin search path in the `client_started` RPC.

This also includes one bug fix around default plugin search paths.

### DIFF
--- a/rust/core-lib/assets/client_example.toml
+++ b/rust/core-lib/assets/client_example.toml
@@ -1,0 +1,20 @@
+#####################################################################
+# User settings must be in $XI_CONFIG/preferences.xiconfig, where   #
+# '$XI_CONFIG' is the Xi configuration directory for your platform. #
+#                                                                   #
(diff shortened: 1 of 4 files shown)


rust/core-lib/assets/client_example.toml
```diff
@@ -0,0 +1,20 @@
...
+# The width of a tab, in spaces.
+tab_size = 4
+# Insert spaces when the tab key is pressed.
+translate_tabs_to_spaces = true
+
+# List of paths to additional plugins
+plugin_search_path = []
+
+font_face = "InconsolataGo"
+# In points
```
raphlinus: This is interesting, my measurements show the units being px, but the [NSFont docs](https://developer.apple.com/documentation/appkit/nsfont/1525977-init) clearly say points. Probably ok to leave it as it is, unlikely to cause real problems.

PR #436, opened 2017-11-22
Fixed deprecations with BLACK

### LINKED ISSUE #435: Fix syntect warnings
Currently building core prints some deprecation warnings from syntect:

```
src/styles.rs:93:54
   |
93 |         let fg = theme.settings.foreground.unwrap_or(BLACK);
   |                                                      ^^^^^

warning: use of deprecated item: use Color::BLACK instead
  --> src/styles.rs:21:81
   |
21 | use syntect::highlighting::{Color, Theme, ThemeSet, ThemeSettings, Highlighter, BLACK};
```

this should be an easy update

bemeurer:
Fixes #435

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1,19 +1,15 @@
-[root]
-name = "xi-core"
-version = "0.2.0"
-dependencies = [
- "serde 1.0.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde_json 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "xi-core-lib 0.2.0",
- "xi-rpc 0.2.0",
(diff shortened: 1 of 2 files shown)


raphlinus:
I think the CI failure is due to a Cargo.lock which still has syntect at 1.7.0. I'm not sure why we're seeing both the deprecation warning and the failure; perhaps the checked-in Cargo.lock is applying to the xi-core binary but not a separate check build of core-lib.

In any case, the fix for this out to be running "cargo update" from rust/ and checking in the resulting Cargo.lock.

bemeurer:
@raphlinus I think I got it all working now.

raphlinus:
Great, many thanks for fixing this!

PR #441, opened 2017-11-25
Start implementation of minimal invalidation

raphlinus:
Start implementation of minimal invalidation

This patch uses a fancy data structure to track the state of the
front-end's line cache on a fine grained basis, and construct a
minimal delta to be sent on update.

As of this patch, the data structure is done, but most changes to
the view just invalidate the whole thing. Thus, the visible effect
is minimal. That said, the core is in control of eviction of the
front-end's cache, so the unbounded growth is not possible, and it
also won't keep resending the same content, which should help with
scrolling performance.

Fixes #270 and is good progress towards #317.

PR #443, opened 2017-11-27
Add cancel operation

### LINKED ISSUE #324: implement `cancel_operation` method
Sent on `esc`. Exact operation is debatable, but should probably collapse multiple cursors, and probably collapse selections.

WesleyAC:
This adds a cancel_operation method that collapses all cursors/selections and dehighlights searches. Let me know if I missed anything - it's my first time poking around in Xi :)

Closes #324

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -901,6 +901,11 @@ impl Editor {
         }
     }
 
+    fn cancel_operation(&mut self) {
+        self.view.unset_find();
+        self.view.collapse_selections();
+    }
+
     fn cmd_prelude(&mut self) {
         self.this_edit_type = EditType::Other;
     }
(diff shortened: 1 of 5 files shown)


WesleyAC:
@googlebot I signed the CLA :)

rust/core-lib/src/view.rs
```diff
@@ -233,6 +233,11 @@ impl View {
         &self.selection
     }
 
+    /// Collapse all selections in this view into a single caret
```
raphlinus: Pretty sure you'll need to set sel_dirty here one way or another. Setting it directly would work on current master, but that won't merge cleanly with #441.

What I'd recommend is cloning the selection, collapsing your clone, and calling set_selection. This will let minimal invalidation compute a good diff, for which it needs both the "before" and "after" values of the selection.

raphlinus:
Oh, also it would be great to add documentation for this to `doc/frontend.md`.

rust/core-lib/src/editor.rs
```diff
@@ -901,6 +901,11 @@ impl Editor {
         }
     }
 
+    fn cancel_operation(&mut self) {
```
cmyr: nit: the convention we've been using is to preface methods that are directly handling an RPC with `do_`. 

Otherwise this looks great! I've added xi-mac support here: https://github.com/google/xi-mac/pull/84

raphlinus:
Thanks!

PR #445, opened 2017-12-05
Finer grained invalidation

### LINKED ISSUE #317: Minimal invalidation
One of the most important performance improvements on our roadmap is "minimal invalidation". There are several ways for invalidation to be minimal, all of which fall under the same code change:

1. When the content changes, the diff sent to the front-end should be the smallest possible.

2. Re-rendering the content that's changed should invalidate as little as possible of the remainder of the line cache.

Per discussion in #270 the core should be responsible for maintaining the state of the line cache. Basically, the core _must_ keep the visible scroll region (plus a small slop, currently 2 lines) up to date. Any change within this region is re-rendered. The "update" method allows changes to some subset of text, cursor, and style spans; if only some of these change (common, for example, with cursor movement), then the re-rendering should only update the elements that changed.

Changes outside this region generally transition to invalid (a probable exception to this is when the change is cursor only, in which case it's just as cheap to update the cursor as it is to send it to invalid).

There is an additional window (say ~1000 lines on either side) where contents are preserved if they do not change. Scrolling will populate this window (ie, lines will transition from invalid to valid). Deltas should be constructed to not disturb this additional window. For example, inserting a line should be an "ins", rather than replacing the line and invalidating everything after it.

A consequence of the above is that, for example, re-highlighting the entire file will, the first time, re-render the visible region and transition the rest to an invalid state. Once the rest is in an invalid state, changes such as rehighlighting outside the visible region will be a no-op to the line cache, so no updates will be sent to the front-end at all. This should help a lot with google/xi-mac#42.

Part of the #270 work is to drop lines when they are more than the ~1000 line limit past the visible scroll region.

A primary motivation behind this logic is to keep individual updates bounded (ie no more than the visible region plus small slop). The idea is that the "additional region" is built up over multiple updates (ie scrolling) and has no protocol cost; the only cost is the RAM to store the additional line cache entries, and costs of maintaining the data structure (on the order of 1000 is small enough that a dense array is a good choice). A reasonable heuristic is that _clearing_ the cursor on a line is cheap enough it should be done in favor of invalidating the line; worst case 1000 cursor clear operations is within the acceptable range for a single update. A consequence is that most cursor movement will avoid invalidation (all single-cursor movement, and multi-cursor movement within the visible region).

Also keep in mind #184, which is not implemented yet, but when it is, the deltas will need to keep the line numbers in sync as well.

This work is not trivial, but has the nice property of being fairly localized (dirty state tracking and re-rendering), and having hopefully no visible changes other than improving performance.

raphlinus:
Mostly working for the base case: finer-grained invalidation on cursor
motion, selection changes, style changes (syntax highlighting), and
edits.

What's still coarse-grained is rewrapping (which is experimental at
this point). Also, cursors with upstream affinity won't always be
included in an invalidation (this is highly related to line wrapping).
These items could be considered part of the ongoing line wrapping
work.

Fixes #317.

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -123,7 +123,7 @@ impl View {
 
     pub fn scroll_to_cursor(&mut self, text: &Rope) {
         let end = self.sel_regions().last().unwrap().end;
-        let (line, _) = self.offset_to_line_col(text, end);
+        let line = self.line_of_offset(text, end);
         if line < self.first_line {
             self.first_line = line;
         } else if self.first_line + self.height <= line {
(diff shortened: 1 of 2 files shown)


rust/core-lib/src/view.rs
```diff
@@ -183,12 +183,49 @@ impl View {
         Some(region.end)
     }
 
+    /// Sets the selection to a new value, without invalidating. Return values is
```
cmyr: "Return values" -> "Return value"

rust/core-lib/src/view.rs
```diff
@@ -183,12 +183,49 @@ impl View {
...
+        // selection state, and for scrolling it into view if needed. This choice can
+        // likely be improved.
+        let region = self.selection.last().unwrap().clone();
+        self.scroll_to_cursor(text);
+        Some(region.end)
+    }
+
     /// Sets the selection to a new value, invalidating the line cache as needed. This
     /// function does not perform any scrolling.
     fn set_selection_raw(&mut self, text: &Rope, sel: Selection) {
```
cmyr: Missed this in the last PR, but I find this name confusing. 😕

Issue #449, opened 2017-12-09
(Docs) What is a base_unit?


Eh2406:
[Metric](https://docs.rs/xi-rope/0.2.0/xi_rope/tree/trait.Metric.html) has a method to_base_units, but no info on what is a base_unit. What is it? How would a Metric use these functions?

Eh2406:
looks like there is more info in the src that is not being displayed by rust doc. Perhaps it is `\\` instead of using `\\\`?

Eh2406:
Conversation happening in irc https://botbot.me/mozilla/xi/2017-12-09/?msg=94473883

cmyr:
resolved in #450

PR #450, opened 2017-12-10
WIP: Clarify docs

Eh2406:
In discussing #449 on irc it became clear that not all comments made it into the docs. Some comments were missing the third `/`. So I went thru and added them. While I was there I rewrote some where I thought I understood, and added `TODO`s for things I had questions about. Fundamentally, I don't yet understand, so all of this may be wrong. But I find it easier to explain how someone is wrong then to explain from scratch. So I thought I would write something up so you can tell me how I am wrong.

### DIFF
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -46,7 +46,7 @@ pub trait NodeInfo: Clone {
     /// The identity of the monoid. Need not be implemented because it
     /// can be computed from the leaf default.
     /// 
-    /// TODO: Why have it if it never needs to be overridden?
+    /// This is hear to demonstrate that this is a monoid.
     fn identity() -> Self {
         Self::compute_info(&Self::L::default())
     }
@@ -61,6 +61,9 @@ pub trait NodeInfo: Clone {
 pub trait Leaf: Sized + Clone + Default {
 
     /// Measurement of leaf in base units.
+    /// A 'base unit' refers to the smallest discrete unit
+    /// by which a given concrete type can be indexed.
+    /// Concretely, for Rust's String type the base unit is the byte.
     fn len(&self) -> usize;
 
     /// Generally a minimum size requirement for leaves.
@@ -131,15 +134,22 @@ enum NodeVal<N: NodeInfo> {
 pub trait Metric<N: NodeInfo> {
     /// Return the number of boundarys in the NodeInfo::Leaf
     /// 
-    /// TODO: What does the second argument do?
+    /// The usize argument is the total size/length of the node, in base units.
     fn measure(&N, usize) -> usize;
 
-    /// Return a offset in base units forwich self.from_base_units will return the argument.
-    /// 
-    /// TODO: Does it have to be the largest, smallest, or just eny offset?
+    /// Returns the smallest offset, in base units, for an offset in measured units.
+    ///
+    /// Invariants:
+    ///
+    /// - `from_base_units(to_base_units(x)) == x` is True for valid `x`
+    /// - `is_boundary(to_base_units(x))` is True for all `x`
     fn to_base_units(l: &N::L, in_measured_units: usize) -> usize;
 
-    /// Return the number of boundarys in the NodeInfo::Leaf befor the offset in_base_units.
+    /// Returns the smallest offset in measured units corresponding to an offset in base units.
+    ///
+    /// Invariants:
+    ///
+    /// - `from_base_units(to_base_units(x)) == x` is True for valid `x`
     fn from_base_units(l: &N::L, in_base_units: usize) -> usize;
 
     /// Return whether the offset in base units is a boundary of this metric.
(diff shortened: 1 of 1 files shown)


Eh2406:
When you have some time, This is wating on you to tell me what is incorrect, and to explain my todo's.

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
 // help separate metrics
+
+/// A trait for quickly processing attributes of a NodeInfo.
+/// 
+/// For the conceptual background see the 
+/// [blog post, Rope science, part 2: metrics](https://github.com/google/xi-editor/blob/master/doc/rope_science/rope_science_02.md).
 pub trait Metric<N: NodeInfo> {
+    /// Return the number of boundarys in the NodeInfo::Leaf
+    /// 
+    /// TODO: What does the second argument do?
```
cmyr: To the best of my understanding, the second argument is the total size/length of the node, in base units.

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
     fn measure(&N, usize) -> usize;
 
+    /// Return a offset in base units forwich self.from_base_units will return the argument.
```
cmyr: Maybe just: "Returns the offset, in base units,  for an offset in measured units."

I believe there can only a single possible result for a given input. (At least, I cannot think of an example where this is not the case.)

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
+    /// 
+    /// TODO: Does it have to be the largest, smallest, or just eny offset?
     fn to_base_units(l: &N::L, in_measured_units: usize) -> usize;
 
+    /// Return the number of boundarys in the NodeInfo::Leaf befor the offset in_base_units.
```
cmyr: "Returns the offset in measured units corresponding to an offset in base units."

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
+    /// If a boundary is at end of a leaf then this method must return true.
+    /// However, A boundary at the beginning of a leaf is optional
+    /// (the previous leaf will be queried).
     fn is_boundary(l: &N::L, offset: usize) -> bool;
 
-    // will be called with offset > 0
+    /// Return the largest usize smaller then offset forwich is_boundary will return true.
+    /// will be called with offset > 0.
+    /// 
+    /// TODO: If offset is a boundary can this return offset, or does it have to return a strictly smaller number?
```
cmyr: "Returns the index of the boundary directly preceding `offset`, or `None` if no such boundary exists. Input and result are in base units."

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
     fn prev(l: &N::L, offset: usize) -> Option<usize>;
 
+    /// Return the smallest usize larger then offset forwich is_boundary will return true.
+    /// will be called with offset > 0.
+    /// 
+    /// TODO: If offset is a boundary can this return offset, or does it have to return a strictly smaller number?
```
cmyr: "Returns the index of the first boundary for which `index > offset`, or `None` if no such boundary exists. Input and result are in base units."

rust/rope/src/tree.rs
```diff
@@ -116,29 +123,48 @@ enum NodeVal<N: NodeInfo> {
...
     fn next(l: &N::L, offset: usize) -> Option<usize>;
 
-    // When can_fragment is false, the ends of leaves are always
-    // considered to be boundaries. More formally:
-    // !can_fragment -> to_base_units(measure) = leaf.len
+    /// When can_fragment is false, the ends of leaves are always
+    /// considered to be boundaries. More formally:
+    /// !can_fragment -> to_base_units(measure) = leaf.len
+    /// 
+    /// TODO: What dose this mean?
```
cmyr: "Returns `true` if the measured units in this metric can span multiple leaves.  As an example, in a metric that measures lines in a rope, a line may start in one leaf and end in another; however in a metric measuring bytes, storage of a single byte cannot extend across leaves."

Eh2406: Ok, but I don't understand:
 - The code fragment in the doc.
 - Why and by whom this would be called.

rust/rope/src/tree.rs
Eh2406: Why have two sources of truth, it is `N.len()` right?

cmyr: Good question. I'm also just reading through the code and trying to figure it out; might it be that only leaves have a `len`, but this is implemented on all nodes?

rust/rope/src/tree.rs
cmyr: I don't know much better than you; I'm mostly just grepping through the code and taking my best guess, so apologies if my explanations aren't the best.

The little shorthand statement in the doc is saying, "'can_fragment() == false' _implies_  the number of base units in this leaf is equal to `leaf.len()`". (Raph is using logical notation; I assume `->` is the [material conditional](https://en.wikipedia.org/wiki/Material_conditional) operator.)

I'm not sure if there's a case where this would be used by a general library consumer; it is [used internally](https://github.com/google/xi-editor/blob/master/rust/rope/src/tree.rs#L521) when doing boundary calculations.

raphlinus:
Just wanted to say I hope to get a chance to review this carefully soon but am currently juggling too many things. Some questions will require me to research the code, but that's a good thing, these things should be written down.

I think the extra len argument is likely redundant, this might have been intended as an optimization, but if it can be safely removed without actually slowing things down, it probably should be.

rust/rope/src/tree.rs
```diff
@@ -45,37 +45,44 @@ pub trait NodeInfo: Clone {
 
     /// The identity of the monoid. Need not be implemented because it
     /// can be computed from the leaf default.
+    /// 
+    /// TODO: Why have it if it never needs to be overridden?
```
raphlinus: It might be worth keeping this for pedagogical reasons. The identity element is one of the fundamental properties of a monoid. It _needn't_ be overridden, but in many monoids it has a simple value (0 for anything that counts). I think it will also be important for property testing.

rust/rope/src/tree.rs
raphlinus: I believe this is actually the largest, but I need to think about it more to make sure. I'd also prefer that this and `from_base_units` not be defined in terms of each other, but I do think it makes a ton of sense to explicitly write down invariants they need satisfy - `from_base_units(to_base_units(x)) = x` is one for valid `x`.

rust/rope/src/tree.rs
raphlinus: I think @cmyr has this right.

rust/rope/src/tree.rs
raphlinus: Same.

rust/rope/src/tree.rs
raphlinus: This is used in `Tree::convert_metrics` as an optimization. The motivating example is definitely counting line ends, which can fragment. Looking at the code, I'm actually not 100% sure how important the optimization is. I _think_ this might be equivalent to saying that it's possible for a leaf to have zero measure.

rust/rope/src/tree.rs
raphlinus: On rereading the code, I think @Eh2406 is correct, the additional argument should be removed. I thought it might be more efficient to pass an argument rather than have the method reach into the `NodeInfo` struct, but on reflection there's no good reason for this.

rust/rope/src/tree.rs
raphlinus: Thinking about it more, I think `can_fragment` is "not atomic" in the terminology of #456. I think it's useful to preserve, but maybe "atomic" (and inverting the boolean) is clearer. I agree, the way I wrote the invariant is confusing.

Eh2406:
I think I have incorporated all the suggestions. I still have a `TODO: What does Interval represent?`

PR #453, opened 2017-12-15
Improve style resolution performance

cmyr:
This patch significantly improves the speed at which we resolve scope spans into styles and greatly improves startup performance, especially noticeable on debug builds.

### before:
screen shot 2017-12-14 at 8 17 09 pm

### after:
screen shot 2017-12-14 at 8 18 00 pm

### DIFF
--- a/rust/core-lib/src/layers.rs
+++ b/rust/core-lib/src/layers.rs
@@ -20,6 +20,7 @@
 
 use std::collections::BTreeMap;
 use syntect::parsing::Scope;
+use syntect::highlighting::StyleModifier;
 
 use xi_rope::interval::Interval;
 use xi_rope::spans::{Spans, SpansBuilder};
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/layers.rs
```diff
@@ -40,8 +41,12 @@ pub struct Scopes {
 pub struct ScopeLayer {
     stack_lookup: Vec<Vec<Scope>>,
     style_lookup: Vec<Style>,
+    // TODO: this might be efficient (in memory at least) if we use
+    // a prefix tree.
+    /// style state of existing scope spans, so we can more efficiently
+    /// compute styles of child spans.
+    style_cache: HashMap<Vec<Scope>, StyleModifier>,
```
trishume: If hashing shows up any significant amount in the time profile, you can switch this to use a FNV hasher, which is what I use in syntect.

cmyr: @trishume okay, good note, thanks!

Issue #323, opened 2017-05-29
multi selection polish


cmyr:
pulling out some todos from #188.

- [x] Copy-paste of multiple lines into an equal number of cursors should paste one line per cursor. Current behavior is to paste all lines (separated by newline) into all cursors

- [x] Transpose (^T) should rotate the contents of selection (non-caret) regions. Currently they're ignored.

- [ ] Selection region merging, when one region is forward and the other backward, always resolves the direction of the merged region to forward. I think better behavior would be to match the other parameter (ie the region added later).

- [x] Update of style spans on edit is based on edit summary rather than fine grained edit, so everything from the beginning of the first region to the end of the last loses style. This is recoverable because (according to the current plugin protocol) the plugin will re-highlight all of that, but it's not ideal.

sp3d:
The first of these is really interesting; naïvely I'd expect multi-cursor pasting to broadcast a single copied string to all cursors rather than splitting on newlines, and would only expect different pastes per cursor if the text had been copied from a set of multi-cursor selections. In either case, what happens when the number of objects to paste isn't the same as the number of cursors?

Are both kinds of functionality (broadcast and pointwise paste) accessible in other editors?

trishume:
@sp3d AFAIK in all common editors with multi-select (Sublime, Atom, VSCode) the way it works is that it broadcasts the copy buffer to all cursors unless the number of cursors is equal to the number of lines, in which case it pastes each line's contents at the corresponding cursor.

The reason is so that when you copy using multiple cursors, it gets put into the clipboard as multiple lines, and then if you move those cursors somewhere else and paste it properly acts like each cursor has its own paste buffer, but without having to make the clipboard actually be a fancy data structure.

I think this is a feature rather than just laziness about making the paste buffer fancy. First of all, I actually use this, I've copied a bunch of lines using a single cursor and then column-selected to paste each line in a different place. It's also not nice to have hidden state in the paste buffer that you can't observe just by pasting, two buffers that paste to the same thing in the single-cursor case should act the same in a multi-cursor case.

sp3d:
Thanks for the info about other editors.

I agree it's handy to be able to paste in a way that splits an N-line selection (or M*N? how do our peers handle copy/paste of multiple multi-line selections?) across N cursors. But I would advocate storing an additional clipboard format that doesn't discard knowledge of how many selections text came from when we copy, and having a way to access paste functionality that explicitly broadcasts or splits.

trishume:
[...]

PR #457, opened 2017-12-21
Add transpose for multi-select

vlovich:
Resolves one of the polish items in #323

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -63,6 +63,16 @@ enum CharacterEncoding {
 
 const UTF8_BOM: &str = "\u{feff}";
 
+fn last_selection_region(regions: &[SelRegion]) -> Option<&SelRegion> {
+    for region in regions.iter().rev() {
+        if !region.is_caret() {
+            return Some(region);
+        }
+    }
+
+    None
+}
+
 pub struct Editor {
     text: Rope,
     encoding: CharacterEncoding,
@@ -877,9 +887,20 @@ impl Editor {
         }
     }
 
+    fn sel_region_to_interval_and_rope(&self, region: &SelRegion) -> (Interval, Rope) {
+        let as_interval = Interval::new_closed_open(region.min(), region.max());
+        let interval_rope = Rope::from(self.text.slice_to_string(
+            as_interval.start(), as_interval.end()));
+        (as_interval, interval_rope)
+    }
+
     fn do_transpose(&mut self) {
         let mut builder = delta::Builder::new(self.text.len());
         let mut last = 0;
+        let mut optional_previous_selection : Option<(Interval, Rope)> =
+            last_selection_region(self.view.sel_regions()).map(
+                |ref region| self.sel_region_to_interval_and_rope(region));
+
         for region in self.view.sel_regions() {
             if region.is_caret() {
                 let middle = region.end;
@@ -895,8 +916,11 @@ impl Editor {
                         last = end;
                     }
                 }
+            } else if let Some(previous_selection) = optional_previous_selection {
+                let current_interval = self.sel_region_to_interval_and_rope(&region);
+                builder.replace(current_interval.0, previous_selection.1);
+                optional_previous_selection = Some(current_interval);
             }
-            // TODO: handle else case by rotating non-caret regions.
         }
         if !builder.is_empty() {
             self.this_edit_type = EditType::Other;
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -114,6 +114,7 @@ enum EditType {
     Delete,
     Undo,
     Redo,
+    Transpose,
 }
 
 impl EditType {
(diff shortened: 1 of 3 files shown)


raphlinus:
Deepest apologies for taking so long to review. I had meant to earlier, but was very focused on the textplane stuff and found it difficult to focus mentally.

I like the fact that this captures the desired semantics (including preserving selections) without being a huge code change. However, the change to `transform_selections` makes me deeply uncomfortable; I don't feel confident I could write down exactly what it's supposed to do after this change. (I absolutely _should_ write down what it's supposed to do now, but that's another question).

My preference would be to have the `keep_selections` logic as it is now, but to have clean separation between the two cases; if `true`, then the edit logic is entirely responsible for computing the new selection (which mostly means moving the carets to the "after" position), if `false` then it's as now, the update is done by OT.

There's an additional invariant which I think should probably be written down explicitly; an edit that's done with `keep_selections` on _must_ appear in its own commit without other edits before or after. In other cases, it's possible to aggregate multiple edits together in a commit, though this isn't exploited in the current code (`cmd_postlude` is called after the handling of each notification).

Thanks again for taking this on!

vlovich:
Thanks for the feedback.  Sorry for the delay in responding as I tried to, apparently successfully :), shutoff my digital life during the break.

Just to double-check that I understood your feedback, are you suggesting that the selection logic fixup for transpose move to cmd_postlude rather than in apply_delta, right?  I think I tried pretty hard to keep it contained within editor.rs & this was the only way I could figure out to make it work.  That being said, it was about 2 weeks ago so I can't recall the exact issues I ran into.  Want to make sure I understand your feedback properly first.

For the additional invariant, where should it be written down?  Is it a documentation thing or something to be programmatically enforced?

Thanks!

raphlinus:
Congrats on managing to disconnect :) I didn't, but I had fun on the fast text renderer.

Good clarifying question. What I have in mind is that do_transpose actually computes the new selection and sets it, and in the transpose case the updating of selections is disabled in `View::after_edit`. The best way to do this might be to have a `new_selection` in the editor state which is an `Option<Selection>`, and have after_edit use that on Some, and call apply_delta on None (as opposed to checking the edit type). It's possible that the edit code actually sets the selection (set_selection_for_edit) and sets a flag to disable it, but for some reason I like the idea of the updated selection being set in one place (the end of after_edit).

One future change that interacts with this is that we're going to have multiple views into a single buffer. In that case, we'll want the newly computed selection to affect the primary view (ie the one that the edit was done from), and the normal selection delta in the others.

Regarding the invariant I suggest documenting it. I think it's not impossible to check programmatically, but it's extra code that's probably not worth it.

Lemme know if you have more questions.

vlovich:
OK.  I think I thought of recomputing the selection in the editor as you suggest but for some reason I thought it would be more complicated (probably the way you're suggesting is simpler).  I'll try it out.

FWIW, I think the multiple views case is exactly why I wanted to keep the selection logic within the view.  I think the way it's written now would allow the multiple-view change to be a minimal change since the view logic would be unchanged.  You just need to (maybe) break the self.view logic in the editor but everything else should remain untouched.  Having the selection set recomputation live within the editor might make the multi-view feature more challenging, no?

raphlinus:
I'm thinking about this more, and see the value of doing it both ways. I'm worried about having special-case logic in the transform just for transpositions, but now that I think about it, it might be an instance of a general problem, which is priorities. I've already been planning to have "before cursor" and "after cursor" priorities for the auto-indent stuff (planned soon, it's in the current roadmap), but it's possible that "after the selection start but before the selection end, if the selection is not a caret" is a valid priority and would be useful.

If so, as you say, extending to multiple views is easy, and it also has the nice property that if all the views have the same selection, they'll end up in the same state.

I need to convince myself that the change you've made to the transformation matches what I'd expect from the priority interpretation. If so, I'm fairly inclined to take this change, and then implement the more general priorities on top of it.

raphlinus:
Ok, I've wrapped my head around this PR and think I understand what's going on, and have a plan. Apologies it's taken so long.

First, I am very uneasy with the changes to Transformer::transform. I haven't run the code, but based on examination there are cases it will give the wrong answer. For example, if the buffer is "ab", the input is 2, and the delta is "insert 'x' at 1" (so result is "axb"), the result will still be 2. With non-ASCII buffers, there's a risk of not respecting code point boundaries, which can easily lead to panics.

That said, I _think_ you can get the effect you're looking for just by passing `false` in as the `after` parameter (so basically `after && !preserve_selection`, keeping the existing logic for the latter). That matches what I was thinking of in the comment above about priorities. I think I'd change the argument to Selection::apply_delta from `after: bool` to a priority enum, as it's more self-descriptive than binaries and is on the path to a more robust priority approach.

If that works (ie reverting the changes to Transformer::transform), then I'd be happy to merge. If it doesn't (ie if it really cares about omitting specifically the last insert) then I'd like to understand what's going wrong. If there will be changes to `Transformer` then I'd like to see tests.

As I say, this is based on reading the code rather than playing with it.

vlovich:
I think the concerns have been addressed in the way you suggest & I agree it's cleaner.  PTAL.

PR #459, opened 2017-12-29
very, very, really quite basic auto indent

cmyr:
This is a first stab at bolting simple auto-indent onto syntect.

This depends on #406, which should go in first.

This is very naive. If a newline is entered, it matches the indentation of the previous line; if the previous line ended in one of '{', '[', ':', or '(', it increases the indentation level by whatever the current indentation preferences indicate.

This is **not** our long-term auto indentation solution, which involves fuller-featured language definitions. However this _is_ hopefully an improvement over the current behaviour!

To dos:

- [ ] be at least a little language aware
- [x] add a config flag to disable this feature
- [x] make it so that all indentation can be removed with a single backspace, where appropriate. I think this should live in core, but it _could_ be done in the plugin.

### DIFF
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -275,7 +275,12 @@ impl<'a> state_cache::Plugin for PluginState<'a> {
     fn update(&mut self, mut ctx: PluginCtx<State>, start: usize, end: usize,
               new_len: usize, rev: usize, text: &Option<String>) -> Option<Value> {
         ctx.schedule_idle(0);
-        self.do_indentation(&mut ctx, start, end, new_len, rev, text)
+        let should_auto_indent = ctx.get_config().auto_indent;
+        if should_auto_indent {
+            self.do_indentation(&mut ctx, start, end, new_len, rev, text)
+        } else {
+            None
+        }
     }
 
     fn did_save(&mut self, ctx: PluginCtx<State>) {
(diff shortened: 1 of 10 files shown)


rust/syntect-plugin/src/main.rs
```diff
@@ -184,6 +193,76 @@ impl<'a> PluginState<'a> {
...
+        }
+    }
+
+    /// Checks if the indent level should be increased. If so, returns the text
+    /// to insert.
+    fn indent_increase(&self, prev_line: &str, use_spaces: bool, tab_size: usize)
+                       -> Option<&str> {
+        let trailing_char = prev_line.trim_right().chars()
+            .rev().next().unwrap_or(' ');
+        // very naive heuristic for inreasing indentation level.
```
cmyr: "increasing"

rust/syntect-plugin/src/main.rs
```diff
@@ -184,6 +193,76 @@ impl<'a> PluginState<'a> {
...
+        if end != start { return None }
+        let is_newline = {
+            let line_ending = &ctx.get_config().line_ending;
+                text.as_ref()
+                .map(|t| t.ends_with(line_ending))
+                .unwrap_or(false)
+        };
+
+        if is_newline {
+            let line_num = ctx.find_offset(start).err();
```
raphlinus: I think this fails to indent if the edit is to insert, say, "{\n" at the beginning of a line. The value should come back in the ok part of the result. This is unlikely to happen (the actual case likely to trigger it is a paste, for which we need to plumb through disabling this).

cmyr: Good spot. I was mentally thinking of a "simple edit" as being one where start == end, but what we really care about is that this action only fires on the "insert_newline" RPC; so we can just do a string comparison on the text, and make this simpler.

cmyr: (I would leave the requirement that start == end; I'm not sure I want to touch indentation in the case where a selection is replaced with a newline.)

PR #460, opened 2017-12-30
Backspace deletes tab_size worth of spaces

cmyr:
With this patch backspace will delete `tab_size` contiguous spaces, if possible.

@raphlinus: I'm earnestly curious to know how you would've implemented this; my implementation feels okay but I have a nagging feeling there's a cleaner one out there?

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -476,16 +476,20 @@ impl Editor {
                 region.min()
             } else {
                 // backspace deletes max(1, tab_size) contiguous spaces
-                let tab_size = self.config.items.tab_size;
+                let (_, c) = self.view.offset_to_line_col(&self.text,
+                                                          region.start);
                 let use_spaces = self.config.items.translate_tabs_to_spaces;
-                let preceded_by_spaces = (0..tab_size)
-                    .map(|i| (region.max() - 1) - i)
-                    .all(|i| self.text.byte_at(i) == b' ');
+                let tab_size = self.config.items.tab_size;
+                let tab_size = if c % tab_size == 0 { tab_size } else { c % tab_size };
+                let preceded_by_spaces = (1..tab_size + 1)
+                    .map(|i| region.start.saturating_sub(i))
+                    .all(|i| self.text.len() > 0 && self.text.byte_at(i) == b' ');
                if preceded_by_spaces && use_spaces {
-                   region.max() - tab_size
+                   region.start - tab_size
                } else {
                    // TODO: implement complex emoji logic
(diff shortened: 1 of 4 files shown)


eyelash:
If you have `tab_size` set to 4 and you have a document that contains just a single line with 6 spaces and a cursor at the end of the document, will it delete 4 spaces if you press backspace? I would expect it to only delete 2.

cmyr:
@eyelash This patch will delete four spaces in the case you mention. I agree the other behaviour would be better; I'll play around a bit.

cmyr:
Okay, updated to match the behaviour @eyelash described.

rust/core-lib/src/editor.rs
```diff
@@ -475,9 +475,24 @@ impl Editor {
...
                 region.min()
             } else {
-                // TODO: implement complex emoji logic
-                self.text.prev_codepoint_offset(region.end).unwrap_or(region.end)
+                // backspace deletes max(1, tab_size) contiguous spaces
+                let (_, c) = self.view.offset_to_line_col(&self.text,
+                                                          region.start);
+                let use_spaces = self.config.items.translate_tabs_to_spaces;
+                let tab_size = self.config.items.tab_size;
+                let tab_size = if c % tab_size == 0 { tab_size } else { c % tab_size };
```
raphlinus: I think this is reasonable, but another way to write it is `(c + tab_size - 1) % tab_size + 1`. Your way is probably clearer though. I actually looked on godbolt and the code is _slightly_ better for my variant, but I think we're well into "this doesn't really matter" territory.

rust/core-lib/src/editor.rs
```diff
@@ -475,9 +475,24 @@ impl Editor {
...
+                let preceded_by_spaces = (1..tab_size + 1)
```
raphlinus: I might be tempted to write this `(region.start.saturating_sub(tab_size) .. region.start).all(|i| self.text.byte_at(i) == b' ')`

cmyr:
@raphlinus updated this with your second suggestion, and also added a config setting for this behaviour.

jansol:
Another thing to consider is how non-code files should be handled (proportional fonts, actual screenspace measured tabstops).

cmyr:
@jansol I don't _think_ this shouldn't impact those scenarios; there are no immediate plans for variable-width tab stops, and proportional fonts don't strike me as a special case, here. If users want to disable this behaviour for certain file types it's at least a user option.

raphlinus:
Still approve. I think the default for non code editing use cases (think email composition) should probably be disabled, but it sounds like that's covered by having use_tab_stops false in that case.

PR #477, opened 2018-01-12
Put ledger (fuchsia) under feature gate

raphlinus:
It's going to take more work to get ledger running on fuchsia, so this
commit puts it under a feature gate.

### DIFF
--- a/rust/core-lib/src/lib.rs
+++ b/rust/core-lib/src/lib.rs
@@ -25,19 +25,17 @@ extern crate toml;
 extern crate notify;
 
 #[cfg(feature = "ledger")]
-extern crate fuchsia_zircon;
+mod ledger_includes {
+    extern crate fuchsia_zircon;
+    extern crate fuchsia_zircon_sys;
+    extern crate mxruntime;
+    #[macro_use]
+    extern crate fidl;
+    extern crate apps_ledger_services_public;
+    extern crate sha2;
+}
 #[cfg(feature = "ledger")]
-extern crate fuchsia_zircon_sys;
-#[cfg(feature = "ledger")]
-extern crate mxruntime;
-#[cfg(feature = "ledger")]
-#[macro_use]
-extern crate fidl;
-#[cfg(feature = "ledger")]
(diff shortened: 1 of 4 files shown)


raphlinus:
@cramertj

As of this change and the recent changes in the Fuchsia build system, we can now build xi-core-lib. Most of the remaining changes (including integration with the new app lifecycle) are going to be on the Fuchsia side.

rust/core-lib/src/lib.rs
```diff
@@ -24,19 +24,19 @@ extern crate syntect;
 extern crate toml;
 extern crate notify;
 
-#[cfg(target_os = "fuchsia")]
```
cramertj: You might be able to save yourself some work here by doing something like this:
```rust
#[cfg(feature = "ledger")]
mod ledger_includes {
    pub extern crate fuchsia_zircon;
    pub extern crate fuchsia_zircon_sys;
    pub extern crate mxruntime;
    pub extern crate fidl;
    ....
}
#[cfg(feature = "ledger")]
use ledger_includes::*;
```

rust/core-lib/src/tabs.rs
```diff
@@ -893,25 +893,25 @@ impl BufferIdentifier {
 // =============== Fuchsia-specific synchronization plumbing
 // We can't move this elsewhere since it requires access to private fields
```
cramertj: Can you make the fields `pub(crate)` or some other visibility that would allow you to move more of the cfg'd code into its own module?

raphlinus: This is going to get refactored; the idea is that we'll define an interface for multi-device sync that can then get provisioned, and the platform-specific implementation for that will live in the binary wrapper. We could move things around a bit before then, but I'm not feeling strongly motivated.

cramertj: SGTM.

rust/core-lib/src/lib.rs
raphlinus: Good call, I'll do that.

PR #480, opened 2018-01-17
Move notify under feature gate

raphlinus:
The "notify" feature is now optional, but on by default, so no behavior change is expected. (It'll temporarily be off for fuchsia).

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -12,7 +12,7 @@ serde_json = "1.0"
 serde_derive = "1.0"
 time = "0.1"
 toml = "0.4"
-notify = "4.0"
+notify = { optional = true, version = "4.0" }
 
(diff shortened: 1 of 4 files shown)


raphlinus:
@cramertj The plan is to actually remove notify altogether from the Cargo.toml in the fuchsia mirror of this repo. With this refactor, that's a two-line delta.

cramertj:
@raphlinus Cool-- I've also been working with acrichto to make it so that you can `cargo vendor` only the dependencies for the features you select, which would solve that problem. That sounds like a good intermediate approach, though.

raphlinus:
@cramertj Yep, I'm looking forward to that, because it would mean not even having to maintain a small diff with upstream.

raphlinus:
Going to merge based on all signs from tests being positive (travis is part-failing because of ancient xcode provisioning)

Issue #466, opened 2018-01-05
Reworked plugin edits tracking issue


cmyr:
For a while we've been talking about a few changes to how plugin edits are handled; this has been brought up again because it is interfering with our ability to implement things like autoindent.

There are a number of distinct changes around plugin edits, which it makes sense to tackle en masse:

- [x] finger grained edits / delta edits: we currently have a unique representation for plugin edits, when it makes more sense to have plugins use the same Delta representation we use elsewhere in `xi-core`.  This entails implementing serialize/deserialize for deltas, and updating the plugin libraries appropriately.
- [x] send line numbers with updates: (#496) this also means updating plugin libraries to track and update these line offsets, and use them to transparently fetch lines as needed.
- [x] async plugin edits (#367); more specifically, we want to deprecate the current synchronous API for plugin edits, and make this async only.

Currently unclear to me: whether deltas can currently represent cursor priority; and what this means for undo gc, specifically around tracking 'oldest active revision'.

cmyr:
@betterclever This could be a good issue for you to close out.

Take a look at #367. This PR adds the ability for plugins to send edits asynchronously; plugins can send a notification to core with an edit. Previously the plugin could only make an edit in response to an update sent from core.

We would like to remove the old behaviour. When a plugin gets an update, it has to send an acknowledgement, but it does not send an actual edit; any edits would be sent as notifications, using the new API.

This would involve some changes in core, and it would also involve updating syntect to the new behaviour. There will be some tricky bits, but I think it will give you a good idea of the overall plugin system.

Issue #424, opened 2017-11-01
Auto-indentation


cmyr:
Auto-indentation is one of the major missing features in xi right now. What is our approach?

1. Is providing indentation a plugin's responsibility? (Seems like it.)
1. If providing indentation is a plugin's responsibility, what does the API for that look like?
2. If providing indentation is a plugin's responsibility, how do we approach the syntect plugin? Do we add support upstream (is @trishume open to this?) or do we add support locally?

raphlinus:
1. Yes absolutely.

2. The existing API for plugin-originated edits is designed for exactly this use case. I do expect some refinement around things like making sure the inserted indentation is in the same undo group as the user-inserted newline.

3. In the end state, I would like to see support in syntect, because [increaseIndentPattern](https://manual.macromates.com/en/appendix#increase_pattern) and friends are part of TextMate syntax definitions and I believe respected by all modern editors. As a short term expedient, I can see adding simpler logic by hand to the syntect plugin before support lands in syntect proper.

A continuation of this work is "electric parentheses" but I think getting solid auto-indent only should be the immediate priority.

purpleP:
I'm not a contributor, but as I've already said being dissatisfied with every editor I tried I've spent some time thinking about architecture of extensible, rpc-based editor like yours. One of the things I have come is that correct indentation and autoclosing parentheses, xml-tags and stuff like that is impossible to do right without full-fledged parser. As a community we should either demand language authors to provide fail-safe lazy parsers (so they wouldn't have to parse whole document, but only relevant part of it) or create infrastructure to do this. Maybe we could start discussion about making this part of language servers.

dlee:
In line with what @purpleP stated, this new incremental parsing library looks promising: https://github.com/tree-sitter/tree-sitter/.

Kwarrtz:
What's the status of this now? I've heard that auto-indentation is close to landing. If so, which approach has been adopted?

cmyr:
I'll be writing more about this in the coming weeks. I have a mostly working implementation that's based on syntect metadata, but this has raised bigger questions about language support more generally, that I think requires some more bottom-up thinking.

Issue #289, opened 2017-05-14
Update plugin lib to support multi-cursor edits


cmyr:
We currently send updates to plugins by using `Delta::summary`, which is lossy. We need to move towards sending full deltas, and we also need to accept real deltas back from plugins.

PR #481, opened 2018-01-18
Send deltas to plugins

cmyr:
Progress towards #289, #466, #424.

Next task will be using deltas in the opposite direction.

It would be great if plugins had line indexes; it would clear up a lot of this logic. I'm not sure what the best implementation of that is, though. @raphlinus?

Deltas are a bit annoying to work with, for this purpose; we'll probably want to add some more helpers.

It's hard to write tests for the plugin lib; I'd like it if there was a better separation of concerns, but the lifetime constraints make this difficult. I'm not sure what the best approach is.


rust/core-lib/src/editor.rs
```diff
@@ -371,15 +371,11 @@ impl Editor {
                 Some(s) => s.to_owned(),
                 None => self.view.view_id.to_string(),
             };
-                let text = match new_len < MAX_SIZE_LIMIT {
-                true => Some(self.text.slice_to_string(iv.start(), iv.start() + new_len)),
-                false => None
-            };
```
cmyr: I'm not sure what to do about this. 😕

raphlinus: Yeah, this is really really the tricky bit. It is important to keep the size of the RPC bounded, as doing something like delete all + undo on a gigabyte buffer is something that can happen. There are two ways this can go wrong; either the sizes of the strings add up to a large amount, or you can have a ton of small deltas (a million blank lines with a cursor on each, then press space).

What I'm inclined to suggest, in the interest of both correctness and simplicity, is make the entire delta optional. If it's too expensive (by a metric that correlates strongly with the size of its serialized repr), then the semantics are "invalidate everything." Probably add a non-optional field for the size of the document after delta. For plugins that compute a pure function of the input (like syntax highlighting) it's pretty easy to do the right thing. For stuff like indentation, we'll effectively drop the potential edit on the floor.

Other than that, it looks pretty good, I don't see any problems.

cmyr:
@raphlinus updated.

rust/rope/src/delta.rs
```diff
@@ -264,6 +264,17 @@ impl<N: NodeInfo> Delta<N> {
         )
     }
 
+    /// Returns the sum length of the inserts of the delta.
+    pub fn inserts_len(&self) -> usize {
+        self.els.as_slice().iter()
```
raphlinus: The `as_slice` doesn't seem to be needed here, you should be able to do `iter` directly on a Vec with the same effect.

rust/core-lib/src/editor.rs
```diff
@@ -361,6 +361,11 @@ impl Editor {
         self.increment_revs_in_flight();
 
         {
+            let new_len = delta.new_document_len();
+            let delta = match delta.inserts_len() > MAX_SIZE_LIMIT {
```
raphlinus: This logic is not a hard bound on the size of the encoded delta, as 1M deletes would come back as 0. Easiest thing to do would be to add 10 * els.len(). I'd also change the name to something like len_estimate, with a comment that it's an estimate of the encoded size of the delta.

rust/plugin-lib/src/state_cache.rs
```diff
@@ -181,12 +181,16 @@ impl<'a, S: Default + Clone> PluginCtx<'a, S> {
...
+        self.state.buf_size = new_len;
         self.state.rev = rev;
-
-        self.update_line_cache(&delta);
-        self.update_chunk(&delta);
+        if let Some(ref delta) = delta {
+            self.update_line_cache(delta);
+            self.update_chunk(delta);
+        } else {
+            // if there's no delta (very large insert) we blow away everything
```
raphlinus: I'd say "edit" rather than "insert" here, for precision, as there are other edits that may qualify.

rust/core-lib/src/editor.rs
cmyr: I'm a bit confused. My thinking here was that we're only concerned with the byte-size of the RPC, and in almost all cases RPC length is determined by the contents of an insert.

I _can_ imagine a scenario where in a very long document, an editor could diligently create a delta that deleted every second character, producing a delta with ~len(docu)/2 delete ops; but it isn't clear to me that we are explicitly trying to defend against a DOS; in any case I think I'm probably misunderstanding your point? 😁

If we're interested, I could definitely write a heuristic that takes into account (say) the number of delete ops, and approximates the actual size of the encoded delta...

cmyr:
Okay updated & rebased, should be gtg.

Issue #471, opened 2018-01-11
Add chrome tracing support


cmyr:
In [xi-mac#103](https://github.com/google/xi-mac/pull/103) we added support for tracing using the [chrome trace format](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview).

This should be rolled out to xi-core, and eventually plugins. I would defer on the plugin work initially, and implement this in xi-core.

So: we should be generating traces on an ongoing basis, and then in response to some new `dump_trace` RPC we should be _either_ writing them to a directory passed as an argument, or else returning them as a response. (It's unclear to me which of these is better; it comes down to whether the tools for viewing traces want a single file or not, probably)

Some questions:
- Do we want to have some global trace repository, or do we want to use some thread-local structure and synchronize only as needed?
- Do we want to include traces _in_ xi-rpc? If yes, this gets a bit more complicated; we'd need to add some mechanism for collecting these.
- Do we write to a file or return a response?

vlovich:
Step 1 I'll be working on is the base tracing API.  It will be a standalone library crate that works as follows:

1. Use time::precise_time_ns from the time crate as the timestamp basis.
2. Events are either instantaneous or a start/stop block.  The API usage would be:
```
{
   let _guard = xi_perf::trace_block("some thing");
   some_expensive_operation();
}
xi_perf::trace_instant("some other thing");
```
Tracing a closure could also be provided if desirable as demonstrated by the flame library.
3. Capturing is done into 1 ring ring buffer shared between threads (1 per process).  This implies the need for locking at every record site.  For now I think this will be OK as it simplifies the design greatly (can have a global per-process limit for collection hit in terms of RAM & it's easy to iterate/dump whereas traversing TLS data structures requires more work) & shouldn't be an issue until there are a lot of threads generating a lot of trace data.
4. It's off by default.  You have to call `xi_perf::start()` to actually have the perf calls record anything.  Otherwise they're not recorded.  Similarly, there's a `xi_perf::stop()` which stops collection.  Start/stop simply set a global atomic boolean that's checked in each function. 
 `xi_perf::clear()` in case it conflicts with performance improvements in the future.

cmyr:
Sounds great. My only note is that in xi-mac we're currently enabled by default; at least in the short term this at least saves us an RPC, and lets us write a trace if/when we see performance problems, instead of having to start a trace and repro?

vlovich:
[...]

PR #489, opened 2018-01-19
Add support for serializing samples into Chrome trace event

vlovich:
Part 2/2 for #471 provides a mechanism to serialize in-memory samples to a stream.

### DIFF
--- a/rust/trace-dump/Cargo.toml
+++ b/rust/trace-dump/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "xi-trace-dump"
+version = "0.1.0"
+license = "Apache-2.0"
+authors = ["Vitali Lovich <vlovich@google.com>"]
+categories = ["development-tools::profiling"]
+repository = "https://github.com/google/xi-editor"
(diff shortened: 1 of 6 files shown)


cmyr:
It looks like this CI failure is a real one?

vlovich:
Yeah sorry.  I must have added a doc & rebased without testing.  Fixed.

raphlinus:
Looks like there's some work needed to pass the CI tests? I'm very confused because 2edd76c seems to be passing but travis is running 41bc809. I haven't dug in to investigate.

```
   Compiling xi-trace-ffi v0.1.0 (file:///home/travis/build/google/xi-editor/rust/trace-ffi)
error: couldn't read "trace-ffi/examples/c/main.rs": No such file or directory (os error 2)
```

I don't think we have any active intention to use this, but it might be a "nice to have" for other uses.

vlovich:
Fixed in theory.  The commit that should be in this pull request should just be 2602e937d2bb89 which shouldn't have any FFI stuff.

PR #490, opened 2018-01-20
Add FFI bindings

vlovich:
Add FFI bindings to generate samples & serialize them for transmission.

### DIFF
--- a/rust/trace-dump/src/ipc.rs
+++ b/rust/trace-dump/src/ipc.rs
@@ -38,6 +38,10 @@ pub fn serialize_to_stream<'a, W>(samples: &Vec<Sample>, output: &mut W)
         .map_err(|e| Error::BincodeError(e))
 }
 
+pub fn serialized_size(samples: &Vec<Sample>) -> u64 {
+    bincode::serialized_size(samples)
+}
+
 pub fn deserialize_from_bytes(encoded: &[u8]) -> Result<Vec<Sample>, Error> {
     bincode::deserialize(encoded)
         .map_err(|e| Error::BincodeError(e))
--- a/rust/trace-ffi-examples/Cargo.toml
+++ b/rust/trace-ffi-examples/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "xi-trace-ffi-examples"
+version = "0.1.0"
+license = "Apache-2.0"
+authors = ["Vitali Lovich <vlovich@google.com>"]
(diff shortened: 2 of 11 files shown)


cmyr:
@vlovich I'm still confused about about the relationship between this and #489. My best guess is that that PR should just be #b30d037, and this should just be #d2919f0? In any case thanks for your patience.

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -0,0 +1,114 @@
...
+/// ```
+/// use xi_trace_dump::chrome_trace;
+///
+/// let samples = xi_trace::samples_cloned_sorted();
+/// let mut serialized = Vec::<u8>::new();
+/// chrome_trace::serialize(samples.iter(), OutputFormat::JsonArray, serialized);
+/// ```
+pub fn serialize<'a, I, W>(samples: I, format: OutputFormat, output: W)
+    -> Result<(), Error> 
+    where I: IntoIterator<Item = &'a Sample>, W: Write
```
raphlinus: Any particular reason this is IntoIterator rather than Iterator? I think serde is fine with its inputs being passed by reference rather than ownership, and it might be handy to keep the data around after serializing.

vlovich:
Yeah, sorry.  Let me fix up the commits.  Not sure what happened.

rust/trace-dump/src/chrome_trace.rs
vlovich: This code snippet doesn't work:

```
        let samples = trace.samples_cloned_unsorted();

        let mut serialized = Vec::<u8>::new();
        let result = chrome_trace::serialize(
            &samples, chrome_trace::OutputFormat::JsonArray, &mut serialized);
        assert!(result.is_ok());
```

Results in 
```
error[E0277]: the trait bound `&std::vec::Vec<xi_trace::Sample>: std::iter::Iterator` is not satisfied
  --> src/lib.rs:58:22
   |
58 |         let result = chrome_trace::serialize(
   |                      ^^^^^^^^^^^^^^^^^^^^^^^ `&std::vec::Vec<xi_trace::Sample>` is not an iterator; maybe try calling `.iter()` or a similar method
   |
   = help: the trait `std::iter::Iterator` is not implemented for `&std::vec::Vec<xi_trace::Sample>`
   = note: required by `chrome_trace::serialize`
```

PR #491, opened 2018-01-20
Fix config issues

cmyr:
We now verify values before accepting a config, and report errors back to the client. This also includes some minor cleanup.

screen shot 2018-01-20 at 2 03 42 pm

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -341,7 +341,7 @@ impl Documents {
                 }
             Plugin(cmd) => self.do_plugin_cmd(cmd),
             ModifyUserConfig { domain, changes } =>
-                self.do_modify_user_config(domain, changes)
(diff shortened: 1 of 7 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -685,28 +681,36 @@ impl Documents {
...
+            Ok((d, t)) => self.set_config(peer, d, t, Some(path.to_owned())),
+            Err(e) => {
+                let err_msg = format!("{}", &e);
+                peer.send_rpc_notification("alert", &json!({"msg": err_msg}));
+            }
         }
     }
 
     /// Sets (overwriting) the config for a given domain. Will fail if the config
     /// fails validation.
```
raphlinus: Is "fails validation" still the most accurate text for this, or should it be slightly more general to cover other potential failures?

PR #498, opened 2018-01-31
Deserialize chrome JSON trace into data structure

vlovich:
This way xi-mac (or any other frontend) can just send the JSON trace to xi_core for dumping to disk instead of needing to use the Rust tracing framework.

### DIFF
--- a/rust/trace-dump/src/chrome_trace.rs
+++ b/rust/trace-dump/src/chrome_trace.rs
@@ -12,12 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use xi_trace::{Sample, SampleType};
-use serde;
-use serde::ser::Serializer;
-use serde::ser::SerializeSeq;
+use xi_trace::{Sample, SampleType, StrCow, TracePayloadT};
 use serde_json;
-use std::io::{Error as IOError, Write};
+use std::io::{Error as IOError, Read, Write};
 use std::iter::Iterator;
 
 pub enum OutputFormat {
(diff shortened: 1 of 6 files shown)


vlovich:
This CL is based on ffi branch but only depends on the serialize branch.

cmyr:
@vlovich Want to merge #498 and rebase this? You have commit access correct?

vlovich:
This probably isn't the cleanest implementation.  The way it works is that chrome_trace provides the mapping from Samples object to chrome trace JSON struct. If we wanted to add flame graph we'd add a flame_graph.rs that implemented that mapping.  So you don't get serialization for free automatically if you add/remove/rename stuff as the serialization/deserialization code for that specific format has to be adjusted.  However, that seems pretty reasonable since the serialization/deserialization format determines how things are serialized/deserialized and I'm not sure how you'd create a way for it to happen automagically.  I guess you could define a custom tag that indicates how structs maps to fields if it's a 1:1 mapping but even the chrome trace stuff isn't 1:1.

Did I answer your question correctly or were you asking something else?

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -29,58 +26,175 @@ pub enum OutputFormat {
...
+impl<'a, A> From<A> for ChromeTraceArrayEntries where A: IntoIterator<Item = &'a Sample> {
+    fn from(samples: A) -> Self {
+        // Worst case is every sample is a duration.  if we were to use an
+        // original capacity of samples.len() then we'd inevitably encounter
+        // at least 1 duration & thus the underlying Vec would double the
+        // capacity anyway.  Alternatively, we could scan the vector to count
+        // the number of durations & add that to samples.len() for the exact
+        // final capacity.
+        let samples_iter = samples.into_iter();
+        let mut result = Vec::with_capacity(samples_iter.size_hint().0 * 2);
```
cmyr: I'm curious to hear your thinking around vec buffer sizes, since I've noticed you go out of your way a few times to try and pick an initial capacity. My personal inclination would be to use `with_capacity` when I know the initial capacity exactly or I know it's large, and otherwise not sweat a few reallocs; and if I'm worried about size I can always `shrink_to_fit` when I'm done.

Do you think there's a tangible advantage to trying to make an accurate estimate up front? This is more for my edification, not really a review comment. 🙂

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -29,58 +26,175 @@ pub enum OutputFormat {
...
-        "ph": event_type(sample, begin),
-        "ts": to_us(if begin { sample.start_ns } else { sample.get_end_ns() }),
-        "pid": sample.pid,
-        "tid": sample.tid,
-        "args": {
-            "payload": sample.payload
+fn sample_to_entry(sample: &Sample, begin: bool) -> ChromeTraceArrayEntry {
+    ChromeTraceArrayEntry {
+        name: sample.name.clone(),
+        cat: StrCow::from(sample.categories.join(",")),
```
cmyr: Thinking about this being a `Cow`... I guess the borrow case is if there's one category and it's a static &'str, we can skip the allocation?

I'm actually confused about `join` being available here, in the docs I have it seems to be nightly only?

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -29,58 +26,175 @@ pub enum OutputFormat {
...
+    #[serde(rename = "ts")]
+    ts_us: u64,
+    pid: u64,
+    tid: u64,
+    args: Option<ChromeTraceArrayEntryArgs>,
+}
+
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+enum ChromeTraceArrayEntries {
```
cmyr: Is there a reason this is an enum, and not just a tuple struct?

rust/trace-dump/src/chrome_trace.rs
vlovich: If I have a good guess or a reasonable upper-bound, then I tend to provide a capacity.  I tend to not do so when I know there's going to be a large variance in the upper bound.  In this specific case I know it's going to end up allocating exactly 2x anyway using the naiive way so I just do it upfront instead of via a realloc.  The reason I know it's going to be 2x anyway is that it's likely going to be at least len + 1 since there's probably going to be a duration sample.  Thus if I allocate a capacity of len I'll end up with 2*len since vector implementations typically 2x when they need to grow (& I just double-checked that rust does indeed).

shrink_to_fit is good when you know you're going to have a lot of garbage (i.e. after clearing a large vector that's sticking around).

rust/trace-dump/src/chrome_trace.rs
vlovich: Well Chrome trace does technically support 2 serialization formats.  I think I'm overthinking things here too but all this should disappear in the cleanup commit where i'll just change the Sample to map 1:1 with chrome trace events.

PR #499, opened 2018-01-31
bytecount update

llogiq:
@Veedrac managed to simplify *and* (on current rustc) speed up bytecount.


raphlinus:
Thanks for the update!

llogiq:
You're welcome!

Issue #507, opened 2018-02-02
crash in rust plugin lib when pasting over a selection


cmyr:
This is a regression in #481. 

Repro:
1. Copy some text to the clipboard
2. select some amount of text shorter than what is in your clipboard
3. paste

backtrace:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', /Users/rofls/dev/projects/xi-mac/xi-editor/rust/plugin-lib/src/state_cache.rs:526:12
stack backtrace:
   0:        0x1002d2ca3 - std::sys::imp::backtrace::tracing::imp::unwind_backtrace::h3421d74c3554d1d6
   1:        0x1002cf64c - std::sys_common::backtrace::_print::h81d2920454574615
   2:        0x1002d4ad1 - std::panicking::default_hook::{{closure}}::h046b82ccd70886b8
   3:        0x1002d47d6 - std::panicking::default_hook::h00ff9adc172e30f9
   4:        0x1002d5015 - std::panicking::rust_panic_with_hook::hb58bab85cf8f73f4
   5:        0x1002d4eae - std::panicking::begin_panic::hf76e4af07ba1dfa4
   6:        0x1002d4d82 - std::panicking::begin_panic_fmt::hf59ff9d53a059895
   7:        0x10009668d - <xi_plugin_lib::state_cache::PluginCtx<'a, S>>::update_line_cache::ha8ce6fc05aba9900
   8:        0x10009856d - <xi_plugin_lib::state_cache::PluginCtx<'a, S>>::do_update::hb81e3c7f1aefeaf9
   9:        0x1000814b1 - <xi_plugin_lib::state_cache::CacheHandler<'a, P> as xi_plugin_lib::plugin_base::Handler>::handle_request::he20fa1e6d4ff94de
  10:        0x100102bfe - <xi_plugin_lib::plugin_base::MyHandler<'a, H> as xi_rpc::Handler>::handle_request::hde6c4a9d85f57650
  11:        0x100079329 - <xi_rpc::RpcLoop<W>>::mainloop::{{closure}}::h9fa2ce7a2e01d170
  12:        0x10009f0db - crossbeam::scoped::scope::h9e09d4ee80ac304d
  13:        0x100078d2c - <xi_rpc::RpcLoop<W>>::mainloop::he4d9c5576edc92fe
  14:        0x1000f21cd - xi_plugin_lib::plugin_base::mainloop::hba368e66445bc3d0
[...]

PR #512, opened 2018-02-02
Workaround for #507

cmyr:
This doesn't fully resolve the underlying question (what is the expected
behaviour of `Delta::iter_deletions`), but it does fix the
crash.

### DIFF
--- a/rust/plugin-lib/src/state_cache.rs
+++ b/rust/plugin-lib/src/state_cache.rs
@@ -521,8 +521,11 @@ impl<'a, S: Default + Clone> PluginCtx<'a, S> {
     /// Updates the line cache to reflect this delta.
     fn update_line_cache(&mut self, delta: &RopeDelta) {
         let (iv, new_len) = delta.summary();
-        if let Some(n) = delta.as_simple_insert() {
-            assert_eq!(iv.size(), 0);
+        // A paste that replaces a selection is treated as a simple insert,
+        // but our logic in line_cache_simple_insert does not properly account
+        // for the deleted regions, so we have to just clear_to_start in that case.
+        // TODO: revisit if a patch to iter_deletions lands.
+        if let (0, Some(n)) = (iv.size(), delta.as_simple_insert()) {
             assert_eq!(new_len, n.len());
 
             let newline_count = n.measure::<LinesMetric>();


rust/plugin-lib/src/state_cache.rs
```diff
@@ -521,8 +521,10 @@ impl<'a, S: Default + Clone> PluginCtx<'a, S> {
     /// Updates the line cache to reflect this delta.
     fn update_line_cache(&mut self, delta: &RopeDelta) {
         let (iv, new_len) = delta.summary();
-        if let Some(n) = delta.as_simple_insert() {
-            assert_eq!(iv.size(), 0);
+        // A paste that replaces a selection is treated as a simple insert,
+        // but it screws up our logic, so we pass if the interval size > 0.
```
vlovich: can you comment specifically how it screws up the logic?

vlovich:
Another comment I'd add is that this code could use some more documentation.  It's not immediately clear what state is being cached & how some of these functions operate.  Might be worth having a little description at the top about what the caching is for (presumably so that plugin can maintain the stateful RPC), how it works (is it caching the document in a local Rope? is it caching other things too?) & how some of these functions modify the cache.  Also identifying which, if any, functions are purely optimizations & the only side-effect of getting it wrong is performance issues then that would be helpful too probably.

raphlinus:
Having read through the deletions iterator I'm pretty sure it's wrong in the case the delta has both insertions and deletions. While this patch seems safe, I think we should fix that.

rust/plugin-lib/src/state_cache.rs
cmyr: Your understanding is correct: the strategy for updating the cache here is that we handle the two common editing cases (typing/deleting) efficiently, and then in the rest of the cases we just blow away the cache.

How this is getting confused is that we were assuming that a 'simple edit' meant an insertion without any deletions, but apparently an insertion that replaces a region was also getting treated as a simple edit. Our logic in `line_cache_simple_insert` is based on the idea that when updating line offsets further down in the cache, the only thing we need to keep track of is the length of the inserted text; but if there's also a deleted region that logic breaks.

cmyr:
@vlovich Agree with the docs comment, opened #527 as a todo for that.

@raphlinus This was my suspicion. I'll clean up my comments in this patch to be a bit clearer, and add a TODO to revisit it once we patch the deletions iter, and then we can merge this as a temporary solution.

raphlinus:
What I think I'd like to do is write a direct is_simple_insert and is_simple_delete on Delta, as these will be more efficient and simpler to understand. That leaves #528 still open, but not on the critical path. I'll prepare a patch shortly.

cmyr:
I actually added `is_simple_delete` and `as_simple_insert` to delta as part of the original patch; they might just need to be touched up?

cmyr:
closing this, #529 resolves

Issue #506, opened 2018-02-01
Incorrect handling of BOM


cmyr:
We currently fail to properly handle utf-8 which includes a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark). This is valid utf-8, and should be handled correctly.

There's probably a meta-issue here around supporting different encodings. Our internal representation will always be utf-8, but we may want to convert from/to another format when reading and writing; and we may want to allow the user to specify a desired encoding.

My longer term vision of this is that the `Editor` type has a new member, `encoding`, which will be an instance of some `CharacterEncoding` enum. Initially this will have two members, `Utf8` and `Utf8WithBom`. When we read a file, we check for the presence of the BOM (the first three bytes being `ef bb bf`); if it's present we strip it, and assign the appropriate enum member. When writing to disk, we pass the enum with the text to write, and switch on it to determine behaviour.

xi-mac issue: https://github.com/google/xi-mac/issues/125

wuweiweiwu:
@cmyr can I work on this? :)

cmyr:
@wuweiweiwu absolutely. If you have any questions, you can ask them here or in the IRC. :)

wuweiweiwu:
@cmyr will do! Thanks!

ed359:
Sorry @wuweiweiwu, I'd already done this PR which implements in its most basic form the UTF8 BOM support @cmyr mentions here. 

It might be fun to extend this to support a wide variety of encodings, that's a much bigger job!

cmyr:
Fixed in #513. @wuweiweiwu sorry, very happy to find you something else though!

wuweiweiwu:
@cmyr That would be awesome! Is there anything you would recommend for someone beginning in Rust?

cmyr:
@wuweiweiwu #444 might be a reasonable first issue? As raph mentions there, the project would be to move to a more abstract 'gesture' api, instead of talking about clicks specifically. This has been implemented in the case of cmd+click, which the frontend sends as a 'toggle selection gesture'. 

Alternatively, #434 is quite contained.

wuweiweiwu:
@cmyr Sounds good! I'd love to take #444

PR #513, opened 2018-02-05
add UTF8 BOM check to the editor

ed359:
Small fix for for #506. This doesn't represent proper handling of a wide range of input encodings as it only deals with UTF8+BOM, but it fixes a crash in xi-mac: google/xi-mac#125.


raphlinus:
+1 from me also. We're going to have to figure out a general approach to encodings, but for the time being this fixes a real problem.

PR #522, opened 2018-02-15
Add tracing RPC APIs

vlovich:
Add APIs that let the frontend control the tracing state of core + all plugins as well as the ability to gather the samples from all the places (frontend, core, + plugins) & write it to a single trace file.

### DIFF
--- a/rust/core-lib/src/plugins/manager.rs
+++ b/rust/core-lib/src/plugins/manager.rs
@@ -14,7 +14,7 @@
 
 //! `PluginManager` handles launching, monitoring, and communicating with plugins.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::io;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex, Weak, MutexGuard};
@@ -25,6 +25,7 @@ use std::fmt::Debug;
 use serde::Serialize;
 use serde_json::{self, Value};
 
+use xi_rpc;
 use xi_rpc::{RpcCtx, Handler, RemoteError};
 
 use tabs::{BufferIdentifier, ViewIdentifier, BufferContainerRef};
@@ -143,6 +144,29 @@ impl PluginManager {
                 }
             }
         }
+    }
+
+    // NOTE: This is a temporary API just for tracing.
+    fn request_trace_rpc_sync(&self, method: &str, params: &Value)
+        -> Vec<Result<Value, xi_rpc::Error>>
+    {
+        let mut gathered_results = Vec::new();
+
+        for plugin in self.global_plugins.values() {
+            let result = plugin.request_trace_rpc_sync(method, &params);
+            gathered_results.push(result);
+        }
+
+        let mut processed_plugins = HashSet::new();
+
+        for plugin in self.buffer_plugins.values().flat_map(|group| group.values()) {
+            // currently each buffer must have its own instance of a given plugin running.
+            assert!(processed_plugins.insert(plugin.get_identifier()));
+            let result = plugin.request_trace_rpc_sync(method, &params);
+            gathered_results.push(result);
+        }
+
+        gathered_results
     }
 
     fn dispatch_command(&self, view_id: ViewIdentifier, receiver: &str,
@@ -403,6 +427,19 @@ impl PluginManagerRef {
 
     }
 
+    pub fn toggle_tracing(&self, enabled: bool) {
+        self.lock().request_trace_rpc_sync(
+            "tracing_config",
+            &json!({"enabled": enabled}),
+        );
+    }
+
+    pub fn collect_trace(&self) -> Vec<Result<Value, xi_rpc::Error>> {
+        self.lock().request_trace_rpc_sync(
+            "collect_trace",
+            &json!({}),
+        )
+    }
 
     /// Called when a new buffer is created.
     pub fn document_new(&self, view_id: ViewIdentifier, init_info: &PluginBufferInfo) {
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -30,6 +30,7 @@ use notify::{RecursiveMode, DebouncedEvent};
 
 use xi_rope::rope::Rope;
 use xi_rpc::{RpcCtx, RemoteError};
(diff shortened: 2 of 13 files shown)


rust/core-lib/src/plugins/manager.rs
```diff
@@ -145,6 +146,41 @@ impl PluginManager {
...
+            } else if let Ok(locals) = self.running_for_view(view_id) {
+                locals.len()
+            } else {
+                0
+            }
+        };
+        let num_plugins = num_global_plugins + num_non_global_plugins;
+        let mut gathered_results = Vec::with_capacity(num_plugins);
+
+        // TODO: query plugins in parallel & gather results.
```
cmyr: Not sure if you saw, but there _is_ a non-blocking request method that takes a callback. Or at least, there used to be. This would require something like using an atomic int type to track how many requests you send, and then decrement that when each returns, firing some finalizer when it hits zero... probably more bookkeeping than is worth doing right now, but at least an option.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -145,6 +146,41 @@ impl PluginManager {
         }
     }
 
+    fn request_plugins_sync(&self, view_id: ViewIdentifier,
```
cmyr: Looking down at the code where you create a dummy view identifier, I don't think we should add this method. It's only used (and probably should only ever be used) for collecting traces, but by presenting itself as a general helper method it includes fields (`view_id` and `only_globals`) that are irrelevant to this use case. Instead the plugin manager should have a method specific to this request, `collect_traces`, which implements this specific logic.

Also given that it's a slow, blocking call, we really don't want anyone else to reuse it.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -403,6 +439,20 @@ impl PluginManagerRef {
...
+        let dummy_view_id = ViewIdentifier::from("view-id-0");
+        let only_globals = true;
+        self.lock().request_plugins_sync(
+            dummy_view_id, only_globals, "tracing_config",
+            &json!({"enabled": enabled}));
+    }
+
+    pub fn collect_trace(&self) -> Vec<Result<Value, xi_rpc::Error>> {
+        let dummy_view_id = ViewIdentifier::from("view-id-0");
+        let only_globals = true;
```
cmyr: if we _do_ go this route, we want `only_globals` to be false. Syntect, for instance, is not currently a global plugin. (Each document has its own instance of syntect running, something we ultimately want to change)

rust/core-lib/src/tabs.rs
```diff
@@ -350,6 +405,7 @@ impl Documents {
         use rpc::CoreRequest::*;
         match cmd {
             NewView { file_path } => {
+                eprintln!("xi-core new view");
```
cmyr: Maybe not intended for this commit?

rust/core-lib/src/plugins/manager.rs
vlovich: Yeah i was concerned about whether it would deadlock (wasn't sure about which thread it would come back on) which is why I went the synchronous route.  I was going to use a condition variable with a vector although the conditional variable has a pretty bad API at the moment for conditional wait (I have a CL out for stdlib to add them back but currently it's unstable & not submitted yet).

rust/core-lib/src/plugins/manager.rs
vlovich: Makes sense.  I'll adjust & put a note that eventually should migrate to an asynchronous API.

rust/core-lib/src/plugins/manager.rs
vlovich: Gotcha.  I didn't understand what that variable meant so I guessed wrong :).

rust/core-lib/src/tabs.rs
vlovich: Definitely not.  I was debugging stuff & accidentally committed.  Thanks for catching.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -145,6 +146,35 @@ impl PluginManager {
         }
     }
 
+    // NOTE: This is a temporary API just for tracing.
+    fn request_trace_rpc_sync(&self, method: &str, params: &Value)
+        -> Vec<Result<Value, xi_rpc::Error>>
+    {
+        let num_global_plugins = self.global_plugins.len();
+        let num_non_global_plugins = self.buffer_plugins.len();
```
cmyr: `self.buffer_plugins` isn't a list of plugins, but a list of buffers which _have_ plugins, so this won't be returning the correct value if any buffers have more than one plugin running.

Your logic below where you actually dispatch to all plugins is correct, though.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -145,6 +146,35 @@ impl PluginManager {
...
+
+        for plugin in self.global_plugins.values() {
+            let result = plugin.request_trace_rpc_sync(method, &params);
+            gathered_results.push(result);
+        }
+
+        let mut processed_plugins = HashSet::new();
+
+        for plugin in self.buffer_plugins.values().flat_map(|group| group.values()) {
+            if !processed_plugins.insert(plugin.get_identifier()) {
```
cmyr: I think we discussed this on IRC; this isn't strictly necessary. It's an invariant that a given plugin exists in either `global_plugins` or in **one** `buffer_plugin` collection.

rust/core-lib/src/plugins/manager.rs
vlovich: Yeah, you're right.  I might just remove the with_capacity then.  There's no easy way a priori of iterating the data structure to know it.

rust/core-lib/src/plugins/manager.rs
vlovich: Did i implement it wrong?  my intent here is to de-dupe plugins that may be for multiple buffers.  i'm not processing global plugins here i *think*.  Did I miss something?

cmyr: A plugin (which refers to an individual process) can only ever exist in one buffer collection. If multiple buffers are running syntect, for instance, each has its own distinct `PluginRef`, communicating with a distinct process.

PR #523, opened 2018-02-16
Add make uppercase/lowercase

scholtzan:
This adds functionality to make text uppercase and lowercase. 
Let me know if anything is missing - I am currently getting familiar with the codebase in preparation for GSoC.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -961,13 +961,13 @@ impl Editor {
         self.view.collapse_selections(&self.text);
     }
 
-    fn transform_text(&mut self, transform_function: &Fn(String) -> String) {
+    fn transform_text<F: Fn(&str) -> String>(&mut self, transform_function: F) {
         let mut builder = delta::Builder::new(self.text.len());
 
         for region in self.view.sel_regions() {
             let selected_text = self.text.slice_to_string(region.min(), region.max());
             let interval = Interval::new_closed_open(region.min(), region.max());
-            builder.replace(interval, Rope::from(transform_function(selected_text)));
+            builder.replace(interval, Rope::from(transform_function(&selected_text)));
         }
         if !builder.is_empty() {
             self.this_edit_type = EditType::Other;
@@ -1052,12 +1052,13 @@ impl Editor {
             DebugRewrap => self.debug_rewrap(),
             DebugPrintSpans => self.debug_print_spans(),
             CancelOperation => self.do_cancel_operation(),
-            Uppercase => self.transform_text(&|s| s.to_uppercase()),
-            Lowercase => self.transform_text(&|s| s.to_lowercase()),
+            Uppercase => self.transform_text(|s| s.to_uppercase()),
+            Lowercase => self.transform_text(|s| s.to_lowercase()),
         };
 
         self.cmd_postlude();
     }
+
 
     pub fn handle_request(&mut self, _view_id: ViewIdentifier,
                           cmd: rpc::EditRequest) -> Result<Value, RemoteError> {
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -961,27 +961,13 @@ impl Editor {
         self.view.collapse_selections(&self.text);
     }
 
-    fn make_uppercase(&mut self) {
+    fn transform_text(&mut self, transform_function: &Fn(String) -> String) {
         let mut builder = delta::Builder::new(self.text.len());
 
         for region in self.view.sel_regions() {
             let selected_text = self.text.slice_to_string(region.min(), region.max());
             let interval = Interval::new_closed_open(region.min(), region.max());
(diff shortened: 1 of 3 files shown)


scholtzan:
I signed it!

cmyr:
@scholtzan Cool, thanks for this! We might need to do some thinking about whether it makes more sense to express this more generally (perhaps having a `transform_text` method that takes an enum for upper/lower/title, e.g.?) but will be happy to get this in in some form, and very much appreciate the initiative.

Not sure if you're on irc, but that's often the best place to find us; #xi on irc.mozilla.org.

scholtzan:
Thank you for the suggestion! I tried to implement it using a general `transform_text` method. Let me know if that's what you had in mind.
And I'll join your irc channel.

rust/core-lib/src/editor.rs
```diff
@@ -961,27 +961,13 @@ impl Editor {
         self.view.collapse_selections(&self.text);
     }
 
-    fn make_uppercase(&mut self) {
+    fn transform_text(&mut self, transform_function: &Fn(String) -> String) {
```
vlovich: cmyr, should the Fn accept a &String?  Not sure what as written implies about the source string being copied.  Also, should this be an FnMut instead?

cmyr: I'm going to hold off on this until tomorrow when I'm bright eyed and caffeinated, because I don't want to suggest something daft and then overrule myself in the morning.

rust/core-lib/src/editor.rs
```diff
@@ -961,6 +961,20 @@ impl Editor {
         self.view.collapse_selections(&self.text);
     }
 
+    fn transform_text(&mut self, transform_function: &Fn(String) -> String) {
+        let mut builder = delta::Builder::new(self.text.len());
+
+        for region in self.view.sel_regions() {
```
cmyr: There's a sort of discussion question, here: what is the granularity of this operation? On the mac, the `uppercaseWord` / `lowercaseWord` methods are supposed to transform any word that overlaps with the selection. What is our desired behaviour with an empty selection region (a caret)? On macOS, if you map a key to the `uppercaseWord:` selector, then (in TextEdit, say) it expands your selection to the full word before uppercasing. In sublime text, uppercase / lowercase commands operate on the selected region, or on the containing word in the caret case.

My personal feeling here is that in the caret case, we should be applying the operation to the containing word. Does that seem reasonable?

vlovich: I concur that it should only do it for the selection within core.  For the caret case I think it shouldn't do anything.  It would be the front-end's job I think to go with whatever makes sense for the platform.  For OSX if we want TextEdit-like behaviour the front-end should adjust the selection(s) to word boundaries.  If you wanted to adjust the words without modifying the selection, the frontend could adjust to word boundaries, make transformation, adjust back.

rust/core-lib/src/editor.rs
```diff
@@ -961,6 +961,20 @@ impl Editor {
...
+            let selected_text = self.text.slice_to_string(region.min(), region.max());
```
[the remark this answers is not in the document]
raphlinus: Agree that we should prioritize code cleanliness over efficiency in this case; it's hard to imagine a use case where it'll make a noticeable difference.

rust/core-lib/src/editor.rs
```diff
@@ -961,6 +961,20 @@ impl Editor {
         self.view.collapse_selections(&self.text);
     }
 
+    fn transform_text(&mut self, transform_function: &Fn(String) -> String) {
```
raphlinus: I think a more idiomatic signature for this function would be:

    fn transform_text<F: Fn(String) -> String>(&mut self, transform_function: F)

I'm not sure about this, and looking at relevant docs. Most examples use this pattern, but perhaps `&Fn` might be appropriate, as it might be smaller compiled code (less duplication on monomorphization). Was wondering your thoughts as you wrote this.

I'm also wondering why the argument is an owned `String` rather than a `&str` ref; the latter might afford an opportunity to optimize, as `slice_to_string` has a TODO to be a `Cow`, which would let it apply the same optimization as the LinesRaw iterator.

scholtzan: I was also a bit unsure when I wrote it and looked at documentation and after reading them thought that `&Fn` might be better since otherwise code might be duplicated. However, now that I read the docs again I am not sure if that is actually a problem for this case.

As for `String` vs `&str` I didn't put much thought into that. But now that you are pointing it out, it might be better to use `&str` instead.

raphlinus:
@scholtzan Thanks. Let us know if you want to make any changes to transform_text and then I'll be happy to merge.

scholtzan:
I changed the signature to `Fn` and the argument to `&str`. I think it is ready to be merged.

cmyr:
@scholtzan Great, looks good!

cmyr:
This is ready, but blocked on #512.

cmyr:
@scholtzan thanks for this!

PR #524, opened 2018-02-17
First version of GitHub page for Xi

ArazAbishov:
- Features xi color theme and icons
- Copied over all GSoC ideas from Google Doc to a dedicated document under tab
- Copied over contribution document under standalone tab

This is what we will get after merging this PR: http://abishov.com/xi-editor/


raphlinus:
Thanks for this, looks great!

PR #526, opened 2018-02-20
Tracepoints

vlovich:
Add a few tracepoints into the RPC layer.

### DIFF
--- a/rust/rpc/src/lib.rs
+++ b/rust/rpc/src/lib.rs
@@ -28,6 +28,7 @@ extern crate serde_json;
 extern crate serde_derive;
 extern crate serde;
 extern crate crossbeam;
+extern crate xi_trace;
 
 mod parse;
 mod error;
@@ -213,6 +214,8 @@ impl<W: Write + Send> RpcLoop<W> {
           RF: Send + FnOnce() -> R,
           H: Handler,
     {
+        let _mainloop_trace = xi_trace::trace_block(
+            "rpc mainloop", &["rpc", "notification"]);
 
         let exit = crossbeam::scope(|scope| {
             let peer = self.get_raw_peer();
@@ -222,6 +225,7 @@ impl<W: Write + Send> RpcLoop<W> {
             scope.spawn(move|| {
                 let mut stream = rf();
                 loop {
+                    let json_decode = xi_trace::trace_block("decode rpc", &["rpc", "json"]);
                     let json = match self.reader.next(&mut stream) {
                         Ok(json) => json,
                         Err(err) => {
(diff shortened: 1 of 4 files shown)


rust/rpc/src/lib.rs
```diff
@@ -249,6 +256,7 @@ impl<W: Write + Send> RpcLoop<W> {
             });
 
             loop {
+                let read_result_trace = xi_trace::trace_block("read & process result", &["rpc"]);
```
cmyr: I think rather than explicitly dropping these my preference would be to add a new block for the scope of the trace.

rust/rpc/src/lib.rs
```diff
@@ -263,15 +271,24 @@ impl<W: Write + Send> RpcLoop<W> {
...
                 match json.into_rpc::<H::Notification, H::Request>() {
                     Ok(Call::Request(id, cmd)) => {
+                        let _request_trace = xi_trace::trace_block("handle_request", &["rpc"]);
                         let result = handler.handle_request(&ctx, cmd);
                         peer.respond(result, id);
                     }
-                    Ok(Call::Notification(cmd)) => handler.handle_notification(&ctx, cmd),
-                    Ok(Call::InvalidRequest(id, err)) => peer.respond(Err(err), id),
+                    Ok(Call::Notification(cmd)) => {
+                        let notification_trace = xi_trace::trace_block("handle_notification", &["rpc"]);
```
cmyr: Needs a _dunder or we get a compilation warning.

rust/rpc/src/lib.rs
```diff
@@ -222,13 +225,17 @@ impl<W: Write + Send> RpcLoop<W> {
             scope.spawn(move|| {
                 let mut stream = rf();
                 loop {
+                    let json_decode = xi_trace::trace_block("decode rpc", &["rpc", "json"]);
```
cmyr: This block will be spending most of it's time waiting on new input, we probably want to be tracing in `rpc/src/parse.rs`. Get that this PR is mostly intended as an illustration, though.

vlovich: Yeah, this CL is just an illustration not intended for merging.  Knowing how much time the process is waiting on input isn't actually a terrible idea but you're right the RPC decoding duration is better measured elsewhere.  I can close the PR to make it more clear if you feel the example is OK to find in the closed PRs.

cmyr: I don't mind keeping it open until something gets merged, up to you. :)

cmyr:
Tracing added in #579, going to close this. Thanks for the illustration!

PR #530, opened 2018-02-23
Fix comment typo

RyanWarnock:


cmyr:
Thanks, but I don't think this is a typo: https://en.wikipedia.org/wiki/If_and_only_if

RyanWarnock:
Ah yep, that should have clicked for me. Time for bed I think.

cmyr:
@RyanWarnock No worries, thanks for the PR!

PR #533, opened 2018-02-27
Moved front-end docs under dedicated page

ArazAbishov:
Start the process of moving docs to xi-editor.io. 

Note, frontend protocol and update documents have been merged into a single page.


ArazAbishov:
Awesome!

PR #534, opened 2018-02-28
Moving over more docs to xi-editor.io - part 1

ArazAbishov:
- Fixed a broken link to the frontend-protocol document in the readme.
- Moved all of the documents over, except rope science related files (will be done in the following PR).


docs/docs/crdt.md
```diff
@@ -1,4 +1,10 @@
-# A CRDT approach to async plugins and undo
+---
+layout: page
+title: CRDT - Approach to async plugins and undo
```
cmyr: Nit: Maybe 'An approach'?

ArazAbishov: Good point, done!

Issue #362, opened 2017-07-09
reload files if changed outside the editor


jansol:
Another very common and very useful feature. Needs a way to have the frontend ask the user if they actually want to reload the file, though.

prasoon2211:
We can probably use a combination of [last modified](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.modified) timestamp and md5 of file contents to check this. If the file has been modified, the backend can emit a `ReloadPrompt` request. Then, it'll be up to the client to handle this. If user says yes, we reload file contents from disk. Otherwise, we set a `reload_chk` flag to `false` (reset as `true` every time file is saved and `true` by default).

I'll try to come up with a solution and send in a PR in a couple of days.

cmyr:
@prasoon2211 we're going to need to do fs monitoring for a bunch of other stuff (watching for settings changes, new plugins/themes etc). I'd encourage you to think about those use cases as well if you're taking a stab at this. I'd also take a look and see if there are any crates available for some of this functionality (https://github.com/passcod/notify)?

prasoon2211:
@cmyr Alright. I'll try to think in a more generic way to implement this although I should warn you - I'm pretty new to both Rust and this project and the PR might require some help. The crate you mentioned looks good at a first glance - I'll see if it'll do for our use cases.

cmyr:
@prasoon2211 very happy to help out however I can, fs monitoring is something that's in the back of my head as needing to get thought about soon but I haven't really gotten around to it yet. Let me know if you have any questions. @raphlinus or @trishume might also have some opinions w/r/t fuchsia. Given our pretty broad platform goals we'll probably want the ability to implement the monitoring mechanics ourselves where necessary, but it would be nice to take advantage of existing work where possible.

prasoon2211:
I have very little knowledge about Fuchsia. The `notify` crate is cross-platform though either by using native OS support (Win/Linux/Mac) (fast) or through polling (slow).
So, I think that we should be able to use this crate. Native support for Fuchsia can probably be added sometime later.

trishume:
@prasoon2211 yah don't worry about supporting Fuchsia. We don't even really support loading files on Fuchsia, so reloading files is kinda a long way out. When we get around to that we'll fix it so it works. As long as you don't do anything that would obviously break the build on Fuchsia, and you don't have to check that yourself, if you're worried you might break Fuchsia just ask me and I can test it.

[...]

PR #538, opened 2018-03-06
Watch open files for changes

cmyr:
This is a first stab at fixing #362.

This is a fairly dumb implementation; no md5, just timestamps.

If we have a file open and it changes on disk we automatically load the changes if we have no local changes, otherwise we set a flag indicating there are changes on disk. In this case, currently, we just don't let you save at that path again.

This behaviour obviously isn't great; I'm still thinking a bit about how this should actually behave. This comes down in general to how we report errors to the client; my feeling is that we should probably define a set of known error codes, (of which `FileHasChangedOnDisk` could be one) and then more nuanced handling of individual errors could be a client concern; for instance in this case the client could prompt the user to overwrite the changed file, or to show a diff, or whatever. (For the former case, `do_save` would have to get an `overwrite` flag.)

Other things:

- I'd rewritten all the file watching stuff a while ago because my first pass wasn't very flexible. That's included here.
- In the _Future_ I think that something at the runloop level should handle watching files, and send notifications up similar to how idle scheduling works.
- Should we be taking an md5 of file contents? Maybe?
- It's debatable that this should be merged as is, depending on whether the ergonomic improvement in the case where there are no changes in the open file is offset by the ergonomic degradation in the case where there _are_ changes.


rust/core-lib/src/tabs.rs
```diff
@@ -1019,6 +1111,18 @@ impl Documents {
     }
 }
 
+
+/// Returns the modification timestamp for the file at a given path,
+/// if present.
+pub fn get_file_mod_time<P>(path: P) -> Option<SystemTime>
```
raphlinus: I'd be curious how resistant all this logic is to races. From my understanding, View.new_view_with_file reads the file, _then_ does set_path which eventually drills down to this call. If the file is changed between these two invocations, I think we can get a situation where it has the old contents but believes the timestamp is fresh.

Getting this right is hard, in such races a lot can go wrong. I'd lean toward not doing anything heroic in this PR, but filing an issue.

rust/core-lib/src/watcher.rs
```diff
@@ -12,108 +12,501 @@
...
 
-    //TODO impl unwatch, when we add in watching of opened files
+impl fmt::Debug for Watchee {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Watchee path: {:?}, r {}, t {} f {}",
+               self.path, self.recursive, self.token.0, self.filter.is_some())
+    }
+}
+
+fn mode_from_bool(b: bool) -> RecursiveMode {
```
raphlinus: `is_recursive` would be a better name than `b`

rust/core-lib/src/watcher.rs
```diff
@@ -12,108 +12,501 @@
...
+
+    // Sleep for `duration` in milliseconds
+    pub fn sleep(duration: u64) {
+        thread::sleep(Duration::from_millis(duration));
+    }
+
+    // Sleep for `duration` in milliseconds if running on OS X
+    pub fn sleep_macos(duration: u64) {
+        if cfg!(target_os = "macos") {
+            thread::sleep(Duration::from_millis(duration));
```
raphlinus: Feels like it would be better to call sleep rather than duplicating the impl.

rust/core-lib/src/watcher.rs
```diff
@@ -12,108 +12,501 @@
...
+        }
+    }
+
+    // Sleep for `duration` in milliseconds
+    pub fn sleep(duration: u64) {
+        thread::sleep(Duration::from_millis(duration));
+    }
+
+    // Sleep for `duration` in milliseconds if running on OS X
+    pub fn sleep_macos(duration: u64) {
```
raphlinus: I think `sleep_if_macos` might be better, the current name suggests it's a mac-specific impl. That said, looking at the CI failures, I wouldn't be surprised if this logic needs to change.

rust/core-lib/src/watcher.rs
```diff
@@ -12,108 +12,501 @@
...
+    }
+
+    impl Notify for mpsc::Sender<bool> {
+        fn notify(&self) {
+            self.send(true).expect("send shouldn't fail")
+        }
+    }
+
+    // Sleep for `duration` in milliseconds
+    pub fn sleep(duration: u64) {
```
raphlinus: I would say `millis` rather than `duration` as a more precise characterization of units.

rust/core-lib/src/tabs.rs
```diff
@@ -694,6 +752,40 @@ impl Documents {
         }
     }
 
+    /// Handles a file system event related to a currently open file
+    fn handle_open_file_fs_event(&mut self, event: DebouncedEvent, _peer: &MainPeer) {
```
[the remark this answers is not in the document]
cmyr: Fair, I wrote the method signature before the implementation, and just matched the signature of the other fs event handling function.

rust/core-lib/src/tabs.rs
cmyr: You're right, we could definitely race here. We could definitely try `flock`ing while we do the read / get the modification date without adding much complexity? Seems reasonable, but would add a dep ([fs2](https://github.com/danburkert/fs2-rs)?). I don't know much about how file locking works in general, possible there's some problem with this approach I'm not aware of.

raphlinus: I'm not expert on `flock` and friends, but my understanding is that they give very weak guarantees (my recollection from years back is that they're especially weak on network filesystems, which these days are rarer but not unheard of). I think figuring out the right thing to do is something of an investigation, as I don't think there are any clear, simple wins.

cmyr: Just musing a bit, but couldn't we also just grab the mod time before and after the read, retrying if there's a change?

raphlinus: That sounds plausible. Another possibility is registering the watcher before loading, so that a notification would come in.

cmyr:
I've been running this the past few days without anything terrible happening, so let's try bringing it in.

vlovich:
Little late to this review so this might be a fairly stupid question, but can you clarify why we need timestamps/md5 at all?  Can't we just translate the filesystem watcher event directly to a UI notification?  What problem do the extra checks solve?

cmyr:
@vlovich in the hashing case, it would be nice if a write that didn't change document contents was ignored. In the timestamp case, the main advantage is that we don't need to do anything special when we write the file ourself; because we update our internal timestamp after the write, we can correctly ignore any notifications that result from writes that we initiate.

vlovich:
I would think hashing would drastically slow down editing large files as you're going to have to hash on every edit.  You might be able to make it smarter enough to do at least once per FS change although that still seems like an expensive operation for large files to me.  Are there any existing editors out there you know of that hash the contents?

For the self-update case, a simple case would be to unregister the watch file saving & re-register after the save is complete.  It has a race condition window but it seems like it would be the same as the last modified timestamp approach.  It's also more robust than the timestamp approach as changing file contents without a changing timestamp seems like a realistic scenario that could come up in the real world.

cmyr:
@vlovich agree, the hashing approach (which is not currently implemented) is only something we would ever do after detecting a write, in which case we would be reading the file anyway; and we wouldn't be comparing to the active buffer contents, but just to a stored hash of that file.

I went with the timestamp approach because that's what NSDocument does, and it seemed like a reasonable middle ground; but it's definitely possible that there could be races there; I suspect much of that will be file system dependent?

For self-update, registering/unregistering a watcher is not an atomic operation, so it suffers similar problems. It's also very dependent on the underlying API (FSEvent / kqueue / inotify).

I'm definitely not a file systems expert, so I can't speak to the likelihood of sequential writes to a file producing identical timestamps. It seems like that would be... bad? If this is actually a common thing that happens, I agree we should look at another approach. I'm hesitant about relying too heavily on the notifier though; I think the filesystem should be the source of truth, and the notifier is just telling us when we might want to go take a look.

vlovich:
So in the typical case of editing a plain file, you are correct that the timestamps are not going to be surprising.  There are actually mount options on Linux which disable modified timestamps too for extra performance so Xi would be broken on those systems - if Xi is at all trying to compete with Vi/Emacs I think that's something to consider, if it's more traditionally Gui stuff then probably not.

Additionally, I see it being totally reasonable for a VCS operation(s) to have the file modified without the timestamp changing or the file stay the same but the timestamp changes.

Similarly, unzipping a file typically restores the last modified time so it's not out of the realm of possibility that you end up with weird stuff about the timestamps.

I think the model for this particular case is erring on the side of assume things changed if the FS event tells you it probably did & let the user decide.  If you want to be more clever then sure, but I think you want to aim for 0 chance of false positive/false negative (e.g. hashing).

Regarding the race condition, yes I agree 100% & the same race condition exists with when you end up reading the timestamp.  I think a reasonable mitigation approach would be to pick a random file name that's a sibling & rename on top of the original.  Then you can tell it's a save because you just ignore the updates for a rename from the random path that no one but you could know.  I would think this way the race condition is pretty effectively eliminated.

PR #539, opened 2018-03-07
prevent syntect plugin autoindent update on empty string

### LINKED ISSUE #492: Syntect plugin shouldn't send an update for autoindent if the previous line is blank
Currently we check if the preceding line only contained whitespace, but the empty string passes this test. We should not send an update in the empty string case.

Related to, (but not necessarily the cause of?) https://github.com/google/xi-mac/issues/115

Anyone interested in taking a stab at this it should be a pretty quick fix in [rust/syntect-plugin/src/main.rs](https://github.com/google/xi-editor/blob/master/rust/syntect-plugin/src/main.rs)

betterclever:
Fixes #492 

@cmyr Seems to be working for me. Is it the intended solution? Please suggest changes if any.

### DIFF
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -206,7 +206,9 @@ impl<'a> PluginState<'a> {
         // don't touch indentation if this is not a simple edit
         if end != start { return None }
 
-        let is_newline = &ctx.get_config().line_ending == text;
+        let line_ending = ctx.get_config().line_ending.clone();
+        let is_newline = line_ending == text;
+
         if is_newline {
             let line_num = ctx.find_offset(start).err();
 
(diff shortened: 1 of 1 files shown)


rust/syntect-plugin/src/main.rs
```diff
@@ -214,6 +214,9 @@ impl<'a> PluginState<'a> {
             let tab_size = ctx.get_config().tab_size;
             let buf_size = ctx.get_buf_size();
             if let Some(line) = line_num.and_then(|idx| ctx.get_line(idx).ok()) {
+                // do not send update if last line is empty string (contains only new line)
+                if line == "\n" { return None; }
```
cmyr: This is basically it, yea, but there's one subtlety: Instead of comparing directly to "\n", we should compare against the line-ending in use by this particular file; it may be using something else. You can see we do something similar above here, when we check `is_newline`.

You'll have to do something to get around the borrow checker here; perhaps cloning the config's `line_ending`.

And a tiny nit: Rust style is to leave off the trailing ';' on return statements.

otherwise looks good. :)

rust/syntect-plugin/src/main.rs
```diff
@@ -206,14 +206,19 @@ impl<'a> PluginState<'a> {
         // don't touch indentation if this is not a simple edit
         if end != start { return None }
 
-        let is_newline = &ctx.get_config().line_ending == text;
+        let line_ending = ctx.get_config().line_ending.clone();
```
betterclever: Not sure if this is the right thing to do but I was not able to access &ctx within if due to an error. 
```
closure requires unique access to `ctx` but `*ctx` is already borrowed
```
I am not sure if there is a better way of doing this.

cmyr: Yep this is reasonable.

Issue #408, opened 2017-10-08
Real logging in xi-core (+ core RPC logging)


cmyr:
Currently xi-core's only logging mechanism is printing to stderr. We will eventually want to move to a proper logging solution; a good example of this is available in the [xi-tui](https://github.com/little-dude/xi-tui/blob/master/src/main.rs) repo.

In addition to this, a small thing we've talked about in the past is adding complete logging of RPC messages, which would be a useful tool for debugging.

betterclever:
I am working on this.

cmyr:
@betterclever awesome, I'd really love to see this land; a goal should be to replace every use of `eprintln!` a log call.

betterclever:
@cmyr I tried using the log4rs module. The problem with it is that , as you pointed, stdout is use for rpc so it leads a panic. We need to direct the logging output to a file probably. Or we first need to fix this, https://github.com/google/xi-editor/issues/434 so that system ignores the unrecognized RPC (log messages) and prints it to console instead.

What is your take?

cmyr:
@betterclever there must be some way to use stderr instead of stdout for logging. That this works for xi-tui means it must be possible, since they would have the same problem.

We definitely don't want to be using stdout for anything but RPC traffic.

little-dude:
In `xi-tui` I use `log4rs` to have multiple loggers logging in multiple files.
- One file contains the regular logs
- One file contains only the RPC calls

That way, I can easily debug `xi-core` issues by copy-pasting RPC requests from the rpc log file, and I don't pollute my actual log file with all the RPC logs.

Here is for instance what my regular log file looks like:

```
 2018-03-12T22:07:34.116785526+01:00 INFO xi_term - starting the event loop
2018-03-12T22:07:34.116924558+01:00 INFO xi_term - starting xi-core
2018-03-12T22:07:34.117583209+01:00 INFO xi_term - starting logging xi-core errors
2018-03-12T22:07:34.117629007+01:00 INFO xi_term - initializing the TUI
2018-03-12T22:07:34.117705815+01:00 INFO xrl::client - >>> notification: method=client_started, params={"config_dir":"/home/corentih/.config/xi/"}
2018-03-12T22:07:34.117911420+01:00 INFO xrl::client - >>> request : method=new_view, params={"file_path":"README.md"}
2018-03-12T22:07:34.118002252+01:00 INFO xrl::client - >>> notification: method=set_theme, params={"theme_name":"base16-eighties.dark"}
[...]

PR #540, opened 2018-03-08
WIP: Added logging support with Fern

betterclever:
First step towards #408 

Work in progress.

## Tasks

- [x] : add logging library
- [ ] : convert existing calls to stderr directly via eprintln macro to log calls.
- [ ] : log all RPC calls.  

@cmyr  Please affirm if I am heading in right direction. Thanks 😄

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -183,6 +183,14 @@ version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "fern"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "log 0.4.3 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "filetime"
 version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


cmyr:
This looks good! The one thing I'm noticing in the diff is that log4rs brings in a bunch of dependencies... I wonder if it might be worth looking at various other logging backends, like [slog](https://github.com/slog-rs/slog) or [fern](https://github.com/daboross/fern)? I don't really know much about the tradeoffs here, but that might be worth investigating.

betterclever:
@cmyr I tried 2-3 implementations, almost all of them have their own dependencies as much as log4rs. Seeing more.

cmyr:
@betterclever Okay, thanks for checking. If you think log4rs is the best fit then let's go with that. It looks like there are a lot of cargo features, so we might be able to turn some stuff off; we probably don't need file-based config, for instance.

betterclever:
[stderrlog-rs](https://github.com/cardoe/stderrlog-rs) has less dependencies and is working fine. Updating PR soon.

cmyr:
@betterclever We would like it to be possible to have logs written to file or to an internal buffer or something at some point. Basically: it should be relatively easy for the client to view the logs in real-time, for troubleshooting etc.

Not sure if stderrlog supports this, but if it _only_ supports logging to stderr it might not be adequate. :(

betterclever:
Yeah. That would be an issue. In that case, a solution like log4rs seems
more fit. Let's implement this. Most log modules in Rust follow same log
facade thus switching later, if felt necessary, won't be a cumbersome task.

cmyr:
@betterclever yep that sounds good to me.

cmyr:
@betterclever re: IRC, I think log4rs is fine!

cmyr:
@betterclever do you think you'll get a chance to revisit this at some point?

betterclever:
Yeah. I would this weekend.

PR #546, opened 2018-03-09
Ensure core shuts down after an error

### LINKED ISSUE #434: core becomes non-responsive when receiving unrecognized RPC
Noticed in the process of updating xi-mac to use some more of the new config stuff.

In the main loop the outer thread joins the inner thread before returning, but the inner thread doesn't know that parsing has failed.

This is a bug; at the very least this should print an error and panic.

cmyr:
Previously if an error occured in the main thread, xi-core would hang
indefinitely because it is waiting on the read thread, which doesn't
know about the error.

With this change, we keep track of crashes via an AtomicBool, which
the read thread checks before each new read operation.

Because reads are blocking, core will not exit if an error occurs
_while_ thre read thread is blocked, until the next RPC is received.
This isn't ideal, but is a limitation of blocking I/O.

Fixes #434

### DIFF
--- a/rust/core-lib/tests/rpc.rs
+++ b/rust/core-lib/tests/rpc.rs
@@ -93,7 +93,7 @@ fn test_malformed_json() {
     let mut state = MainState::new();
     let write = io::sink();
     let mut rpc_looper = RpcLoop::new(write);
-    // malformed json, no id: should not receive a response, and connection should close.
+    // malformed json: method should be in quotes.
     let read = make_reader(r#"{method:"client_started","params":{}}
 {"id":0,"method":"new_view","params":{}}"#);
     match rpc_looper.mainloop(|| read, &mut state).err()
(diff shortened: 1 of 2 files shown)


cmyr:
Been putting this off for a while but it's blocking https://github.com/google/xi-mac/pull/151.

vlovich:
Is there a reason not to just shutdown the process non-gracefully?  It seems like that's a simpler strategy than trying to gracefully shutdown & it's not a bad idea in general for processes to assume sudden death is possible at any time, especially for mobile/embedded environments where the OS will actively do this kind of stuff.

cmyr:
@vlovich the problem I was encountering is that even when we panic the main thread we just don't exit at all, because the read thread is blocking.

This also has the advantage that we will correctly return errors over the RPC channel when appropriate, before exiting, that's definitely a minor gain at runtime but does help us write better tests against failure conditions.

PR #549, opened 2018-03-10
Prepare doc comments for pulldown migration

cmyr:
This was motivated by CI failure on nightly; a code example that wasn't being correctly identified under hoedown was being identified and run when parsed with pulldown.


cmyr:
I'm not that familiar with the spec, such as it is, but I had thought that lists had to have a preceding newline? I guess not.

Rebased onto #548, will merge on green.

raphlinus:
I'm referring to the CommonMark spec, which is still not at 1.0 but is expected to be very close. At one point, pulldown was 100% in compliance with the draft spec, but that's drifted as the result of spec changes.

cmyr:
Ah gotcha.

Issue #558, opened 2018-03-15
2018 Spring Cleaning


cmyr:
_tl;dr: a very loosely structured overview of a bunch of mostly non-functional changes planned for the next little while._

# 2018 Spring Cleaning

In anticipation of GSoC and other upcoming work, I would like to do some refactoring over the next few months. Much of this is long planned/previously discussed; I'll link to existing issues where possible.

I'm going to take a fairly high-level view, here; some of these items will need to be fleshed out in more detail in subsequent issues.

## First (preliminary architecture changes)

There are a number of changes I would like to land in the near future, which will pave the way for future work:
  
1. _Global plugins_. (#472) Currently (in the general case) a plugin is bound to a particular view; multiple views which want to use the same plugin each launch their own instance. I would like to make plugins global-by-default, where a single plugin process is expected to handle multiple views.on

2. _Buffer tags and plugin lifespans_. I would like to essentially adopt the model proposed by @trishume in #374. Buffers will be assigned 'tags', which will supercede fixed syntax definitions. Plugins can provide new tags; plugins and the frontend can add/remove tags; plugins can declare the tags that they should be activated for.

2. (2b) _Remove support for non-global plugins?_. This needs a bit more discussion, but I it is unclear to me that the possible advantages of non-global plugins (where each buffer has its own running process) outweigh the increase in API complexity.

## Second (RPC refactor)

At some later point, I would also like to do some work on the RPC library, with the goal of adding support for async I/O where possible, and for handling multiple peers in a single runloop.

### How it works now

Currently, each plugin has its own instance of the runloop, running in a thread; since each runloop starts a distinct thread to handle reading, that means each plugin is currently starting two threads _in addition_ to the plugin process itself.

In addition, since each plugin is running in its own thread, access to buffers has to go through a mutex; this makes a lot of the API around interacting with buffers more complicated than it needs to be.

### Goals

I don't have a coherent design right now, but i do have an outline of what I would like to achieve:

- all connections to peers (client, plugins, possibly others) should be handled by a single runloop.

- where possible, connections to peers should use async APIs under the hood; where that isn't possible, we can use a blocking read thread as a shim.

- All messages should be handled on a single main thread; this removes a bunch of complexity around sharing buffers between threads.

- we should be able to add and remove peers during execution, by passing a handle to the runloop

- the runloop should support timers, so we can do things like wait momentarily for plugin actions to arrive before updating the client

- the runloop should be easy to deploy as a library, and provide FFI bindings

[...]

Issue #472, opened 2018-01-11
Make syntect plugin (and rust plugin lib) 'global'


cmyr:
Currently each document has its own copy of syntect running in a distinct process. This causes a delay before highlighting kicks in on a newly opened document, and increases overhead.

We have the concept of global plugins, but they aren't supported by the rust plugin lib.

A relevant question at this point, though: do we ever really need to support _local_ plugins? @trishume's #374 outlines a reasonable approach to having at most a single process per plugin, which seems reasonable.

Issue #184, opened 2017-04-02
Include line numbers when sending updates from core


cmyr:
Since line wrapping happens in the core, the client needs to be informed whether a line is a real/logical line or part of a wrapped line.

The easiest place to put this is probably in the line structure, which already includes cursor positions and selection information. The most trivial solution might be to just send the logical line number with each line, and the client can decide how to render line numbers. Alternatively, the client could do more of the work, and there could be some optional field on each line, "newline", which if presents instructs the client to increment an internal counter. 

I err towards being more explicit, but I'm not sure how valuable minor optimizations are, here. If they're valuable then a slightly more complicated protocol may be worth slightly more complexity.

raphlinus:
I think the answer is to include line number as effectively an Option<usize> with every line. This presents no problem when rendering new lines. Currently, deciding what to re-render is crude, so that _all_ lines get rendered.

However, I really want to get minimal update (see #152). In that case, when you (say) insert a line, then the line numbers of all lines following that increment by one. I don't want to explicitly send the line numbers for all the lines that otherwise do not change.

I think I've worked out the answer (after struggling with more complex ideas involving maintaining a delta). Essentially, the "copy" operation is augmented to contain an initial line number (technically, the count of present line numbers for all lines preceding), call it `ln`. When copying a line, if a line number is present, then `ln` is incremented, and the line number is replaced with `ln`.

The "update" op could be extended to allow flipping the present state for a line number, but I'm more inclined not to do that, and just require the core to do a "skip" + "ins" with the new line state.

Note also that the update protocol needs to send a maximum line number, for calculation of the gutter width.

raphlinus:
Ok, thinking about it some more, the proposal above _almost_ works, but doesn't handle the case where valid lines in the line cache may be non-contiguous, i.e. there are valid lines, invalid lines, then more valid lines.

That said, what I think will work is sending a delta with each "copy" and "update" operation. In the current implementation, these deltas will always be 0 (because it blows away existing lines and re-renders everything), so as a consequence it would be possible to get a first cut implemented quickly (calculating the line number deltas will simply be an additional responsibility for minimal update when that happens).

PR #560, opened 2018-03-16
Cleanup plugin lib, continued

cmyr:
based off #559; that should be reviewed first.

This is an initial attempt to separate out the document handling portions of the state cache into their own type, which can be used on its own by simple plugins.

This has reminded me of the need for #184; getting that implemented would make our cache implementations much easier. I intend to revisit that now, it seems like a good time.

Eventually this new cache should implement `get_line`, `offset_for_line`, and `line_for_offset`, but this is a good checkpoint.

progress towards #472 and #558


cmyr:
yep bits of this are a bit clunky; i'm focusing on doing the refactor incrementally, with intermediate steps still building / usable. This is definitely one of those slightly ungainly middle steps.

Issue #562, opened 2018-03-19
Change behaviour of get_data plugin RPC


cmyr:
Working around the plugin lib right now, and have some thoughts about this API.

1. `get_data` should allow chunk position to be specified as an offset _or_ a line number; because plugin clients are generally going to be line oriented, I think this would simplify things on the client side with very little cost. Where a line number is used, it would be converted to an offset internally. We would not guarantee that whole lines are returned.

2. Augment return value with line info: `get_data` should return line information in addition to a chunk of the document. In particular, it should return the offset of the chunk, the position in the chunk of newlines (as an array of offsets) and the line number of the first newline in the chunk. It might also be useful to include the number of newlines that exist _after_ the chunk, or the total number of lines in the document, which are informationally equivalent.

Making this change would go along way to simplifying writing a plugin library.

Earlier, I had been imaging augmenting plugin _updates_ to include, with deltas, the number of newlines in any deleted regions. This is only useful when deletions occur outside of the plugin's cache, though; and it might be simpler in that case to just blow away the cache, and know that the next `get_data` request will include the correct line info? This would only be bad in some use case where a plugin was primarily working in one part of a (very) large document, and deletions were occurring frequently in another part of the document, which seems like maybe not a case we need to be too defensive about.

@raphlinus does this make sense?

raphlinus:
First point sounds good. I am definitely swayed by the argument of having one round-trip rather than having two separate ones for resolving line number to offset and then getting the data.

Second point I'm partly convinced. Returning both line number and offset of the returned chunk sounds great, no problem (this is consistent with the way RopeInfo is implemented). Adding data for newlines _within_ the chunk does not sound great to me - it seems likely to me that the cost of serializing, sending, and deserializing this info is more than finding the newlines on the plugin side.

And, for total number of lines in the document, it feels like this should be sent in delta rather than get_data. That way, we avoid states where we have partial information about the number of lines (ie not konwing whether a reference to a particular line is valid). I'd lean toward sending absolute line count, it feels less stateful.

Absolutely agree with the goal of simplification and not trying to optimize for the validity of line number to offset correspondence in the face of large document changes. As you point out, the cache in those cases is probably lukewarm at best.

cmyr:
Okay, good point. As long as we know what the starting line in a chunk is, finding the rest of the offsets is trivial.

Implementation plan:

- `get_data` has fields `{ offset: Option<usize>, line: Option<usize>, rev: usize, max_size: usize }`
[...]

PR #568, opened 2018-03-21
get data + finish initial base_cache

cmyr:
Implements the changes proposed in #562, using them as the basis of a simple line-based cache.

It feels like a bunch of this was more an exercise than anything, but I'll be building on this over the rest of the week to roll out global plugin support.


rust/core-lib/src/plugins/rpc.rs
```diff
@@ -219,11 +244,42 @@ impl PluginBufferInfo {
...
+                } else {
+                    if text.is_codepoint_boundary(offset) {
+                        offset.into()
+                    } else {
+                        text.prev_codepoint_offset(offset).into()
+                    }
+                }
+            }
+            TextUnit::Line => {
+                let num_lines = text.measure::<LinesMetric>() + 1;
```
raphlinus: I would either name this "max_line_number", or move the `+ 1` to the next line, probably preferring the latter.

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -238,6 +294,7 @@ mod tests {
             "view_id": "view-id-42",
             "delta": {"base_len": 6, "els": [{"copy": [0,5]}, {"insert":"rofls"}, {"copy": [5,6]}]},
             "new_len": 11,
+            "new_line_count": 1,
```
raphlinus: If we change the "+ 1" logic, this will need to be updated.

rust/plugin-lib/src/base_cache.rs
```diff
@@ -27,49 +30,270 @@ const CHUNK_SIZE: usize = 1024 * 1024;
...
+    pub fn offset_of_line<DS>(&mut self, source: &DS, line_num: usize) -> Result<usize, Error>
+        where DS: DataSource
+    {
+        if line_num > self.num_lines { return Err(Error::BadRequest) }
+        match self.cached_offset_of_line(line_num) {
+            Some(offset) => Ok(offset + self.offset),
+            None => {
+                let resp = source.get_data(line_num, TextUnit::Line, CHUNK_SIZE, self.rev)?;
+                self.reset_chunk(resp);
+                self.offset_of_line(source, line_num)
```
raphlinus: My preference would be a loop rather than recursion, but it's not a biggie either way.

rust/plugin-lib/src/base_cache.rs
```diff
@@ -27,49 +30,270 @@ const CHUNK_SIZE: usize = 1024 * 1024;
...
+    /// it is present in the chunk.
+    fn cached_offset_of_line(&self, line_num: usize) -> Option<usize> {
+        if line_num < self.first_line { return None }
+
+        let rel_line_num = line_num - self.first_line;
+
+        if rel_line_num == 0 && self.first_line_offset == 0 {
+            return Some(0)
+        }
+        if rel_line_num <= self.line_offsets.len() {
```
raphlinus: So we've got a problem if rel_line_num is 0 and first_line_offset is not 0; this code will panic because the `- 1` will underflow. We actually know the answer; it's `-self.first_line_offset`, but that won't fit in a `usize` as it's negative.

rust/plugin-lib/src/base_cache.rs
```diff
@@ -27,49 +30,270 @@ const CHUNK_SIZE: usize = 1024 * 1024;
...
+        // first find out if offset is a line offset, and set first_line / first_line_offset
+        let (new_line, new_line_off) = match self.line_offsets.binary_search(&offset) {
+            Ok(idx) => (self.first_line + idx + 1, 0),
+            Err(0) => (self.first_line, self.first_line_offset + offset),
+            Err(idx) => (self.first_line + idx, offset - self.line_offsets[idx - 1]),
+        };
+
+        // then clear line_offsets up to and including offset
+        self.line_offsets = self.line_offsets.iter()
+            .filter(|i| **i > offset)
```
raphlinus: Odd that this requires two `*`'s and the following none, but maybe I just don't understand the types...

rust/plugin-lib/src/base_cache.rs
```diff
@@ -27,49 +30,270 @@ const CHUNK_SIZE: usize = 1024 * 1024;
...
+            true
+        }
+    }
+
+    /// Patches up `self.line_offsets` in the simple insert case.
+    fn simple_insert(&mut self, text: &Rope, ins_offset: usize) {
+        let has_newline = text.measure::<LinesMetric>() > 0;
+        let self_off = self.offset;
+        assert!(ins_offset >= self_off);
+        //let ins_offset = ins_offset + self.offset;
```
raphlinus: This comment should probably be removed.

rust/core-lib/src/plugins/rpc.rs
cmyr: Again my understanding is that to actually get the number of lines in the rope I need the + 1? Seems like I'm probably confused.

rust/plugin-lib/src/base_cache.rs
cmyr: Principally for performance reasons? My motivation here was that we know we can only ever recurse once, and felt this was the elegant solution. Would be curious to know your reasoning though.

rust/plugin-lib/src/base_cache.rs
cmyr: good catch. fixed this up and added a test case.

rust/plugin-lib/src/base_cache.rs
cmyr: Indeed. As far as I understand it, the issue is mostly that `usize - &size` is okay but `usize > &usize` isn't.

rust/core-lib/src/editor.rs
```diff
@@ -438,6 +438,7 @@ impl Editor {
         let keep_selections = self.this_edit_type == EditType::Transpose;
         self.scroll_to = self.view.after_edit(&self.text, &last_text, &delta, is_pristine, keep_selections);
         let (iv, new_len) = delta.summary();
+        let new_line_count = self.text.measure::<LinesMetric>() + 1;
```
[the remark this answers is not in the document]
cmyr: This may be a naming problem? this is supposed to represent "new count of total lines". `LinesMetric` returns 0 if there are no newline characters in the rope, although in this case the number of lines is still 1.

rust/plugin-lib/src/base_cache.rs
```diff
@@ -69,13 +69,13 @@ impl ChunkCache {
         }
 
         // We now know that the start of this line is contained in self.contents.
-        let mut start_off = self.cached_offset_of_line(line_num).unwrap();
+        let mut start_off = self.cached_offset_of_line(line_num).unwrap() - self.offset;
 
         // Now we make sure we also contain the end of the line, fetching more
         // of the document as necessary.
         loop {
             if let Some(end_off) = self.cached_offset_of_line(line_num + 1) {
```
raphlinus: I think I'd prefer the `- self.offset` on both here, it feels asymmetrical to have different coordinates for `start_off` and `end_off`.

cmyr: it's on the next line? Do you mean you'd like it declared before use?

cmyr: I do see another bug here though, in comparing start_off to 0 in the next block

raphlinus: Sorry if my comment was confusing. Just I'd like to start_off and end_off have the same units, either by subtracting off `self.offset` from both at the time they're bound, or neither, and have that applied at time of use. I see how doing the first is more than a little annoying because of the `if let`.

PR #569, opened 2018-03-22
rust global plugin lib [WIP]

### LINKED ISSUE #472: Make syntect plugin (and rust plugin lib) 'global'

cmyr:
This is based off of #568, which should be resolved first.

This isn't quite usable at the moment, but I wanted to PR in case there was feedback on the overall approach.

This introduces essentially a new plugin lib, rooted in `plugin-lib/src/global`. This new lib allows for a single plugin to manage events in multiple views.

Todos:
- [ ] break `StateCache` out into a stand alone component (without breaking existing lib)
- [ ] implement the new `Cache` trait for `StateCache` and `ChunkCache`
- [ ] implement outstanding features from current lib (`request_is_pending`, maybe a bit more)
- [ ] implement syntect on top of new API

I think (famous last words) most of the tricky stuff is out of the way, so I hope to have the rest of this work done by the weekend.

- closes #472 
- progress on #558

### DIFF
--- a/rust/plugin-lib/src/global/mod.rs
+++ b/rust/plugin-lib/src/global/mod.rs
@@ -1,0 +1,17 @@
+mod view;
+
+use std::collections::HashMap;
+
+use xi_core::{ViewIdentifier, PluginPid, plugin_rpc};
+use xi_rpc::{self, RpcLoop, RpcCtx, RemoteError, ReadError, Handler as RpcHandler};
+use self::view::{Plugin, View};
+
+/// Handles raw RPCs from core, updating documents and bridging calls
+/// to the plugin,
+pub struct Dispatcher<'a, P: 'a + Plugin> {
+    //TODO: when we add multi-view, this should be an Arc+Mutex/Rc+RefCell
+    views: HashMap<ViewIdentifier, View<P::Cache>>,
+    pid: PluginPid,
+    plugin: &'a mut P,
+}
+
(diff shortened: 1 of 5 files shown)


rust/plugin-lib/src/global/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn clear(&mut self);
+}
+
+/// An interface for plugins.
+///
+/// Users of this library must implement this trait for some type.
+pub trait Plugin {
+    type Cache: Cache;
+
+    //TODO: async edits only; this is here for feature paritiy during initial hacking
```
raphlinus: parity

rust/plugin-lib/src/global/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Called when an edit has occured in the remote view. If the plugin wishes
```
raphlinus: I think we wanted to move away from this, in favor of plugin edits being async, so not sure why I'm seeing it in new code. Perhaps it's related to the fact we'll need more sophisticated CRDT so a sequence of async edits appear to the plugin as if it's editing its own copy of the doc.

I guess this is more of a repackaging of existing stuff, so I'm willing to be convinced those changes should happen separately.

PR #570, opened 2018-03-23
Remove old Rope impl, migrate tests to src/rope.rs

cmyr:
There's more work to do here for sure, but this is a small step in the
right direction.


cmyr:
Might be worth figuring out what we want to re-export while we're making the change. Also lots more docs to write. But: progress.

cmyr:
I'm going to reopen this from a branch on `origin`.

PR #573, opened 2018-03-23
global syntect

cmyr:
based off of #569.

This should... just work?

Known issues:

- the idle handler has no concept of priority, so if a big file is being
processed in the background text entered in another buffer can end up
not being processed immediately. In some instances it may not be
processed at all.


rust/syntect-plugin/src/main.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All rights reserved.
+// Copyright 2018 Google Inc. All rights reserved.
```
raphlinus: nit: we generally don't update years, it's the date of original creation

rust/syntect-plugin/src/main.rs
```diff
@@ -273,54 +296,70 @@ impl<'a> PluginState<'a> {
...
-    type State = State;
 
-    fn initialize(&mut self, ctx: PluginCtx<State>, _buf_size: usize) {
-        self.do_highlighting(ctx);
+impl<'a> Plugin for Syntect<'a> {
+    type Cache = StateCache<LineState>;
+
+    fn new_view(&mut self, view: &mut View<Self::Cache>) {
+        let _t = trace_block("Syntect::new_view", &["syntect"]);
+        eprintln!("added view {:?}", view.get_id());
```
raphlinus: there's random logging in the original, and this seems very relevant, but at some point we probably want to make this more systematic

PR #575, opened 2018-03-25
Remove old Rope impl, migrate tests to src/rope.rs

cmyr:
There's more work to do here for sure, but this is a small step in the
right direction.


cmyr:
Opened #583 as a note about the possible optimizations.

PR #576, opened 2018-03-25
Use Range instead of `start` & `end` in Rope impl

### LINKED ISSUE #571: Modify various Rope iter-returning methods to use range arguments
There are a number of methods implemented on `Rope` that return iterators. Currently these methods all take `start` and `end` arguments for specifying the range of the rope to iterate over. It would be more ergonomic if this could be a `Range` argument instead. For a guide, you can check out [`Vec::drain`](https://doc.rust-lang.org/1.24.1/src/alloc/vec.rs.html#1089) in the stdlib.

This will want to wait until #570 has been merged; part of this work can be removing the `_all` variants of these iterators that were added there.

This will also involve fixing various consumers of the existing API.

LiHRaM:
Closes #571 
Related to #575 

Realized that `slice_to_string` isn't a function that returns an iterator, as per the request in #571 - maybe I should change that back?

Hoping I'm on the right track here. :)
cc @cmyr 

---
TODO:
- [x] Change functions returning iterators, so they now use `RangeBounds` instead of `start` and `end`
- [x] Update tests
- [x] Update external calls to changed functions
- [x] Wait for RangeBounds to become stable

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -602,7 +602,7 @@ impl Editor {
 
     fn sel_region_to_interval_and_rope(&self, region: SelRegion) -> (Interval, Rope) {
         let as_interval = Interval::new_closed_open(region.min(), region.max());
-        let interval_rope = Rope::from(self.text.subseq(as_interval));
+        let interval_rope = self.text.subseq(as_interval);
         (as_interval, interval_rope)
     }
 
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -424,7 +424,7 @@ impl Editor {
         let mut saved = None;
         for region in sel_regions {
             if !region.is_caret() {
-                let val = self.text.slice_to_string(region.min(), region.max());
+                let val = self.text.slice_to_string(region);
                 match saved {
                     None => saved = Some(val),
                     Some(ref mut s) => {
@@ -519,8 +519,7 @@ impl Editor {
             let tab_offset = view.line_col_to_offset(&self.text, line,
                                                      tab_text.len());
             let interval = Interval::new_closed_open(offset, tab_offset);
-            let leading_slice = self.text.slice_to_string(interval.start(),
-                                                          interval.end());
+            let leading_slice = self.text.slice_to_string(interval.start()..interval.end());
             if leading_slice == tab_text {
                 builder.delete(interval);
             } else if let Some(first_char_col) = leading_slice.find(|c: char| !c.is_whitespace()) {
@@ -603,8 +602,7 @@ impl Editor {
 
     fn sel_region_to_interval_and_rope(&self, region: SelRegion) -> (Interval, Rope) {
         let as_interval = Interval::new_closed_open(region.min(), region.max());
-        let interval_rope = Rope::from(self.text.slice_to_string(
-            as_interval.start(), as_interval.end()));
+        let interval_rope = Rope::from(self.text.subseq(as_interval));
         (as_interval, interval_rope)
     }
 
@@ -624,8 +622,8 @@ impl Editor {
                 if let Some(end) = self.text.next_grapheme_offset(middle) {
                     if start >= last {
                         let interval = Interval::new_closed_open(start, end);
-                        let swapped = self.text.slice_to_string(middle, end) +
-                                      &self.text.slice_to_string(start, middle);
+                        let swapped = self.text.slice_to_string(middle..end) +
+                                      &self.text.slice_to_string(start..middle);
                         builder.replace(interval, Rope::from(swapped));
                         last = end;
                     }
@@ -677,8 +675,7 @@ impl Editor {
         let mut builder = delta::Builder::new(self.text.len());
 
         for region in view.sel_regions() {
-            let selected_text = self.text.slice_to_string(region.min(),
-                                                          region.max());
+            let selected_text = self.text.slice_to_string(region);
             let interval = Interval::new_closed_open(region.min(), region.max());
             builder.replace(interval, Rope::from(transform_function(&selected_text)));
         }
@@ -704,7 +701,7 @@ impl Editor {
             let mut cursor = Cursor::new(&self.text, line_start);
 
             if let Some(line_end) = cursor.next::<LinesMetric>() {
-                let duplicated_text = self.text.slice_to_string(line_start, line_end);
+                let duplicated_text = self.text.slice_to_string(line_start..line_end);
                 let iv = Interval::new_closed_open(line_start, line_start);
 
                 // last line does not have new line character so it needs to be manually added
@@ -814,7 +811,7 @@ impl Editor {
             end_off = text.prev_codepoint_offset(end_off + 1).unwrap();
         }
 
-        let chunk = text.slice_to_string(offset, end_off);
+        let chunk = text.slice_to_string(offset..end_off);
         let first_line = text.line_of_offset(offset);
         let first_line_offset = offset - text.offset_of_line(first_line);
 
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -211,7 +211,7 @@ fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding)
             CharacterEncoding::Utf8 => (),
         }
 
-        for chunk in text.iter_chunks(0, text.len()) {
+        for chunk in text.iter_chunks(..text.len()) {
             f.write_all(chunk.as_bytes())?;
         }
         Ok(())
--- a/rust/core-lib/src/selection.rs
+++ b/rust/core-lib/src/selection.rs
@@ -350,14 +350,14 @@ impl SelRegion {
     }
 }
 
-// Returns `[start..end)`
+// Returns `[min..max)`
 impl<'a> RangeBounds<usize> for &'a SelRegion {
     fn start_bound(&self) -> Bound<&usize> {
-        Bound::Included(&self.start)
+        Bound::Included(min(&self.start, &self.end))
     }
     
     fn end_bound(&self) -> Bound<&usize> {
-        Bound::Excluded(&self.end)
+        Bound::Excluded(max(&self.start, &self.end))
     }
 }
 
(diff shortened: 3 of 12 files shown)


cmyr:
Interesting, I hadn't noticed that this wasn't stabilized. I think in that case we should probably wait for it to land. :(

cmyr:
So let's hold this now, waiting for https://github.com/rust-lang/rust/issues/30877

cmyr:
Okay, I believe this is on track to stabilize soon; let's wait until that happens, and then can give this a proper review. :)

LiHRaM:
Fairly sure this was stabilized in 1.28 - any thoughts on when we'll be moving on to that version of rust for the CI?

cmyr:
@LiHRaM good question. If this is stable in 1.28 that's a reasonably good argument to bump shortly; and then our next major bump will probably be for the rust 2018 release in early december? I don't believe there's any blocking reason for us to stay on 1.26, so if going to 1.28 would get this merged then that sounds good to me. @raphlinus?

rust/core-lib/src/editor.rs
```diff
@@ -423,7 +423,7 @@ impl Editor {
         let mut saved = None;
         for region in sel_regions {
             if !region.is_caret() {
-                let val = self.text.slice_to_string(region.min(), region.max());
+                let val = self.text.slice_to_string(region.min()..region.max());
```
cmyr: I wonder if we can just implement `RangeBounds` for `SelRegion`, and then we'll be able to pass it in to things like this directly?

rust/core-lib/src/file.rs
```diff
@@ -211,7 +211,7 @@ fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding)
             CharacterEncoding::Utf8 => (),
         }
 
-        for chunk in text.iter_chunks(0, text.len()) {
+        for chunk in text.iter_chunks(0..text.len()) {
```
cmyr: when we go to `RangeBounds` this can just be `..text.len()`.

rust/rope/src/rope.rs
```diff
@@ -571,33 +572,33 @@ impl Rope {
...
         ChunkIter {
-            cursor: Cursor::new(self, start),
-            end,
+            cursor: Cursor::new(self, range.start),
+            end: range.end,
         }
     }
 
     //TODO: implement iter_chunks using ranges and delete this
     pub fn iter_chunks_all(&self) -> ChunkIter {
```
cmyr: as part of this work we can remove this fn, since `iter_chunks(..)` will let us iter all chunks easily.. (see the TODO comment above)

rust/rope/src/rope.rs
```diff
@@ -633,15 +637,42 @@ impl Rope {
...
         let mut result = String::new();
-        for chunk in self.iter_chunks(start, end) {
+        for chunk in self.iter_chunks(range) {
             result.push_str(chunk);
         }
         result
     }
 }
 
+// Helper methods for RangeBounds implementation
```
LiHRaM: This is what I'm a little concerned about. I hope it's idiomatic enough. Would love suggestions if it doesn't seem to fit in right.

rust/rope/src/rope.rs
```diff
@@ -726,7 +757,7 @@ impl From<Rope> for String {
 
 impl<'a> From<&'a Rope> for String {
     fn from(r: &Rope) -> String {
-        r.slice_to_string(0, r.len())
+        r.slice_to_string(0..r.len())
```
LiHRaM: Lots of little examples like this that can be changed to `slice_to_string(..)` now, I might add that in this PR, what do you think @cmyr?

rust/core-lib/src/selection.rs
```diff
@@ -16,6 +16,9 @@
 
 use std::cmp::{min, max};
 use std::ops::Deref;
+use std::ops::Bound;
+use std::ops::Bound::*;
```
cmyr: I prefer to avoid glob imports of enums. If you really need to use them a bunch somewhere where they're annoying or cluttery to type out, I'll do something like `use std::ops::Bound as B` or something, in the scope of the particular function where I need it.

I feel this might be kind of weird of me, so 🤷‍♂️. I think it's okay to just use things like `Bound::Included(_)` directly.

rust/core-lib/src/view.rs
```diff
@@ -940,13 +940,13 @@ impl View {
         let search_query = match self.selection.last() {
             Some(region) => {
                 if !region.is_caret() {
-                    text.slice_to_string(region.min(), region.max())
+                    text.slice_to_string(region.min()..region.max())
```
cmyr: can't we just pass in `region` now? A few other places, too..

rust/rope/src/find.rs
```diff
@@ -316,7 +316,7 @@ mod tests {
     fn find_small() {
         let a = Rope::from("Löwe 老虎 Léopard");
         let mut c = Cursor::new(&a, 0);
-        let mut raw_lines = a.lines_raw(0, a.len());
+        let mut raw_lines = a.lines_raw(0..a.len());
```
cmyr: no `0` needed.

rust/rope/src/rope.rs
```diff
@@ -490,7 +495,12 @@ impl Rope {
     }
 
     /// Returns a slice of the string from the byte range [`start`..`end`).
-    pub fn slice(&self, start: usize, end: usize) -> Rope {
+    pub fn slice<T>(&self, range: T) -> Rope 
+        where T: RangeBounds<usize>
+    {
+        let start = self.get_start(&range);
+        let end = self.get_end(&range);
```
cmyr: seeing this everywhere makes me wonder if we'd rather have a fn,

```rust
    self.extract_range(range: T) -> (usize, usize) { .. }
```

and then you could have,
```rust
let (start, end) = self.extract_range(&range);
```

rust/rope/src/rope.rs
cmyr: I think this is fine. As mentioned before, you could also consider having a single fn that returns both start and end.

rust/rope/src/rope.rs
cmyr: ah yea, saw that earlier, I think that change makes sense where appropriate.

cmyr:
@LiHRaM something else before this gets merged: we'll need to update the .travis.yml file to build with rust 1.28, and we'll need to update the readme to note the new minimum version.

LiHRaM:
Alright. Pretty sure I've made all changes as requested. I think it's ready to merge?

@cmyr

rust/core-lib/src/editor.rs
```diff
@@ -424,7 +424,7 @@ impl Editor {
         let mut saved = None;
         for region in sel_regions {
             if !region.is_caret() {
-                let val = self.text.slice_to_string(region.min(), region.max());
+                let val = self.text.slice_to_string(region);
```
cmyr: this is cool. 😎

rust/core-lib/src/editor.rs
```diff
@@ -604,7 +603,7 @@ impl Editor {
     fn sel_region_to_interval_and_rope(&self, region: SelRegion) -> (Interval, Rope) {
         let as_interval = Interval::new_closed_open(region.min(), region.max());
         let interval_rope = Rope::from(self.text.slice_to_string(
-            as_interval.start(), as_interval.end()));
+            as_interval.start()..as_interval.end()));
```
cmyr: we should be able to just use `region` here.

cmyr: Also there's another little issue here, which is that we should be using `Rope::subseq` instead of `Rope::slice_to_string` here; subseq will clone existing nodes where possible. This might be out of scope for this PR, though.

rust/rope/src/rope.rs
```diff
@@ -490,7 +494,11 @@ impl Rope {
     }
 
     /// Returns a slice of the string from the byte range [`start`..`end`).
```
cmyr: this doc is now out of date. maybe just, 'Returns a new `Rope` with the contents of the provided range.`

rust/rope/src/rope.rs
```diff
@@ -478,7 +480,9 @@ impl Rope {
     /// Note: `edit` and `edit_str` may be merged, using traits.
     ///
     /// Time complexity: O(log n)
-    pub fn edit_str(&mut self, start: usize, end: usize, new: &str) {
+    pub fn edit_str<T>(&mut self, range: T, new: &str) where T: RangeBounds<usize> {
```
cmyr: Nit: I would consistently put the `where` clause on the next line.

rust/rope/src/rope.rs
LiHRaM: :+1:

rust/rope/src/rope.rs
LiHRaM: :+1:

rust/core-lib/src/editor.rs
LiHRaM: `subseq` sounds good to me.

rust/core-lib/src/editor.rs
raphlinus: It would be, but I'm unsure about something. What if the region is "reversed," which can happen for example by pressing shift-left-arrow?

rust/core-lib/src/selection.rs
```diff
@@ -348,6 +350,17 @@ impl SelRegion {
     }
 }
 
+// Returns `[start..end)`
+impl<'a> RangeBounds<usize> for &'a SelRegion {
+    fn start_bound(&self) -> Bound<&usize> {
+        Bound::Included(&self.start)
```
raphlinus: As I mentioned before, I'm not sure this is the right behavior, as start > end is allowed in a SelRegion (it means the region is reversed), but I'm expecting violates expectations of a RangeBounds. I could be wrong about this though.

cmyr: Good point; this should be resolved by making the start bound should be `self.start.min(self.end)`, and doing the inverse for the end bound?

LiHRaM: Yeah, implementing `start_bound` as `Bound::Included(min(&self.start, &self.end))` works just fine.

Issue #444, opened 2017-11-28
Ctrl-DoubleClick replaces selection instead of adding


WesleyAC:
Not sure if this is a bug in xi-gtk or xi-core, but holding control while double-clicking selects the word and removes other cursors - IMO it should add a selection (this is sublime's behaviour). There is the same issue with triple-click.

Filing this on xi-core since it looks like all the clicking logic is in `do_click` in `editor.rs`, but I'm not sure if the frontend is supposed to send a different RPC on ctrl-click.

raphlinus:
It straddles both. What should happen is that this mouse gesture gets migrated from `do_click` to `do_gesture`, and that `do_gesture` grows to be able to handle it. Actually I'd like to migrate _all_ gestures off of `do_click` but have been lazy.

This feels like a good thing to work on :)

wuweiweiwu:
@cmyr @raphlinus On it!

rscircus:
@wuweiweiwu did you start working on this?

Else, I'd like to collaborate and/or take over here, as suggested in https://twitter.com/cmyr/status/975728008546963456 :-)

Meanwhile, I'll try to grok @raphlinus suggestion to migrate all gestures off of `do_click`.

wuweiweiwu:
@rscircus Please do! I've been swamped at work unfortunately :(

rscircus:
Alright, onboarding. I'll come back to you, when this hits review-time maybe? :yum:

scholtzan:
Oh, I added a fix for this issue without seeing that you (@rscircus) wanted to work on this issue. Sorry! In case you already started working on this, feel free to give feedback and suggestions to my pull-request. I think there might be room for improvements.

rscircus:
Haha, @scholtzan, actually I grabbed this issue from somebody else already. :) Now you just finished it off. Actually, I'm midway in solving it. Had to fix some issues with xi-gtk first as my platform didn't like anything and I'm learning Rust and thought this to be a nice motivation. Having said that, I'm maybe not the best reviewer. I'll have a look at your solution and would be happy if you answer my questions?! Maybe that helps in some kind of pair programming way...

scholtzan:
That sounds good! I'd be happy to answer your questions.

rscircus:
Alright :) I added a few remarks in your PR without starting a review.

cmyr:
@rscircus if you've already been working in this area, we do still have the related issue of migrating the `do_click` RPC to be a gesture, if you'd like to work on that?

PR #585, opened 2018-03-29
Add gestures for selecting multiple lines and multiple words

scholtzan:
This should fix https://github.com/google/xi-editor/issues/444

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -867,28 +867,10 @@ impl Editor {
                 return;
             }
         } else if click_count == 2 {
-            let (start, end) = {
-                let mut word_cursor = WordCursor::new(&self.text, offset);
-                word_cursor.select_word()
-            };
-            self.set_sel_single_region(SelRegion{
-                start: start,
-                end: end,
-                horiz: None,
-                affinity: Affinity::default(),
-            });
-            self.view.start_drag(offset, start, end);
+            self.view.select_word(&self.text, offset, false);
             return;
         } else if click_count == 3 {
-            let start = self.view.line_col_to_offset(&self.text, line as usize, 0);
-            let end = self.view.line_col_to_offset(&self.text, line as usize + 1, 0);
-            self.set_sel_single_region(SelRegion{
-                start: start,
-                end: end,
-                horiz: None,
-                affinity: Affinity::default(),
-            });
-            self.view.start_drag(offset, start, end);
+            self.view.select_line(&self.text, offset, line as usize, false);
             return;
         }
         self.view.start_drag(offset, offset, offset);
@@ -904,6 +886,8 @@ impl Editor {
         let offset = self.view.line_col_to_offset(&self.text, line as usize, col as usize);
         match ty {
             GestureType::ToggleSel => self.view.toggle_sel(&self.text, offset),
+            GestureType::MultiLineSelect => self.view.select_line(&self.text, offset, line as usize, true),
+            GestureType::MultiWordSelect => self.view.select_word(&self.text, offset, true)
         }
     }
 
--- a/rust/core-lib/src/rpc.rs
+++ b/rust/core-lib/src/rpc.rs
@@ -289,6 +289,8 @@ pub struct EditCommand<T> {
 #[serde(rename_all = "snake_case")]
 pub enum GestureType {
     ToggleSel,
+    MultiLineSelect,
+    MultiWordSelect,
 }
 
 /// An inclusive range.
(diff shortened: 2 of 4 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -904,6 +886,8 @@ impl Editor {
         let offset = self.view.line_col_to_offset(&self.text, line as usize, col as usize);
         match ty {
             GestureType::ToggleSel => self.view.toggle_sel(&self.text, offset),
+            GestureType::MultiLineSelect => self.view.select_line(&self.text, offset, line as usize, true),
+            GestureType::MultiWordSelect => self.view.select_word(&self.text, offset, true)
```
rscircus: Would it be more consistent to name these `GestureType`s to:
```
SelectLine
SelectWord
```?

And here comes the newbie question:
Why is this necessary at all?

File: rust/core-lib/src/rpc.rs [hunk 2]
```diff
@@ -289,6 +289,8 @@ pub struct EditCommand<T> {
 #[serde(rename_all = "snake_case")]
 pub enum GestureType {
     ToggleSel,
+    MultiLineSelect,
+    MultiWordSelect,
```
rscircus: See above.

rust/core-lib/src/view.rs
```diff
@@ -242,6 +243,55 @@ impl View {
...
+        let mut selection = Selection::new();
+
+        if multi_select {
+            selection = self.selection.clone();
+        }
+
+        selection.add_region(region);
+        self.set_selection(text, selection);
+
+        self.start_drag(offset, start, end);
```
rscircus: So basically, 260-269 is the code to create the multi-selection.

Where is the `multi_select` flag set?

What would be a better option?

1. Create a function for this (which is repeated in the other function) and and add it to the original `editor.rs`. Or
2. Leave it as it is.
3. Create a function for the repeated identical part?

I'm wondering here...

cmyr: @rscircus your intuition is correct, the client sends a different message to core depending on whether or not ctrl/cmd is present during the click.

@scholtzan I do agree that there's possibly room for some code reuse, but I think this is fine as-is. Most of the ways this could be cleaner would require some api additions around `Selection` and `SelRegion`.

rust/core-lib/src/editor.rs
scholtzan: Regarding the why: I added support for multi selection of lines and words in xi-mac (see this pull-request: https://github.com/google/xi-mac/pull/159). So every time the gestures are executed in xi-mac (either by pressing <kbd>CMD</kbd> and clicking twice or three times) then xi-mac sends the RPCs to the core. The RPC messages contain as gesture type `multi_line_select` or `multi_word_select`. For message handling xi-editor will eventually deserialize the messages from JSON and `multi_line_select` will become `MultiLineSelect` and `multi_word_select` will become `MultiWordSelect`. Depending on the gesture type, which is determined at this specific point in the code, the corresponding actions (`self.view.select_line` or `self.view.select_word`) are performed. 

Regarding the name, these events are triggered when users press <kbd>CMD</kbd> + double click or <kbd>CMD</kbd> + triple click (on OSX <kbd>CMD</kbd>, on other platforms maybe <kbd>CTRL</kbd>). So I thought when users perform these gestures their intent is specifically to select multiple words/lines. 
`self.view.select_word` and `self.view.select_line` are also used for selecting single lines/words in `do_click`. The last parameter of these functions indicates whether multiple or single lines/words can be selected.

rust/core-lib/src/view.rs
```diff
@@ -242,6 +243,55 @@ impl View {
...
+            end: end,
+            horiz: None,
+            affinity: Affinity::default(),
+        };
+
+        let mut selection = Selection::new();
+
+        if multi_select {
+            selection = self.selection.clone();
+        }
```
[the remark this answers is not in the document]
scholtzan: @cmyr Yes, the second one looks much better. Thanks for pointing that out. I updated it in my pull request.

rust/core-lib/src/view.rs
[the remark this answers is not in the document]
scholtzan: @cmyr If the `start_drag` is not there then the behaviour is a little weird. For example, if I add a word to my selection by pressing <kbd>CMD</kbd> + double click then the entire word gets selected. But the moment I start dragging, the part of the word before the location where I clicked gets deselected again. So I'd say `start_drag` should stay here.

rust/core-lib/src/editor.rs
rscircus: @scholtzan thanks for the detailed explanation. Now it makes perfectly sense. I'll add the corresponding gestures to xi-gtk. :)

PR #586, opened 2018-03-30
Implemented indent/outdent

nangtrongvuon:
Added an indent and outdent selector that uses the user's tab settings to insert and delete the tab character respectively.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -755,8 +755,9 @@ impl Editor {
             let offset = self.view.line_col_to_offset(&self.text, line, 0);
             let interval = Interval::new_closed_open(offset, offset);  
             builder.replace(interval, Rope::from(tab_text));
-            self.this_edit_type = EditType::InsertChars;
+          
         }      
+        self.this_edit_type = EditType::InsertChars;
         self.add_delta(builder.build());   
     }
 
@@ -774,8 +775,8 @@ impl Editor {
                 let interval = Interval::new_closed_open(offset, first_char_offset);
                 builder.delete(interval);
             }
-            self.this_edit_type = EditType::Delete;
         }
+        self.this_edit_type = EditType::Delete;
         self.add_delta(builder.build());   
     }
 
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -759,8 +759,8 @@ impl Editor {
             for line in first_line..last_line {
                 let tab_offset = self.view.line_col_to_offset(&self.text, line, tab_text.len());
                 let offset = self.view.offset_of_line(&self.text, line);
-                let interval = Interval::new_closed_open(offset, tab_offset);
-                if self.text.slice_to_string(interval.start(), interval.end()).contains(tab_text) {
+                let interval = Interval::new_closed_open(offset, tab_offset);         
+                if self.text.slice_to_string(interval.start(), interval.end()).starts_with(tab_text.chars().next().unwrap())  {
                     builder.delete(interval);
                 }
             }
@@ -1104,8 +1104,10 @@ impl Editor {
     }
 
     fn do_cancel_operation(&mut self) {
-        self.view.unset_find(&self.text);
         self.view.collapse_selections(&self.text);
+        self.view.unset_find(&self.text);
+        self.view.set_dirty(&self.text);
+        self.render();
     }
 
     fn transform_text<F: Fn(&str) -> String>(&mut self, transform_function: F) {
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -755,13 +755,13 @@ impl Editor {
                 last_line
             } else {
                 last_line + 1
-            };    
+            };
             for line in first_line..last_line {
-                let tab_offset = self.view.line_col_to_offset(&self.text, line, tab_text.len());             
+                let tab_offset = self.view.line_col_to_offset(&self.text, line, tab_text.len());
                 let offset = self.view.offset_of_line(&self.text, line);
-                let interval = Interval::new_closed_open(offset, tab_offset);  
+                let interval = Interval::new_closed_open(offset, tab_offset);
                 if self.text.slice_to_string(interval.start(), interval.end()).contains(tab_text) {
-                    builder.delete(interval); 
+                    builder.delete(interval);
                 }
             }
         }
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -723,7 +723,6 @@ impl Editor {
     /// Tries to have behavior consistent with other editors like Atom, Sublime and VSCode,
     /// with non-caret selections not being modified.
     fn modify_indent(&mut self, direction: IndentDirection) {
-        let mut builder = delta::Builder::new(self.text.len());
         let mut lines = BTreeSet::new();
         let tab_text = if self.config.items.translate_tabs_to_spaces {
                 let tab_size = self.config.items.tab_size;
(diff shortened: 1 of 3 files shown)


cmyr:
Very cool, thanks for this!

Having trouble focusing enough for a full review, but some initial observations:

1. This can get a companion PR in `xi-mac` that includes support for these two commands. These should be new menu items, with new default hotkeys. You can look at existing editors to decide what hotkeys make sense, but my default would be cmd+[ and cmd+]; items can be in the edit menu.

2. In the case where we're decreasing the indent, we should decrease it even if the current indent is < then the indent size. That is, if the line is indented 3 spaces and the indent size is 4 spaces, we should still indent it to zero.

3. It looks like these two functions share a lot of code. It would be nice if we could refactor this so that they share code, if possible.

This is awesome though, this is a very nice feature that I'm excited to use. :)

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,62 @@ impl Editor {
...
+        for region in self.view.sel_regions() {
+            let (first_line, _) = self.view.offset_to_line_col(&self.text, region.min());
+            let (last_line, last_col) =
+                self.view.offset_to_line_col(&self.text, region.max());
+            let last_line = if last_col == 0 && last_line > first_line {
+                last_line
+            } else {
+                last_line + 1
+            };    
+            for line in first_line..last_line {
```
cmyr: There's a bug here: the same line can have multiple regions, and in this case you will change the indentation once per region, so multiple times for those lines.

I'm not sure what the best approach is. It is guaranteed that regions are in order, so you should be able to keep track of the last line in the last region, and skip that line if it occurs in the next region? Or you could build a set of lines, and then iterate over them.

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,62 @@ impl Editor {
...
+                let offset = self.view.offset_of_line(&self.text, line);
+                match direction {
+                    IndentDirection::In => {
+                        let interval = Interval::new_closed_open(offset, offset);  
+                        builder.replace(interval, Rope::from(tab_text));
+                        self.this_edit_type = EditType::InsertChars;
+                    },
+                    IndentDirection::Out => {
```
cmyr: this is also starting to drift pretty far to the right. This will be quite a better better if you end up just having a collection of lines and you iterate them, but I might also consider just adding `indent_line` and `outdent_line` fns that take a line number and contain the indent/outdent logic?

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,65 @@ impl Editor {
...
+            let (first_line, _) = self.view.offset_to_line_col(&self.text, region.min());
+            let (last_line, last_col) =
+                self.view.offset_to_line_col(&self.text, region.max());
+            let last_line = if last_col == 0 && last_line > first_line {
+                last_line - 1
+            } else {
+                last_line   
+            };
+            lines.insert(first_line);
+            lines.insert(last_line);
```
cmyr: This only inserts the first and last lines, where we want all the lines in between as well.

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,65 @@ impl Editor {
         */
     }
 
+    /// Indents or outdents lines based on selection and user's tab settings.
+    /// Uses a BTreeSet to holds the collection of lines to modify.
+    /// Preserves cursor position and current selection as much as possible.
+    /// Tries to have behavior consistent with other editors like Atom, Sublime and VSCode,
+    /// with non-caret selections not being modified.
+
```
cmyr: extra newline here

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,65 @@ impl Editor {
...
+        let mut lines = BTreeSet::new();
+        let tab_text = if self.config.items.translate_tabs_to_spaces {
+                let tab_size = self.config.items.tab_size;
+                n_spaces(tab_size)
+            } else { "\t" };
+        for region in self.view.sel_regions() {
+            let (first_line, _) = self.view.offset_to_line_col(&self.text, region.min());
+            let (last_line, last_col) =
+                self.view.offset_to_line_col(&self.text, region.max());
+            let last_line = if last_col == 0 && last_line > first_line {
```
cmyr: just to clarify: this is handling the case where a selection is actually just a full line, but the cursor is on the subsequent line?

nangtrongvuon: Yes. I plucked the old commented code for indent and modified it a bit.

cmyr:
@eyelash `InsertTab` should have distinct behaviour; it inserts the tab at the cursor location(s) (not at the start of the line) and it also respects tab stops. `Indent` ignores tab stops and always inserts at start-of-line.

eyelash:
@cmyr both Sublime Text and Atom have the tab key bound to indent by default and I heavily rely on that feature. In fact the alternative shortcut <kbd>ctrl</kbd>+<kbd>[</kbd> does not even work with my keyboard layout.

cmyr:
@eyelash playing around with what is (I think) a vanilla installation of ST3 and the tab key inserts a tab, replacing the selection, unless the selection spans multiple lines?

I actually vaguely remember reading some release notes about changing this behaviour somewhere. So maybe it has changed recently? It might also be platform specific.

I agree that tab/shift+tab is a good binding for this, but it isn't clear to me that the RPCs themselves should be unified. If you have your cursor in the middle of a line, you should still be able to indent that whole line if you have an appropriate key binding. If we unify the RPC there's no way to distinguish between that action and the action of inserting a tab at the cursor position.

eyelash:
@cmyr good point. maybe the `InsertTab` command should be changed to indent if a selection spans multiple lines.

cmyr:
@eyelash yea I agree. Very simple to implement now too, just a matter of checking the selection size and forwarding the call as necessary. Will open an issue for that when this gets merged.

cjohansson:
Does it support indentation with space characters? Does xi-editor have a function that indents/outdents based on code styling and language? I'm new to this project so I haven't yet understood what features are implemented

In a code-editor, indentation-features are one of the most important features for me

nangtrongvuon:
@cjohansson This feature supports space indent and outdent, but it gets the settings from the user’s config file. 

Indent/outdent based on language isn’t supported yet, but is a feature soon to be implemented in summer.

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,71 @@ impl Editor {
...
+            let interval = Interval::new_closed_open(offset, tab_offset);         
+            let leading_slice = self.text.slice_to_string(interval.start(), interval.end());
+            if leading_slice == tab_text {
+                builder.delete(interval);
+            } else if let Some(first_char_col) = leading_slice.find(|c: char| !c.is_whitespace()) {
+                let first_char_offset = self.view.line_col_to_offset(&self.text, line, first_char_col);
+                let interval = Interval::new_closed_open(offset, first_char_offset);
+                builder.delete(interval);
+            }
+            self.this_edit_type = EditType::Delete;
```
cmyr: we only need to set this once, not on each iteration of the loop. (same as up top)

rust/core-lib/src/editor.rs
```diff
@@ -712,6 +717,71 @@ impl Editor {
...
+            } else {
+                last_line   
+            };
+            let line_range = first_line..(last_line + 1);
+            for line in line_range {
+                lines.insert(line);
+            }
+        }    
+        match direction {
+            IndentDirection::In => {
```
cmyr: If you can express the match action in a single line, you don't need to use the braces; that is, this can be `IndentDirection::In => self.indent(lines, tab_text),`.

rust/core-lib/src/editor.rs
nangtrongvuon: I’m not sure what you mean here, do you mean the variables in the loop? They seem dependent on the line to me.

cmyr: I mean you set `self.this_edit_type` for each line you adjust, when you only need to set it once for the whole operation.

cmyr:
Awesome, I think this turned out really well!

PR #587, opened 2018-03-30
fix(comments): remove comment about invariant

mqzry:
It was stated that:
"`is_boundary(to_base_units(x))` is True for all `x`"

This is not the case for BaseMetric. 
For BaseMetric there can be several 'measured units'(=utf8 units) between boundaries.

This comment confused me poking around the code, so I felt like opening this trivial PR.


mqzry:
I signed it!

PR #588, opened 2018-03-30
Cleanup trace library

### LINKED ISSUE #581: Polish and cleanup in xi-trace
I know that @vlovich was already considering/planning some of these changes, but after actually using the tracing tools I wanted to note down some of my observations:

~- [ ] _Add macros to wrap various tracing functions?_ currently even if tracing is disabled, the trace function is processed, which could include the allocation of a payload. This could be replaced with a macro that checks if tracing is enabled and noops if not. It's possible we would only want to do this for the `_payload` variants of the trace functions, since allocating payloads is the thing that could be potentially expensive in some cases.~

- [ ]  _Add ability to specify process/thread names_ The chrome tracing format allows you to add certain special [metadata traces](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#bookmark=id.iycbnb4z7i9g), which can be used to associate a name with a process or thread. Adding support for this would make interpreting traces easier.

- [ ] _Settle on a payload type_ we'd talked about this at some point; the various payload types and related features is confusing to library consumers. My preference is once more for a `serde_json::Value`, since that's what chrome_trace will want, but a `CowStr` works too.

vlovich:
Build a more direct 1:1 mapping of the internal data structure to the
Chrome trace event format.  Avoids the overcomplication of assuming we
need to support something else.  Also provides for a mechanism to do
things like annotating thread/process names much more naturally &
simplifies the encoding/decoding a lot (just direct automatic 1:1
mapping to the JSON via serde instead of needing a manual translation
layer).

Fixes #581


rust/core-lib/src/tabs.rs
```diff
@@ -349,16 +349,23 @@ impl Documents {
...
-            .into_iter()
-            .map(|json| chrome_trace::decode(&json).unwrap())
-            .collect::<Vec<_>>();
+        let mut plugin_traces_json = self.plugins.collect_trace();
+        let mut plugin_traces = plugin_traces_json.drain(0..)
+            .map(|json| {
+                chrome_trace::decode(json).map_err(|e| {
+                    // TODO(vlovich): transport the name of the plugin along with
+                    // the sample so we can print a better diagnostic about which
+                    // plugin failed to send well-formed JSON.
```
cmyr: one way we could do this is to decode close to the return site; the plugin manager has access to the plugin's name.

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -31,172 +23,10 @@ pub enum Error {
...
-
-    /// Converting Chrome trace event back into a Sample is a bit more work
-    /// because a Sample can represent a duration so we have to merge Chrome
-    /// trace events that indicate start/end.
-    fn try_from(entries: ChromeTraceArrayEntries) -> Result<Self, Error> {
-        match entries {
-            ChromeTraceArrayEntries::Array(samples_array) => Vec::try_from(samples_array),
-        }
-    }
+    Array(Vec<Sample>)
```
cmyr: Is this enum necessary? If this is just about serialization we can always just put a `Vec<Sample>` on the wire, we don't need a wrapper type.

rust/trace-dump/src/chrome_trace.rs
```diff
@@ -216,44 +46,27 @@ impl XiTryFrom<ChromeTraceArrayEntries> for Vec<Sample> {
 /// ```norun
 /// let samples = xi_trace::samples_cloned_sorted();
 /// let mut serialized = Vec::<u8>::new();
-/// chrome_trace::serialize(samples.iter(), OutputFormat::JsonArray, serialized);
+/// chrome_trace::serialize(samples.iter(), serialized);
 /// ```
-pub fn serialize<'a, I, W>(samples: I, _format: OutputFormat, output: W)
+pub fn serialize<'a, W>(samples: &Vec<Sample>, output: W)
```
cmyr: Is there any advantage to calling these functions instead of just using `serde_json` directly?

rust/trace/src/lib.rs
```diff
@@ -240,12 +240,78 @@ impl Config {
 static SAMPLE_COUNTER: AtomicUsize = ATOMIC_USIZE_INIT;
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
-pub enum SampleType {
-    /// This is an instantaneous sample (i.e. X occurred)
+pub enum SampleEventType {
```
cmyr: I'm assuming this is 1:1 with the chrome trace spec?

rust/core-lib/src/tabs.rs
vlovich: Can't seem to reply to a top-level review comment so I'm putting it here.  Yes, you observed correctly that trace_dump is itself probably not needed anymore & you can just invoke serde directly.  I'm leaving it for now because trace-ffi depends on it - that dependency could be removed pretty easily but I didn't want the scope to spiral out (I started by removing trace_dump but too many things broke that I felt punting was the correct answer for now).

vlovich: Regarding your comment, yes, we could.  Something to investigate in a future CL I think rather than tackling it all at once.

rust/trace-dump/src/chrome_trace.rs
vlovich: Yeah, I couldn't figure out how to get the code to compile without this dummy enum.  If you have suggestions, happy to implement it.

rust/trace-dump/src/chrome_trace.rs
vlovich: Not much.  I think it's mildly easier to use but probably not worth it.  Should be taken care of when trace-dump is removed.

rust/trace/src/lib.rs
vlovich: I tried to make it so, so if it doesn't it's a bug.

vlovich:
Also, it looks like thread & process names aren't captured but I figure I'll do that as a separate CL to avoid shoving everything into 1 mega CL.

rust/trace-dump/src/chrome_trace.rs
cmyr: No worries, especially if we're going to get rid of this crate at some point.

rust/core-lib/src/tabs.rs
cmyr: Okay that sounds reasonable. This seems pretty close to ready then, just the CI issue. Happy to take another look whenever it's ready.

rust/trace/src/lib.rs
```diff
@@ -393,18 +482,22 @@ impl<'a> SampleGuard<'a> {
...
 }
 
 impl<'a> Drop for SampleGuard<'a> {
     fn drop(&mut self) {
-        if let Some(ref mut sample) = self.sample {
-            sample.set_end_ns(time::precise_time_ns());
-            self.trace.unwrap().record(sample);
+        if let Some(ref mut trace) = self.trace {
+            let sample = self.sample.as_mut().unwrap();
+            sample.event_type = SampleEventType::DurationEnd;
```
cmyr: Don't we still need to set the end time here? It's confusing to me that we treat guard blocks differently than we treat closure blocks.

vlovich: There's no more end time.  Duration is different from start/stop in the Chrome Trace format.  There's no good clarification in the docs on where you would use either, but using the duration type for closures should be slightly more efficient (cpu & memory) since it only creates 1 Sample entry & inserts it once.

cmyr: In that case I think we should use the duration type for blocks as well? Certainly the way I've been using them I expect them to show up as spans in the trace viewer. There are a lot of situations where `trace_block` is more ergonomic than `trace_closure`, but conceptually I think they perform the same task.

In either case I'm confused by what's going on in this bit of code; my understanding is that we're adding a new sample, that has a distinct `SampleEventType` but otherwise is the same as the first. They even have the same `sample_id`?

cmyr: Also following this back to `Trace::record`, and noticed that we're always cloning the sample. I think instead `Trace::record` should be taking ownership of the sample, and if the caller needs to retain a copy they should `clone()` explicitly. In the general case this shouldn't be necessary, so it's currently an extra clone on every trace.

vlovich: It's a little less than straightforward (i.e. i'm likely to accidentally insert a bug) to have the block sample path coalesce the sample into a duration event so my instinct is to punt on it for now & leave a TODO.  In the UI I don't believe there's a difference between trace_block & trace_closure so the only difference should be the slightly larger JSON depending on how much more frequently trace_block is used.

I've corrected the perf issue you noted with Trace::record.

vlovich:
I think this CL is ready for integration.  Here's a screenshot of how it looks like.  Chrome trace event format doesn't do so well with thread pools which is why all the threads get the same name.  Chrome UI also is unfortunate in that it doesn't let you resize the label column so long names are illegible without digging into the HTML.

image

vlovich:
What's missing from this is capturing thread names from Rust but we can punt for later.

cmyr:
Okay cool will take a look shortly. Happy to punt on thread names; my thinking is that we should allow the user to manually name a thread (at least in the rust lib).

PR #597, opened 2018-04-05
Move do_scroll from Editor to View

cmyr:


rust/core-lib/src/view.rs
```diff
@@ -283,7 +277,7 @@ impl View {
             );
             sel
         });
-        new_sel.and_then(|new_sel| self.set_selection(text, new_sel))
+        new_sel.map(|sel| self.set_selection(text, sel));
```
raphlinus: Total style point, but I'd probably write this out as
```
if let Some(sel) = new_sel {
    self.set_selection(text, sel);
}
```

rust/core-lib/src/view.rs
```diff
@@ -486,6 +480,11 @@ impl View {
         let height = self.line_of_offset(text, text.len()) + 1;
         let plan = RenderPlan::create(height, self.first_line, self.height);
         self.send_update_for_plan(text, tab_ctx, style_spans, &plan);
+        if let Some(new_scroll_pos) = self.scroll_to {
```
raphlinus: It's tempting to write `self.scroll_to.take() and save a line, but this is probably easier to understand.

rust/core-lib/src/view.rs
```diff
@@ -586,7 +585,7 @@ impl View {
     ///
     /// Return value is a location of a point that should be scrolled into view.
```
raphlinus: Need to update comment, something like "Requests the point that should be scrolled into view."

rust/core-lib/src/view.rs
```diff
@@ -774,22 +773,29 @@ impl View {
...
     /// determines whether the search should stop at the first found occurrence (does only apply
     /// to forward search, i.e. reverse = false). If `allow_same` is set to `true` the current
     /// selection is considered a valid next occurrence.
     pub fn select_next_occurrence(&mut self, text: &Rope, reverse: bool, wrapped: bool,
-                                  stop_on_found: bool, allow_same: bool) -> Option<usize>
+                                  stop_on_found: bool, allow_same: bool)
     {
         if self.search_string.is_none() {
-            return None;
+            return
```
raphlinus: I personally would include the trailing `;`

cmyr:
Addressed your comments.

I've been noticing just how much of the code in editor.rs is just plumbing to access the view; I'm going to look into clearing that up more tomorrow.

Issue #456, opened 2017-12-18
Clean up and document metrics and boundaries


raphlinus:
I wrote this in response to more carefully thinking through the questions raised in #450. It's an issue rather than a doc PR because there are non-trivial code changes required to clean everything up. The resolution of this issue is to fix the code and adapt this text into documentation (ideally preferring documentation inside the xi-rope crate as opposed to a standalone file in `doc`).

# Metrics and Boundaries

The xi-rope code tries to capture general underlying mathematical concepts without being overly abstract. This can be confusing (especially to newcomers) and there are definitely ways in which the current code doesn't live up to this ideal.

## Monoid homomorphisms

A rope is a data structure for representing strings, so that many operations (especially editing) are efficient even as the string grows. Xi-rope implements strings, but also generalizes to other structures.

The mathematical generalization of a string is a [monoid](https://en.wikipedia.org/wiki/Monoid). The operation is string concatenation, and the identity element is the empty string.

A string has a length. The empty string has zero length, and the length of the concatenation of two strings is the sum of their lengths. The mathematical generalization of this idea is a monoid homomorphism, which basically says that the function preserves the monoid structure.

The concept of monoid homomorphism doesn't cover everything we care about regarding strings. In particular, it's missing any way to compute substrings, and we care about that deeply.

## Metrics

A _metric_ is a monoid homomorphism ranging over nonnegative integers, and also satisfying a form of the triangle inequality. If we write the monoid operation as ⊕, then this is simply a ⊕ b ≤ a + b. String length is an important metric, but there are other interesting examples.

A metric is _additive_ if its operation is addition. Any additive metric trivially satisfies the triangle inequality.

A good example of a nonadditive metric is the "nonascii" metric, defined as being 0 for a string that is entirely ASCII, and 1 if it contains any non-ASCII character. Here, the monoid operator could be written a ⊕ b = min(a + b, 1). (Note: maybe this is confusing, and metrics should be reserved for nonnegative integers with addition, and some other concept. That said, as we define boundaries in terms of metric it will probably be useful.)

An _atomic metric_ is one that is nonzero for any element of the original monoid other than the identity element. A good way to explore atomicity is to look at representations of Unicode strings. (Note: given how often I say "additive atomic" and that I can't think of a good use case for a non-additive atomic metric, it might be worth not separating these concepts out so much)

[...]

PR #600, opened 2018-04-07
docs(rope): copied documentation from issue 456 into file

mqzry:
As mentioned in https://github.com/google/xi-editor/issues/456 and as I discussed with @cmyr on IRC, it would be better to not have the explanation only in the issues. The issues are sort of a stack and having this explanation only in a 'we need to refactor' issue hides it from any potential contributor.

### DIFF
--- a/rust/rope/docs/MetricsAndBoundaries.md
+++ b/rust/rope/docs/MetricsAndBoundaries.md
@@ -1,0 +1,80 @@
+#Metrics and Boundaries
+The xi-rope code tries to capture general underlying mathematical concepts without being overly abstract. This can be confusing (especially to newcomers) and there are definitely ways in which the current code doesn't live up to this ideal.
+
+## Monoid homomorphisms
+A rope is a data structure for representing strings, so that many operations (especially editing) are efficient even as the string grows. Xi-rope implements strings, but also generalizes to other structures.
+
+The mathematical generalization of a string is a monoid. The operation is string concatenation, and the identity element is the empty string.
+
+A string has a length. The empty string has zero length, and the length of the concatenation of two strings is the sum of their lengths. The mathematical generalization of this idea is a monoid homomorphism, which basically says that the function preserves the monoid structure.
+
+The concept of monoid homomorphism doesn't cover everything we care about regarding strings. In particular, it's missing any way to compute substrings, and we care about that deeply.
+
+## Metrics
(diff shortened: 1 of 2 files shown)


rust/rope/docs/MetricsAndBoundaries.md
```diff
@@ -0,0 +1,80 @@
+#Metrics and Boundaries
```
eyelash: missing space after `#`

mqzry: thanks for catching this, should be fixed now

mqzry:
For my macbook(2013, 16gig mem, 2.3ghz i7) I got the following benchmarks.
Benchmarks without memrchr:

`test benchmark_few_big_lines ... bench:       3,305 ns/iter (+/- 295)                                                                                            
test benchmark_short_lines   ... bench:  20,055,720 ns/iter (+/- 3,623,465)
test benchmark_triangle      ... bench:  23,026,709 ns/iter (+/- 2,518,993)`

Benchmark with memrchr:
`test benchmark_few_big_lines ... bench:       3,232 ns/iter (+/- 1,025)
test benchmark_short_lines   ... bench:  17,075,772 ns/iter (+/- 3,927,781)
test benchmark_triangle      ... bench:  21,615,568 ns/iter (+/- 3,111,899)`

cmyr:
Awesome, thanks for following up on this. I am however wondering where this should live; in particular I'm worried about discoverability. I'm not sure anyone will know to look in `rope/docs`. This is almost-but-not-quite part of the rope science series in the top level docs, but I'm not sure.

Maybe my suggestion right now would be that we go with your plan here, but explicitly mention this file somewhere in `rope.rs`, so people who are curious have a better chance of finding it?

mqzry:
I think this is a step in the right direction for discoverability from a contributors perspective.
Listing the contents of the root of the rope crate now will have a `docs` dir, which is the obvious place to look further if interested.

The fact that you have to look in rope/docs AND in the science series to get a clue what is happening in the rope crate is not  optimal.  Ideally the rope crate should have all documentation contained within for understanding it, with links to some resources(papers/specs/) that explain everything in great detail for the interested. These resources should not be globally linked to, but whenever some topic is used we should point to that specific section. So instead of 'For info about x, see science_series_url' on the top of the file,  it would be better to say 'For more info about atomic metrics, see science_series_url#metrics'.

cmyr:
Okay, that sounds reasonable. I'm going to leave this up for @raphlinus because this is very much his domain, and he might have some other ideas, but lgtm.

PR #601, opened 2018-04-07
refactor(rope): use memrchr for reverse byte lookup

mqzry:
The crate [memchr](https://docs.rs/memchr/2.0.1/memchr/) offers byte lookup in both directions. Until now only `memchr` was used, which is the lookup in positive direction. I replaced all `as_bytes().iter().rposition` calls with memrchr.

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -238,7 +238,7 @@ impl Metric<RopeInfo> for LinesMetric {
     }
 
     fn prev(s: &String, offset: usize) -> Option<usize> {
-        memrchr(b'\n', s.as_bytes()[..offset])
+        memrchr(b'\n', &s.as_bytes()[..offset])
             .map(|pos| pos + 1)
     }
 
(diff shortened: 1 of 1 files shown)


mqzry:
For my macbook(2013, 16gig mem, 2.3ghz i7) I got the following benchmarks.
Benchmarks without memrchr:

`test benchmark_few_big_lines bench:  3,305 ns/iter (+/- 295)`
`test benchmark_short_lines  bench:  20,055,720 ns/iter (+/- 3,623,465)`
`test benchmark_triangle bench:  23,026,709 ns/iter (+/- 2,518,993)`

Benchmark with memrchr:
`test benchmark_few_big_lines  bench:       3,232 ns/iter (+/- 1,025)`
`test benchmark_short_lines   bench:  17,075,772 ns/iter (+/- 3,927,781)`
`test benchmark_triangle      bench:  21,615,568 ns/iter (+/- 3,111,899)`

PR #603, opened 2018-04-07
Some extra documentation in rope.rs

mqzry:
Put some more documentation around the two metrics(Line+Base) that are defined in rope.rs.
For the BaseMetric I had some difficulties explaining that it was atomic and why. My explanation felt a bit like [hand-waving](https://en.wikipedia.org/wiki/Hand-waving). Any comments would be appreciated.

Added ref to utf8 RFC for people to find out where magic bytes like 0x80 come from.

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -147,12 +147,12 @@ impl NodeInfo for RopeInfo {
 #[derive(Clone, Copy)]
 pub struct BaseMetric(());
 
-/// Measured unit is utf8 code unit
-/// Base unit is utf8 code unit
+/// Measured unit is utf8 code unit.
+/// Base unit is utf8 code unit.
 /// Boundary is atomic and determined by codepoint boundary.
 /// Atomicity is implicit, putting the offset
 /// between two utf8 code units that form a code point is considered invalid.
-/// For example, take a string that starts with a 0x80 byte.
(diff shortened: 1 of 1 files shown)


rust/rope/src/rope.rs
```diff
@@ -147,6 +147,13 @@ impl NodeInfo for RopeInfo {
 #[derive(Clone, Copy)]
 pub struct BaseMetric(());
 
+/// Measured unit is utf8 code unit
+/// Base unit is utf8 code unit
+/// Boundary is atomic and determined by codepoint boundary.
+/// Atomicity is implicit, putting the offset
+/// between two utf8 code units that form a code point is considered invalid.
+/// For example, take a string that starts with a 0x80 byte.
```
raphlinus: I would use `0xc2` as an example because no valid UTF-8 string ever starts with 0x80.

rust/rope/src/rope.rs
```diff
@@ -208,6 +216,9 @@ pub fn len_utf8_from_first_byte(b: u8) -> usize {
 #[derive(Clone, Copy)]
 pub struct LinesMetric(usize);  // number of lines
 
+/// Measured unit is newline amount
```
raphlinus: I think you'll want periods at the end of these lines, otherwise the markdown formatting will make it all flow together.

cmyr:
Great, thanks again @mqzry!

Issue #591, opened 2018-03-31
remove do_click, migrate to do_gesture


cmyr:
I thought there was an existing issue for this, but apparently not?

mouse clicks were previously handled in `Editor::do_click`. We would like to move all mouse/touch related events into the the `do_gesture` method / RPC.

This involves adding new gesture types for the events currently handled by `do_click`, and handling the new events in `do_gesture`, with most of the code moving to `view.rs`. #585 is a good example of how gestures should work. Ideally a bunch of code can be shared between the various gesture types.

scholtzan:
I will work on this.

cmyr:
closed in #605

PR #605, opened 2018-04-08
Remove do_click and migrate to do_gesture

scholtzan:
This addresses https://github.com/google/xi-editor/issues/591. I moved all the click gestures to `do_gesture` and removed `do_click`. I am not entirely sure if the way I did it is the best one.
This also fixes the issue in https://github.com/google/xi-editor/pull/604/files

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -881,7 +881,7 @@ impl Editor {
     fn do_gesture(&mut self, line: u64, col: u64, ty: GestureType) {
         let offset = self.view.line_col_to_offset(&self.text, line as usize, col as usize);
         match ty {
-            GestureType::Click => {
+            GestureType::PointSelect => {
                 self.view.set_selection(&self.text, SelRegion::caret(offset));
                 self.view.start_drag(offset, offset, offset);
             },
@@ -1113,6 +1113,20 @@ impl Editor {
             RequestLines(LineRange { first, last }) => self.do_request_lines(first, last),
             Yank => self.yank(),
             Transpose => self.do_transpose(),
+            Click(MouseAction {line, column, flags, click_count} ) => {
+                // Deprecated (kept for client compatibility): should be removed in favor of do_gesture
+                println!("Usage of click is deprecated and should be replaced by gestures");
+
+                if (flags & FLAG_SELECT) != 0 {
+                    self.do_gesture(line, column, GestureType::RangeSelect)
+                } else if click_count == Some(2) {
+                    self.do_gesture(line, column, GestureType::WordSelect)
+                } else if click_count == Some(3) {
+                    self.do_gesture(line, column, GestureType::LineSelect)
+                } else {
+                    self.do_gesture(line, column, GestureType::PointSelect)
+                }
+            }
             Drag (MouseAction {line, column, flags, ..}) => {
                 self.do_drag(line, column, flags);
             }
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -874,37 +874,6 @@ impl Editor {
         self.view.request_lines(&self.text, &self.doc_ctx, self.styles.get_merged(), first as usize, last as usize);
     }
 
-    fn do_click(&mut self, line: u64, col: u64, flags: u64, click_count: u64) {
-        // TODO: calculate affinity
-        let offset = self.view.line_col_to_offset(&self.text, line as usize, col as usize);
-        if (flags & FLAG_SELECT) != 0 {
-            if !self.view.is_point_in_selection(offset) {
-                let sel = {
-                    let (last, rest) = self.view.sel_regions().split_last().unwrap();
-                    let mut sel = Selection::new();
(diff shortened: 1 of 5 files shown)


rust/core-lib/src/view.rs
```diff
@@ -220,6 +220,26 @@ impl View {
...
+          for &region in rest {
+            sel.add_region(region);
+          }
+          // TODO: small nit, merged region should be backward if end < start.
+          // This could be done by explicitly overriding, or by tweaking the
+          // merge logic.
+          sel.add_region(SelRegion::new(last.start, offset));
+          sel
+        };
+        self.set_selection(text, sel);
```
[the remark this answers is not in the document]
cmyr: This is moved over from the deleted `do_click` fn in `editor.rs`. In any case the rationale is that `View::set_selection` has a few side effects (well: _effects_) such as modifying the view's scroll position if necessary. 

There are definitely other ways of handling this, but some sort of copying is currently necessary because you can't manipulate the selection in place.

rust/core-lib/src/editor.rs
```diff
@@ -1137,9 +1113,6 @@ impl Editor {
             RequestLines(LineRange { first, last }) => self.do_request_lines(first, last),
             Yank => self.yank(),
             Transpose => self.do_transpose(),
-            Click(MouseAction {line, column, flags, click_count} ) => {
```
cmyr: This will break all frontends that aren't xi-mac, so maybe we don't want to remove it right away? What would be ideal is if we can keep it in place and just call `do_gesture` with the appropriate gesture for a given click event.

We could also consider logging a deprecation warning, maybe pointing people to this PR?

This actually reminds me that, ideally, this PR would include updates to the docs to include the newly added gestures. I'd been forgetting this, we've had a few recent PRs without doc updates.

rust/core-lib/src/rpc.rs
```diff
@@ -288,7 +288,11 @@ pub struct EditCommand<T> {
 #[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
 #[serde(rename_all = "snake_case")]
 pub enum GestureType {
+    Click,
```
cmyr: Maybe `Click` could be `SetSelection`? Or `Select` or `PointSelect`? Not sure, open to other names. My thinking is that one nice thing about moving to this gesture model is that it lets us abstract away the particulars of the hardware; users on touch devices don't 'click', but they do still send whatever this gesture is.

rust/core-lib/src/rpc.rs
```diff
@@ -288,7 +288,11 @@ pub struct EditCommand<T> {
...
     ToggleSel,
+    RangeSelect,
+    LineSelect,
+    WordSelect,
```
cmyr: Noticing that there's a bit of inconsistency with your names (..Selection) and @raphlinus's ToggleSel. I prefer your names, but at some point when we do a breaking update we might want to standardize these. Fine for now though.

rust/core-lib/src/rpc.rs
```diff
@@ -372,7 +376,6 @@ pub enum EditNotification {
     RequestLines(LineRange),
     Yank,
     Transpose,
-    Click(MouseAction),
```
cmyr: see earlier re: breaking clients

eyelash:
Maybe instead of 6 gesture types we could have 3 plus a boolean that indicates whether you want to modify or replace the current selection (basically whether the cmd/ctrl key is pressed or not). Just a thought.

rust/core-lib/src/editor.rs
scholtzan: That makes sense. I will leave it there for now to prevent other clients from breaking.

I will add it to the documentation.

rust/core-lib/src/rpc.rs
scholtzan: I agree. I couldn't come up with a better name at the time but I like `PointSelect`.

scholtzan:
@eyelash I also thought about that. But I thought for the sake of simplicity it might be better to have different gesture types.

cmyr:
We could definitely do it either way, I don't have strong feelings. @scholtzan you have a conflict from a little change I made earlier in the week, should be a simple fix though.

rust/core-lib/src/editor.rs
```diff
@@ -1113,6 +1113,20 @@ impl Editor {
             RequestLines(LineRange { first, last }) => self.do_request_lines(first, last),
             Yank => self.yank(),
             Transpose => self.do_transpose(),
+            Click(MouseAction {line, column, flags, click_count} ) => {
+                // Deprecated (kept for client compatibility): should be removed in favor of do_gesture
+                println!("Usage of click is deprecated and should be replaced by gestures");
```
cmyr: you can't use `println` in xi-core because it prints to stdout, which is interpreted as an RPC by the client, which it won't be able to parse. To print to stderr, use the `eprintln!` macro.

scholtzan:
I accidentally broke something on my first attempt to get your changes. However, it should be fixed now.

Issue #602, opened 2018-04-07
InsertTab should adjust indentation when multiple lines are selected


cmyr:
Breaking this out from #586: currently the `insert_tab` RPC will replace any selections with `\t`. In many editors, the tab key will increase indentation of lines when multiple lines are selected.

I think this makes sense as the default behaviour; it matches the default behaviour of ST3 on macOS. 

As of #586 this should be easy to implement; in `Editor::insert_tab` check to see if the selection contains multiple regions, or if a single region spans multiple lines, and in that case forward the call to `Editor::modify_indent`.

Reificator:
> in `Editor::insert_tab` check to see if the selection contains multiple regions, or if a single region spans multiple lines

This is a bit different from the behavior of Sublime (and VS Code).

In Sublime, if any region spans multiple lines, it will indent rather than insert a tab.  However, if none of the regions contain a newline, then a tab will be inserted in place of each region.

In VS Code, only regions that span multiple lines will have their indentation changed.  This is consistent even if one region spans multiple lines and another does not.

I'm personally a fan of the VS Code approach, but then I use VS Code as my daily editor.

cmyr:
@Reificator cool, thanks for the more complete tests.

I agree that inserting a tab makes sense when none of the regions span a newline. I'm curious about your preference, though; if I'm undertanding then if I have a caret on one line and elsewhere I have a selection that spans multiple lines, hitting tab will insert a tab in the first location and increase indentation in the second? The mixed behaviour feels weird to me, hard to imagine it being something anyone would use (although I refuse to be surprised by how people use their text editors 😏).

Reificator:
You're understanding correctly.

The choice is to be consistent in the action performed by one keypress or to be consistent in the action performed in one context.  I prefer the latter, but it's a mild preference personally, and I can see the other point of view.

virattara:
I too think this will be the correct approach. 
@cmyr making changes in #609 to handle this case.

virattara:
Updated #609 @cmyr

cmyr:
@virattara okay thanks! I'll take a look tomorrow at some point. :)

cmyr:
closed in #609

PR #609, opened 2018-04-10
adjust indentation when multiple lines are selected

virattara:
#602


virattara:
Thanks for pointing that out, have updated the PR.

virattara:
Yes, there is. Can try to do that in a separate PR.

rust/core-lib/src/editor.rs
```diff
@@ -649,17 +649,36 @@ impl Editor {
...
-            let tab_text = if self.config.items.translate_tabs_to_spaces {
-                    let (_, col) = self.view.offset_to_line_col(&self.text, region.start);
-                    let tab_size = self.config.items.tab_size;
-                    let n = tab_size - (col % tab_size);
-                    n_spaces(n)
-            } else {
-                "\t"
-            };
-            builder.replace(iv, Rope::from(tab_text));
+            let mut sel_text = self.text.slice(region.min(), region.max());
```
raphlinus: It would be more efficient to measure the text at the endpoints and compute the difference, than to allocate the slice.

rust/core-lib/src/view.rs
```diff
@@ -882,6 +883,19 @@ impl View {
             self.set_selection(text, occ);
         }
     }
+
+    pub fn get_line_range(&self, text: &Rope, region: &SelRegion) -> Range<usize> {
```
raphlinus: Would appreciate a doc comment, even if brief.

rust/core-lib/src/view.rs
```diff
@@ -882,6 +883,19 @@ impl View {
...
+        let (first_line, _) = self.offset_to_line_col(text, region.min());
+        let (last_line, last_col) =
+            self.offset_to_line_col(text, region.max());
+        let last_line = if last_col == 0 && last_line > first_line {
```
raphlinus: Just as point of style, this reads cleaner to me to make `last_line` mut and just do `-= 1` if the condition holds.

rust/core-lib/src/view.rs
```diff
@@ -882,6 +883,19 @@ impl View {
...
+            last_line - 1
+        } else {
+            last_line
+        };
+        let line_range = first_line..(last_line + 1);
```
raphlinus: Another style point, can just return rather than binding to a variable, esp. since the name of the variable is pretty much the same as the name of the function.

rust/core-lib/src/editor.rs
```diff
@@ -649,17 +649,36 @@ impl Editor {
...
+
+            if nb_lines > 1 {
+                let line_range = self.view.get_line_range(&self.text, region);
+
+                for line in line_range {
+                    let offset = self.view.line_col_to_offset(&self.text, line, 0);
+                    let iv = Interval::new_closed_open(offset, offset);
+                    builder.replace(iv, Rope::from(const_tab_text));
+                }
+            }
```
raphlinus: Minor style nit: keep `else` on same line.

cmyr:
@virattara merging this now to get it in ahead of #613. That's going to get merged sometime tomorrow, so be aware of that if you're doing any followup. Thanks again! :)

virattara:
Okay, will keep that in mind. Also, I think I need to make changes suggested by Raph too.

PR #612, opened 2018-04-13
remove page_down and page_up from docs

bvinc:
I'm not sure if it was intentional, but the `page_down` and `page_up` edit commands have not worked for a while.  Now, serde auto-converts the variants to snake_case strings, and there is only `scroll_page_up` and `scroll_page_down`.

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -211,10 +211,8 @@ move_left_and_modify_selection
 move_right
 move_right_and_modify_selection
 scroll_page_up
-page_up
 page_up_and_modify_selection
 scroll_page_down
-page_down
 page_down_and_modify_selection
 ```
 


cmyr:
Interesting, I think these were removed before I was involved in the project. Docs cleanup is always welcome, and maybe @raphlinus can let us know if there should be a TODO here or if these are just gone/renamed.

raphlinus:
I think ok to remove. macOS does make a distinction between `scrollPageUp:` (page up key also ctrl-up, but the latter seems to be shadowed by the window manager) and `pageUp:` (alt-page up key) and I think I was trying to preserve that. But I think it's subtle enough we can not care.

eyelash:
They were removed in 32cad53de205f24b221eacfe6dfe6120442fe27b BTW.

cmyr:
Ah interesting. I they weren't removed on purpose, then. It would be easy enough to put them back, if we think there's a case. Thanks @eyelash!

PR #615, opened 2018-04-17
Small serde cleanup

dtolnay:
Minor things I noticed as I was skimming the Serde-related code.

- `as_tup` can be serialized to the serializer directly, does not need to be converted to a Value first.
- Conversion from usize to Value is infallible.
- Value is typically accessed as serde_json::Value.


raphlinus:
Thanks, and thanks for taking a closer look at the code size issue.

cmyr:
Thanks for this @dtolnay

PR #622, opened 2018-04-18
Finer grained multi-cursor style invalidation

cmyr:
I suspect there is room for efficiency gains here; if we had typed base units (and if `SpansInfo<T>` was `T: Default`?) we wouldn't need to convert the delta back and forth.

Maybe a better option that occurs to me now is to just have a special `apply_with_blanks` method on `Spans` that takes a `Delta`, and just duplicates the logic in `Delta::apply`? Happy to make that change if we think it's worth it.


rust/rope/src/spans.rs
```diff
@@ -291,6 +291,23 @@ impl<T: Clone + Default> Spans<T> {
             ix: 0,
         }
     }
+
+    /// Applies a generic delta to `self`, inserting empty spans for any
+    /// added regions.
+    ///
+    /// This is intended to be used to keep spans up to date with a `Rope`
+    /// as edits occur.
+    pub fn apply_simple<M: NodeInfo>(&mut self, delta: &Delta<M>) {
```
raphlinus: I'd quibble with the name, to me "simple" means an insert / delete / replace of a single contiguous region. Would "apply_delta" be accurate?

cmyr: My thinking was that it doesn't actually apply the delta, because it ignores the contents of ignored regions. I agree the name isn't great, though. `apply_ignoring_contents`? `apply_shape_only`? `make_fit`? `update_shape`?

PR #630, opened 2018-04-22
modified logic to count no. of lines in insert_tab

virattara:
Mods in #609 P.R
Removed slice logic to count the number of lines.
Made a common function of tab_text.
Some style changes.

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -1082,18 +1082,17 @@ impl View {
         }
     }
 
+    //get the line range of a selected region
     pub fn get_line_range(&self, text: &Rope, region: &SelRegion) -> Range<usize> {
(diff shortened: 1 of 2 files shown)


rust/core-lib/src/view.rs
```diff
@@ -1082,18 +1082,17 @@ impl View {
         }
     }
 
+    //get the line range of a selected region
```
cmyr: Nit: Rust doc comments begin with three `/`; I'd also add a space after the last `/`, and use a full sentence (that is, Capitalize the first word and include a terminal `.`)

virattara: Thanks for pointing this out. Fixed.

cmyr:
I've tested, lgtm

PR #633, opened 2018-04-24
Allow new_view into non-existent file (fix #627)

### LINKED ISSUE #627: `new_view` returns an error for non-existing file
When sending a `new_view` request with a non-existing file as argument the core responds with a message like
```json
{"error":{"code":5,"message":"failed to load file"},"id":0}
```
which xi-gtk is not prepared to handle. I'm not sure if this is intended to be this way but I would prefer the core to just open a view and create the file on save. At the very least this behaviour is inconsistent with the [documentation](https://github.com/google/xi-editor/blob/master/docs/docs/frontend-protocol.md#new_view).

cmyr:
cc @eyelash 

This does have one issue, which is that we cannot monitor non-existent files for changes; with this patch, if the file we're pointed to is written by another process after we open the view, a subsequent write by us will overwrite the file without warning.

We currently trust the client to know what it wants when it asks us to save, so I'm not sure what the best approach is here. 😕

### DIFF
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -110,9 +110,15 @@ impl FileManager {
     pub fn open(&mut self, path: &Path, id: BufferId)
         -> Result<Rope, FileError>
     {
-        let (rope, info) = try_load_file(path)?;
+        let (rope, info) = if path.exists() {
+            try_load_file(path)?
+        } else {
+            new_for_path(path)
+        };
+
         self.open_files.insert(path.to_owned(), id);
         if self.file_info.insert(id, info).is_none() {
+            #[cfg(feature = "notify")]
             self.watcher.watch(path, false, OPEN_FILE_EVENT_TOKEN);
         }
         Ok(rope)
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/file.rs
```diff
@@ -110,9 +110,15 @@ impl FileManager {
     pub fn open(&mut self, path: &Path, id: BufferId)
         -> Result<Rope, FileError>
     {
-        let (rope, info) = try_load_file(path)?;
+        let (rope, info) = if path.exists() {
```
raphlinus: This is a race if the file is deleted, but maybe it's ok for that to be an error.

eyelash:
> This does have one issue, which is that we cannot monitor non-existent files for changes; with this patch, if the file we're pointed to is written by another process after we open the view, a subsequent write by us will overwrite the file without warning.

Do we know how other editors handle this case?

rust/core-lib/src/file.rs
cmyr: Yea that sounds okay in this case, but there are definitely some other bits of possible strangeness here; see my top level comment.

cmyr:
@eyelash Vim (still my daily driver) has the same behaviour as this patch.

Thinking about this though, it should be relatively easy to handle without any major changes; if the buffer has an associated path but there was no saved modification date, if the path exists at save time we can error as we would in the conflicting-write case.

eyelash:
Another possible solution might be to just create an empty file.

cmyr:
Yea that's appealingly simple, I like.

PR #642, opened 2018-05-07
Documentation cleanup/fixup

cmyr:
A bunch of little cleanup / additions to the docs, in the ongoing battle against entropy.

@ArazAbishov I'm not sure how I should test this locally?

### DIFF
--- a/docs/docs.md
+++ b/docs/docs.md
@@ -7,7 +7,8 @@ site_nav_category: docs
 ---
 
 ### Table of Contents:
- - [Notes on writing frontends](docs/frontend-protocol.html)
+ - [Notes on writing frontends](docs/frontend-notes.html)
+ - [The Xi Protocol](docs/frontend-protocol.html)
  - [Plugin architecture](docs/plugin.html)
  - [The config system](docs/config.html)
  - CRDT Docs:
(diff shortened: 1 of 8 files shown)


cmyr:
@raphlinus oh that was an out of date commit. The update doc is now included inline, at the bottom of this page, there should be a link from the update method. Something might've got mangled, though

ArazAbishov:
@cmyr to build page locally, you need to do following:

Ensure that you have Ruby and Bundler installed.
```
gem install bundler
```

One-time setup

```
bundle install --path vendor/bundle
```

> Note: If you’re on Mac OS and this fails installing nokogiri, run brew unlink xz, install, and then brew link xz

Running the site

```
bundle exec jekyll serve
```

Taken from here: https://android.github.io/kotlin-guides/contribute.html 

I can make a PR with similar instructions for xi page as well (in case if that is useful).

ArazAbishov:
@cmyr Yep, we can do that. If we leave it under docs/ directory, it will be visible on the web page as well (among tabs). If that's alright, I can create a PR :)

cmyr:
@ArazAbishov whatever you think is best!

PR #646, opened 2018-05-09
[RFC] language support

cmyr:
An initial sketch. Feedback welcome!

### DIFF
--- a/rfcs/language-support.md
+++ b/rfcs/language-support.md
@@ -1,0 +1,87 @@
+
+## Motivation
+
+For an overview of the general problem, see [#645](https://github.com/google/xi-editor/issues/645). For an earlier RFC on a similar problem, see [#374](https://github.com/google/xi-editor/blob/plugin-tag-rfc/doc/rfcs/plugin-lifetimes.md). To quickly recap: we would like plugins to be able to add support for new languages, but we do not want to put any constraints on how that support is implemented; and we would like some mechanism by which we can ensure that only a single plugin is providing core language support for a given language.
+
+## Proposal
+
+### A plugin can add new language definitions in its manifest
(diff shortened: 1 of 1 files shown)


Kwarrtz:
I'm totally new to this project, and I don't know whether this RFC is even up to date anymore, but a few questions jumped out at me reading through it:

1. In addition to syntect, which you addressed in the RFC, there seems to be some talk about tree-sitter support (#743, #424). Do you have plans for what integration might look like?
2. You list autocomplete and hover definitions as features which could be provided by plugins other than the language providers and thus would lack the uniqueness constraints imposed on providers, but it seems that having multiple plugins attempting to provide these features for the same language could be highly problematic. This also points to a broader concern, which is that many plugins and packages, even those having nothing in particular to do with language support, can easily conflict with one another, especially once front-end integration becomes a possibility. Do we expect to be able to design the API in such a way as to prevent the possibility of conflicts? If not, should it be the responsibility of the user and/or plugin author to detect and prevent them? Or should Xi offer some sort of general framework for handling plugins whose features overlap in problematic ways?

cmyr:
Good questions! This RFC is definitely out of date; exactly how out of date is to  be determined.

1. There is no intention for syntect to have any special privileges; whatever features syntect is able to provide should also be usable for other plugins, such as a tree-sitter wrapper.

2. The api for things like autocomplete and hover will support multiple providers; the user should be able to also have fine-grained control over what providers are active for a given project/scope. But I could imagine a compiler-backed auto complete provider that knew about symbols in other files, and I could also imagine a simple auto-complete provider that just naively matched against symbols in the active file, and these should be able to work fine together. Similarly with say some snippet provider.

PR #649, opened 2018-05-09
[WIP] Find enhancements

scholtzan:
Outsourced find logic from `view`

### DIFF
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -1,3 +1,19 @@
+// Copyright 2018 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Module for searching text.
+
 use std::cmp::{min,max};
 
 use serde_json::Value;
@@ -22,12 +38,10 @@ use xi_rope::tree::Metric;
 //  query_id: usize,      // id or reference to SearchQuery?
 //  highlight: Selection
 //}
-//
-
-
 
 const BACKWARDS_FIND_CHUNK_SIZE: usize = 32_768;
 
+/// Contains logic to search text
 pub struct Find {
     // todo: support multiple queries
     //  search_queries: Vec<SearchQuery>,
@@ -47,8 +61,6 @@ pub struct Find {
     /// Set of ranges that have already been searched for the currently active search string
     valid_search: IndexSet,
 }
-
-
 
 impl Find {
     pub fn new() -> Find {
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -38,7 +38,7 @@ pub struct Find {
     /// The case matching setting for the currently active search
     case_matching: CaseMatching,
     /// The set of all known find occurrences (highlights)
-    occurrences: Option<Selection>,
+    occurrences: Selection,
     /// Set of ranges that have already been searched for the currently active search string
     valid_search: IndexSet,
 }
@@ -49,12 +49,12 @@ impl Find {
             hls_dirty: true,
             search_string: None,
             case_matching: CaseMatching::CaseInsensitive,
-            occurrences: None,
+            occurrences: Selection::new(),
             valid_search: IndexSet::new(),
         }
     }
 
-    pub fn occurrences(&self) -> &Option<Selection> {
+    pub fn occurrences(&self) -> &Selection {
         &self.occurrences
     }
 
@@ -70,23 +70,20 @@ impl Find {
         // update search highlights for changed regions
         if self.search_string.is_some() {
             self.valid_search = self.valid_search.apply_delta(delta);
-            let mut occurrences = self.occurrences.take().unwrap_or_else(Selection::new);
 
             // invalidate occurrences around deletion positions
             for DeltaRegion{ old_offset, new_offset, len } in delta.iter_deletions() {
                 self.valid_search.delete_range(new_offset, new_offset + len);
-                occurrences.delete_range(old_offset, old_offset + len, false);
+                self.occurrences.delete_range(old_offset, old_offset + len, false);
             }
 
-            occurrences = occurrences.apply_delta(delta, false, false);
+            self.occurrences = self.occurrences.apply_delta(delta, false, false);
 
             // invalidate occurrences around insert positions
             for DeltaRegion{ new_offset, len, .. } in delta.iter_inserts() {
                 self.valid_search.delete_range(new_offset, new_offset + len);
-                occurrences.delete_range(new_offset, new_offset + len, false);
+                self.occurrences.delete_range(new_offset, new_offset + len, false);
             }
-
-            self.occurrences = Some(occurrences);
 
             // update find for the whole delta (is going to only update invalid regions)
             let (iv, _) = delta.summary();
@@ -117,7 +114,7 @@ impl Find {
     /// Unsets the search and removes all highlights from the view.
     pub fn unset(&mut self) {
         self.search_string = None;
-        self.occurrences = None;
+        self.occurrences = Selection::new();
         self.hls_dirty = true;
         self.valid_search.clear();
     }
@@ -155,7 +152,6 @@ impl Find {
         // extend the search by twice the string length (twice, because case matching may increase
         // the length of an occurrence)
         let slop = if include_slop { self.search_string.as_ref().unwrap().len() * 2 } else { 0 };
-        let mut occurrences = self.occurrences.take().unwrap_or_else(Selection::new);
         let mut invalidate_from = None;
 
         for (_start, end) in self.valid_search.minus_one_range(start, end) {
@@ -173,8 +169,8 @@ impl Find {
                 let end = cursor.pos();
 
                 let region = SelRegion::new(start, end);
-                let prev_len = occurrences.len();
-                let (_, e) = occurrences.add_range_distinct(region);
+                let prev_len = self.occurrences.len();
+                let (_, e) = self.occurrences.add_range_distinct(region);
                 // in case of ambiguous search results (e.g. search "aba" in "ababa"),
                 // the search result closer to the beginning of the file wins
                 if e != end {
@@ -188,14 +184,14 @@ impl Find {
                 // add_range_distinct() above removes ambiguous regions after the added
                 // region, if something has been deleted, everything thereafter is
                 // invalidated
-                if occurrences.len() != prev_len + 1 {
+                if self.occurrences.len() != prev_len + 1 {
                     invalidate_from = Some(end);
-                    occurrences.delete_range(end, text_len, false);
+                    self.occurrences.delete_range(end, text_len, false);
                     break;
                 }
             }
         }
-        self.occurrences = Some(occurrences);
+
         if let Some(invalidate_from) = invalidate_from {
             self.valid_search.union_one_range(start, invalidate_from);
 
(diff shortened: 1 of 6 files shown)


cmyr:
Just checking in on this, and I'm wondering if it might make sense to do some of this in separate PRs? For instance the very first step of separating the existing find functionality out into its own file is something that we could review and merge on its own, and further work could build from there. This has the advantage of giving us smaller, more focused diffs to read.

scholtzan:
Yes, I guess that would make it easier to review. I will try to put the regex support in a separate one.
I was just unsure what to do about pull requests that depend on changes made in another pull request. But I guess, ideally I should wait for the changes to get approved before implementing more features that depend on them.

cmyr:
@scholtzan github isn't great at supporting that workflow, unfortunately, but there are a number of possible approaches. The easiest would be to open your first PR as usual, and then open your second PR _including_ the commits from your first PR, and some flag like [WIP], and a little note at the start of the PR that it includes extra commits and is dependent on #xxx; then after the first gets merged you can rebase the second, and repeat as necessary. This lets you keep working, well also making it clearer for reviewers, at the cost of a bit of extra work.

cmyr:
@scholtzan Whenever you think this is ready for a look, let me know.

We should also discuss what UI you will need in order to test this, and what you would like the RPC protocol to look like? The protocol stuff is a good thing to open a separate issue for.

scholtzan:
I'd say you can take a look. One thing that is not ideal in the current implementation is that all the entire line cache is invalidated after each find operation. I think, it would be better to only invalidate visible lines.

I will open a separate issue in xi mac for the UI and another one for the RPC here.

cmyr:
@scholtzan sounds good, thanks!

rust/core-lib/src/find.rs
```diff
@@ -0,0 +1,304 @@
...
+//  query_id: usize,      // id or reference to SearchQuery?
+//  highlight: Selection
+//}
+//
+
+
+
+const BACKWARDS_FIND_CHUNK_SIZE: usize = 32_768;
+
+pub struct Find {
```
cmyr: Nit: any public type should have at least a short /// doc comment.

rust/core-lib/src/find.rs
```diff
@@ -0,0 +1,304 @@
+use std::cmp::{min,max};
```
cmyr: Also, underneath the copyright notice, all of our modules (hopefully) have a very short (often one line) //! top level doc comment; here that would be something like "Logic and state management for search and replace"

rust/core-lib/src/lib.rs
```diff
@@ -76,6 +76,7 @@ pub mod internal {
     pub mod edit_types;
     pub mod event_context;
     pub mod file;
+    pub mod find;
```
cmyr: 👍 for alphabetical order :)

rust/core-lib/src/view.rs
```diff
@@ -20,17 +20,14 @@ use std::ops::Range;
 use serde_json::Value;
 
 use xi_rope::rope::{Rope, LinesMetric, RopeInfo};
-use xi_rope::delta::{Delta, DeltaRegion};
-use xi_rope::tree::{Cursor, Metric};
+use xi_rope::delta::{Delta};
```
cmyr: No need for braces on a single item import.

You might also try running clippy on this, to make sure we're keeping things clean(ish).

rust/core-lib/src/view.rs
```diff
@@ -73,20 +69,11 @@ pub struct View {
...
-    /// The currently active search string
-    search_string: Option<String>,
-    /// The case matching setting for the currently active search
-    case_matching: CaseMatching,
-    /// The set of all known find occurrences (highlights)
-    occurrences: Option<Selection>,
-    /// Set of ranges that have already been searched for the currently active search string
-    valid_search: IndexSet,
+    /// The state for finding text for this view.
+    find: Find,
```
cmyr: Semantically, I wonder if this shouldn't be an `Option<Find>`?

rust/core-lib/src/view.rs
```diff
@@ -819,6 +789,7 @@ impl View {
 
     pub fn do_find(&mut self, text: &Rope, chars: Option<String>,
                    case_sensitive: bool) -> Value {
+        // todo
```
cmyr: ?

rust/core-lib/src/find.rs
scholtzan: I can add it. It seems there are several modules and public types that don't have the doc or top level doc comment. Like `View` and `Editor`.

rust/core-lib/src/find.rs
```diff
@@ -0,0 +1,304 @@
...
+//  highlight: Selection
+//}
+//
+
+
+
+const BACKWARDS_FIND_CHUNK_SIZE: usize = 32_768;
+
+pub struct Find {
+    // todo: support multiple queries
```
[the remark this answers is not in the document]
scholtzan: Good idea. That would change the logic behind getting the next occurrence a bit. But should be possible.

rust/core-lib/src/view.rs
scholtzan: If we have one `Find` instance for each query then they would be stored in `Vec` otherwise `Option<Find>`.

rust/core-lib/src/find.rs
cmyr: Yea, this may be missing from some older files, but we've been trying to get more consistent with it with new files.

scholtzan:
I fixed the issues you mentioned and multiple queries are now represented as multiple instances of the `Find` struct.

scholtzan:
I updated the find logic so all occurrences are determined and not only the ones that are visible.

rust/core-lib/src/find.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// The case matching setting for the currently active search
+    case_matching: CaseMatching,
+    /// The set of all known find occurrences (highlights)
+    occurrences: Option<Selection>,
+    /// Set of ranges that have already been searched for the currently active search string
+    valid_search: IndexSet,
+}
+
+impl Find {
+    pub fn new() -> Find {
```
cmyr: This is a place where you might consider the `Default` trait, which I personally prefer anytime my constructor has no arguments. Definitely not a big deal though, and probably comes down to preference.

rust/core-lib/src/view.rs
```diff
@@ -163,12 +146,10 @@ impl View {
             Gesture { line, col, ty } =>
                 self.do_gesture(text, line, col, ty),
             GotoLine { line } => self.goto_line(text, line),
-            FindNext { wrap_around, allow_same } =>
-                self.find_next(text, false,
-                               wrap_around.unwrap_or(false),
-                               allow_same.unwrap_or(false)),
+            FindNext { wrap_around, allow_same: _ } =>
+                self.find_next(text, false, wrap_around.unwrap_or(false)),
```
cmyr: not for this PR, but an observation: instead of having `wrap_around` be an `Option<bool>`, I would have serde give it a default value if it's missing in the RPC.

rust/core-lib/src/view.rs
```diff
@@ -612,9 +601,7 @@ impl View {
                     } else {
                         let start_line = seg.our_line_num;
                         let end_line = start_line + seg.n;
-                        if self.hls_dirty {
-                            self.update_find_for_lines(text, start_line, end_line);
-                        }
```
cmyr: I don't totally understand the logic here.. is this related to how we now find all instances, instead of just for the active region?

cmyr:
Oh @scholtzan just noticed that I'm getting some errors in the console log when I run this on xi-mac:

```javascript
unknown json from core: ["id": 11, "error": {
    code = "-32600";
    data = "missing field `case_sensitive`";
    message = "Invalid request";
}]
```
so not totally sure what that's about, I do recall maybe seeing something like this at some point previously?

scholtzan:
@cmyr It seems to be a bug in xi-mac. When the search field is closed another `find` command is sent but for this one the `case_sensitive` field is missing (see `Search.swift` in `func find`).

rust/core-lib/src/find.rs
```diff
@@ -0,0 +1,253 @@
...
+    // todo: link to search query so that search results can be correlated back to query
+
+    /// The occurrences, which determine the highlights, have been updated.
+    hls_dirty: bool,
+    /// The currently active search string
+    search_string: Option<String>,
+    /// The case matching setting for the currently active search
+    case_matching: CaseMatching,
+    /// The set of all known find occurrences (highlights)
+    occurrences: Option<Selection>,
```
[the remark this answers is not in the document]
scholtzan: Yes, true

rkusa:
> I updated the find logic so all occurrences are determined and not only the ones that are visible.

Sorry for dropping in without having the whole context. I am just wondering about the reasoning for searching the whole file instead of just the visible area? I recall multiple pros and cons for both approaches and just curious to know what triggered the change ☺️

cmyr:
@rkusa Primarily we want to be able to show a count of the total results. iirc our primary motivation for doing it in chunks was performance, but we should be able to adapt this in the future to still be incremental, but we can do that using the idle handler, so the result count will still converge on the correct result eventually.

There was also a drawback to the current approach that we stumbled on during some refactors, which is that the logic for continuing the find operation was tightly coupled with the logic for rendering the view; by trying to keep that more separate it will be easier to work on one without potentially breaking the other. :)

rust/core-lib/src/find.rs
```diff
@@ -226,31 +222,29 @@ impl Find {
...
+            let next_occurrence = self.occurrences.regions_in_range(sel_end + 1, text.len()).first();
 
-                next_occurrence
+            if next_occurrence.is_none() && !wrapped {
+                return self.occurrences.regions_in_range(0, text.len()).first().cloned();
             }
-        }).cloned()
+
+            next_occurrence.cloned()
+        }
```
scholtzan: I wonder if there is a more elegant way than adding `cloned()` to each occurrence

rust/core-lib/src/view.rs
```diff
@@ -785,30 +773,9 @@ impl View {
...
-                occurrences.delete_range(new_offset, new_offset + len, false);
-            }
-
-            self.occurrences = Some(occurrences);
-
-            // update find for the whole delta (is going to only update invalid regions)
-            let (iv, _) = delta.summary();
-            self.update_find(text, iv.start(), iv.end(), true, false);
+        // update only find highlights affected by change
+        for find in self.find.iter_mut() {
```
raphlinus: Tiny style nit, could this be written as `for find in &mut self.find`? I tend to prefer that.

rust/rope/src/find.rs
```diff
@@ -76,7 +76,7 @@ pub fn find(cursor: &mut Cursor<RopeInfo>, cm: CaseMatching, pat: &str) -> Optio
 /// 
 /// [find]: fn.find.html
 pub fn find_progress(cursor: &mut Cursor<RopeInfo>, cm: CaseMatching, pat: &str,
-    num_steps: usize) -> FindResult
+                     num_steps: usize) -> FindResult
```
raphlinus: These are just whitespace changes? Maybe we should institute rustfmt so we don't have to have these discussions in PR's. It looks like this is a change from "block" to "visual" for fn args, in the [rustfmt lingo](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#function-arguments). The former is the default, and also my preference, as it allows more room for arguments.

rust/core-lib/src/find.rs
raphlinus: I'd be fine with `SelRegion` being Copy, if that works out.

rust/rope/src/find.rs
scholtzan: I got confused here. I saw that in many files the "visual" style is used. Now that I look at some other files, I also see the "block" style is used. I could certainly change it back if that is wished.

rust/core-lib/src/find.rs
scholtzan: `Copy` is already derived for `SelRegion`. The problem was that instead of `Option<SelRegion>` the derived clone method will create a reference `Option<&SelRegion>` (I think it has something to do with this: https://github.com/rust-lang/rust/issues/26925). I could implement `Copy` manually, but not sure if that is better.

rust/rope/src/find.rs
cmyr: Yea, I tend to use visual indenting, which I prefer for oblique aesthetic reasons, although as a general rule I  try not to change whitespace in a file unless it's just clearly _wrong_.

 I have a bunch of similar rust style idiosyncrasies that I would be happy to abandon if we want to just go ahead and punt to rustfmt.

rust/core-lib/src/view.rs
cmyr: Just want to say for the record that I prefer the original, I think this is exactly what `iter_mut()` is for. But I 100% don't mind raph's approach either. 😏

rust/rope/src/find.rs
scholtzan: Okay, should I change it back for this to get merged?

raphlinus: I don't have a strong feeling and am ok with it being merged either way.

PR #651, opened 2018-05-10
[WIP: Review Needed] Made Plugin Updates Async

betterclever:
This patch removes the Synchronous Plugin Updates API in favour of new Async API using Notifications added in #367 

Tracking Issue: #466

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -23,7 +23,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "byteorder 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.43 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -82,7 +82,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cfg-if"
-version = "0.1.2"
+version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -131,7 +131,7 @@ name = "filetime"
 version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "redox_syscall 0.1.37 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -216,7 +216,7 @@ name = "log"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -279,7 +279,7 @@ name = "net2"
 version = "0.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cfg-if 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cfg-if 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.40 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -331,7 +331,7 @@ dependencies = [
  "base64 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.43 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
  "xml-rs 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -371,18 +371,18 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.43"
+version = "1.0.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "serde_derive"
-version = "1.0.43"
+version = "1.0.45"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "proc-macro2 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "quote 0.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive_internals 0.23.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 0.13.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.13.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -391,7 +391,7 @@ version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "proc-macro2 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "syn 0.13.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "syn 0.13.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -401,7 +401,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "dtoa 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "itoa 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "serde 1.0.43 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -422,7 +422,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "syn"
-version = "0.13.4"
+version = "0.13.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "proc-macro2 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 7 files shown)


rust/core-lib/src/core.rs
```diff
@@ -144,8 +144,8 @@ impl WeakXiCore {
     /// Handles the result of an update sent to a plugin.
     ///
     /// All plugins must acknowledge when they are sent a new update, so that
-    /// core can track which revisiions are still 'live', that is can still
-    /// be the base revision for a delta. Once a plugin has acknowleged a new
```
cmyr: 👍 this is embarrassing :)

rust/plugin-lib/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub trait Plugin {
     /// Called when an edit has occured in the remote view. If the plugin wishes
     /// to add its own edit, it may return `Some(edit)`.
     fn update(&mut self, view: &mut View<Self::Cache>, delta: Option<&RopeDelta>,
-              edit_type: String, author: String) -> Option<PluginEdit>;
+              edit_type: String, author: String) -> Option<u64>;
```
cmyr: This can now return nothing; the plugin lib can handle acking the update without the plugin needing to know about it.

rust/syntect-plugin/src/main.rs
```diff
@@ -223,10 +222,10 @@ impl<'a> Syntect<'a> {
     /// Checks if a newline has been inserted, and if so inserts whitespace
     /// as necessary.
     fn do_indentation(&mut self, view: &mut MyView, start: usize,
-                      end: usize, text: &str) -> Option<PluginEdit> {
+                      end: usize, text: &str) -> Option<u64> {
```
cmyr: ditto, since we don't need to return an edit we don't need to return anything.

rust/syntect-plugin/src/main.rs
```diff
@@ -249,15 +249,24 @@ impl<'a> Syntect<'a> {
...
                 let mut builder = EditBuilder::new(buf_size);
                 builder.replace(interval, indent.into());
+
                 let delta = builder.build();
-                return Some(PluginEdit {
-                    rev: rev,
-                    delta: delta,
+                Some(delta)
+            } else {
+                None
```
cmyr: could we not just return in the case that there's no delta? Then `result` doesn't need to be an option, and the following code is a bit cleaner.

rust/syntect-plugin/src/main.rs
```diff
@@ -317,20 +326,19 @@ impl<'a> Plugin for Syntect<'a> {
         self.do_highlighting(view);
     }
 
-    fn config_changed(&mut self, _view: &mut View<Self::Cache>, _changes: &ConfigTable) {
-    }
+    fn config_changed(&mut self, _view: &mut View<Self::Cache>, _changes: &ConfigTable) {}
 
     fn update(&mut self, view: &mut View<Self::Cache>, delta: Option<&RopeDelta>,
-              _edit_type: String, _author: String) -> Option<PluginEdit> {
+              _edit_type: String, _author: String) -> Option<u64> {
```
cmyr: ditto not needing to return anything here

rust/Cargo.lock
```diff
@@ -23,7 +23,7 @@ version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
cmyr: I would remove the Cargo.lock update from this PR, we can update it separately later.

rust/core-lib/src/event_context.rs
```diff
@@ -368,14 +368,8 @@ impl<'a> EventContext<'a> {
     pub(crate) fn do_plugin_update(&mut self, update: Result<Value, RpcError>,
```
cmyr: we can get rid of the TODO above :)

rust/core-lib/src/event_context.rs
```diff
@@ -368,14 +368,8 @@ impl<'a> EventContext<'a> {
...
 
-        match update.map(serde_json::from_value::<UpdateResponse>) {
-            Ok(Ok(UpdateResponse::Edit(edit))) => {
-                let author = edit.author.clone();
-                self.editor.borrow_mut().apply_plugin_edit(edit, Some(undo_group));
-                self.after_edit(&author);
-                self.render_if_needed();
-            }
-            Ok(Ok(UpdateResponse::Ack(_))) => (),
+        match update.map(serde_json::from_value::<u64>) {
```
cmyr: here we don't actually need to parse the value at all; as long as this is _any_ `Ok(_)` we should consider the message ack'd.

rust/plugin-lib/src/view.rs
```diff
@@ -131,6 +131,15 @@ impl<C: Cache> View<C> {
         self.peer.send_rpc_notification("add_scopes", &params);
     }
 
+    pub fn edit(&self, edit: PluginEdit) {
```
cmyr: instead of taking a `PluginEdit`, this should take just the arguments determined by the plugin (such as the delta, priority, etc) and then construct the `PluginEdit` in place.

rust/syntect-plugin/src/main.rs
betterclever: I thought that we needed to return an Int acknowledgement here, but I guess that is not the case. The plugin_lib sends an acknowledgement when plugin receives any event to the plugin. Right?

rust/syntect-plugin/src/main.rs
betterclever: I could not call the edit function on the view while the view is mutably borrowed for "get_line", thus I used 'if expression' to return the result instead and utilize it later. The same thing I discussed in IRC.

rust/syntect-plugin/src/main.rs
cmyr: Correct: whenever the plugin receives a new revision from core (via the `update` rpc) the plugin sends an acknowledgement back. The contract here is that the plugin will not subsequently send any edits to core that are based of revisions older than the most recently ack'd.

This is an Int for convenience, but the plugin lib can just fill that in on that side, and the return value has no meaning beyond an ack. It could just as easily be any other type.

cmyr: To clarify: because the `View` object actually constructs any edits, we know that once the plugin lib has received the update, and has updated it's internal `rev`, it will no longer be using any previous value. The actual plugin doesn't need to know any of this; it's just bookkeeping. Therefore the plugin-lib can update the plugin without needing a response; the plugin-lib will always respond to the core RPC to acknowledge, independent of any action taken by the particular plugin.

rust/core-lib/src/plugins/manager.rs
```diff
@@ -99,7 +99,8 @@ impl PluginManager {
                 let buffers = self.buffers.clone().to_weak();
                 let mut plugin_ref = plugin.clone();
 
-                plugin.update(&update, move |response| {
+                plugin.update(&update);
+                /*plugin.update(&update, move |response| {
```
betterclever: @cmyr What was being done in this method earlier? Are similar changes required somewhere else now.

cmyr: No, this is not currently used and will be removed shortly.

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -75,6 +75,10 @@ pub struct PluginUpdate {
     /// The total number of lines in the document after applying this delta.
     pub new_line_count: usize,
     pub rev: u64,
+    /// The undo_group associated with this update. Each PluginEdit sent by plugin is associated
```
cmyr: I would reword the second sentence a little, to say:

"The plugin may pass this value back to core when making an edit, to associate the plugin's edit with this undo group." (then continue with your sentence that starts, "Core uses ...")

rust/plugin-lib/src/lib.rs
```diff
@@ -107,9 +107,9 @@ pub trait Plugin {
 
     //TODO: async edits only; this is here for feature parity during initial hacking
     /// Called when an edit has occured in the remote view. If the plugin wishes
-    /// to add its own edit, it may return `Some(edit)`.
+    /// to add its own edit, it must do so using asynchronously via edit notification.
```
cmyr: Good stuff, it's important to keep docs up to date. I would say "via **the** edit notification".

rust/plugin-lib/src/lib.rs
```diff
@@ -107,9 +107,9 @@ pub trait Plugin {
 
     //TODO: async edits only; this is here for feature parity during initial hacking
```
cmyr: we can delete this TODO

rust/plugin-lib/src/view.rs
```diff
@@ -131,7 +131,10 @@ impl<C: Cache> View<C> {
         self.peer.send_rpc_notification("add_scopes", &params);
     }
 
-    pub fn edit(&self, edit: PluginEdit) {
+    pub fn edit(&self, delta: RopeDelta, priority: u64, after_cursor: bool,
+                undo_group: Option<usize>, author: String) {
```
cmyr: This is just about right, but one little change: instead of `undo_group: Option<usize>` I'd have an argument `new_undo_group: bool`; then we just keep track of the undo group internally, like we do with `rev`, and send it to core whenever `new_undo_group` is `false`.

rust/syntect-plugin/src/main.rs
```diff
@@ -222,10 +222,10 @@ impl<'a> Syntect<'a> {
     /// Checks if a newline has been inserted, and if so inserts whitespace
     /// as necessary.
     fn do_indentation(&mut self, view: &mut MyView, start: usize,
-                      end: usize, text: &str) -> Option<u64> {
+                      end: usize, text: &str, undo_group: Option<usize>) {
```
cmyr: related to my previous comment: we know in this case that we always want an undo group, so we don't need to pass this through.

rust/syntect-plugin/src/main.rs
```diff
@@ -329,19 +320,18 @@ impl<'a> Plugin for Syntect<'a> {
     fn config_changed(&mut self, _view: &mut View<Self::Cache>, _changes: &ConfigTable) {}
 
     fn update(&mut self, view: &mut View<Self::Cache>, delta: Option<&RopeDelta>,
-              _edit_type: String, _author: String) -> Option<u64> {
+              undo_group: Option<usize>, _edit_type: String, _author: String) {
```
cmyr: ditto earlier, all of syntect's edits will be part of the active undo group so we don't need to pass the argument around.

betterclever:
@cmyr Did the suggested changes. 😄

rust/syntect-plugin/src/main.rs
```diff
@@ -249,17 +247,18 @@ impl<'a> Syntect<'a> {
...
-                    author: "syntect".to_owned(),
-                })
+                Some(delta)
+            } else {
+                None
+            };
+
+            if let Some(delta) = result {
+                view.edit(delta, INDENTATION_PRIORITY, false,
+                          false, String::from("syntect".to_owned()));
```
cmyr: Didn't notice this earlier, but this is redundant; you can just do `"syntect".into()` or `"syntect".to_string()` _or_ `String::from("syntect")`. Alternatively you could let `view.edit(..)` take a `&str`, and then you can do the conversion internally. Alternatively there's a pattern we use a lot where the function takes a generic `S: AsRef<str>`, see [this nice explanation](https://deterministic.space/elegant-apis-in-rust.html#use-conversion-traits).

rust/plugin-lib/src/lib.rs
```diff
@@ -129,7 +128,7 @@ pub trait Plugin {
                       view: &mut View<Self::Cache>,
                       changes: &ConfigTable);
 
-    /// Called when the runloop is idle, if the plugin has prevoiusly
+    /// Called when the runloop is idle, if the plugin has preciously
```
raphlinus: This should be "previously".

betterclever:
Fixed it. Also resolved merge conflicts.

cmyr:
Great, thanks!

Issue #652, opened 2018-05-10
Weird undo behaviour


cmyr:
I suspect this is a regression from #613: I'm seeing some unexpected behaviour when using undo. It's hard to point at particular examples because I don't think the expected behaviour is documented anywhere, but in particular entering text in one location and then clicking in another location and entering more text causes both edits to be part of the same undo group, which feels broken.

Undo behaviour is tricky and subtle (for instance I think there's a reasonable case that we should have a timeout that creates a new undo group; if you type some stuff and then pause for 30s or a minute before resuming, it might be nice if this created a new undo group, because it likely indicates a context switch of some kind.)

Undo behaviour is also one of the things in the editor that feels the most 'magic', in that the state transitions are scattered all over and are hard to ever fully have in memory. I would be curious to see an approach to undo group handling that used an explicit state machine, with well defined conditions and transitions.

raphlinus:
Minimal repro: type "ab" then left arrow, then "c". Undo goes back to empty string, expected result is "ab", ie the "ab" and the "c" should be in different undo groups.

It's pretty easy to understand what changed: before #613 every editor command started with setting `this_edit_type` to `Other` ([code](https://github.com/google/xi-editor/pull/613/files#diff-cdd62803d32b5bd56e7e05ba4fce41abL1077). Thus, on insert of the "c", `last_edit_type` is `Other` (from the cursor move command), while `this_edit_type` is `InsertChars`. Then, when it gets into the logic to compute undo groups ([code](https://github.com/google/xi-editor/pull/613/files#diff-cdd62803d32b5bd56e7e05ba4fce41abL356) decides that there should be two different undo groups.

In the new code, commands dispatched to the View don't touch the editor at all ([code](https://github.com/google/xi-editor/blob/1df4c48fe5c0fc6bc2b112ab5b6876a82efee694/rust/core-lib/src/event_context.rs#L97)). I think there's a super simple fix, which is to make `EventContext::do_edit` reset the `last_edit_type` field for `E::View` events.

cmyr:
fixed in #654

PR #654, opened 2018-05-10
Fix broken undo behavior

LiHRaM:
See #652. Can confirm that this behaves as expected in the minimal repro described by raph, 
i.e. typing "ab", pressing left arrow button and typing "c", undoing will only remove c.


cmyr:
Okay cool, I think this is probably the correct fix. I won't be able to test until after the weekend, but this _looks_ right to me.

PR #658, opened 2018-05-16
[WIP] Find using regular expressions

scholtzan:
Preliminary support of regular expressions when searching for text.

This depends on changes in https://github.com/google/xi-editor/pull/649

### DIFF
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -1,0 +1,320 @@
+use std::cmp::{min,max};
+use std::mem;
+use std::cell::RefCell;
+use std::ops::Range;
+
+use serde_json::Value;
+
+use index_set::IndexSet;
+use xi_rope::delta::{Delta, DeltaRegion};
+use xi_rope::find::{find, CaseMatching};
+use xi_rope::rope::{Rope, LinesMetric, RopeInfo};
+use xi_rope::tree::{Cursor, Metric};
+use xi_rope::breaks::{Breaks, BreaksInfo, BreaksMetric, BreaksBaseMetric};
+use xi_rope::interval::Interval;
+use xi_rope::spans::Spans;
+use xi_trace::trace_block;
+use client::Client;
+use edit_types::ViewEvent;
+use line_cache_shadow::{self, LineCacheShadow, RenderPlan, RenderTactic};
+use movement::{Movement, region_movement, selection_movement};
+use rpc::{GestureType, MouseAction};
+use styles::{Style, ThemeStyleMap};
+use selection::{Affinity, Selection, SelRegion};
+use tabs::{ViewId, BufferId};
+use width_cache::WidthCache;
+use word_boundaries::WordCursor;
+
+pub struct SearchQuery {
+  id: usize,   // necessary?
+  query: String,
+  case_matching: CaseMatching // todo: add regex
+}
+
+pub struct SearchOccurrence {
+  query_id: usize,      // id or reference to SearchQuery?
+  highlight: Selection
+}
+
+
+
+
+const BACKWARDS_FIND_CHUNK_SIZE: usize = 32_768;
+
+pub struct Find {
+  // todo: support multiple queries
+  //  search_queries: Vec<SearchQuery>,
+
+  // todo: occurrences in separate type that references back to the search query (required for coloring, ...)
+  //  occurrences: Vec<SearchOccurrence>,
+
+  // todo: the following will be deprecated after adding support for multiple queries
+  /// The occurrences, which determine the highlights, have been updated.
+  hls_dirty: bool,
+  /// The currently active search string
+  search_string: Option<String>,
+  /// The case matching setting for the currently active search
+  case_matching: CaseMatching,
+  /// The set of all known find occurrences (highlights)
+  occurrences: Option<Selection>,
+  /// Set of ranges that have already been searched for the currently active search string
+  valid_search: IndexSet,
+}
+
+
+
+impl Find {
+  pub fn new() -> Find {
+    Find {
+//      search_queries: Vec::new(),
+//      occurrences: Vec::new(),
+      hls_dirty: true,
+      search_string: None,
+      case_matching: CaseMatching::CaseInsensitive,
+      occurrences: None,
+      valid_search: IndexSet::new(),
+    }
+  }
+
+  pub fn occurrences(&self) -> &Option<Selection> {
+    &self.occurrences
+  }
+
+  pub fn hls_dirty(&self) -> bool {
+    self.hls_dirty
+  }
+
+  pub fn set_hls_dirty(&mut self, is_dirty: bool) {
+    self.hls_dirty = is_dirty
+  }
+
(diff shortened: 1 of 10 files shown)


rust/rope/src/find.rs
```diff
@@ -163,50 +177,111 @@ fn find_core(cursor: &mut Cursor<RopeInfo>, pat: &str,
...
         }
     }
 }
 
+/// Compare whether the substring beginning at the cursor location matches
+/// the provided regular expression. The substring begins at the beginning
+/// of the leave/start of the line.
+/// If the regular expression can match multiple lines then all leaves are
+/// consumed and matched against the regular expression. Otherwise only the
+/// current leave is matched.
```
eyelash: Did you mean `leaf` here? Also on line 228 three lines above.

scholtzan: Yes, right. Thanks!

scholtzan:
This is ready for review

rust/core-lib/src/find.rs
```diff
@@ -120,14 +120,18 @@ impl Find {
     }
 
     /// Sets find parameters and search query.
-    fn set_find(&mut self, search_string: &str,
-                case_sensitive: bool) {
-        let case_matching = if case_sensitive {
-            CaseMatching::Exact
+    fn set_find(&mut self, search_string: &str, case_sensitive: bool, regex: bool) {
+        let case_matching = if regex {
+            CaseMatching::RegularExpression
```
raphlinus: This is related to another comment - I'm wondering why we're throwing away case_sensitive in the regex case, because it is a very reasonable option to pass along to regex.

rust/rope/src/find.rs
```diff
@@ -162,51 +176,116 @@ fn find_core(cursor: &mut Cursor<RopeInfo>, pat: &str,
...
+    if pat.is_empty() {
+        return Some(orig_position);
+    }
+
+    // create regex from untrusted input
+    match RegexBuilder::new(pat).size_limit(REGEX_SIZE_LIMIT).build() {
+        Ok(regex) => {
+            let mut text = String::new();
+
+            match cursor.get_leaf() {
```
raphlinus: So this will work in the common case, but isn't fully correct - there's no guarantee that a line won't be fragmented between multiple leaves. It will definitely happen when lines are long (and if there's direct access to leaves, that case should probably be reflected in tests).

My sense of the best thing to do here is to use a lines iterator, which will abstract over the details of leaves, while being efficient and not allocating when line boundaries line up with leaf boundaries.

I think that means that you don't want to use the same Cursor<RopeInfo> for both string match and regex match; it's fine for string match because you can do a string match on a line fragment, so the efficiency gain in dealing with individual leaves is probably worth it.

raphlinus: I was thinking we could use [`lines_raw`](https://docs.rs/xi-rope/0.2.0/src/xi_rope/rope.rs.html#318), which is probably going to be more efficient than a Cursor-based approach if it's strictly doing a forward scan. Is there some reason that won't work?

raphlinus: Thinking about this, one difference between Cursor and LinesRaw is that in the latter case the client will have to keep track of the offset. But that can be done by adding up `line.len()` as we go along.

I can also see how to make a `Cursor` work, but it would at the very least require duplicating some of the logic we already have implemented in LinesRaw.

scholtzan: I updated this pull request so that for regex search a line iterator is used. I am not sure if that is the best way to implement it.

cmyr:
@scholtzan what's the status of this? There are some CI failures that should get cleared up, but is it otherwise ready to go? I think we should get some of this backlog merged, but I'm not sure if this should go in before #687 & #688. This also needs a xi-mac companion, but I'll do that.

scholtzan:
I will fix the test. This should go in after https://github.com/google/xi-editor/pull/687 and maybe also after https://github.com/google/xi-editor/pull/688 (though it should be independent from that one).

scholtzan:
One thing I noticed: if the highlights get updated while the search query is a multiline regex, then for larger text there is a noticable delay. 
The reason for this is that all the highlights after the edit position get recalculated since they might have changed. I am not sure if there is a good way to solve this problem.

cmyr:
@scholtzan maybe if we're in regex mode the frontend should only send the query on <kbd>return</kbd>?

cmyr:
Good point, I wasn't thinking about updating the highlights during a document edit.

This will be easier when search is incremental, and we can just clear the edited region immediately and schedule the runloop to re-search the rest of the document.

I'm not sure what our best bet is for now; it might be that in the regex case we want to clear the highlights while typing. Alternatively, we could go ahead and implement incremental search. There's definitely a performance issue on a debug build though, enough that i wonder if there isn't something else going on.

rust/rope/src/find.rs
```diff
@@ -137,81 +152,164 @@ fn find_progress_iter(cursor: &mut Cursor<RopeInfo>, pat: &str,
...
+        .size_limit(REGEX_SIZE_LIMIT)
+        .case_insensitive(cm == CaseMatching::CaseInsensitive)
+        .build();
+
+    match regex {
+        Ok(regex) => {
+            let mut text = String::new();
+
+            if is_multiline_regex(pat) {
+                // consume all of the text if regex is muli line matching
```
raphlinus: typo: "multi"

rust/rope/src/find.rs
```diff
@@ -137,81 +152,164 @@ fn find_progress_iter(cursor: &mut Cursor<RopeInfo>, pat: &str,
...
+                text.extend(lines);
+            } else {
+                match lines.next() {
+                    Some(line) => text.push_str(&line),
```
raphlinus: This seems ok for now, but is a copy of every byte of the source text. What I'd prefer is that `text` is a `Cow<str>`, and that in the single line case, it just copies the `Cow` returned by the lines iterator (meaning that if it's a borrowed reference, it just copies the reference, not the actual bytes).

Since you're reusing the buffer, at least you're not going to get an allocation per line, only a copy, which is not too bad. I wouldn't be surprised if the cost of the regex evaluation dominated in most cases.

What I think I'm most comfortable with is accepting this code as-is, and having a future work item to avoid the copy based on `Cow`.

scholtzan: Okay, that sounds reasonable. I will make a separate PR for this later.

scholtzan:
The regex is now cached in `Find` and gets only rebuilt when the search query changes. This makes find with regex _a lot_ faster.

coin723:
It'd be a foolish question, but is it possible to search with regex on the editor which is built from the latest master branch? If so, how can I use in the find feature?
I'm using xi-mac.

jansol:
This is the wrong repo for that, but you can switch to regex search by clicking on the magnifying glass icon on the left side of the find bar.

cmyr:
@coin723 no worries, the situation is a bit confusing, but xi-mac now lives in its own repo: https://github.com/google/xi-mac

PR #660, opened 2018-05-16
Put back multi-plugin support

cmyr:
This is a quick temporary (famous last words) fix to allow support for
running multiple plugins. This is chiefly intended to enable the GSoC
LSP work.

With this patch, all plugins are launched on startup, and are global to
all views. Plugins can be manually stopped and started via the debug
menu.


cmyr:
@betterclever Can I get a review of this from you if you have a few minutes?

betterclever:
This looks good for the initial hacking with LSP for me. Thanks.

PR #671, opened 2018-05-24
Send rpc notification to plugins on config_changed

nangtrongvuon:
This was done with help from @cmyr during a session, where we found that plugins weren't getting sent an RPC notification when the config changes. This PR fixes that.

### DIFF
--- a/rust/core-lib/src/plugins/mod.rs
+++ b/rust/core-lib/src/plugins/mod.rs
@@ -30,6 +30,7 @@ use xi_rpc::{self, RpcPeer, RpcLoop};
 use xi_trace;
 
 use WeakXiCore;
+use config::Table;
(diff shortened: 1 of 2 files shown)


rust/core-lib/src/plugins/mod.rs
```diff
@@ -117,7 +118,13 @@ where F: FnOnce(Result<Value, xi_rpc::Error>) + Send + 'static
     }
 
     pub fn collect_trace(&self) -> Result<Value, xi_rpc::Error> {
-        self.peer.send_rpc_request("collect_trace", &json!({}))
+       self.peer.send_rpc_request("collect_trace", &json!({}))
```
cmyr: we accidentally changed this whitespace :(

Issue #645, opened 2018-05-09
Language support tracking issue


cmyr:
## Overview

This is intended as a tracking issue for 'language support', that is the ability for the editor to provide language-specific functionality, and specifically the ability for plugins to add support for _new_ languages.

A while ago we added `syntax.rs` as a simple placeholder for language support, so that we could write a first pass of the config/preferences system. In the longer term, however, we would like languages and language support to be provided by plugins.

How does a user go about adding support for a new language to xi? How are languages assigned to buffers? How are plugins activated in response to language changes? How do we determine which plugin is responsible for providing a set of features, when multiple such plugins are available?

### Language != syntax/grammar

A _language_ is a way of identifying the structure of a particular document. A _grammar_ or a _syntax_ (used interchangeably here) is a set of rules that describe a language, particularly in the context of providing syntax highlighting. A _language_ does not require a grammar, e.g. in the case of the 'Plaintext' language.

In most editors languages and syntaxes are defined together; for instance in ST3 via `sublime-syntax` files or in Textmate (and many other editors) via `.tmLanguage` files.

This works in those editors because those editors have a single central engine for doing syntax highlighting, based on rules that can be described in files.

Unlike in these editors, xi does not treat syntax highlighting as a feature of the core editor; instead it is provided by plugins. Clients (such as xi-mac) _may_ include a general purpose syntax highlighting plugin, based on syntect, which _can_ use these files, but that is external to xi-core itself.

In these other editors, adding support for a new language is a matter of placing a new `.tmLanguage`/`.sublime-syntax` package in a particular location where it can be discovered and loaded by the editor.

This doesn't work in xi, because there's no single mechanism for describing a language's grammar.

One editor which _does_ take a different approach is VSCode. In Code, languages and grammars are treated as [separate entities in a plugin manifest](https://code.visualstudio.com/docs/extensions/themes-snippets-colorizers#_add-to-your-language-support-extension), although it looks like grammars are still always `.tmLanguage` files.

I think this offers a good starting point for xi; like in Code, xi will probably use [_language definitions_](https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributeslanguages) that are distinct from a grammar/syntax.

PR #672, opened 2018-05-24
 Replace hard coded 'SyntaxDef's with plugin provided LanguageDefinitions

cmyr:
All current language support is now provided by the syntect plugin; the syntect plugin's manifest is generated by a small program (`syntect-plugin/examples/make_manifest.rs`) that  parses them out of syntect itself.

This should 'just work' when syntect is installed via the Makefile; for xi-mac I'll have a patch shortly that handles this in its build script.

I've tried to keep this PR fairly minimal; there's some cleanup possible, but I'll keep that for follow-on work.

Also: this bumps our minimum supported rustc to 1.26. I could've avoided that, but I think we'll want to get there pretty soon anyway, so why not now.

cc @eyelash: let me know if this actually works for xi-gtk?
cc @betterclever: you might want to give this a quick look, it's plugin related

progress on #645

### DIFF
--- a/rust/syntect-plugin/Makefile
+++ b/rust/syntect-plugin/Makefile
@@ -4,12 +4,27 @@ XDG_CONFIG_HOME ?= $(HOME)/.config
 XI_CONFIG_DIR ?= $(XDG_CONFIG_HOME)/xi
 XI_PLUGIN_DIR ?= $(XI_CONFIG_DIR)/plugins
 
-target/release/xi-syntect-plugin:
+lang_configs = assets/*.toml
+
+out/syntect: $(lang_configs) out/gen_manifest.toml target/release/xi-syntect-plugin
+	mkdir -p out/syntect/bin
+	cp $(lang_configs) out/syntect
+	install target/release/xi-syntect-plugin out/syntect/bin
+	cp out/gen_manifest.toml out/syntect/manifest.toml
+
+out/gen_manifest.toml: Cargo.toml Cargo.lock
+	cargo run --release --example make_manifest
+
+target/release/xi-syntect-plugin: src/*.rs Cargo.toml Cargo.lock
 	cargo build --release
 
-install: manifest.toml target/release/xi-syntect-plugin
-	install -d $(XI_PLUGIN_DIR)/syntect/bin
-	install manifest.toml $(XI_PLUGIN_DIR)/syntect
-	install target/release/xi-syntect-plugin $(XI_PLUGIN_DIR)/syntect/bin
+install: out/syntect
+	mkdir -p $(XI_PLUGIN_DIR)
+	cp -r out/syntect $(XI_PLUGIN_DIR)/syntect
+
+clean:
+	rm out/gen_manifest.toml
+	rmdir out
+	cargo clean
 
 .PHONY: install
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ below.
 ### Building the core
 
 Xi targets 'recent stable Rust'. We recommend installing via [rustup](https://www.rustup.rs).
-The current minimum supported version is 1.25, although we expect to require 1.26 shortly.
+The current minimum supported version is 1.26.
 
 To build the xi editor from the root directory of this repo:
 
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -37,7 +37,7 @@ mod defaults {
 
     /// A cache of loaded defaults.
     lazy_static! {
-        static ref LOADED: Mutex<HashMap<ConfigDomain, Option<Table>>> = {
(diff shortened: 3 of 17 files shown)


rust/syntect-plugin/Makefile
```diff
@@ -4,12 +4,27 @@ XDG_CONFIG_HOME ?= $(HOME)/.config
...
+out/syntect: $(lang_configs) out/gen_manifest.toml target/release/xi-syntect-plugin
+	mkdir -p out/syntect/bin
+	cp $(lang_configs) out/syntect
+	install target/release/xi-syntect-plugin out/syntect/bin
+	cp out/gen_manifest.toml out/syntect/manifest.toml
+
+out/gen_manifest.toml: Cargo.toml Cargo.lock
+	cargo run --release --example make_manifest
+
+target/release/xi-syntect-plugin: src/*.rs Cargo.toml Cargo.lock
```
eyelash: Thanks for changing this. I noticed that the makefile would never rebuild the syntect plugin if it was already built but I never got around to fixing this.

rust/syntect-plugin/Makefile
```diff
@@ -4,12 +4,27 @@ XDG_CONFIG_HOME ?= $(HOME)/.config
...
 	cargo build --release
 
-install: manifest.toml target/release/xi-syntect-plugin
-	install -d $(XI_PLUGIN_DIR)/syntect/bin
-	install manifest.toml $(XI_PLUGIN_DIR)/syntect
-	install target/release/xi-syntect-plugin $(XI_PLUGIN_DIR)/syntect/bin
+install: out/syntect
+	mkdir -p $(XI_PLUGIN_DIR)
+	cp -r out/syntect $(XI_PLUGIN_DIR)/syntect
```
eyelash: This line will only work the first time it is run. The second time it will copy the `syntect` folder into the already existing `syntect` folder so you end up with `$(XI_PLUGIN_DIR)/syntect/syntect/bin/xi-syntect-plugin`.

cmyr:
@eyelash Patched up the makefile, I think? Please sign off though, writing make for me still feels like I'm performing some magic incantation.

eyelash:
Looks good to me now. I wonder whether there's a difference between
```bash
mkdir -p $(XI_PLUGIN_DIR)/syntect
cp -r out/syntect/* $(XI_PLUGIN_DIR)/syntect
```
and
```bash
mkdir -p $(XI_PLUGIN_DIR)
cp -r out/syntect $(XI_PLUGIN_DIR)
```
If there isn't then I might prefer the latter.

rust/syntect-plugin/Makefile
```diff
@@ -4,12 +4,26 @@ XDG_CONFIG_HOME ?= $(HOME)/.config
 XI_CONFIG_DIR ?= $(XDG_CONFIG_HOME)/xi
 XI_PLUGIN_DIR ?= $(XI_CONFIG_DIR)/plugins
 
-target/release/xi-syntect-plugin:
+lang_configs = assets/*.toml
+
+out/syntect: $(lang_configs) out/gen_manifest.toml target/release/xi-syntect-plugin
+	mkdir -p out/syntect/bin
+	cp $(lang_configs) out/syntect
+	install target/release/xi-syntect-plugin out/syntect/bin
```
eyelash: Why not use `cp` here as well for the sake of consistency?

cmyr: @eyelash I believe install also sets +x?

eyelash: Shouldn't that already be set by cargo?

eyelash:
In order to make thing work with paths that contain spaces, you need to quote these paths. I just tried with
```makefile
install: out/syntect
	mkdir -p "$(XI_PLUGIN_DIR)"
	cp -r out/syntect "$(XI_PLUGIN_DIR)"
```
and it seemed to work. I guess we're all just learning.

One other alternative would be to replace the makefile with a bash script or even a python script.

cmyr:
On mac, if I use quotes and `$XI_PLUGIN_DIR` includes a `~` it isn't expanded. This is currently working for me _without_ the `"`, though, with just

```makefile
install: out/syntect
	mkdir -p $(XI_PLUGIN_DIR)
	cp -r out/syntect $(XI_PLUGIN_DIR)
```

And this is working even when there's a space in `$XI_PLUGIN_DIR`, so I'm not sure exactly what problem I was having, earlier? In any case, updated.

cmyr:
@eyelash is this resolved to your satisfaction? The manifest is no longer generated by the makefile, so we're back to about the status quo.

eyelash:
I guess it's ok for now, especially if it enables us to make progress. I'm also not a big fan of the current solution, though. I wish we had an easy cross-platform solution for installing the syntect plugin.

Issue #637, opened 2018-05-04
Add custom theme support


cmyr:
This isn't super high priority but it's adjacent to some other things i'm looking at right now so wanted to add an issue in case someone is interested, it would be a fun little project.

Approximate work involved:

1. When the editor receives the `client_started` RPC, check if the `config_dir` argument is present; if it is, look for theme files in a `themes` subdirectory, and if any are present attempt to load them into the `ThemeSet` (owned by `ThemeStyleMap` in `styles.rs`).
2. sent the client the `available_themes` notification.
3. Ideally we would also use the file watcher to be notified on the fly if new themes are added; check how we do this already for config files.
4. Ensure that these loading operations show up in traces; see other uses of `trace_block` in the codebase.

Eventually we would also want to be caching binary dumps of all custom themes, so we can load them more quickly.

virattara:
Interested!

cmyr:
@virattara great, let me know if you have any questions!

scholtzan:
This has been resolved in https://github.com/google/xi-editor/pull/675

PR #675, opened 2018-05-27
Load custom theme files

virattara:
#637 
Completed:
* Load custom themes on startup.
* Handle fs events inside the themes directory.
* Caching

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -25,7 +25,7 @@ xi-rpc = { path = "../rpc", version = "0.2.0" }
 [dependencies.syntect]
 version = "2.0"
 default-features = false
-features = ["assets","dump-load-rs"]
+features = ["assets","dump-load-rs","dump-create"]
 
 [target."cfg(target_os = \"fuchsia\")".dependencies]
 sha2 = "0.7"
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -487,6 +487,15 @@ impl ConfigManager {
         let _: BufferItems = serde_json::from_value(defaults.into())?;
         Ok(())
     }
+
+    /// Path to themes sub directory inside config directory.
+    pub(crate) fn get_themes_dir(&self) -> Option<PathBuf> {
+        let themes_dir = self.config_dir.as_ref()
+                                .map(|p| p.join("themes"));
+        let exists = themes_dir.as_ref()
+                                .map_or(false, |p| p.exists());
+        if exists { themes_dir } else { None }
+    }
 }
 
 impl TableStack {
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -489,12 +489,18 @@ impl ConfigManager {
     }
 
     /// Path to themes sub directory inside config directory.
+    /// Creates one if not present.
     pub(crate) fn get_themes_dir(&self) -> Option<PathBuf> {
         let themes_dir = self.config_dir.as_ref()
             .map(|p| p.join("themes"));
-        let exists = themes_dir.as_ref()
-            .map_or(false, |p| p.exists());
-        if exists { themes_dir } else { None }
+
+        if let Some(p) = themes_dir {
+            if p.exists() { return Some(p); }
+            if fs::DirBuilder::new().create(&p).is_ok() {
+               return Some(p);
+            }
+        }
+        None
     }
 }
 
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -491,9 +491,9 @@ impl ConfigManager {
     /// Path to themes sub directory inside config directory.
     pub(crate) fn get_themes_dir(&self) -> Option<PathBuf> {
         let themes_dir = self.config_dir.as_ref()
-                                .map(|p| p.join("themes"));
+            .map(|p| p.join("themes"));
         let exists = themes_dir.as_ref()
-                                .map_or(false, |p| p.exists());
+            .map_or(false, |p| p.exists());
         if exists { themes_dir } else { None }
     }
 }
--- a/rust/core-lib/src/styles.rs
+++ b/rust/core-lib/src/styles.rs
@@ -14,15 +14,16 @@
 
 //! Management of styles.
 
-use std::collections::HashMap;
-use std::path::{Path, PathBuf};
+use std::collections::{HashMap, HashSet};
 use std::ffi::OsStr;
 use std::fs;
+use std::iter::FromIterator;
+use std::path::{Path, PathBuf};
 
 use serde_json::{self, Value};
+use syntect::dumps::{dump_to_file, from_dump_file};
 use syntect::highlighting::StyleModifier as SynStyleModifier;
-use syntect::highlighting::{Color, Theme, ThemeSet, Highlighter};
-use syntect::dumps::{from_dump_file, dump_to_file};
+use syntect::highlighting::{Color, Highlighter, Theme, ThemeSet};
 use syntect::LoadingError;
 
 pub use syntect::highlighting::ThemeSettings;
@@ -157,8 +158,12 @@ pub struct ThemeStyleMap {
     // It's not obvious we actually have to store the style, we seem to only need it
     // as the key in the map.
     styles: Vec<Style>,
+    themes_dir: Option<PathBuf>,
     cache_dir: Option<PathBuf>,
     caching_enabled: bool,
+
+    // Maintaining all theme paths for comparison on an fs event.
+    state: HashSet<PathBuf>,
 }
 
 impl ThemeStyleMap {
@@ -167,7 +172,8 @@ impl ThemeStyleMap {
         let theme_name = DEFAULT_THEME.to_owned();
         let theme = themes.themes.get(&theme_name).expect("missing theme").to_owned();
         let default_style = Style::default_for_theme(&theme);
-        let (caching_enabled, cache_dir) = init_cache_dir(themes_dir);
+        let cache_dir = None;
+        let caching_enabled = true;
 
         ThemeStyleMap {
             themes,
@@ -176,8 +182,10 @@ impl ThemeStyleMap {
             default_style,
             map: HashMap::new(),
             styles: Vec::new(),
+            themes_dir,
             cache_dir,
             caching_enabled,
+            state: HashSet::new(),
         }
     }
 
@@ -233,12 +241,14 @@ impl ThemeStyleMap {
         result
     }
 
-    /// Delete key from the themes map.
+    /// Delete key and the corresponding dump file from the themes map.
     pub(crate) fn remove_theme(&mut self, path: &Path) -> Option<String> {
+        validate_theme_file(path).ok()?;
+
         let theme_name = path.file_stem().and_then(OsStr::to_str)?;
         self.themes.themes.remove(theme_name);
+        self.state.remove(path);
 
-        //Delete dump file
         let dump_p = self.get_dump_path(theme_name)?;
         if dump_p.exists() {
             let _ = fs::remove_file(dump_p);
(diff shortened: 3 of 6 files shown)


rust/core-lib/src/config.rs
```diff
@@ -328,6 +328,12 @@ impl ConfigManager {
     pub fn default_buffer_config(&self) -> BufferConfig {
         self.get_buffer_config(None, None)
     }
+
+    /// Path to themes sub directory inside config directory
+    pub fn get_themes_dir(&self) -> PathBuf {
+        let themes_dir = self.config_dir.as_ref().unwrap().join("themes");
```
cmyr: there is no guarantee that `self.config_dir` exists, so we can't safely unwrap here.

rust/core-lib/src/styles.rs
```diff
@@ -15,6 +15,7 @@
 //! Management of styles.
 
 use std::collections::HashMap;
+use std::path::{Path};
```
cmyr: Nit: no need for `{}` for a single item import

rust/core-lib/src/styles.rs
```diff
@@ -220,4 +221,11 @@ impl ThemeStyleMap {
         self.styles.push(style.clone());
         result
     }
+
+    /// Load all themes inside themes folder and append it to
+    /// the default themes set
+    pub fn load_theme_set(&mut self, themes_dir: &Path) {
```
cmyr: Nit: for new API I'm trying to use `pub(crate)` for fns that are only going to be used internally. (One advantage of this is that `rustc` can emit a warning if a fn marked `pub(crate)` is unused).

rust/core-lib/src/config.rs
virattara: Ok, for this I can check for `config_dir` before calling this method or return a `Result` from here and then handle the error in the calling fn. I think the last method is better.

rust/core-lib/src/config.rs
```diff
@@ -328,6 +328,12 @@ impl ConfigManager {
     pub fn default_buffer_config(&self) -> BufferConfig {
         self.get_buffer_config(None, None)
     }
+
+    /// Path to themes sub directory inside config directory
+    pub(crate) fn get_themes_dir(&self) -> PathBuf {
+        let themes_dir = self.config_dir.as_ref().unwrap().join("themes");
```
cmyr: this will crash clients that don't use a config dir, of which several exist.

virattara: Yeah, I'm working on it. Haven't pushed its fix yet.

virattara: Fixed.

virattara:
@cmyr 
Just wanted to know if I am headed in the right direction. 
Particularly the last commit.
I did try to handle all the fs events but I don't think `Write` makes much sense in usual cases.
Currently what I did in `Write` is to simply load the theme back so that it replaces its predecessor, here what I think is we need to refresh the state of the editor(Something like a reload popup? Which restarts the editor.) 

Also, I wanted to know am I delaying this feature too much? Didn't get much time before but trying to cover up now.

Thanks.

cmyr:
I think your code looks good! For the `Write` case, I think we should check to see if the modified theme is the current theme, and if it is we can follow the same codepath we follow when the theme gets changed. We shouldn't need a popup or anything, I think this can all happen automatically.

It's possible I'm missing something, but i think this should work? You should be able to test it out locally.

Don't worry at all about taking your time; this is a nice-to-have feature, it isn't interfering with any other work. :)

virattara:
@cmyr can you please review these changes. 
I have tried to cover all the points mentioned in #637 

Thanks.

rust/core-lib/src/config.rs
```diff
@@ -487,6 +487,15 @@ impl ConfigManager {
         let _: BufferItems = serde_json::from_value(defaults.into())?;
         Ok(())
     }
+
+    /// Path to themes sub directory inside config directory.
+    pub(crate) fn get_themes_dir(&self) -> Option<PathBuf> {
+        let themes_dir = self.config_dir.as_ref()
+                                .map(|p| p.join("themes"));
```
cmyr: this whitespace seems a bit wonky to me?

rust/core-lib/src/styles.rs
```diff
@@ -220,4 +222,65 @@ impl ThemeStyleMap {
...
+                    self.load_theme(theme);
+                }
+            },
+            Err(e) => eprintln!("Error while loading themes dir: {:?}",e),
+        }
+    }
+
+    fn load_theme(&mut self, path: &Path) {
+        //Check for the dump file first
+        let theme_cache = path.with_extension("tmdump");
```
cmyr: A minor possible issue with this (I believe) is that the cache could be stale if the theme file was modified while the editor was closed.

One possible solution would be to associate the cache file with the theme file's modification time?

rust/core-lib/src/styles.rs
```diff
@@ -220,4 +222,65 @@ impl ThemeStyleMap {
...
+    }
+
+    /// Insert theme into the map.
+    /// Store binary dump in a file with `tmdump` extension.
+    pub(crate) fn insert_theme(&mut self, path: &Path) 
+        -> Option<String>
+    {
+        match ThemeSet::get_theme(path) {
+            Ok(theme) => {
+                dump_to_file(&theme, path.with_extension("tmdump")).unwrap();
```
cmyr: I'd prefer to keep caches somewhere that isn't so user facing. A simple solution for now would be to just use $theme_dir/cache? At some point we might want the client to specify where caches should be written.

rust/core-lib/src/tabs.rs
```diff
@@ -601,6 +625,52 @@ impl CoreState {
...
+            }
+            _ => ()
+        }
+        let theme_names = self.style_map.borrow().get_theme_names();
+        self.peer.available_themes(theme_names);
+    }
+
+    /// Load a single theme file. Updates if already present.
+    fn load_theme_file(&mut self, path: &Path) {
+        let _t = trace_block("CoreState::load_theme_file", &["core"]);
```
cmyr: excellent, thanks for including this. 👍

cmyr:
Okay I tracked down my issue where I themes weren't being added: to reproduce, open a theme in safari (like https://raw.githubusercontent.com/Colorsublime/Colorsublime-Themes/master/themes/GitHub_Dark.tmTheme), and then save it to the themes dir directly.

Safari is downloading the file to a temporary location, and then moving it to the themes dir; because of the way FSEvents works on mac, we get `NoticeRemove` and `Chmod` events, but no `Create` event.

A reasonable fix would be to check for `Chmod`, and if it occurs in a file we don't have loaded, to load that file?

cmyr:
Also: taking a look at this PR in my editor and I noticed a bunch of trailing whitespace. Could you clean that up as well before we merge?

virattara:
@cmyr Sure will check these out.

virattara:
Replacing an existing theme file does not work properly. Looking into it.

virattara:
While trying to fix the inconsistency caused while replacing existing theme files, I noticed that the fsEvents were different at times. 
Chrome: [`noticeRemove`, `chmod`]
Safari: [`noticeRemove`, `Remove`] (Editor ends up removing the file!)
Terminal: [`noticeRemove`, `create`]
Justified by [this](https://lists.apple.com/archives/filesystem-dev/2016/Mar/msg00004.html)?

cmyr:
This will be caused by different applications having slightly different ways of implementing an atomic save, and yes it's related to some of the implementation details of the `FSEvent` api. See this issue here: https://github.com/passcod/notify/issues/147

as that issue mentions, the _best_ way to handle this would be to maintain your own record of what is in that directory, and then when an event happens to compare against your record. Alternatively if any event happens in a file that you've previously loaded, or an event occurs for a file you haven't seen before, you could try to reload.

virattara:
Fixed replacing existing theme files.

cmyr:
@virattara you think this is ready to go?

virattara:
Yes.

cmyr:
@virattara Okay amazing, I'll give it another close look later in the week, thanks again for keeping with this, very exciting! 😃

rust/core-lib/Cargo.toml
```diff
@@ -24,7 +24,7 @@ xi-rpc = { path = "../rpc", version = "0.2.0" }
 [dependencies.syntect]
 version = "2.0"
 default-features = false
-features = ["assets","dump-load-rs"]
+features = ["assets","dump-load-rs","dump-create"]
```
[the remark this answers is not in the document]
cmyr: These are syntect's names, and they're used for loading and writing caches of syntax definitions as well as themes.

rust/core-lib/src/styles.rs
```diff
@@ -220,4 +240,159 @@ impl ThemeStyleMap {
...
+        let dump_p = self.get_dump_path(theme_name)?;
+
+        if !&dump_p.exists() { return None; }
+
+        let mod_t = fs::metadata(&dump_p)
+            .and_then(|md| md.modified()).ok()?;
+        let mod_t_org = fs::metadata(theme_p)
+            .and_then(|md| md.modified()).ok()?;
+
+        match mod_t.duration_since(mod_t_org) {
```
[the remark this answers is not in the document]
virattara: Actually, the `duration_since` returns `ok(time_difference)` if the `self` time is greater than the argument. Otherwise this returns an error. As I'm not using the `time_difference` here so I must replace it with an if-else branch.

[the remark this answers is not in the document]
jansol: I think `if let ... {} else {}` or `unwrap_or_else` would be nice rusty ways to do this

rust/core-lib/Cargo.toml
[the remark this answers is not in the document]
jansol: My bad, didn't read the context well enough.

rust/core-lib/src/tabs.rs
```diff
@@ -418,10 +433,18 @@ impl CoreState {
...
 
     fn do_set_theme(&self, theme_name: &str) {
-        if self.style_map.borrow_mut().set_theme(&theme_name).is_err() {
-        //TODO: report error
-            return;
+        //Set only if requested theme is different from the 
+        //current one.
+        if theme_name != self.style_map.borrow().get_theme_name() {
+           if self.style_map.borrow_mut().set_theme(&theme_name).is_err() {
+                //TODO: report error
```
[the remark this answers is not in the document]
cmyr: I agree that it makes sense to do this now instead of later, but it might be worth waiting for #761, so we can use real logging.

rust/core-lib/src/styles.rs
```diff
@@ -220,4 +240,152 @@ impl ThemeStyleMap {
...
+        let dump_p = self.get_dump_path(theme_name)?;
+
+        if !&dump_p.exists() { return None; }
+
+        let mod_t = fs::metadata(&dump_p)
+            .and_then(|md| md.modified()).ok()?;
+        let mod_t_org = fs::metadata(theme_p)
+            .and_then(|md| md.modified()).ok()?;
+
+        if mod_t >= mod_t_org {
```
jansol: 'org' usually means 'organization' whereas 'orig' is used for 'original' but this is now *really* in nit-picking territory and this section is pretty obvious otherwise. What is an interesting thing to consider, though, is how the case should be handled where a dump exists but the original file has been deleted.

virattara: The iterator runs over the existing ".tmTheme" files in the "themes_dir" directory.
So, if a file is not present or is deleted it will never be loaded indirectly(via its dump) onto the editor's `theme` map.

jansol: Never trust files to exist until you have actually read them 😉. I probably should have said 'where the original file is deleted between the creation of the iterator and the running of this fn on said file.' Either way, this code does the (arguably) reasonable thing of returning None in case of such situation. Consider leaving a note about it though.

virattara: Something similar to hardlink.

cmyr:
@virattara this needs a rebase, and then we should be good to go. It might also be a good time to log the error we were talking about earlier, since #761 has landed.

virattara:
Rebased, added error logging.

Issue #664, opened 2018-05-16
RPC for search and replace


scholtzan:
First draft of how the RPC for search and replace should look like:

```
find {[{"case_sensitive":false,"regex":false,"chars":"m",id:0}, {"case_sensitive":false,"regex":true, "chars":".*",id:1}]}
find_next {"wrap_around":true}
find_previous {"wrap_around":true}
find_all {}
replace {"chars":"n"}
replace_all {"chars":"n"}
select_next {}               // CMD + D
```

### Open questions
* What is a good way to correlate search results to search queries? Should `id` be used or something else instead?
* RPC for search highlights: https://github.com/google/xi-editor/issues/650

### Related issues
* select_next: https://github.com/google/xi-editor/issues/325

cmyr:
So the idea is that we can have multiple active search queries, which we will highlight differently? You mentioned this in your proposal but I don't think I fully understood what you had in mind. I think that makes sense.

What's the distinction between `find` and `find_all`? How do they interact?

When is `wrap_around` set? 

Thinking a bit more about multiple active find operations, I wonder if it doesn't make sense to send them separately? I think I'd need to better understand what the UX will be for this. It really isn't a feature I've used in another editor. I also haven't read the xi-mac issue yet, will take a look at that now. :)

scholtzan:
`find` will send all the parameters (search strings, flags, ...) of the search to the backend. Then occurrences are determined and highlighted in the frontend. `find_next`/`find_previous` will select search occurrences that are closest to the current cursor position while `find_all` will select all occurrences in the file.

I forgot the `wrap_around` option in the UI. But if this flag is set, then the search will automatically continue at the beginning after the end of the file is reached.

Sending them separately might also be an option, though I am not sure if that has advantages compared to sending them together. 
I haven't seen this in any other editor but it might be useful, for example when searching for different patterns in log files.

cmyr:
Okay, I have some exploratory questions, some of them are concrete and some of them are more just abstract things I've had in mind when thinking about search long term.

[...]

PR #676, opened 2018-05-28
Keep find result selected on close

cmyr:
This fixes an issue where the find result wasn't selected after closing the find view.

Questions:
- is `cancel_operation` the correct RPC to be sending when the user dismisses the find view? Maybe we want a new RPC specific to this case?
- is it correct to call `clear()` when unsetting find? 

@scholtzan let me know if this change makes sense to you, or if you have a better idea.

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -193,16 +193,19 @@ impl View {
     }
 
     fn do_cancel(&mut self, text: &Rope) {
-        self.collapse_selections(text);
-        for mut find in self.find.iter_mut() {
-            find.unset();
+        // if we have active find highlights, we don't collapse selections
+        if self.find.is_empty() {
+            self.collapse_selections(text);
+        } else {
+            self.unset_find();
         }
     }
 
-    pub fn unset_find(&mut self) {
+    pub(crate) fn unset_find(&mut self) {
         for mut find in self.find.iter_mut() {
             find.unset();
         }
+        self.find.clear();
     }
 
     fn goto_line(&mut self, text: &Rope, line: u64) {


scholtzan:
* I think it would be better to have a new RPC for this case. Otherwise, for example, in case there is no match for the current search query but users select some text and then close the search dialog, `cancel_operation` will remove all the selections. 
Perhaps this command can also be send (with a different attribute) when find is opened, so that the core sends the currently stored query to the frontend (in the last comment of https://github.com/google/xi-editor/issues/664 I proposed `set_search_dialog {"open": false}` for this. I don't like the command name though).

* Sending `clear` would mean that the search query is removed. So if the find view is opened, a search query is entered, the find view is closed and then opened again, then the view will be empty. Maybe it would be better to keep the last search query? Then if the find view is opened again the last query could be automatically inserted into the search field.

cmyr:
I think In the first case, if there is no match there should still be a `Find` item, so `find.is_empty()` will return false, and we won't clear the selection.

I was also confused about the second case, but what you're describing doesn't happen in xi-mac. My understanding is that the frontend is storing the state of the last query it made, so when it gets opened it resends the search query. I'm not sure this is the _right_ behaviour, but it is the current behaviour; and this patch is definitely focused on the current behaviour.

We should definitely think through exactly who should be storing what state, and when those changes land we'll have to change a bunch of this.

scholtzan:
Oh right, then that would work.

I would prefer storing the query in the core but that is a discussion for somewhere else. For the current behaviour this should work.

cmyr:
@scholtzan yea I don't have a very sold grasp on the state storage stuff right now, but agree we don't need to resolve that immediately.

PR #677, opened 2018-05-28
Added status bar rpcs to send to front end

nangtrongvuon:
This PR is a companion PR to xi-mac's [#183](https://github.com/google/xi-mac/pull/183). This PR adds necessary RPC items and commands to support status bar items, along with a sample plugin that  makes use of these features.

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -380,13 +380,13 @@ Adds a status item, which will be displayed on the frontend's status bar. The al
 
 #### update_status_item
 
-`update_status_item { key: "my_key", value: "hello"}
+`update_status_item { key: "my_key", value: "hello"}`
 
 Update a status item with the specified key with the new value.
 
 #### remove_status_item
 
-`remove_status_item { key: "my_key" }
+`remove_status_item { key: "my_key" }`
 
 Removes a status item from the front end.
 
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -12,7 +12,9 @@ site_nav_category: docs
     - [Backend](#from-front-end-to-back-end)
         - [Edit Commands](#edit-namespace)
         - [Plugin Commands](#plugin-namespace)
+
     - [Frontend](#from-back-end-to-front-end)
+        - [Status Bar Commands](#status-bar-commands)
 
 ----
 
@@ -367,6 +369,27 @@ this writing, the following is valid json for a `Command` object:
         ]
     }
 ```
+
+### Status Bar Commands
+
+#### add_status_item
+
+`add_status_item { key: "my_key", value: "hello", alignment: "left" }`
+
+Adds a status item, which will be displayed on the frontend's status bar. The alignment key dictates whether this item appears on the left side or the right side of the bar. This alignment can only be set when the item is added.
+
+#### update_status_item
+
+`update_status_item { key: "my_key", value: "hello"}
+
+Update a status item with the specified key with the new value.
+
+#### remove_status_item
+
+`remove_status_item { key: "my_key" }
+
+Removes a status item from the front end.
+
 
 ## Other future extensions
 
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -162,9 +162,12 @@ impl<'a> EventContext<'a> {
             Edit { edit } => self.with_editor(
                 |ed, _, _| ed.apply_plugin_edit(edit)),
             Alert { msg } => self.client.alert(&msg),
-            AddStatusItem { key, value, alignment }  => self.client.add_status_item(viewID, &key, &value, &alignment),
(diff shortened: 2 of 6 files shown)


rust/core-lib/src/event_context.rs
```diff
@@ -160,6 +160,9 @@ impl<'a> EventContext<'a> {
             Edit { edit } => self.with_editor(
                 |ed, _, _| ed.apply_plugin_edit(edit)),
             Alert { msg } => self.client.alert(&msg),
+            AddStatusItem { key, value, alignment }  => self.client.add_status_item(viewID, &key, &value, &alignment),
```
cmyr: Nit: I'd break these lines up by inserting a newline after the `=>`.

rust/plugin-lib/src/view.rs
```diff
@@ -170,6 +170,36 @@ impl<C: Cache> View<C> {
         self.peer.request_is_pending()
     }
 
+    pub fn add_status_item(&self, key: &str, value: &str, alignment: &str) {
```
cmyr: These three methods should be added to the docs, alongside the other core->frontend methods.

rust/sample-plugin/src/main.rs
```diff
@@ -70,6 +71,16 @@ impl Plugin for SamplePlugin {
                 let _ = self.capitalize_word(view, iv.end());
             }
         }
+        self.0 += 1;
+        view.update_status_item("my_key", &format!("hello {}", self.0));
```
cmyr: after you get to 100 and remove "my_key", this stops doing anything. Overall I like the example, but maybe it could be tweaked a bit?

docs/docs/frontend-protocol.md
```diff
@@ -368,6 +370,27 @@ this writing, the following is valid json for a `Command` object:
...
+
+#### add_status_item
+
+`add_status_item { key: "my_key", value: "hello", alignment: "left" }`
+
+Adds a status item, which will be displayed on the frontend's status bar. The alignment key dictates whether this item appears on the left side or the right side of the bar. This alignment can only be set when the item is added.
+
+#### update_status_item
+
+`update_status_item { key: "my_key", value: "hello"}
```
cmyr: this ` is never closed.

rust/core-lib/src/event_context.rs
```diff
@@ -147,7 +147,7 @@ impl<'a> EventContext<'a> {
     pub(crate) fn do_plugin_cmd(&mut self, plugin: PluginId,
                                  cmd: PluginNotification) {
         use self::PluginNotification::*;
-
+        let viewID = self.view.borrow().view_id;
```
cmyr: Rust convention is to give identifiers `snake_case` names. (You should get a warning about this when you compile; in general warnings should be addressed unless there's a good reason.)

rust/core-lib/src/event_context.rs
```diff
@@ -160,6 +160,12 @@ impl<'a> EventContext<'a> {
             Edit { edit } => self.with_editor(
                 |ed, _, _| ed.apply_plugin_edit(edit)),
             Alert { msg } => self.client.alert(&msg),
+            AddStatusItem { key, value, alignment }  =>
+            self.client.add_status_item(viewID, &key, &value, &alignment),
+            UpdateStatusItem { key, value }
+            => self.client.update_status_item(viewID, &key, &value),
```
cmyr: rust style nit: the `=>` should be on the same line as the match case; and if the statement for that case is on a newline it should be indented.

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -181,6 +181,9 @@ pub enum PluginNotification {
     UpdateSpans { start: usize, len: usize, spans: Vec<ScopeSpan>, rev: u64 },
     Edit { edit: PluginEdit },
     Alert { msg: String },
+    AddStatusItem { key: String, value: String, alignment: String },
```
cmyr: There's an argument that there should be a new `Alignment` type used here, but I don't think that needs to be blocking.

rust/sample-plugin/src/main.rs
```diff
@@ -33,7 +33,7 @@ use xi_plugin_lib::{Plugin, ChunkCache, View, mainloop, Error};
 /// intended to demonstrate how to edit a document; when the plugin is active,
 /// and the user inserts an exclamation mark, the plugin will capitalize the
 /// preceding word.
-struct SamplePlugin;
+struct SamplePlugin(usize);
```
cmyr: I'm not sure I'd like to include this update to the sample plugin in this PR. I'm not exactly sure what the role of the `sample-plugin` project is, and I do imagine updating that project to show examples of various plugin features, but this particular addition doesn't feel like it is a very useful illustration.

What might make the most sense for things like this, where we want some simple plugin just to demo a new feature, would be to copy `sample-plugin` into a new folder (out side of `xi-editor`) and then make a new repo for your new example, put that on your github, or something like this?

raphlinus:
Thanks!

Issue #650, opened 2018-05-09
Missing Search Highlights


scholtzan:
I wrote a simplified WIP version (https://github.com/google/xi-editor/pull/649) of find that, for now, searches the entire text and all search occurrences (it is not much different from the old find functionality). However, only the first one is highlighted (tested in xi-mac). I tried to find out the reason for this and narrowed it down to `send_update_for_plan`. It seems that only the line with the current cursor is rendered/updated, after xi receives a `Find` event and performs the search. I suspect it has something to do with the line cache shadow of the view. 

I am currently trying to find out how rendering in `View` and the line cache shadow should behave during find. Or if this should all be done in a completely different way. Any thoughts on this are appreciated!

https://github.com/google/xi-editor/issues/634

scholtzan:
This might be a problem on the xi mac side. I forced a full re-rendering after `find` using `set_dirty` in `View`. Only the first search occurrence is highlighted. However, if I change the style for search highlights to 0 (= style for selections) in `render_styles` then _all_ search occurrences are correctly selected in xi mac. So my guess is that it is a xi mac bug.

raphlinus:
Right, xi-mac is currently not rendering style id=1 (reserved for search highlights) at all. This would be fairly straightforward to fix, but the main reason it hasn't been is that there is a design decision to be made about where in the pipeline "special" styles get resolved into their actual appearance. For styles from plugins (syntax highlighting) this is done in the core, but not so for the two reserved styles.

The main reason style=0 (selection) is handled specially is that the appearance is different depending on whether the window is focused. I'll outline 3 possible approaches, which basically represent different places in the pipeline where resolving occurs.

1. (basically status quo). We continue to have reserved styles, and logic in the front-end to render them. Re-implement logic in xi-mac to render style 1 correctly.

2. We continue to have style 0 be reserved because of its special focus behavior, but selection highlights (and any other future spans that represent core behavior) are resolved in the core to their actual appearance, and set as a non-reserved style. There's still xi-mac work, I think the `bg` attribute is not respected.

3. We resolve _all_ styles in the core. The special focus behavior of style 0 is handled by extending def_style so that can specify two different background colors, for focused and non-focused state.

Each of these has advantages and disadvantages. I think the main advantage of (3) is that it's less special-case work by the front-ends. On the other hand, it's less flexibility in case we want the front-end to use the semantic knowledge about the meaning of the span to produce special appearance or behavior.

[...]

PR #678, opened 2018-05-29
Fix find highlight drawing

cmyr:
The active selection is included in both 'selections' and 'highlights', whichever of these occurs last in the list of styles will be rendered. By making sure this is the 'selection', we can make the active highlight render in the selection color.

This isn't ideal, but it's an improvement on the current behaviour.


cmyr:
I agree, this touches on #650. I'll add a comment for now.

PR #682, opened 2018-05-31
Refactor/configs in config manager

cmyr:
(includes #680, which should go in first.)

This is another batch of cleanup & refactoring. In pointform:

- the language assigned to a buffer is tracked in the ConfigManager
- the buffer's config is tracked in the ConfigManager. This means that
we don't need to have a config in order to init an `Editor`, which means
we now only have one code path for creating a new view.
- configs are passed into `Editor` and `View` as needed; this means the
`View` doesn't need to reach in to the `Editor` to access the config.

This will enable and/or simplify some additional feature work:
- allowing the client to set a view's language without saving
- implementing width wrapping via the config system
- auto launching plugins based on language (and other) tags

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -110,12 +110,24 @@ pub struct ConfigPair {
     cache: Arc<Table>,
 }
 
+/// The language associated with a given buffer; this is always detected
+/// but can also be manually set by the user.
+#[derive(Debug, Clone)]
+struct LanguageTag {
+    detected: LanguageId,
+    user: Option<LanguageId>,
+}
+
 #[derive(Debug)]
 pub struct ConfigManager {
     /// A map of `ConfigPairs` (defaults + overrides) for all in-use domains.
     configs: HashMap<ConfigDomain, ConfigPair>,
     /// The currently loaded `Languages`.
     languages: Languages,
+    /// The language assigned to each buffer.
+    buffer_tags: HashMap<BufferId, LanguageTag>,
+    /// The configs for any open buffers
+    buffer_configs: HashMap<BufferId, BufferConfig>,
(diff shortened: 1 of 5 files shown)


rust/core-lib/src/config.rs
```diff
@@ -465,6 +567,31 @@ impl ConfigDomain {
     }
 }
 
+impl LanguageTag {
+    fn new(detected: LanguageId) -> Self {
+        LanguageTag { detected, user: None }
+    }
+
+    fn resolve(&self) -> LanguageId {
```
raphlinus: I believe this could be written slightly more concisely as

```
self.user.as_ref().unwrap_or(&self.detected).clone()
```

rust/core-lib/src/config.rs
```diff
@@ -465,6 +567,31 @@ impl ConfigDomain {
...
+        self.user.as_ref().map(LanguageId::clone)
+            .unwrap_or_else(|| self.detected.clone())
+    }
+
+    fn set_detected(&mut self, detected: LanguageId) -> bool {
```
raphlinus: Would be helpful to explain return value.

rust/core-lib/src/editor.rs
```diff
@@ -467,15 +437,15 @@ impl Editor {
         saved
     }
 
-    fn insert_newline(&mut self, view: &View) {
+    fn insert_newline(&mut self, view: &View, config: &BufferItems) {
         self.this_edit_type = EditType::InsertChars;
-        let text = self.config.items.line_ending.clone();
-        self.insert(view, &text);
+        //let text = self.config.items.line_ending.clone();
```
raphlinus: Probably best to just remove this now.

PR #684, opened 2018-05-31
Track view size and use for word wrapping

cmyr:
(based off of #682; this PR is really just 8abe978)
 
This adds an rpc, `resize`, which sets the size of the view (I don't like this name, and I'm not sure if this RPC shouldn't be combined with 'scroll' or something; for discussion).

It also adds a 'word_wrap' config setting; when set to `true`, the width of the view is used for determining line breaks.

It also deprecates the `debug_wrap_width` and `debug_rewrap` methods.

todos:
- [x] docs

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -53,6 +53,7 @@ pub struct View {
     /// We attempt to reduce duplicate renders by setting a small timeout
     /// after an edit is applied, to allow batching with any plugin updates.
     pending_render: bool,
+    size: Size,
     /// The selection state for this view. Invariant: non-empty.
     selection: Selection,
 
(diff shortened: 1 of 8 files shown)


rust/core-lib/src/view.rs
```diff
@@ -98,6 +99,13 @@ enum FindStatusChange {
     Matches
 }
 
+/// A size, in pixels.
```
raphlinus: I'd change this to "px units".

rust/core-lib/src/view.rs
```diff
@@ -252,6 +261,11 @@ impl View {
         self.set_selection(text, SelRegion::caret(offset));
     }
 
+    pub fn set_size(&mut self, size: Size) {
+        eprintln!("set size {:?}", size);
```
raphlinus: Do we want to leave this debugging in?

PR #685, opened 2018-06-01
Rustfmt

Lapz:
* Added rustfmt to the CI.
* Updated the contributing guide
* Ran rustfmt on the whole codebase


cmyr:
@Lapz cool, thanks for this! I'm going to take some time to look through this and figure out if there's any other options we might want to set, etc, and I'll give it a proper review then. 👍

cmyr:
Oh I think this is related to our weird module structure in core, with the `internal` module controlling visibility.

Lapz:
Yeah the rustfmt problem meant that I couldn't format the core-lib

cmyr:
Okay, so I'm not sure if we want to get this in before we can use it on core. One option would be to have our own `rustfmt.sh` script that just looks for *.rs files and passes them in to rustfmt individually, or we could try and wait on an upstream fix? My impression is that rustfmt isn't a super high priority at the moment, though, so we might be waiting a while.

cmyr:
@Lapz Any thoughts on this? A very reasonable option, to me, is that we just add a script of some kind that manually runs rustfmt on any `.rs` files contained in `core-lib/src/*/*`. That should be relatively simple, but would give us full coverage, and would be a solution until this issue is addressed upstream.


Also just glancing at the current travis config, will this actually produce an error of the formatting is wrong? It looks like it will just format the code, but I'm not sure that this exits with an error if formatting is necessary.

Lapz:
I got exams right now so I wont be able to do anything for at least one week. I could implement a script that manualy formats all the files in core. Currently on nightly rustfmt has a flag called `check` that can be used to check if fmt is sucessful. So I'm leaning towards checking the exitcode on stable and using the check flag on nightly

cmyr:
@Lapz okay cool no rush, just wanted to make sure we weren't stuck in PR purgatory. Good luck with your exams!

cmyr:
I think we're most concerned about stable. I'm surprised nightly is different, are we using nightly specific flags or something?

Lapz:
@cmyr Most config flags for rustfmt don’t work on stable so its not due to something that we are doing. 
It fails on nightly because methods chains are formatted differently to when on stable
i.e
```diff
-        let plugin_desc = self.catalog
+        let plugin_desc = self
+            .catalog
             .get_named(plugin_name)
             .ok_or(Error::Other(format!("no plugin found with name {}", plugin_name)))?;
```

cmyr:
Is this just a flag that we can turn off in our config file?

jansol:
Doesn't seem to be a flag. Either way I'd say it's reasonable to only run rustfmt on stable (if possible), since that's what xi officially targets.

cmyr:
And it looks like rustfmt 1.0 will be on [stable september 13th[(https://www.ncameron.org/blog/rustfmt-1-rc/), so we're getting closer and closer to having no excuses to merge this. @Lapz, how's school going? 😄 😄

cmyr:
@Lapz another bump, would be happy to help you get this finished...

Lapz:
I'll try and finish this later today

Lapz:
Closing in favour of #846

Issue #670, opened 2018-05-23
Improve plugin-lib edit API


cmyr:
as illustrated by #661, the process of creating an edit from a plugin using the rust `plugin-lib` is... elaborate.

There's definitely a better API possible. The main headaches are around calculating and keeping track of indices, and having to manually scan through document's buffer to do any sort of analysis.

There are a bunch of ways we could make this cleaner, but one thing currently sticks out to me, which is introducing some `BufferRegion` struct, which wraps a string slice while also keeping track of indices, and which provides methods for finding additional `BufferRegion`s representing substrings.

Something like:

```rust
struct BufferRegion<'a> {
    contents: &'a str,
    rev: usize, // maybe not necessary
    start: usize, // utf-8
    end: usize, // utf-8
}

impl<'a> AsRef<str> for BufferRegion<'a> { .. }

// Some methods on `BufferRegion` that return other `BufferRegion`s;
impl<'a> BufferRegion<'a> {
    fn word_at(&'a self, idx: usize) -> BufferRegion<'a> { .. }
    fn words(&'a self) -> impl Iterator<Item=BufferRegion<'a>> { .. }
    fn find<P: Pattern>(&'a self, pattern: P) -> Option<BufferRegion<'a>> { .. }
    // and maybe things like this? would require `BufferRegion` to have a ref to the view, probably??
    fn preceding_word(_) -> Option<BufferRegion<'a>> { .. }
    fn next_word(_) -> Option<BufferRegion<'a>> { .. }
    fn preceding_line(_) -> Option<BufferRegion<'a>> { .. }
    fn next_line(_) -> Option<BufferRegion<'a>> { .. }
}

// Some methods on `plugin-lib::view::View` that return `BufferRegion`s

impl View {
    fn get_line<'a>(&'a self, line_num: usize) -> Result<BufferRegion<'a>, Error> { .. }
    fn get_word<'a>(&'a self, offset: usize) -> Result<BufferRegion<'a>, Error> { .. }
}
```

In addition, `BufferRegion` would be the basis of the edit API:

```rust 
// in your plugin:
impl YourPlugin {
// ...
    fn update(&mut self, view: &mut View<Self::Cache>, delta: Option<&RopeDelta>,
                    _edit_type: String, _author: String) {
// maybe the plugin should have another method we call when the delta is missing?
// it feels bad that every plugin would have to do this check on every update.
        if let Some(delta) = delta {
            if let Some(text) = delta.map(Delta::as_simple_insert) { 
[...]

Issue #659, opened 2018-05-16
[GSoC] Language Server Protocol Plugin Support


betterclever:
This is a roadmap of the the steps planned by me to add support for the Language Server Protocol (LSP) Plugin Support in the Xi Editor.

### Roadmap
_Will be refined over time_

- [ ] Get a basic plugin working which can access the document state.
- [ ] Discover improvements needed in current Plugin API to enable a very simple LSP usecase i.e. JSON.
- [ ] Get the JSON Plugin Working. The basic idea is to develop a crate which is abstraction over the plugin_lib, which JSON Plugin will use to communicate with the Editor. The validation error messages can be dumped to stderr to begin with.
- [ ] Figure out a plan for the features in a specific order in the LSP Plugin including:
    - [ ] Status Information
    - [ ] Inline Error and Warning Display 
    - [ ] Hover Definitions
    - [ ] Go to Definition
    - [ ] Intelligent Autocompletion
    - [ ] Highlight References
    - [ ] Refactoring Support
- [ ] Figure out a test language (probably Rust?) to begin with to implement and test all
features.

**TODO**: More refined steps after the 4th step.

### Planning

The basic architecture currently planned (needs review) can be described by the following picture.

![lsp-archtecture](https://user-images.githubusercontent.com/16309034/40114310-f57da422-58fb-11e8-9325-74c34ff2f5b5.png)

### Roadblocks / Open Questions.

- **Managing Workspaces**. Who should manage the workspaces: Xi-Core or the Plugin itself?

skyne98:
@cmyr, hey, any progress? What is the state of LSP support in Xi? Thanks!

jansol:
Many (most?) bits have PRs that at least were in a relatively working state. However they got kind of stuck in a discussion of how these things should be communicated to the client... The annotations PR (#1010) is one step towards specifying that.

cmyr:
@skyne98 there was a bunch of exploratory work on this this summer, but the conclusion of that was that we should step back and work on incrementally adding support for the features that will _enable_ LSP support in the future. The issue linked above is a first step there.

ShalokShalom:
@cmyr  And is there a list which we can follow about those spoken features?

PR #686, opened 2018-06-01
Added method to get whole document from Cache

betterclever:
Needed for Language Server Plugin #659  and a step towards #670

### DIFF
--- a/rust/plugin-lib/src/base_cache.rs
+++ b/rust/plugin-lib/src/base_cache.rs
@@ -112,6 +112,21 @@ impl Cache for ChunkCache {
 
     }
 
+    fn get_document<DS: DataSource>(&mut self, source: &DS) -> Result<String, Error> {
+        let mut result = String::new();
+        let mut cur_idx = 0;
+        while cur_idx < self.buf_size {
+            if self.contents.len() == 0 || cur_idx != self.offset {
+                let resp = source.get_data(cur_idx, TextUnit::Utf8,
+                                           CHUNK_SIZE, self.rev)?;
+                self.reset_chunk(resp);
+            }
+            result.push_str(&self.contents);
+            cur_idx = self.offset + self.contents.len();
+        }
+        Ok(result)
+    }
+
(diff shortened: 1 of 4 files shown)


rust/plugin-lib/src/lib.rs
```diff
@@ -74,6 +74,10 @@ pub trait Cache {
     /// [`DataSource`]: trait.DataSource.html
     fn get_line<DS: DataSource>(&mut self, source: &DS, line_num: usize)
         -> Result<&str, Error>;
+    
+    /// Returns the whole document in the current View
```
cmyr: This documentation is for the `Cache` trait, which doesn't have any concept of a view. Instead here I would say something like,

"Returns the entire contents of the remote document, fetching as needed."

rust/plugin-lib/src/base_cache.rs
```diff
@@ -112,6 +112,14 @@ impl Cache for ChunkCache {
 
     }
 
+    fn get_document<DS: DataSource>(&mut self, source: &DS) -> Result<String, Error> {
```
cmyr: If we're actually going to implement this in the cache itself, it is probably worth doing it a bit more efficiently, e.g.:

```rust
    fn get_document<DS: DataSource>(&mut self, source: &DS) -> Result<String, Error> {
        let mut result = String::new();
        let mut cur_idx = 0;
        while cur_idx < self.buf_size {
            if self.contents.len() == 0 || cur_idx != self.offset {
                let resp = source.get_data(cur_idx, TextUnit::Utf8,
                                           CHUNK_SIZE, self.rev)?;
                self.reset_chunk(resp);
            }
            result.push_str(&self.contents);
            cur_idx = self.offset + self.contents.len();
        }
        Ok(result)
    }
```

The solution here, of doing it one line at a time, would make more sense if you were just wanted to fetch the whole document from within your plugin.

Also ideally this should be returning a `Cow`, since we frequently already have the entire document in the cache and don't need to allocate at all. So maybe you could also add a `//TODO:` mentioning that?

PR #687, opened 2018-06-01
find_status

scholtzan:
After the find status (search query, search options, ...) change the core will send `find_status` to the frontend. `find_status` provides information about the number of matches but also about the query stored in the core (this will allow to only store the query in the core and not in the frontend = future work).

I will also update the documentation to add `find_status` to the RPC commands after getting some feedback on this.

### DIFF
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -137,8 +137,11 @@ impl<'a> EventContext<'a> {
         let result = match cmd {
             Cut => Ok(self.with_editor(|ed, view, _| ed.do_cut(view))),
             Copy => Ok(self.with_editor(|ed, view, _| ed.do_copy(view))),
-            Find { chars, case_sensitive } => Ok(self.with_view(
-                |view, text| view.do_find(text, chars, case_sensitive))),
+            Find { chars, case_sensitive } => {
+                let result = Ok(self.with_view(|view, text| view.do_find(text, chars, case_sensitive)));
+                self.send_find_status();
+                result
+            },
             // Replace
         };
         self.after_edit("core");
@@ -410,6 +413,10 @@ impl<'a> EventContext<'a> {
         view.request_lines(ed.get_buffer(), self.client, self.style_map,
                            ed.get_layers().get_merged(), first, last,
                            ed.is_pristine())
+    }
+
+    fn send_find_status(&mut self) {
+        self.view.borrow_mut().send_find_status(&self.client)
     }
 }
 
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -89,11 +89,11 @@ impl<'a> EventContext<'a> {
     /// to the current text. This is common to most edits that just modify
     /// selection or viewport state.
     fn with_view<R, F>(&mut self, f: F) -> R
-        where F: FnOnce(&mut View, &Rope, &Client) -> R
+        where F: FnOnce(&mut View, &Rope) -> R
     {
         let editor = self.editor.borrow();
         let mut view = self.view.borrow_mut();
-        f(&mut view, editor.get_buffer(), &self.client)
+        f(&mut view, editor.get_buffer())
     }
 
     pub(crate) fn do_edit(&mut self, cmd: EditNotification) {
@@ -101,7 +101,7 @@ impl<'a> EventContext<'a> {
         let event: EventDomain = cmd.into();
         match event {
             E::View(cmd) => {
-                    self.with_view(|view, text, client| view.do_edit(text, cmd, client));
+                    self.with_view(|view, text| view.do_edit(text, cmd));
                     self.editor.borrow_mut().update_edit_type();
                 },
             E::Buffer(cmd) => self.with_editor(
@@ -115,7 +115,7 @@ impl<'a> EventContext<'a> {
     fn do_special(&mut self, cmd: SpecialEvent) {
         match cmd {
             SpecialEvent::DebugRewrap => self.with_view(
-                |view, text, _| {
+                |view, text| {
                     view.rewrap(text, 72);
                     view.set_dirty(text);
                 }),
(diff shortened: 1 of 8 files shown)


rust/core-lib/src/event_context.rs
```diff
@@ -89,19 +89,19 @@ impl<'a> EventContext<'a> {
...
+        f(&mut view, editor.get_buffer(), &self.client)
     }
 
     pub(crate) fn do_edit(&mut self, cmd: EditNotification) {
         use self::EventDomain as E;
         let event: EventDomain = cmd.into();
         match event {
             E::View(cmd) => {
-                    self.with_view(|view, text| view.do_edit(text, cmd));
+                    self.with_view(|view, text, client| view.do_edit(text, cmd, client));
```
cmyr: maybe instead of passing in the client here, we could track if the find status is 'dirty', and then update the client during the normal render cycle? In general we've managed to avoid having either the `Editor` or the `View` having access to the client directly, which I think makes the code easier to reason about; you know that the only messages being sent to core are those that are either explicitly in some `EventContext` fn, or else are part of rendering the view.

An alternative to this would be to move the find commands into `EventDomain::Special`, which are handled individually in the `EventContext`, instead of just being forwarded to the `Editor` or `View`. I think I prefer the former, though? Updating the find state feels like it fits well with other types of client updates.

rust/core-lib/src/event_context.rs
```diff
@@ -137,8 +137,12 @@ impl<'a> EventContext<'a> {
         let result = match cmd {
             Cut => Ok(self.with_editor(|ed, view, _| ed.do_cut(view))),
             Copy => Ok(self.with_editor(|ed, view, _| ed.do_copy(view))),
-            Find { chars, case_sensitive } => Ok(self.with_view(
-                |view, text| view.do_find(text, chars, case_sensitive))),
+            Find { chars, case_sensitive } =>
```
cmyr: Is the thinking here that existing clients will continue to work, but new clients will get the new behaviour?

One change with going async is that this should become a JSON-RPC 'notification', which means that the client should no longer be sending an `id` field and no longer expecting a response.

This comes down to how worried we are about breaking existing clients. I don't think we should be _too_ worried, but it's nice to be a _little_ worried.

Can we introduce the new behaviour with a new name? If we want we can even call it `find2` for the time being, and then we can return an error from the current method. Later on we could change `find` to point to the new method, keeping `find2` as an alias, and then get rid of that.

Or we can just put up an issue warning clients that find is about to break, which might be the better option.

rust/core-lib/src/find.rs
```diff
@@ -37,6 +37,8 @@ pub struct Find {
     search_string: Option<String>,
     /// The case matching setting for the currently active search
     case_matching: CaseMatching,
+    /// /// The search query should be considered as regular expression
```
cmyr: extra ///

rust/core-lib/src/view.rs
```diff
@@ -644,6 +644,23 @@ impl View {
         }
     }
 
+    /// Determines the current number of find results and search parameters to send them to
+    /// the frontend.
+    pub fn send_find_status(&self, client: &Client) {
```
cmyr: as above I think I'd prefer if this just returned some `FindStatus` type, which could be sent from the parent context.

rust/core-lib/src/event_context.rs
scholtzan: I would also prefer the last option

rust/core-lib/src/view.rs
```diff
@@ -106,6 +110,22 @@ struct DragState {
...
+    chars: Option<String>,
+
+    /// Whether the active search is case matching.
+    case_sensitive: Option<bool>,
+
+    /// Whether the search query is considered as regular expression.
+    is_regex: Option<bool>,
+
+    /// Total number of matches.
+    matches: usize
```
cmyr: do we also want this to include the index of the currently active highlight?

rust/core-lib/src/view.rs
```diff
@@ -106,6 +110,22 @@ struct DragState {
     max: usize,
 }
 
+/// Information about search queries and number of matches for find
+#[derive(Serialize, Deserialize, Debug)]
```
cmyr: maybe this should be in `find.rs`?

rust/core-lib/src/view.rs
scholtzan: If there are multiple queries then only one active highlight would exist for one of the `Find`. It might make sense to get the index if all occurrences of all queries are considered, however, that will be a bit tricky. It also seems that most editors only show the total number of matches.

rust/core-lib/src/view.rs
```diff
@@ -673,12 +658,7 @@ impl View {
         self.find_changed = false;
 
         self.find.iter().map(|find| {
-            FindStatus {
-                chars: find.search_string().clone(),
-                case_sensitive: Some(find.is_case_sensitive()),
-                is_regex: Some(find.is_regex()),
-                matches: find.occurrences().len(),
-            }
+            find.find_status()
```
cmyr: Not a blocker, but I suspect clippy would nag to write this as `.map(Find::find_status)`.

rust/core-lib/src/find.rs
```diff
@@ -62,6 +79,15 @@ impl Find {
         self.hls_dirty
     }
 
+    pub fn find_status(&self) -> FindStatus {
```
raphlinus: Short docstring?

scholtzan:
Right. That would require to store the old field values and then compare them with the new ones. Currently, for indicating that the status should be sent a flag is set (`find_changed`). I guess it would be possible to store the new field values instead of the boolean value and then use those to determine which ones should be sent. Though, I am not sure if that makes things too complicated?

cmyr:
A possible solution: afaict there are two scenarios under which this message is sent. The first is when the client is opening a find dialog or equivalent; in this case if we have existing find state we should send it all (including regex etc). The second case is when the _user_ is doing some find action; in this case we should only be sending the result count etc.

Does this cover everything, or am I confused? (very possible I'm confused)

scholtzan:
That could work. Will try it out.

scholtzan:
Instead of having just a simple flag to track the find changes, I added `enum FindStatusChange` to determine whether only the number of matches or all find parameters should be sent.

cmyr:
Okay, I like that change. I think this is ready to go; my only thought right now is whether we have frontend support. Your xi-mac branch currently crashes core, because it's still sending 'search_dialog'. Would you like to put up a small PR in xi-mac that supports the new behaviour, or should I? We should also let other frontend maintainers know that the current `find` RPC is going to break.

scholtzan:
I pushed my xi-mac changes to my branch and uncommented the `highlight_find` RPCs.

scholtzan:
The last update fixes the stack overflow when text is deleted and find occurrences need to be updated. While fixing this I also noticed that find occurrences are not correctly updated when text is inserted. Also, the find update was only applied to the visible area and not the entire document (resulted in wrong number of matches).

PR #688, opened 2018-06-02
Find in core and `search_dialog` RPC command

scholtzan:
This depends on https://github.com/google/xi-editor/pull/687

This pull request implements the `search_dialog` RPC command to signal to the core whether the search dialog is open or closed. This enables to store the search query only in the core.

I also added this to xi-mac for testing: https://github.com/scholtzan/xi-mac/tree/async-find

### DIFF
--- a/rust/core-lib/src/client.rs
+++ b/rust/core-lib/src/client.rs
@@ -124,6 +124,14 @@ impl Client {
         self.0.send_rpc_notification("def_style", &style)
     }
 
+    pub fn find_status(&self, view_id: ViewId, queries: &Value) {
+        self.0.send_rpc_notification("find_status",
+                                     &json!({
+                                         "view_id": view_id,
+                                         "queries": queries,
+                                     }));
+    }
+
     /// Ask front-end to measure widths of strings.
     pub fn measure_width(&self, reqs: &[WidthReq])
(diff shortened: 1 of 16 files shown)


cmyr:
Bikeshed:

What exactly is the behaviour that changes based when this is sent? Is it entirely related to the rendering of inactive highlights, or are there other changes?

In general I have a preference for RPC names that communicate behaviour, rather than UI state. In other words, if this flag indicates whether a partial find match should be shown, we might want to call it `highlight_partial_find`.

scholtzan:
In addition to that, the core responds with the `find_status` when this is sent.

scholtzan:
But in general, I agree, showing/hiding the highlights for the find matches is the main thing that happens. Though, why not call it `highlight_find`?

scholtzan:
Okay, I renamed it to `highlight_find`

cmyr:
@scholtzan okay this diff got a bit messy: it's really just adding the `highlight_find` RPC, correct? What's a good strategy for merging this? Do we want to do it all at once, along with the xi-mac PR, with the comments removed?

scholtzan:
This should be merged after https://github.com/google/xi-editor/pull/687. I am currently trying to fix the stack overflow problem but after that `find_status` should be ready to go.

Issue #695, opened 2018-06-07
crash when using selection for find


cmyr:
Using <kbd>cmd</kbd>+<kbd>E</kbd> on xi-mac I get a crash when the selection is a caret.

There's a related issue where the search bar isn't correctly populated when the selection _isn't_ a caret.

@scholtzan maybe this is one for you?

scholtzan:
I ignored <kbd>CMD</kbd> + <kbd>E</kbd> for now, but it is the next thing on my list.

scholtzan:
This is fixed in https://github.com/google/xi-editor/pull/701 and can be closed.

PR #701, opened 2018-06-11
Implement selection for find

scholtzan:
This addresses https://github.com/google/xi-editor/issues/695

### DIFF
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -78,6 +78,10 @@ impl Find {
         self.hls_dirty
     }
 
+    pub fn case_matching(&self) -> bool {
+        self.case_matching == CaseMatching::Exact
+    }
+
     pub fn find_status(&self, matches_only: bool) -> FindStatus {
         if matches_only {
             FindStatus {
@@ -173,7 +177,7 @@ impl Find {
     }
 
     /// Sets find parameters and search query.
-    fn set_find(&mut self, search_string: &str, case_sensitive: bool, is_regex: bool) {
+    pub fn set_find(&mut self, search_string: &str, case_sensitive: bool, is_regex: bool) {
         let case_matching = if case_sensitive {
             CaseMatching::Exact
         } else {
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -150,13 +150,8 @@ impl Find {
     }
 
     /// Set search parameters and executes the search.
-    pub fn do_find(&mut self, text: &Rope, search_string: Option<String>, case_sensitive: bool,
+    pub fn do_find(&mut self, text: &Rope, search_string: String, case_sensitive: bool,
                    is_regex: bool) {
-        if search_string.is_none() {
-            self.unset();
-        }
-
-        let search_string = search_string.unwrap();
         if search_string.len() == 0 {
             self.unset();
         }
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/find.rs
```diff
@@ -78,6 +78,10 @@ impl Find {
         self.hls_dirty
     }
 
+    pub fn case_matching(&self) -> bool {
+        self.case_matching == CaseMatching::Exact
```
jansol: Should this support a "smart" mode like for example ripgrep does?

rust/core-lib/src/view.rs
```diff
@@ -874,8 +875,44 @@ impl View {
...
+        // todo: for now only a single search query is supported however in the future
+        // todo: the correct Find instance needs to be updated with the new parameters
+        if self.find.is_empty() {
+            self.find.push(Find::new());
+        }
+
+        let find = self.find.first_mut().unwrap();
+        let cm = find.case_matching();
+
+        if self.highlight_find {
```
cmyr: What if we do <kbd>cmd</kbd>+<kbd>E</kbd> without the find window open, and then open it? Will we correctly update the highlights?

rust/core-lib/src/view.rs
```diff
@@ -874,8 +875,44 @@ impl View {
...
+            find.do_find(text, Some(search_query), cm, false);
+        } else {
+            find.set_find(&search_query, cm, false);
+        }
+    }
+
+    pub fn do_find(&mut self, text: &Rope, chars: Option<String>, case_sensitive: bool, is_regex: bool) {
```
cmyr: does `chars` still need to be an `Option<_>`? I thought this was the case because `do_find` was doing double-duty, and deriving the find string from the selection in the `None` case.

I think we can probably clean up `do_find` to remove this logic, now?

rust/core-lib/src/find.rs
cmyr: There's a reasonable case that `SelectionForFind` should always be case sensitive, although this would be different from the system behaviour on macOS at least, and different from the behaviour of ST3.

(which is to say that I'm not sure?)

scholtzan: I took Sublime as reference for this but I think it might be best to add an option `case_sensitive` to the RPC command.

rust/core-lib/src/view.rs
scholtzan: Right, it doesn't need to be an `Option` anymore.

rust/core-lib/src/view.rs
```diff
@@ -874,22 +876,37 @@ impl View {
...
+        self.set_dirty(text);
+
+        // todo: this will be changed once multiple queries are supported
+        // todo: for now only a single search query is supported however in the future
+        // todo: the correct Find instance needs to be updated with the new parameters
+        if self.find.is_empty() {
+            self.find.push(Find::new());
+        }
+
+        self.find.first_mut().unwrap().do_find(text, search_query, case_sensitive, false);
```
raphlinus: I'm curious about whether the `selection_for_find` action should always initiate a find. It seems to me that in our investigations a few weeks ago, the editor we were looking at as models didn't always. But I don't think this is a critical question, it seems we can come back to this after the functionality is in place.

scholtzan: I added it because, `selection_for_find` could be followed by a `find_next`. But there might be a better way to also handle special cases. I can see if I can come up with something in a future PR.

Issue #325, opened 2017-05-29
implement `select_next` command


eyelash:
I'm talking about the command that is triggered with ctrl-D or cmd-D on most editors. It selects words if the last selection is a caret and otherwise searches for the last selection and adds a new selection for the next occurrence.

callodacity:
I think it would be important to create a distinction like Sublime Text does:

* If the user starts with an empty selection, then `cmd+d`/`ctrl-d` will search by whole words. 
* If the user starts with a non-empty selection, then it will search for any instance of the text.

For more information, see these:
* https://stackoverflow.com/a/18422707/5552584
* https://forum.sublimetext.com/t/bounded-quick-add-next/8609/6

trishume:
@eyelash whoah that's super weird that there's a difference between selecting `bcd` using `ctrl+d` and double-clicking, interesting. That seems unintuitive and I have no idea what difference in internal state is going on there. Thanks for pointing it out!

I wonder if that's an intentional design choice or just a quirk of the way they implemented it. I can actually kind of see that behavior being handy in some cases. If I don't think about consistency it does what I think I'm most likely to want in those two cases.

That seems like something that can be an extension of the behavior later and isn't necessary to replicate for an initial implementation, if we do want to replicate it.

eyelash:
Both Atom and Sublime Text also have a [command bound to Ctrl+K Ctrl+D](https://stackoverflow.com/q/11548308/799849) that selects the next find but skips the current one. It would be nice to have this command in Xi as well.

cmyr:
That could maybe be a flag? In any case, good feature.

PR #704, opened 2018-06-12
Add next to selection

scholtzan:
This depends on https://github.com/google/xi-editor/pull/701 and resolves https://github.com/google/xi-editor/issues/325

### DIFF
--- a/rust/core-lib/src/edit_types.rs
+++ b/rust/core-lib/src/edit_types.rs
@@ -36,8 +36,8 @@ pub(crate) enum ViewEvent {
     Gesture { line: u64, col: u64, ty: GestureType },
     GotoLine { line: u64 },
     Find { chars: String, case_sensitive: bool, regex: Option<bool> },
-    FindNext { wrap_around: Option<bool>, allow_same: Option<bool> },
-    FindPrevious { wrap_around: Option<bool> },
+    FindNext { wrap_around: Option<bool>, allow_same: Option<bool>, add_to_selection: Option<bool> },
+    FindPrevious { wrap_around: Option<bool>, allow_same: Option<bool>, add_to_selection: Option<bool> },
     Cancel,
     HighlightFind { visible: bool },
     SelectionForFind { case_sensitive: Option<bool> },
@@ -199,10 +199,10 @@ impl From<EditNotification> for EventDomain {
             Redo => BufferEvent::Redo.into(),
             Find { chars, case_sensitive, regex } =>
                 ViewEvent::Find { chars, case_sensitive, regex }.into(),
-            FindNext { wrap_around, allow_same } =>
-                ViewEvent::FindNext { wrap_around, allow_same }.into(),
-            FindPrevious { wrap_around } =>
-                ViewEvent::FindPrevious { wrap_around }.into(),
+            FindNext { wrap_around, allow_same, add_to_selection } =>
+                ViewEvent::FindNext { wrap_around, allow_same, add_to_selection }.into(),
+            FindPrevious { wrap_around, allow_same, add_to_selection } =>
+                ViewEvent::FindPrevious { wrap_around, allow_same, add_to_selection }.into(),
             DebugRewrap => SpecialEvent::DebugRewrap.into(),
             DebugWrapWidth => SpecialEvent::DebugWrapWidth.into(),
             DebugPrintSpans => SpecialEvent::DebugPrintSpans.into(),
--- a/rust/core-lib/src/edit_types.rs
+++ b/rust/core-lib/src/edit_types.rs
@@ -19,7 +19,7 @@
 //! the editor or view as appropriate.
 
 use movement::Movement;
-use rpc::{GestureType, LineRange, EditNotification, MouseAction};
+use rpc::{GestureType, LineRange, EditNotification, MouseAction, SelectionModifier};
 use view::Size;
 
 
@@ -36,8 +36,8 @@ pub(crate) enum ViewEvent {
     Gesture { line: u64, col: u64, ty: GestureType },
     GotoLine { line: u64 },
     Find { chars: String, case_sensitive: bool, regex: Option<bool> },
-    FindNext { wrap_around: Option<bool>, allow_same: Option<bool>, add_to_selection: Option<bool> },
-    FindPrevious { wrap_around: Option<bool>, allow_same: Option<bool>, add_to_selection: Option<bool> },
+    FindNext { wrap_around: Option<bool>, allow_same: Option<bool>, modify_selection: Option<SelectionModifier> },
+    FindPrevious { wrap_around: Option<bool>, allow_same: Option<bool>, modify_selection: Option<SelectionModifier> },
     Cancel,
     HighlightFind { visible: bool },
     SelectionForFind { case_sensitive: Option<bool> },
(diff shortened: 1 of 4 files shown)


cmyr:
What're your thoughts on the 'remove current selection and add next' variant?

scholtzan:
I would be possible to add another optional parameter for this.

cmyr:
Yea I don't want to go too crazy with options, but yea that could work. Another option might be to have a 'ModifySelection' enum, which could be one of, say, `Add` and `AddRemovingCurrent`? (Just brainstorming, either approach works for me)

scholtzan:
Yes, that also works. 
A bit trickier might be tracking the usage of <kbd>cmd</kbd> + <kbd>k</kbd>. In Atom, it changes the behaviour of quite a few keystrokes, not only <kbd>cmd</kbd> + <kbd>d</kbd>. But in general, it should be possible to implement the 'remove current selection and add next' behaviour.

cmyr:
Yea I don't think we'll bother with <kbd>cmd</kbd><kbd>K</kbd> yet; we can map it to some other key for now (like <kbd>cmd</kbd><kbd>shift</kbd><kbd>D</kbd>) and worry about the UX later.

cmyr:
Just preliminary playing with this, and there's a subtlety we're missing: when using <kbd>cmd</kbd><kbd>D</kbd>, sublime only matches _whole_ words.

It may be that this requires some general changes/additions to the find system, in which case I'd be happy to have an issue open for it and it can be follow-up work; but it's probably the behaviour we want.

scholtzan:
It would be necessary to add an option to only search for `whole_words`. That is probably also a useful option for find in general. 
I'd say it is best to open a separate issue for this and then I can create a separate PR for this to keep things easier to review.

cmyr:
@scholtzan rebase this when you get a chance?

scholtzan:
@cmyr Done

rust/core-lib/src/edit_types.rs
```diff
@@ -36,8 +36,8 @@ pub(crate) enum ViewEvent {
     Gesture { line: u64, col: u64, ty: GestureType },
     GotoLine { line: u64 },
     Find { chars: String, case_sensitive: bool, regex: Option<bool> },
-    FindNext { wrap_around: Option<bool>, allow_same: Option<bool> },
-    FindPrevious { wrap_around: Option<bool> },
+    FindNext { wrap_around: Option<bool>, allow_same: Option<bool>, modify_selection: Option<SelectionModifier> },
```
cmyr: since `SelectionModifier` has a `None` variant, we can maybe avoid having it be an `Option` here. One of the annotations serde lets us add to individual fields is `#[serde(default)]`, which will use the type's implementation of the `Default` trait if the field is missing. Then you just need to implement `Default` for `SelectionModifier`, and you can lose the `Option` while still making the field optional on the client side.

cmyr: (we should probably also be doing this for `wrap_around`. Maybe it makes sense to make these changes together in a follow-on PR)

rust/core-lib/src/view.rs
```diff
@@ -920,24 +924,36 @@ impl View {
...
         // select occurrence closest to last selection
         let sel = match self.sel_regions().last() {
-            Some(sel) => (sel.min(), sel.max()),
+            Some(sel) => {
+                if allow_same && sel.is_caret() {
+                    let (start, end) = {
+                        let mut word_cursor = WordCursor::new(text, sel.max());
+                        word_cursor.select_word()
+                    };
+                    (end, start)
```
cmyr: Why is this (end, start) and not (start, end)? 😕❔

rust/core-lib/src/view.rs
```diff
@@ -920,24 +924,36 @@ impl View {
...
+                } else {
+                    (sel.min(), sel.max())
+                }
+            },
             None => return,
         };
-
+
```
cmyr: whitespace

rust/core-lib/src/edit_types.rs
scholtzan: Yeah, probably best in separate PR.

rust/core-lib/src/view.rs
cmyr: Okay, cool. I'd welcome a short comment; reading the code it definitely looked intentional but I couldn't figure out the logic.

scholtzan:
Yes, it is an issue with wrapping. I am currently looking into the best way of how to solve this.

scholtzan:
I fixed the wrap around issue. 

The <kbd>cmd</kbd><kbd>shift</kbd><kbd>D</kbd> is a bit more tricky. I think it would be required to track the last selection that was added, so that only that one is removed. Currently, the last selection is removed which is not always correct.

cmyr:
Now I have some weird behaviour around cursor positions:

with the cursor at the end of a word, <kbd>cmd</kbd><kbd>D</kbd> seems to use the preceding word to populate the find request, but then selects the _subsequent_ occurrence. Also interesting is that this doesn't seem to wrap around; there's no way for the original occurrence to be selected?

I think we should clarify what the behaviour is in this case. I think there's an argument for only using the current word when the caret is in the interior of a word, but I could also see allowing the caret to be on either boundary; in this case though I think the word that is being used to populate the find request should always be selected?

scholtzan:
I fixed the bug. 
The <kbd>cmd</kbd><kbd>shift</kbd><kbd>D</kbd> behaviour is currently the same as in Sublime, which is that the last selection (closest to the bottom of the document) gets removed. It feels a bit buggy but otherwise we'd need a way to track the selection that was added most recently.

cmyr:
@scholtzan I'd like clarity on my last comment (https://github.com/google/xi-editor/pull/704#issuecomment-400677572); I think this current behaviour is a bug; that's my last blocker.

rust/core-lib/src/find.rs
```diff
@@ -263,29 +263,39 @@ impl Find {
...
             next_occurrence.cloned()
         } else {
             let next_occurrence = self.occurrences.regions_in_range(sel_end + 1, text.len()).first();
 
             if next_occurrence.is_none() && !wrapped {
-                return self.occurrences.regions_in_range(0, text.len()).first().cloned();
+                // get first unselected occurrence
+                return self.occurrences.regions_in_range(0, text.len()).iter().cloned().filter(|o| {
+                    sel.regions_in_range(o.min(), o.max()).is_empty()
+                }).collect::<Vec<SelRegion>>().first().cloned();
```
raphlinus: I think `.next().cloned()` is a more efficient approach than collecting. However, that's not as easy to do for the last, and this code is more parallel to the stanza above, so I'm not suggesting changing it.

scholtzan:
Oh, I forgot to push the fix yesterday. It should be there now.

rust/core-lib/src/view.rs
```diff
@@ -920,27 +924,22 @@ impl View {
...
-            Some(sel) => (sel.min(), sel.max()),
-            None => return,
-        };
-
+    pub fn select_next_occurrence(&mut self, text: &Rope, reverse: bool, wrapped: bool,
+                                  _allow_same: bool, modify_selection: &SelectionModifier) {
         // multiple queries; select closest occurrence
         let closest_occurrence = self.find.iter().flat_map(|x|
-            x.next_occurrence(text, reverse, wrapped, sel)
+            x.next_occurrence(text, reverse, wrapped, &self.selection.clone())
```
raphlinus: Pretty sure this clone is not needed.

rust/core-lib/src/rpc.rs
```diff
@@ -322,6 +322,17 @@ pub struct MouseAction {
     pub click_count: Option<u64>,
 }
 
+/// Represents how the current selection is modified (used by find
+/// operations).
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+#[serde(rename_all = "snake_case")]
+pub enum SelectionModifier {
+    None,
```
cmyr: In your follow on PR about the Option/None stuff, it'd be nice if we could add an explanation for each of these cases.

PR #705, opened 2018-06-12
pin travis to 1.26

leomikezee:


.travis.yml
```diff
@@ -2,17 +2,12 @@ language: rust
 osx_image: xcode9.2
 
 rust:
-  - stable
-  - nightly
```
cmyr: Thanks for this! I think we'd like to keep nightly, and just replace `stable` with 1.26.

PR #706, opened 2018-06-13
Added Initialize method to Plugin Trait

betterclever:
Added an Initialize method to the Plugin Trait. The Plugin can use this method to stash the RpcCtx to send 
RPCs asynchronously. 

An example use case is Language Server Plugin. The Language Server sends diagnostics to the Plugin asynchronously as JSON RPC Spec 2.0 Notifications, they must be forwarded to Core -> Client for it to display. But currently RPCs can be sent only using View Proxy that holds an RpcPeer reference which is in scope when the specified methods for new view/ update are dispatched by the Dispatcher. 
But plugin needs raw access to RPC peer to send notifications asynchronously as they arrive. So it can now stash a reference to the RPC Peer in the initialize method.

### DIFF
--- a/rust/plugin-lib/src/core_proxy.rs
+++ b/rust/plugin-lib/src/core_proxy.rs
@@ -1,0 +1,31 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A proxy for the methods on Core
+
+use xi_rpc::{RpcCtx, RpcPeer};
+
+#[derive(Clone)]
+pub struct CoreProxy {
+    peer: RpcPeer
+}
+
+impl CoreProxy {
+
+    pub fn new(rpc_ctx: &RpcCtx) -> Self {
+        CoreProxy {
+            peer: rpc_ctx.get_peer().clone()
+        }
+    }
+}
--- a/rust/plugin-lib/src/lib.rs
+++ b/rust/plugin-lib/src/lib.rs
@@ -30,6 +30,7 @@ mod state_cache;
 mod base_cache;
 mod view;
 mod dispatch;
+mod core_proxy;
 
 use std::io;
 use std::path::Path;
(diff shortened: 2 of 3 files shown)


rust/plugin-lib/src/lib.rs
```diff
@@ -109,6 +110,11 @@ pub trait Cache {
 pub trait Plugin {
     type Cache: Cache;
 
+    /// Called when the Plugin is initialized. RpcCtx is supplied to
+    /// the plugin so that it can stash a reference to peer in case it 
+    /// wants to send async notifications to the core later
```
cmyr: Nit: let's end this sentence with a `.`.

rust/plugin-lib/src/lib.rs
```diff
@@ -109,6 +110,11 @@ pub trait Cache {
...
+    fn initialize(&mut self, ctx: &RpcCtx);
```
cmyr: You can include a default implementation in the trait itself, which means plugins aren't required to implement it at all.

rust/sample-plugin/src/main.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Google Inc. All rights reserved.
+// Copyright 2018 Google LLC
```
raphlinus: There's no need to update copyright notices on existing files, only to make sure they're right on new files.

rust/plugin-lib/src/lib.rs
```diff
@@ -109,6 +111,11 @@ pub trait Cache {
 pub trait Plugin {
     type Cache: Cache;
 
+    /// Called when the Plugin is initialized. The plugin receives CoreProxy
+    /// object that is a wrapper around the RPC Peer and can be used to call
+    /// related methods on the Core in a type-safe manner.
+    fn initialize(&mut self, _core: CoreProxy) {}
```
raphlinus: We prefer `core` (no underline prefix) and `#[allow(unused)]`, because it makes the docs cleaner.

rust/plugin-lib/src/core_proxy.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+impl CoreProxy {
+
+    pub fn new(rpc_ctx: &RpcCtx) -> Self {
+        CoreProxy {
+            peer: rpc_ctx.get_peer().clone()
+        }
+    }
+}
```
raphlinus: Trailing newline

rust/sample-plugin/src/main.rs
cmyr: @betterclever were talking about licences and he pointed out that this file, which was added recently, has the old licence. Not sure if that matters, but this file _was_ added in 2018.

cmyr:
@betterclever there are a few little nits to resolve here, then we can get this in :)

betterclever:
@raphlinus @cmyr Done. Sorry it took so long. The notification about this PR got somewhere lost in my pile of mails.

PR #707, opened 2018-06-14
Clean up dead code

Ephemera:


scholtzan:
Thanks for this! The `allow_same` option is used for adding the next occurrence to the selection (= <kbd>cmd</kbd>+<kbd>d</kbd>) which is introduced in PR https://github.com/google/xi-editor/pull/704. So I'd say we should not remove it since it will actually be used.

Ephemera:
Fair enough, I will close this. Thanks!

cmyr:
@Ephemera Thanks for the PR in any case, always appreciate cleanup.

PR #711, opened 2018-06-15
Find whole words

scholtzan:
This adds a find option to only match whole words. This depends on https://github.com/google/xi-editor/pull/704


cmyr:
Looks good! still need  to give the mac stuff a quick pass, but let's get this in now.

PR #718, opened 2018-06-21
Preliminary cursor info status item support

cmyr:
This shouldn't be merged in its current state, and is intended as a way of getting a better feel for the statusbar added in https://github.com/google/xi-mac/pull/183.

@raphlinus you can give this + that a try at some point.

@nangtrongvuon you might want to give this a read too?

### DIFF
--- a/rust/core-lib/assets/client_example.toml
+++ b/rust/core-lib/assets/client_example.toml
@@ -36,3 +36,6 @@ wrap_width = 0
 
 # If true, wraps lines at the edge of the view. Overrides 'wrap_width'.
 word_wrap = false
+
+# If true, shows the cursor position or selections count as a status item.
+show_cursor_status = false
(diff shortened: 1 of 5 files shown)


nangtrongvuon:
Now that the status bar is more refined (in my opinion!), do you think this could get in?

jansol:
Having played around with this for a bit, I have somewhat mixed feelings about it.

* I'm not really convinced that it's actually useful. Maybe as an example plugin, but it doen't feel necessary as core functionality
* Multiple cursors and multiple selections should be distinguished
* Byte count of selections is nice buuut it's hard to justify why it should be counting bytes specifically and not graphemes/words/whatever
* Column position should be calculated in "navigation units" instead of bytes. I wasn't entirely sure how this should count fullwidth characters, e.g. emojis but given that this is a text editor and not a terminal, those should logically count as one unit, because the character grid of a terminal doesn't exist here (nor does 'monospace' as a concept beyond display implementation details)

jansol:
A more useful item might be showing the current file format/language, tab and newline settings? I still think this is useful in its current form - but only in some cases, and not even in all cases when using xi as a code editor.

And as I mentioned earlier, as a practical example for plugin developers it's definitely a good start.

cmyr:
🤷‍♂️

Issue #712, opened 2018-06-16
RPC for replace


scholtzan:
# RPC for Replace

This is for discussing the RPC for replace. I imagine it to be quite similar to the find RPC, but perhaps a bit simpler.

## Actions related to replace

* Show replace bar with find: <kbd>⌥</kbd> + <kbd>⌘</kbd> + <kbd>F</kbd>
    * Filled with last replacement string 
* Replace next: <kbd>⌘</kbd> + <kbd>⌥</kbd> + <kbd>E</kbd>
    * If there is a selected occurrence: `[insert, find_next]`
    * No occurrence is selected: `find_next`
    * Option to preserve case: <kbd>⌘</kbd> + <kbd>⌥</kbd> + <kbd>A</kbd>
* Replace all: <kbd>^</kbd> + <kbd>⌥</kbd> + <kbd>return</kbd>
    * `[find_all, insert]`
* Use selection for replace: <kbd>⇧</kbd> + <kbd>⌘</kbd> + <kbd>E</kbd>
    * Always case insensitive

## Proposed RPC

```
replace { "chars": "n", "preserve_case": true }
replace_next { }
replace_all { }
selection_for_replace { }
replace_status { "chars": "n", "preserve_case": true }
```

* Replacement string and options are stored in the core similarly to find
* Replacement string and options get sent to the core while typing

cmyr:
I'll admit I'm not super deep in find/replace land right now, but this sounds like a very good place to start. I think with some of this work it makes sense to get it working, and then we'll have a chance to play around more and see if we should refine.

scholtzan:
Implemented in https://github.com/google/xi-editor/pull/720

PR #720, opened 2018-06-22
Replace

scholtzan:
This is a WIP for replace. Feedback is welcome. This PR depends on https://github.com/google/xi-editor/pull/711. This is an implementation of the protocol discussed in https://github.com/google/xi-editor/issues/712

## Tasks

- [x] Find all 
- [x] Replace next
- [x] Replace all
- [x] Set replace in core
- [x] Replace status
- (Preserve case (possible future work))

### DIFF
--- a/rust/core-lib/src/client.rs
+++ b/rust/core-lib/src/client.rs
@@ -24,6 +24,7 @@ use config::Table;
 use styles::ThemeSettings;
 use plugins::rpc::ClientPluginInfo;
 use plugins::Command;
+use internal::view::Replace;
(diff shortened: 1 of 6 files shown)


scholtzan:
Currently, the `preserve_case` option is ignored. But since it is quite tricky to implement, I think it is a feature that could be added in the future.

cmyr:
@scholtzan This has a small conflict, then we can get it in

scholtzan:
@cmyr Conflicts should be resolved now

PR #723, opened 2018-06-23
combine view protocol document with main docs

euclio:
This commit also tweaks the language a bit:

- Use "ARGB" instead of "RGBA", which is more accurate.
- Mention the specific notifications that are used by the protocol.
- Use the correct name for the `def_style` notification.
- Clarify the use of the `scroll_to` notification.


cmyr:
Thanks for this, pardon the delay!

Issue #210, opened 2017-04-18
Chokes on Large Files


wallywyyoung:
Using xi-mac, I opened a relatively large XML file (2.1MB) and was quickly met with the spinny beachball of death.

Given the rope data structure design decision, I thought I'd use this as a sort of ground-truth test for large file handling.

Let me know if this ticket isn't appropriate, or if it isn't in the appropriate place.

Including file for independent verification.

[Jobe_142274871.txt](https://github.com/google/xi-editor/files/930618/Jobe_142274871.txt)

cmyr:
This isn't about filesize as much as it is inefficiencies around handling extremely long lines. Some of this is definitely the client, where the drawing code isn't very efficient, but there's some slowdown in the core as well. Maybe rename the issue to "choking on long lines"?

wallywyyoung:
Updated! Anytime! :)

raphlinus:
Dup of #80, but this should probably be filed against the xi-mac project because as far as I can tell the core is not the problem here.

In profiling, the call to CTLineGetOffsetForStringIndex dominates (this is to position the cursor). Ultimately this should probably be filed against Apple. In any case, might be interesting to do some performance analysis; because it's this slow I suspect it might be O(n^2). Can't wait to see the blog post at accidentallyquadratic :)

wallywyyoung:
Cross-filed @ https://github.com/google/xi-mac/issues/20

It's certainly a sufficiently-large n. :)

SCdF:
I've seen an interesting interaction with the syntax highlighter with very long lines as well. For our project we generate a massive JS file that ends with a huge single line comment at the end containing the base64'd source map.

This is not normally commited, but I've generated and uploaded a copy here: https://github.com/SCdF/medic-webapp-inbox

If you open that file in Xi, and search for `module.exports = widgets;` it will scroll to just before the giant single line. You can then wait for syntax highlighting to catch up, and when it does try to scroll down. You'll also notice that even though the syntax highlighter only took a few seconds to generate syntax for 200,000 lines, it's stuck at 100% CPU on the last half a dozen lines.

PR #725, opened 2018-06-23
Include plugin name with AddStatusItem RPC

nangtrongvuon:
This PR makes adding the status item include the plugin name from which it originates, so that client can manage items that come from the same plugin.

companion to xi-mac [#210](https://github.com/google/xi-mac/pull/210)

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -380,11 +380,9 @@ this writing, the following is valid json for a `Command` object:
 
 #### add_status_item
 
-`add_status_item { plugin, key: "my_key", value: "hello", alignment: "left" }`
+`add_status_item { source: "status_example", key: "my_key", value: "hello", alignment: "left" }`
 
-Adds a status item, which will be displayed on the frontend's status bar. The alignment key dictates whether this item appears on the left side or the right side of the bar. This alignment can only be set when the item is added.
-
-When sent, the RPC also sends the plugin name for which the status item belongs to. 
+Adds a status item, which will be displayed on the frontend's status bar. Status items have a reference to whichever plugin added them. The alignment key dictates whether this item appears on the left side or the right side of the bar. This alignment can only be set when the item is added.
 
 #### update_status_item
 
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -380,7 +380,7 @@ this writing, the following is valid json for a `Command` object:
 
 #### add_status_item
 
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/client.rs
```diff
@@ -19,7 +19,7 @@ use std::time::Instant;
 use serde_json::{self, Value};
 use xi_rpc::{self, RpcPeer};
 
-use tabs::ViewId;
```
cmyr: Is this from rustfmt or something? Normally we don't include braces on a single-item import.

rust/core-lib/src/client.rs
```diff
@@ -147,9 +147,10 @@ impl Client {
         self.0.send_rpc_notification("alert", &json!({ "msg": msg.as_ref() }));
     }
 
-    pub fn add_status_item(&self, view_id: ViewId, key: &str, value: &str, alignment: &str) {
+    pub fn add_status_item(&self, view_id: ViewId, plugin: &str, key: &str, value: &str, alignment: &str) {
```
cmyr: I was thinking about this a bit more, and I think instead of calling this `plugin` we should call it `source`. I think it's likely that core will end up using this API, so we should be more general.

rust/core-lib/src/event_context.rs
```diff
@@ -163,8 +163,11 @@ impl<'a> EventContext<'a> {
             Edit { edit } => self.with_editor(
                 |ed, _, _, _| ed.apply_plugin_edit(edit)),
             Alert { msg } => self.client.alert(&msg),
-            AddStatusItem { key, value, alignment }  => self.client.add_status_item(
-                                                        self.view_id, &key, &value, &alignment),
+            AddStatusItem { key, value, alignment }  => {
+            let plugin_name = &self.plugins.iter().find(|p| p.id == plugin).unwrap().name;
+            self.client.add_status_item(
```
cmyr: Formatting here is a bit messed up.

docs/docs/frontend-protocol.md
```diff
@@ -384,6 +384,8 @@ this writing, the following is valid json for a `Command` object:
```
cmyr: the signature for the method needs to be updated as well.

rust/core-lib/src/TEST
```diff
@@ -0,0 +1 @@
+  2. Press  v  and move the cursor to the fifth item below.  Notice that the
```
cmyr: ?

cmyr:
@nangtrongvuon can you take another pass over this, and then I'll take a look?

Issue #719, opened 2018-06-21
Go-To Definition Support


betterclever:
The next feature we aim towards adding for Language Support is Go-to definition of a symbol. The RPCs 
Client -> Core and Core -> Plugin and vice versa are very similar to Hover Definition but we need a way to instruct a client to open a document in a view itself given the filesystem Url.

The RPCs for Language features are being discussed in Issue #693.

cmyr:
I'm working on adding a core-initiated `new_view` RPC, will put something up in a little bit.

cmyr:
Yea, so I've been thinking about this too.

There are two options.

The first is the way I was thinking about this that led to #727; when core receives a goto event from the plugin, it tells the client to open a new view, and then sends the `scroll_to` command, with the contents of the `Location`.

The second option is one that works without #727; when it receives a goto event, core could just send the client some message like `resolve_goto`, that includes a path and a line/col; then the _client_ would drive the opening of the file and scrolling to the necessary position.

I'm honestly not sure what the best approach is. We probably _do_ need the functionality of #727 at _some_ point, for things like opening diffs, or displaying other transient files.

To do it the #727 way you would handle the plugin result is to handle the RPC in `CoreState`, call `do_new_view_async` (which would need to return a result that includes the new view's `ViewId`, which I overlooked) and then use this could then be used to send the `scroll_to` RPC.

betterclever:
Got it. I have completed work on Plugin Side. Can you add the `ViewId` return fix ? I would rebase my work on that PR then.

cmyr:
done!

betterclever:
To wrap up things that were discussed yesterday.
- The location returned from Core-> Client will be a Vec<Location> where Location is as defined [here](https://github.com/google/xi-editor/pull/717/files#diff-b935638a9ebd01651660a7f96654b66fR235):

The client works this way:
- Vec<Location> is empty: Display "no definition found" or any similar message (maybe popup).
- Vec<Location> size is 1: Move to the location if present in the current view or in other open view, else it must open a new view and display it appropriately. The reason of handling it in the client is that, core does not know the exact behaviour of client. For example, it might want to display things in a split view if CLI.
- Vec<Location> size > 1: Display a window for user to select the location in a GUI client.

There are a lot of open questions like, the RPC needed from Client -> Core to resolve the ViewId given the path to a file, how should the UI for selection look like in frontends, say Xi-mac. 

[...]

PR #727, opened 2018-06-24
Add core->client new_view RPC

cmyr:
This adds a core->client RPC for opening a new view. It also adds a
client->core RPC, `new_view_async`, that is mostly intended to test that
thew new core RPC works.

Regardless this needs docs before it should be merged.

requires https://github.com/google/xi-mac/pull/209
progress on #719


cmyr:
I'm going to close this, we don't need it right now and it isn't clear it's the best approach.

PR #730, opened 2018-06-28
Tweak CI config

fkorotkov:
By leveraging `--all` flag so everything can be executed in parallel.

It makes CI builds about 2 times faster from ~10 minutes to ~5 minutes.


cmyr:
No worries, good to know. 😎

PR #732, opened 2018-07-03
Configure Cirrus CI

fkorotkov:
Since you liked my tweaks to CI config in #730 I decided to propose a more radical approach which gives  **almost 3x performance improvement**. Switch the CI (for now just to run in parallel with Travis CI for evaluation). Feel free to close the PR if it's not interesting.

Here are a few wins of using Cirrus CI:
* CI times for Linux builds improved from ~9 minutes to ~3 minutes thanks to fast caching and using containers with 4CPUs and 8Gb of memory.
* Cirrus CI uses the **official** Docker Images for Rust and Rust Nightly.
* macOS builds are pretty much the same.
* This PR configures only Linux and macOS builds but Cirrus CI supports Windows for when you'll start testing on Windows.

Here is an example of [a build for this change](https://cirrus-ci.com/build/5738995405291520):

image

And [a particular task](https://cirrus-ci.com/task/5666411531730944) of the build:

image

Cirrus CI also shows a more fine grained status checks for PRs:

image

Please feel free to close the PR. But in case you'll decide to merge, and admin of GitHub Google Organization should configure [Cirrus CI App](https://github.com/marketplace/cirrus-ci) for free for this public repository.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,35 @@
+env:
+  CARGO_HOME: /tmp/cargo
+
+task:
+  matrix:
+    - install_cmake_script: apt-get update && apt-get install -y cmake
+      matrix:
+        - name: test (linux)
+          container:
+            image: rust:latest
+            cpu: 4
+            memory: 8G
+        - name: test (linux nightly)
+          allow_failures: true
+          container:
+            image: rustlang/rust:nightly
+            cpu: 4
+            memory: 8G
+    - osx_instance:
+        image: high-sierra-xcode-9.4
+      env:
+        PATH: $PATH:$CARGO_HOME/bin
+      install_cmake_script: brew install cmake
+      matrix:
+        - name: test (macOS)
+          install_rust_script: curl https://sh.rustup.rs -sSf | sh -s -- -y
+        - name: test (macOS nightly)
+          allow_failures: true
+          install_rust_script: curl https://sh.rustup.rs -sSf | sh -s -- -y --default-toolchain nightly
+  cargo_cache:
+    folder: $CARGO_HOME/registry
+    fingerprint_script: cat rust/Cargo.lock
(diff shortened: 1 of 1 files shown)


cmyr:
Cool, I know that @raphlinus has expressed past interested in having CI for windows, so this is exciting. I'll dig in a bit more when I have a moment, but certainly welcome the suggestion!

.cirrus.yml
```diff
@@ -0,0 +1,35 @@
+env:
+  CARGO_HOME: /tmp/cargo
+
+task:
+  matrix:
+    - install_cmake_script: apt-get update && apt-get install -y cmake
+      matrix:
+        - name: test (linux)
+          container:
+            image: rust:latest
```
raphlinus: Would we prefer rust:latest here or a specific pin? I'm actually open to it being latest here but a fixed in in Travis, but it should be a conscious decision.

cmyr: we could optionally limit this to the windows build. I don't have strong feelings here, but I miiiight lean towards keeping it pinned; if travis is down or we want to just switch at some point it would be nice to make sure we're kept honest.

cmyr:
okay, let's try this out. Thanks @fkorotkov!

fkorotkov:
@cmyr awesome! But seems you haven't installed [Cirrus CI App](https://github.com/marketplace/cirrus-ci) for `xi-editor` org. It's free but it's need to be installed so GitHub will send webhook events.

Please don't hesitate to ping me if you'll have any issues.

cmyr:
hm, I thought I had. I see one build running on cirrus-ci.com, and I do see it listed in 'apps and integrations'. Is there something else I need to do?

fkorotkov:
Oh. You merged and installed. That's why I didn't see a build for master. That makes sense. Sorry for the noise. 🙌

PR #741, opened 2018-07-10
Add default serializer for SelectionModifier in FindNext and FindPrevious

scholtzan:
This is a followup PR for https://github.com/google/xi-editor/pull/704. I am not sure if it would make sense to have default serializers for `wrap_around ` and `allow_same` too? 
Currently, serde does not support default literals such as `false` (https://github.com/serde-rs/serde/issues/368) so it would be necessary to write some workaround with functions.


cmyr:
`bool` implements `Default`, so you should be able to just use `#[serde(default)]`, see [here](https://gist.github.com/rust-play/c17ef82b93cc6cbd90839445d714e1a4).

(maybe I'm misunderstanding the problem?)

cmyr:
@scholtzan this needs a rebase as well.

scholtzan:
@cmyr Done

rust/core-lib/src/view.rs
```diff
@@ -208,16 +208,11 @@ impl View {
...
             GotoLine { line } => self.goto_line(text, line),
             Find { chars, case_sensitive, regex, whole_words } =>
-                self.do_find(text, chars, case_sensitive, regex.unwrap_or(false),
-                             whole_words.unwrap_or(false)),
+                self.do_find(text, chars, case_sensitive, regex, whole_words),
             FindNext { wrap_around, allow_same, modify_selection } =>
-                self.find_next(text, false, wrap_around.unwrap_or(false),
-                               allow_same.unwrap_or(false),
-                               &modify_selection.unwrap_or(SelectionModifier::Set)),
+                self.find_next(text, false, wrap_around, allow_same, &modify_selection),
```
cmyr: Ah I actually have one more nit while we're in here; by convention, we preface the names of functions that are directly handling RPCs with `do_`, so this should be `do_find_next`, and below should be `do_selection_for_find`, etc.

scholtzan: Done

PR #745, opened 2018-07-13
Multiple Search Queries

scholtzan:
First steps towards supporting multiple search queries.

I will add more information while I progress. More information, ideas and discussion is collected in https://github.com/google/xi-editor/issues/748

### DIFF
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -25,9 +25,6 @@ use selection::{Selection, SelRegion};
 use xi_rope::tree::Metric;
 use regex::{RegexBuilder, Regex};
 use word_boundaries::WordCursor;
-use serde::de::{Deserialize, Deserializer};
-use serde::ser::{Serialize, Serializer};
-use std::fmt;
 
 const REGEX_SIZE_LIMIT: usize = 1000000;
 
@@ -35,7 +32,7 @@ const REGEX_SIZE_LIMIT: usize = 1000000;
 #[derive(Serialize, Deserialize, Debug)]
 pub struct FindStatus {
     /// Identifier for the current search query.
-    id: FindId,
+    id: usize,
 
     /// The current search query.
     chars: Option<String>,
@@ -53,54 +50,10 @@ pub struct FindStatus {
     matches: usize
 }
 
-/// ID to uniquely identify a search query.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct FindId(pub(crate) usize);
-
-// todo: mostly duplication from ViewId
-impl<'a> From<&'a str> for FindId {
-    fn from(s: &'a str) -> Self {
-        let ord = s.trim_left_matches("find-id-");
-        let ident = usize::from_str_radix(ord, 10)
-            .expect("FindId parsing should never fail");
-        FindId(ident)
-    }
-}
-
-impl From<String> for FindId {
-    fn from(s: String) -> Self {
-        s.as_str().into()
-    }
-}
-
-impl fmt::Display for FindId {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "find-id-{}", self.0)
-    }
-}
-
-impl Serialize for FindId {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-        where S: Serializer
-    {
-        serializer.serialize_str(&self.to_string())
-    }
-}
-
-impl<'de> Deserialize<'de> for FindId
-{
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-        where D: Deserializer<'de>
-    {
-        let s = String::deserialize(deserializer)?;
-        Ok(s.into())
-    }
-}
-
 /// Contains logic to search text
 pub struct Find {
     // Uniquely identifies this search query.
-    id: FindId,
+    id: usize,
     /// The occurrences, which determine the highlights, have been updated.
     hls_dirty: bool,
     /// The currently active search string.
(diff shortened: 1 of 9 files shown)


rust/core-lib/src/find.rs
```diff
@@ -47,27 +53,72 @@ pub struct FindStatus {
     matches: usize
 }
 
+/// ID to uniquely identify a search query.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct FindId(pub(crate) usize);
```
scholtzan: This is quite similar to `ViewId`. I am not sure if there is a more generic solution for this since some code is duplicated?

scholtzan:
This is ready for review.

@cmyr @raphlinus

rust/core-lib/src/find.rs
cmyr: `ViewId` is a string for historical reasons; you could consider having the find id just be a usize?

rust/core-lib/src/view.rs
```diff
@@ -113,6 +116,18 @@ pub struct Replace {
     pub preserve_case: bool
 }
 
+// todo: also used in tab.rs
+#[derive(Debug, Default)]
+struct Counter(Cell<usize>);
```
cmyr: Assuming this is duplicated and not moved, I would just use reuse the existing type (making it pub(crate))?

rust/core-lib/src/view.rs
```diff
@@ -980,13 +1005,30 @@ impl View {
...
+        };
+
         // multiple queries; select closest occurrence
         let closest_occurrence = self.find.iter().flat_map(|x|
             x.next_occurrence(text, reverse, wrapped, &self.selection)
         ).min_by_key(|x| {
             match reverse {
-                true => x.end,
-                false => x.start
+                true => {
```
cmyr: I might choose to write this,

```rust
match reverse {
    true if x.end > cur.start => ...,
    true => ...,
    false if x.start < cur_end => ...,
    false => ...,
}
```

with the same caveat as before, but maybe more so, i.e. I suspect more people disagree with me on this one. :)

rust/core-lib/src/view.rs
```diff
@@ -980,13 +1005,30 @@ impl View {
     /// indicates a search for the next occurrence past the end of the file.
     pub fn select_next_occurrence(&mut self, text: &Rope, reverse: bool, wrapped: bool,
                                   _allow_same: bool, modify_selection: &SelectionModifier) {
+        let (cur_start, cur_end) = match self.selection.last() {
```
cmyr: I might choose to write this,

```rust 
... self.selection.last().map(|s| (s.min(), s.max()).unwrap_or((0, 0))
```

but it's purely a style thing, and it's totally reasonable as written.

rust/core-lib/src/styles.rs
```diff
@@ -22,7 +22,7 @@ use syntect::highlighting::{Color, Theme, ThemeSet, Highlighter};
 
 pub use syntect::highlighting::ThemeSettings;
 
-const N_RESERVED_STYLES: usize = 2;
+const N_RESERVED_STYLES: usize = 8;
```
cmyr: I feel like style 2 is hardcoded somewhere in the word wrap code, maybe take a quick look? It might need to be updated.

rust/core-lib/src/view.rs
```diff
@@ -927,29 +935,46 @@ impl View {
...
         self.find_changed = FindStatusChange::Matches;
 
-        // todo: this will be changed once multiple queries are supported
-        // todo: for now only a single search query is supported however in the future
-        // todo: the correct Find instance needs to be updated with the new parameters
-        if self.find.is_empty() {
-            self.find.push(Find::new());
-        }
+        // remove deleted queries
+        self.find.retain(|f| queries.iter().position(|q| q.id == Some(f.id())).is_some());
```
cmyr: I did not until now know that this method existed on `Vec<_>`. I think in other places I've done a filter and collect to achieve this.

rust/core-lib/src/view.rs
scholtzan: This looks for sure cleaner 👍

rust/core-lib/src/view.rs
scholtzan: Personally, I find it as is easier to read. I would leave it unchanged

scholtzan:
This is ready for another review.

cmyr:
@scholtzan hihi, I admit I've been neglecting this, it feels big and so it feels like it demands a large chunk of focused attention and I haven't felt like I've had so much of that. But it isn't fair to leave it in limbo, that's for sure.

How about: if I haven't gotten to this by the weekend, ping me again and we can maybe pick a time where you can walk me through it on IRC?

scholtzan:
@cmyr Sure, no problem. I just wasn't sure if it had simply been forgotten. Feel free to review it whenever you want, doesn't necessarily have to be this weekend. I'd also be happy to walk you through it on IRC.

scholtzan:
The `find` RPC is a bit different: `find [{"id": 1, "chars": "a", "case_sensitive": false, "regex": false, "whole_words": true}]` vs `find {"chars": "a", "case_sensitive": false, "regex": false, "whole_words": true}`. It might make sense to support both, but I am not sure if it might get confusing if there are two commands for almost the same thing?

jansol:
Won't the id field be necessary when multiple views of the same file are supported anyway?

scholtzan:
Each view would have its own find.

cmyr:
yea this is a slightly tricky one. I'm not sure how many clients have find implemented currently, so I'm not sure how many clients this would break. I don't think supporting two approximately identical calls makes a ton of sense.

PR #747, opened 2018-07-15
Add paste RPC and multi-selection paste

cmyr:
This adds a new RPC, 'paste', and modifies the behaviour of paste in the
multi-selection case. Now, if the number of lines to be pasted matches
the number of selection regions, lines will be spread across cursors. If
not, the behaviour is the same.

This also fixes a long standing issue where paste was treated the same
as keyboard input for the purposes of undo. Now, a paste action will
always create its own undo group.

Progress on #323


rust/core-lib/src/editor.rs
```diff
@@ -818,3 +832,12 @@ fn n_spaces(n: usize) -> &'static str {
     assert!(n <= spaces.len());
     &spaces[..n]
 }
+
+/// Counts the number of lines in the string, not including any trailing newline.
+fn count_lines(s: &str) -> usize {
+    let mut newlines = count_newlines(s);
+    if s.as_bytes().last() == Some(&0xa) {
```
jansol: Do we want to/should we bother with CR/LF/CRLF?

cmyr: This handles LF and CRLF. I don't believe we're treating CR on its own as a newline? At least we aren't doing it elsewhere in the codebase that I can determine.

raphlinus: Short answer, yes. Longer answer, I'm hoping that CR-only files will be usable with word-wrapping on, as the line break iterator will always indicate them as hard breaks. That's not quite the same, because it won't count in line numbering, but I think is good enough considering the likelihood of the files in the wild.

rust/core-lib/src/editor.rs
```diff
@@ -554,13 +554,26 @@ impl Editor {
...
 
-    // TODO: insert from keyboard or input method shouldn't break undo group,
-    // but paste should.
     fn do_insert(&mut self, view: &View, chars: &str) {
         self.this_edit_type = EditType::InsertChars;
         self.insert(view, chars);
     }
 
+    fn do_paste(&mut self, view: &View, chars: &str) {
+        if view.sel_regions().len() == 1
```
raphlinus: This means that it will get an EditType of Other? That seems right to me, but want to think it through.

cmyr: Yes, it gets `Other`; previously it was `Insert`.

PR #749, opened 2018-07-16
Add documentation for find and replace commands

scholtzan:
Some documentation of the find and replace RPC commands.

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -36,11 +36,11 @@ from core: {"id":0,"result": "view-id-1"}
 `client_started {"config_dir" "some/path"?, "client_extras_dir":
 "some/other/path"?}`
 
- Sent by the client immediately after establishing the core connection. This is
- used to perform initial setup. The two arguments are optional; the `config_dir`
- points to a directory where the user's config files and plugins live, and the
- `client_extras_dir` points to a directory where the frontend can package
- additional resources, such as bundled plugins.
+Sent by the client immediately after establishing the core connection. This is
+used to perform initial setup. The two arguments are optional; the `config_dir`
+points to a directory where the user's config files and plugins live, and the
+`client_extras_dir` points to a directory where the frontend can package
+additional resources, such as bundled plugins.
 
 ### new_view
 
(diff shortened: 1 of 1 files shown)


docs/docs/frontend-protocol.md
```diff
@@ -491,22 +559,35 @@ this writing, the following is valid json for a `Command` object:
...
+
+#### find_status
+
+`find_status {"view_id": "view-id-1", "queries": [{"chars": "a", "case_sensitive": false, "is_regex": false, "whole_words": true, "matches": 6}]}`
+
+Notifies the client about the current search queries and search options.
+
+#### replace_status
+
+`replace {"view_id": "view-id-1", "status": {"chars": "a", "preserve_case": false}}`
```
eyelash: `replace_status`?

docs/docs/frontend-protocol.md
```diff
@@ -241,6 +241,74 @@ Sends a custom rpc command to the named receiver. This may be a notification
...
+
+`find {"chars": "a", "case_sensitive": false, "regex": false, "whole_words": true}`
+Parameters `regex` and `whole_words` are optional and by default `false`.
+
+Sets the current search query and options.
+
+#### find_next and find_previous
+
+`find_next {"wrap_around": true, "allow_same": false, "modify_selection": "set"}`
+`find_previous {"wrap_around": true, "allow_same": false, "modify_selection": "set"}`
```
cmyr: I think at least `allow_same` should get a bit of additional documentation, since I don't think it's usage is always clear.

docs/docs/frontend-protocol.md
```diff
@@ -73,7 +73,7 @@ note for `new_view`. Errors are not currently reported.
 
 `set_theme {"theme_name": "InspiredGitHub"}`
 
-Requests that core change the theme. If the change succeeds the client
+Requests that core changes the theme. If the change succeeds the client
```
cmyr: This was correct before, although I appreciate if it isn't very clear. 'Change' is the future tense, and 'changes' is the present; we're requesting that the core change the theme at some point in the (immediate) future.

It's probably more confusing because 'Requests' is present. I think this makes sense though; you could read it like "She asks that he turn down the music."

docs/docs/frontend-protocol.md
scholtzan: Done :)

PR #751, opened 2018-07-17
Added Metric to count UTF-16 code units in Rope

### LINKED ISSUE #750: Add UTF-16 Metric in Xi-Rope.
Currently the Rope has implementations for BaseMetric and LinesMetric. 
The language server protocol specification works on UTF-16 based Line and Code-Units offset.

Thus, it makes sense to add a UTF-16 Metric in Rope and do conversions there rather than doing it elsewhere.  

I am working on this.

betterclever:
Fixes #750 

I am not sure about the implementation of `next` and `prev` which I think should be same for both UTF-8 and UTF-16 measure since the text is actually UTF-8 encoded.

@raphlinus Can you affirm/correct this?

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -1157,4 +1157,40 @@ mod tests {
         let rope = Rope::from("hi\ni'm\nfour\nlines");
         rope.offset_of_line(5);
     }
+
+    #[test]
+    fn utf16_code_units_metric() {
+        let rope = Rope::from("hi\ni'm\nfour\nlines");
+        let utf16_units = rope.measure::<Utf16CodeUnitsMetric>();
+        assert_eq!(utf16_units, 17);
+
+        // position after 'f' in four
+        let utf8_offset = 9;
+        let utf16_units = rope.convert_metrics::<BaseMetric, Utf16CodeUnitsMetric>(utf8_offset);
+        assert_eq!(utf16_units, 9);
+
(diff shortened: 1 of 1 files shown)


rust/rope/src/rope.rs
```diff
@@ -265,12 +269,82 @@ impl Metric<RopeInfo> for LinesMetric {
...
+        s.is_char_boundary(offset)
+    }
+
+    fn to_base_units(s: &String, in_measured_units: usize) -> usize {
+        let mut cur_len_utf16 = 0;
+        let mut cur_len_utf8 = 0;
+        for u in s.chars() {
+            cur_len_utf16 += u.len_utf16();
+            cur_len_utf8 += u.len_utf8();
+            if cur_len_utf16 >= in_measured_units {
```
betterclever: @cmyr I need this condition to get it to work properly. After using it works perfectly for cases I tested i.e. text with and without non-ASCII characters. However, I am not sure why should the `>=` be needed and not just `==`.

cmyr: Where are those test cases??

cmyr: In any case: `==` would fail if, (say) `s` was `😈` and `in_measured_units` was `1`, because `"😈".len_utf16() == 2`.

raphlinus: Yes, this sounds right to me, at least it's consistent with the lines metric. When the measured units are not on a boundary, I think `to_base_units` should "round down". I'm not sure if this is documented anywhere, but as I say being consistent with lines is good.

Issue #746, opened 2018-07-14
Plugins and the representation of positions (utf8 and utf16, absolute and line/column)


cmyr:
## Represent positions

We currently have two principal ways of representing positions in a buffer; either as absolute utf-8 offsets, or as a line number + a utf-8 offset column position.

The Language Server Protocol uses utf-16 line + column pairs to represent all positions. To talk to language servers, we _have_ to convert whatever representation we're using into utf-16 line/col. This is true for things like the position of click events, but it's also true for things like the offsets in a delta.

### From core to plugins
There are a number of ways we could handle this:
#### Convert in the plugin

We could continue sending most locations to plugins as utf-8 offsets, and leave it up to the plugin to handle the conversion. This keeps core conceptually simple, but complicates the writing of plugin libs; almost all plugin libs are going to want to be able to convert offsets to line/col positions of _some_ kind, and so this work will be duplicated by every plugin library.

#### Convert in core, per plugin

We could allow each plugin to specify how it would like to receive locations. Before sending a message to a plugin, we could convert any position information in that message to the plugin's desired format.

#### Convert in core, aggressively

We could _always_ send _all_ representations of a position to the plugin. That is, plugins could receive positions as 4 separate offsets; the absolute utf-8 offset, the line number, and the column in both utf-8 and utf-16 unit spaces. (I'm not sure there is an argument for including the absolute utf-16 offset or not.)

This will minorly increase the size of the RPC messages, but will also remove the need for any conversion to happen in the plugin libs, in the general case. It also has the advantage of only requiring these conversions to be done once per event, and they can be done performantly in core. It will also be easier to write unit tests for the conversion, and if any plugins need to implement their own conversion logic they will have a good source of validated test data.

It is also... inelegant.

### Going the other way (plugins to core)

What do we do with plugins that _send_ us positons in utf-16? This is likely only a concern when interfacing with external code, as with language servers.

If we've chosen to do all the original conversions in the plugin lib, then it makes sense that we would do _these_ conversions there as well. If we're doing conversion in core, though, this gets a bit ugly:

If we're converting positions in core when we send them _to_ plugins, do we allow plugins to send us positions in various coordinate spaces, and normalize them in core? This _also_ makes core a bit messier, but seems to be a direct consequence of converting in core on the way out; if we want to keep conversion logic in one place, we need to do that fully.

## So...

[...]

Issue #693, opened 2018-06-05
RPC for Language Features


betterclever:
As we proceed towards adding Language specific features to the Xi editor, we need to have more methods to RPC for incorporating methods that are needed for IDE like features in Xi. These will be utilized Language Server Plugin but are not specific to it. 

The initial list which we can with are:
- Completion Request: For requesting auto-complete suggestions at a point.
- Hover Definition: For requesting Definitions of a symbol to be displayed as a hover dialog.
- Find references: For requesting project-wide references of a symbol.

We will add more methods as we progress.
For all the 3 methods, we only need the offset of the symbol where/whose the completion/definition/references are requested. Thus, RPC from the core the plugin can have a definition like:

## Hover Definition:

**RPC Notification**

**Method**: hover_request
**Params**:
```
{
    "view_id": number,
    // current document revision to keep track of the returned result
    "rev": number,
    // request identifier
    "request_id": number
    // offset of the position where hover definition is requested
    "offset": number
}
```

It can be argued that request_id is maybe irrelevant since view-id, rev and offset trio make a unique combination for each request. I tend to include it since it will probably be easy for the core to remove the pending request later using a single id. 

In response to this, an RPC Notification can be sent back to the client

**Method**: hover_result

```
{
    "request_id": number,
    "result" : {
        contents: MarkedString | MarkedString[] | MarkupContent,
        range?: {
            offset_start: number,
            offset_end: number
        }  
    }
}
```

MarkedString is a representation of the Markup Text in either Markdown or a language block. Refer here: https://microsoft.github.io/language-server-protocol/specification#textDocument_hover

This is one approach. Other can be to use an async request from core to plugin but I am not sure what are the advantages/disadvantages or having it.

cmyr:
Each of these additions is going to require two RPCs, client<->core and core<->plugin.

I think starting with hover is good, and I think your RPC is a reasonable place to start.

One question that comes up is, when does the client actually send the hover RPC? Is it only when the mouse is actually hovering over an item, or can there be a command that sends the hover RPC, using the current cursor location? These will have to be handled differently, because the client doesn't really know where the cursor is. So one possibility is that `hover_request` has an optional `position` argument, which is a line/column position, like what is used for mouse clicks; when that argument is missing and there is a single cursor we would use the cursor location, otherwise we can ignore the message.

[...]

PR #753, opened 2018-07-18
Added hover support

betterclever:
Tracking issues/ PRs: #717 #693 #746

This PR adds support for Hover via Language Server only. It is currently based on work from #750 but will be rebased to master once that is resolved and merged.

Meanwhile this is working with the Client work done by @nangtrongvuon with these changes. The RPC for client has been updated. See the relevant files in this PR.

The changes required in Xi-mac client PR is:

```swift
document.sendRpcAsync("request_hover", params: ["request_id": hoverRequestID, "position": ["line": hoverPosition.line, "column": hoverPosition.column]])
```

PS: I am yet to test out the Range conversion successfully yet since RLS sends it as `null` and we are not using it on the client either.


rust/lsp-lib/src/conversion_utils.rs
```diff
@@ -0,0 +1,100 @@
...
+        HoverContents::Markup(content) => content.value,
+    };
+    if res.is_empty() {
+        Err(LanguageResponseError::FallbackResponse)
+    } else {
+        Ok(res)
+    }
+}
+
+/// Counts the number of utf-16 code units in the given string.
```
cmyr: can we delete all this now?

cmyr:
Is this ready for review?

rust/lsp-lib/src/language_server_client.rs
```diff
@@ -217,10 +274,59 @@ impl LanguageServerClient {
...
+                uri: self.opened_documents.get(&view_id).unwrap().clone(),
+            },
+            position,
+        };
+
+        let params = Params::from(serde_json::to_value(text_document_position_params).unwrap());
+        self.send_request("textDocument/hover", params, Box::new(on_result))
+    }
+
+    pub fn request_definition<CB>(&mut self, view_id: ViewId, position: Position, on_result: CB)
```
cmyr: these two functions probably don't belong here?

rust/lsp-lib/src/conversion_utils.rs
betterclever: Actually, I currently handled the conversion for Hover only. This is still being used for the update methods. I plan to address them in a separate PR to keep it clean.

rust/lsp-lib/src/language_server_client.rs
betterclever: Removing this bit.

rust/core-lib/src/client.rs
```diff
@@ -22,7 +22,7 @@ use xi_rpc::{self, RpcPeer};
 use tabs::ViewId;
 use config::Table;
 use styles::ThemeSettings;
-use plugins::rpc::ClientPluginInfo;
```
cmyr: ~{...}~

rust/core-lib/src/rpc.rs
```diff
@@ -26,7 +26,7 @@ use serde::de::{self, Deserialize, Deserializer};
 use serde::ser::{self, Serialize, Serializer};
 
 use config::{Table, ConfigDomainExternal};
-use plugins::PlaceholderRpc;
+use plugins::{PlaceholderRpc};
```
cmyr: braces not needed

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -183,7 +184,49 @@ pub enum PluginNotification {
     Alert { msg: String },
     AddStatusItem { key: String, value: String, alignment: String },
     UpdateStatusItem { key: String, value: String  },
-    RemoveStatusItem { key: String }
+    RemoveStatusItem { key: String },
+    ShowHover { request_id: usize, result: Result<Hover, LanguageResponseError>, rev: u64 }
```
cmyr: nit: we can add a trailing `,` here

rust/core-lib/src/rpc.rs
```diff
@@ -322,6 +322,12 @@ pub struct MouseAction {
     pub click_count: Option<u64>,
 }
 
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct Position {
```
cmyr: I'd remove this, and just include `line` and `col` params like we do with the gesture RPC, if only to keep the protocol _somewhat_ consistent.

(in fact, it might make sense for us to be just adding a `hover` `GestureType`?)

rust/lsp-lib/src/lsp_plugin.rs
```diff
@@ -311,6 +288,15 @@ impl Plugin for LspPlugin {
 
     fn did_close(&mut self, view: &View<Self::Cache>) {
         eprintln!("close view {}", view.get_id());
+
+        if let Some(view_info) = self.view_info.get(&view.get_id()) {
+            let ls_client = self
```
cmyr: I'm noticing you have to write these ~4 lines _a lot_..

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -183,7 +184,49 @@ pub enum PluginNotification {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub struct Hover {
+    pub content: String,
+    pub range: Option<Range>
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub struct Position {
```
cmyr: These two types will be confusing for people. I've been trying to come up with better names without a lot of luck. maybe `CorePosition` and `ClientPosition`?

rust/core-lib/src/client.rs
```diff
@@ -36,6 +36,18 @@ pub struct WidthReq {
     pub strings: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize)]
+pub struct Range {
```
cmyr: What is this used for, again? As in, if this is present what is the client supposed to do with that information?

rust/core-lib/src/event_context.rs
```diff
@@ -415,6 +436,75 @@ impl<'a> EventContext<'a> {
...
+    // If position is `None`, it tries to get the current Caret Position and use
+    // that instead
+    fn get_resolved_position(&mut self, position: Option<ClientPosition>) -> Option<Position> {
+        position.map(|p|
+            self.with_view(|view, text| view.line_col_to_offset(text, p.line, p.column)
+        )).or_else(|| self.view.borrow().get_caret_offset())
+        .map(|offset| self.get_position(offset))
+    }
+
+    fn get_position(&mut self, utf8_offset: usize) -> Position {
```
cmyr: I would call this something like `offset_to_plugin_position`.

rust/core-lib/src/event_context.rs
```diff
@@ -415,6 +436,75 @@ impl<'a> EventContext<'a> {
...
+    }
+
+    fn get_range_in_utf8_offset(&mut self, range: Range) -> client::Range {
+        client::Range {
+            start: self.get_utf8_offset_from_position_enum(range.start),
+            end: self.get_utf8_offset_from_position_enum(range.end),
+        }
+    }
+
+    fn get_utf8_offset_from_position_enum(&mut self, position: PositionEnum) -> usize {
```
cmyr: maybe `offset_from_plugin_position`? I appreciate that this is a different type than above, but at least this fn name communicates what we're doing.

rust/core-lib/src/rpc.rs
betterclever: I added this Position struct since the whole parameter i.e. line/col is optional, it uses caret position as a fallback (this makes more sense for terminal clients). However, adding this as a gesture type is a good possibility but how do terminal client invoke this then?

cmyr: Okay, good point.

rust/lsp-lib/src/lsp_plugin.rs
betterclever: I added a helper function that do take rest of the code in a closure to clean this up (see below). Will use that here as well.

rust/core-lib/src/client.rs
betterclever: The client is supposed to highlight this area while hover is active. Thinking again, it makes sense to convert this to Line/Col according to View which client can use to have a tint over that area. What do you think?

betterclever:
@cmyr I fixed most of the things. @nangtrongvuon PR does not have the above change in RPC request yet so be sure to make a change if running now.

rust/core-lib/src/client.rs
betterclever: @cmyr Views on this?

betterclever:
@cmyr this failed due to a storage full error, I think it is unrelated to this work. Maybe restart the build?

rust/core-lib/src/client.rs
```diff
@@ -36,6 +36,18 @@ pub struct WidthReq {
     pub strings: Vec<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Range {
```
cmyr: let's not send the range to the client.

rust/core-lib/src/event_context.rs
```diff
@@ -99,6 +100,26 @@ impl<'a> EventContext<'a> {
         f(&mut view, editor.get_buffer())
     }
 
+    fn with_rope<R, F>(&mut self, f: F) -> R
```
cmyr: I'd either just use `with_view` or rename this to `with_text`.

rust/core-lib/src/event_context.rs
```diff
@@ -99,6 +100,26 @@ impl<'a> EventContext<'a> {
...
+        where F: FnOnce(&Rope) -> R 
+    {
+        let editor = self.editor.borrow();
+        f(editor.get_buffer())
+    }
+
+    fn with_rope_rev<R, F>(&mut self, rev: u64, f: F) -> Option<R>
```
cmyr: "with_rev"?

rust/core-lib/src/event_context.rs
```diff
@@ -169,7 +192,16 @@ impl<'a> EventContext<'a> {
             }
             UpdateStatusItem { key, value } => self.client.update_status_item(
                                                         self.view_id, &key, &value),
-            RemoveStatusItem { key } => self.client.remove_status_item(self.view_id, &key)
+            RemoveStatusItem { key } => self.client.remove_status_item(self.view_id, &key),
+            ShowHover { request_id, result, rev } => {
```
cmyr: I would put all the login in something like `do_show_hover`.

rust/core-lib/src/event_context.rs
```diff
@@ -415,6 +447,75 @@ impl<'a> EventContext<'a> {
...
                            ed.is_pristine())
     }
+
+    fn do_request_hover(&mut self, request_id: usize, position: Option<ClientPosition>) {
+        if let Some(position) = self.get_resolved_position(position) {
+            self.with_each_plugin(|p| p.get_hover(self.view_id, request_id, &position))
+        }
+    }
+    
+    // Gives the requested position in UTF-8 offset format to be sent to plugin
```
cmyr: three /// ?

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -183,7 +184,49 @@ pub enum PluginNotification {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub struct Range {
+    pub start: PluginPosition,
+    pub end: PluginPosition
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum LanguageResponseError {
```
cmyr: let's use `xi_rpc::RemoteError`

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -183,7 +184,49 @@ pub enum PluginNotification {
...
+#[serde(rename_all = "snake_case")]
+pub struct Hover {
+    pub content: String,
+    pub range: Option<Range>
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+pub struct CorePosition {
+    pub utf8_offset: usize,
```
cmyr: maybe just `offset`? utf8 offsets are our default representation. You could mention that it's utf8 in a doc comment.

PR #756, opened 2018-07-21
 Improve EditType and undo grouping

### LINKED ISSUE #742: improvements to EditType
xi-core has the concept, in `editor.rs`, of 'edit types', represented by the `EditType` enum. `EditType` has two distinct purposes that I can determine: it is used to decide how user edits should be combined into undo groups, and it is also used to communicate information about edits to plugins.

I think it would be helpful if plugins had additional information about the cause of updates; in particular I would like for plugins to know if a newline is the result of pressing `return`, or if it was pasted in with other text, or any other possible scenario.

I think the easiest way to do this is to expand `EditType`. I propose adding at least `InsertNewline`, `Paste`, and `InsertTab` variants, and possibly more. 

As this will increase the complexity of managing undo groups, we might also want to move the undo group logic into some fn on `EditType`, that might look like:

```rust
fn break_undo_group(&self, other: EditType) -> bool { .. }
```

Currently, if a plugin wants to know whether or not an update is the result of the user hitting <kbd>return</kbd>, we have to manually inspect the string contents of the delta, and even that doesn't tell us anything with any certainty. While the `update` RPC is an efficient way of keeping the plugin's state up to date, it is missing some information that is important to other use cases.

This will also require us to add some way for the client to indicate that an insert is a `paste`, which is a separate but related issue.

cmyr:
This is based on #747, which should be approved first.

EditType now has distinct members for 'insert newline' and 'adjust
indentation', and we now use serde instead of a manual fn when putting
EditType on the wire.

closes #742

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -804,10 +804,9 @@ pub enum EditType {
 impl EditType {
     /// Checks whether a new undo group should be created between two edits.
     fn breaks_undo_group(self, previous: EditType) -> bool {
-        if self == EditType::Other || self == EditType::Transpose {
-            return true;
-        }
-        self != previous
+        self == EditType::Other
+        || self == EditType::Transpose
+        || self != previous
     }
 }
 
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -18,7 +18,7 @@ use std::collections::BTreeSet;
 
 use serde_json::Value;
 
-use xi_rope::rope::{Rope, RopeInfo, LinesMetric};
+use xi_rope::rope::{Rope, RopeInfo, LinesMetric, count_newlines};
 use xi_rope::interval::Interval;
 use xi_rope::delta::{self, Delta, Transformer};
 use xi_rope::engine::{Engine, RevId, RevToken};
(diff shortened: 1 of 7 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -761,29 +782,36 @@ impl Editor {
...
-        match *self {
-            EditType::InsertChars => "insert",
-            EditType::Delete => "delete",
-            EditType::Undo => "undo",
-            EditType::Redo => "redo",
-            EditType::Transpose => "transpose",
-            _ => "other",
+    /// Checks whether a new undo group should be created between two edits.
+    fn breaks_undo_group(self, previous: EditType) -> bool {
+        if self == EditType::Other || self == EditType::Transpose {
```
raphlinus: I'd personally be inclined to write this `self == EditType::Other || self == EditType::Transpose || self != previous` but it's a style judgment call.

rust/core-lib/src/event_context.rs
```diff
@@ -211,14 +211,20 @@ impl<'a> EventContext<'a> {
         let delta = if approx_size > MAX_SIZE_LIMIT { None } else { Some(delta) };
 
         let undo_group = ed.get_active_undo_group();
+        //TODO: we want to just put EditType on the wire, but don't want
+        //to update the plugin lib quite yet.
+        let edit_type_str = serde_json::to_string(&ed.get_edit_type())
```
raphlinus: This is janky, but I understand it won't be the final state of the code. Would it be possible to serde to a json::Value instead of all the way to a string?

PR #758, opened 2018-07-22
Fix incorrect adjustment of offsets on delete

cmyr:
We were comparing the offset of the newline itself as if it were the
offset of the newline character, which is newline - 1.

In particular, this allowed the cached offsets to become polluted in instances where the last char in a deleted range was a newline. This had a bunch of nefarious consequences; `get_line` would misbehave, returning incorrect ranges of the document in certain cases, which would cause syntect to fail to detect when spans had converged, and abandon highlighting; I'm sure there were other problems as well.

cc @betterclever, does this fix anything for you?


cmyr:
Indeed, I didn't understand what was happening until I stepped through with the debugger. A hundred lines of tests for a 6 character change tells the story pretty well...

PR #759, opened 2018-07-22
Fix incorrectly trying to fetch implicit last line

### LINKED ISSUE #721: plugin refetching whole document on some inserts
Not totally sure what's going on here but in a few trace files I've seen the plugin fetching the entire document in situations where that should not be necessary.

cmyr:
self.num_lines is a count, and line_num is an index; if line_num ==
num_lines, it means line_num is the 'implicit' last line, which doesn't
actually exist.

I haven't thought about this too much; there's possibly an argument for
returning the empty string here, but I'm not sure that's right.

I suspect that this fixes #721.


cmyr:
Okay I've compared some traces and this does look to have fixed #721.

PR #761, opened 2018-07-23
Added Logging support using fern

### LINKED ISSUE #408: Real logging in xi-core (+ core RPC logging)

betterclever:
Fixes #408

### DIFF
--- a/rust/plugin-lib/src/lib.rs
+++ b/rust/plugin-lib/src/lib.rs
@@ -27,6 +27,11 @@ extern crate rand;
 extern crate memchr;
 extern crate languageserver_types;
 
+#[macro_use]
+extern crate log;
+extern crate fern;
+extern crate chrono;
+
 mod state_cache;
 mod base_cache;
 mod view;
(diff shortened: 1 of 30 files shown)


cmyr:
So it looks like the approach we'll have to take here is to keep `fern` behind a feature flag, and not expose it on fuchsia. For fuchsia, we will declare our own versions of these macros that just print to stderr.

betterclever:
@cmyr You may have a look. I don't know how to handle it for Fuschia yet. You said something about cfg flags.

cmyr:
Awesome, will need to give this a spin but it looks good.

It looks like Fuchsia won't be a problem; because `fern` is only used in the main binary, which fuchsia doesn't use, there's no dependency issue. In the fuchsia binary we can pick use the fuchsia system logger, which will be a nice improvement there as well, so this looks like a clean win.

rust/plugin-lib/src/lib.rs
```diff
@@ -165,7 +168,7 @@ pub enum Error {
 }
 
 /// Run `plugin` until it exits, blocking the current thread.
-pub fn mainloop<P: Plugin>(plugin: &mut P) -> Result<(), ReadError> {
+pub fn mainloop<P: Plugin>(plugin: &mut P) -> Result<(), ReadError> {
```
cmyr: whitespace

rust/rpc/Cargo.toml
```diff
@@ -13,3 +13,4 @@ serde_derive = "1.0"
 crossbeam = "0.3.2"
 
 xi-trace = { path = "../trace", version = "0.1.0" }
+log = "0.4.3"
```
cmyr: ditto the earlier cargo comment

rust/src/main.rs
```diff
@@ -15,6 +15,11 @@ use std::io;
 
 extern crate xi_core_lib;
 extern crate xi_rpc;
+extern crate fern;
```
cmyr: again, by convention (and also by alphabetical ordering) I'd prefer the non-xi crates above the internal crates

betterclever:
Done.

cmyr:
Thanks @betterclever, this was long overdue. 😄

Issue #583, opened 2018-03-29
Outstanding rope optimizations


cmyr:
Broken out from https://github.com/google/xi-editor/pull/575#pullrequestreview-107817261: there are two optimizations missing from the current rope impl:

> 
> For the purpose of institutional memory, the old rope implementation has a couple of optimizations that the new one lacks:
> 
> It is more aggressive about editing in-place when a unique reference exists
> 
> The RopeBuilder is O(n) in the number of blocks, using a height-sorted stack, while the newer TreeBuilder just does pairwise concat, each of which is O(log n), so O(n log n) total.
> 
> Neither of these things is going to make a measurable difference for ordinary text editing tasks, but if xi-rope is to be considered a "gold-plated" rope library they should probably get some attention.
> 
> I think the best thing at this point is to have an open issue for the potential improvements, pointing at the deleted code as a possible inspiration.

the code in question is [here](https://github.com/google/xi-editor/blob/608b5329adef1997922b67f4c74fddc0c0204e30/rust/rope/src/lib.rs).

This is not especially high priority, but could be fun for someone interested in the rope data structure.

raphlinus:
Agree that this is a good issue for people interested in the underlying data structures. All of the rope code so far was built on in-my-head algorithmic complexity analysis, but a really good basis for going forward would be benchmarks. That would help quantify the amount of improvement. Part of the reason I had kept the old code around is in case it could be benchmarked, but I don't think that's very important - the benchmarking (and testing) focus can be on the new code.

mqzry:
The PR of @vlovich reminded me that I started working on this.
Have already added benchmarks for rope creation and reimplemented the height-sorted stack approach.

Last time I went into the old code, I could not find where 
> It is more aggressive about editing in-place when a unique reference exists

happens. Can someone point me here?

cmyr:
cc @raphlinus

vlovich:
Sweet.  I only started on the benchmarks so I can just let you finish it.
I too couldn’t find the old optimization’s in the referenced CL

raphlinus:
[...]

PR #762, opened 2018-07-25
Faster tree builder

mqzry:
Reintroduces one of the optimisations mentioned in https://github.com/google/xi-editor/issues/583.
This optimisation uses a height sorted stack to push a string on an existing tree.

I also added some benches. Benches with 'stack' in the name are with the new method:

`test build_tree_few_big_lines           ... bench:      90,491 ns/iter (+/- 6,478)`
`test build_tree_few_big_lines_stack ... bench:      80,863 ns/iter (+/- 4,891)`

`test build_tree_short_lines                     ... bench:     322,070 ns/iter (+/- 21,143)`
`test build_tree_short_lines_stack        	... bench:     209,952 ns/iter (+/- 9,921)`

`test build_tree_triangle                           ... bench:     503,447 ns/iter (+/- 36,269)`
`test build_tree_triangle_stack             	... bench:     299,227 ns/iter (+/- 23,092)`


cmyr:
Cool, those are significant gains. Probably won't get around to giving this a thorough review for a few days, but happy to have you plugging away at perf stuff!

mqzry:
I feel like there should be some tests.

I will add on sunday a test to see if the new method yields exectly the same rope as the old method.

Let me know if there are any other scenarios I should.

rust/rope/src/tree.rs
```diff
@@ -478,7 +471,29 @@ impl<N: NodeInfo> TreeBuilder<N> {
     pub fn push(&mut self, n: Node<N>) {
         match self.0.take() {
             None => self.0 = Some(n),
-            Some(buf) => self.0 = Some(Node::concat(buf, n))
+            Some(buf) => self.0 = Some(Node::concat(buf, n)),
+        }
+    }
+
+    pub fn push_leaves(&mut self, leaves: Vec<N::L>) {
```
raphlinus: Being a pub fn, this should have at least a short docstring. I know that's not upheld in the existing code, but trying to raise the standard...

rust/rope/src/tree.rs
```diff
@@ -478,7 +471,29 @@ impl<N: NodeInfo> TreeBuilder<N> {
...
+        let mut stack: Vec<Vec<Node<N>>> = Vec::new();
+        for leaf in leaves {
```
raphlinus: I found this a bit hard to follow (and I probably wrote the original logic). What I think would be most helpful is a description of invariants. I _believe_ the following is true, and could be adapted into comments:

The stack is a list of list of nodes. Each list of nodes is of uniform height, and of length at least one and at most MAX_CHILDREN - 1. The heights are strictly decreasing.

Having gone through it, this looks right.

rust/rope/benches/rope.rs
```diff
@@ -0,0 +1,88 @@
+// Copyright 2017 Google Inc. All rights reserved.
```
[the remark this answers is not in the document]
jansol: Since this is being changed for the rest of the project already in #778 I think this should also be updated to the new line.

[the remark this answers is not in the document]
mqzry: Will change it to the new line as soon as I get to it

rust/rope/src/tree.rs
mqzry: Those invariants are correct. I added a description of those.

I was/am doubting between writing just a short doc string documenting the invariants and a full blown explanation why these invariants lead to amortized linear time.

mqzry:
Ready for next review or merge

rust/rope/src/tree.rs
jansol: More detailed documentation almost never hurts. If you're not sure I'd say write both, much like you would write a long git commit message, i.e. a short oneliner, an empty line and a long description.

jansol: Aaaand I see that's exactly what you did. These github change views can be confusing at times.

rust/rope/src/rope.rs
```diff
@@ -677,6 +680,26 @@ impl TreeBuilder<RopeInfo> {
             s = &s[splitpoint..];
         }
     }
+
+    pub fn push_str_stacked(&mut self, s: &str) {
```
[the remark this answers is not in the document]
mqzry: This is only used for testing and benching. As far as I can tell, #[cfg(test)] also works for `cargo bench` so I've added it.

rust/rope/src/rope.rs
```diff
@@ -822,7 +845,13 @@ impl<'a> Iterator for LinesRaw<'a> {
             if self.fragment.is_empty() {
                 match self.inner.next() {
                     Some(chunk) => self.fragment = chunk,
-                    None => return if result.is_empty() { None } else { Some(result) }
+                    None => {
```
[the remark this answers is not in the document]
mqzry: Sorry for that. Will rework the commit next time, if that is the case. Would love to see rustfmt incorporated indeed.

mqzry:
@raphlinus, this needs to be approved again. Incorporated the small nits of cmyr.

cmyr:
Awesome, thanks for the comments and sorry for the delay. For what it's worth I've been spending a lot more time in the rope crate recently, I should be better able to review changes going forward.

cmyr:
@mqzry sounds good!

PR #766, opened 2018-07-27
Add Edit type (alternative delta representation)

cmyr:
This is tentative work, motivated by wanting to improve plugin
ergonomics. The idea is that a delta can be cheaply converted into
a sequence of edits, which more intuitively represent the changes to a
document.

Problems with this, as currently written:

- the implementation is a bit gross, although this is fixable.
- There's a lot of overlap with the 'iter_inserts' and 'iter_deletions'
methods.
- currently the 'Edit' struct only makes sense in the coordinate space
of the pre-edit document, although in the plugin-lib we only notify the
plugin _after_ applying the delta, so a sequence of edits is not going
to be super meaningful. This is an interesting design problem. It would
not be difficult to include a 'delta_position' value with an edit that
could be used to determine the start/end in the coordinate space of the
post-edit document.

@betterclever this `Edit` type is quite similar to how the language servers represent edits, and shows how you can convert fairly easily from one to the other. Perhaps it's one way to make working with the language servers easier?

In any case, putting this up for consideration, I'm definitely not sure about it though...

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -579,7 +579,7 @@ impl<N: NodeInfo> Builder<N> {
     pub fn delete(&mut self, interval: Interval) {
         // TODO: doesn't handle interval types other than closed_open
         let (start, end) = interval.start_end();
-        assert!(start >= self.last_offset, "Delta builder: intervals not properly sorted");
+        assert!(start >= self.last_offset, "Delta builder: intervals not properly sorted ({} >= {})", start, self.last_offset);
         if start > self.last_offset {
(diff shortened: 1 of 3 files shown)


jansol:
@betterclever thoughts on this?

cmyr:
I'm also unsure if this is an improvement. I'm hoping to pick up work on auto-indent again shortly, which will be a good workout for the plugin edit system, and which will probably motivate me to improve that API (either by finishing this work or coming up with something else)

cmyr:
Going to close this for now, intend to revisit at some point though.

PR #777, opened 2018-08-08
Definition support in the Language Server Plugin

### LINKED ISSUE #719: Go-To Definition Support

betterclever:
Fixes #719 

This adds definition support using Definition RPC of Language Server Protocol. Currently, it was to be triggered by `RequestDefinition` RPC by Client which then gets back the response in `HandleDefinition` request. The client is supposed to check if it wants to open a new file or move caret to the given offset in the same file, it can be done by checking the `file_uri` field of Location.

The value sent to client is a `Location[]`. As discussed in the issue, following behaviour is ideally expected.
- If Location is empty, show a message like "Definition not found"
- If size is 1, go to the location. It could be in the same view or other, `new_view` with Location offset rpc work was done by @cmyr sometimes ago in a PR right? 
- If multiple choices, the client can be given a dialog to choose from. 

cc: @nangtrongvuon

### DIFF
--- a/rust/core-lib/src/client.rs
+++ b/rust/core-lib/src/client.rs
@@ -22,7 +22,7 @@ use xi_rpc::{self, RpcPeer};
 use tabs::ViewId;
 use config::Table;
 use styles::ThemeSettings;
-use plugins::rpc::ClientPluginInfo;
+use plugins::rpc::{ClientPluginInfo, Location};
 use plugins::Command;
 
 /// An interface to the frontend.
@@ -180,6 +180,16 @@ impl Client {
 
     pub fn show_hover(&self, view_id: ViewId, request_id: usize, result: String) {
         self.0.send_rpc_notification("show_hover", &json!(
+            {
+                "view_id": view_id,
+                "request_id": request_id,
+                "result": result
+            }
+        ))
+    }
+
+    pub fn handle_definition(&self, view_id: ViewId, request_id: usize, result: Vec<Location>) {
+        self.0.send_rpc_notification("handle_definition", &json!(
             {
                 "view_id": view_id,
                 "request_id": request_id,
--- a/rust/lsp-lib/src/conversion_utils.rs
+++ b/rust/lsp-lib/src/conversion_utils.rs
@@ -15,15 +15,14 @@
 //! Utility functions meant for converting types from LSP to Core format
 //! and vice-versa
 
-use url::Url;
-use xi_core::ViewId;
 use lsp_types::*;
 use std::fs::File;
 use std::io::Read;
 use types::{Definition, LanguageResponseError};
+use xi_core::ViewId;
 use xi_plugin_lib::{
-    Cache, Error as PluginLibError, Hover as CoreHover, Location as CoreLocation,
-    PluginContext, Range as CoreRange, View,
+    Cache, Error as PluginLibError, Hover as CoreHover, Location as CoreLocation, PluginContext,
+    Range as CoreRange, View,
 };
 use xi_rope::rope::{BaseMetric, LinesMetric, Utf16CodeUnitsMetric};
 use xi_rope::Rope;
@@ -118,7 +117,9 @@ pub(crate) fn core_hover_from_hover<C: Cache>(
     view_id: ViewId,
     hover: Hover,
 ) -> Result<CoreHover, LanguageResponseError> {
-    let view = plugin_context.get_view(&view_id).ok_or(PluginLibError::Other("View not found".to_owned()))?;
+    let view = plugin_context
+        .get_view(&view_id)
+        .ok_or(PluginLibError::Other("View not found".to_owned()))?;
     Ok(CoreHover {
         content: markdown_from_hover_contents(hover.contents)?,
         range: match hover.range {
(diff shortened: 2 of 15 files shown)


rust/plugin-lib/src/dispatch.rs
```diff
@@ -212,8 +260,8 @@ impl<'a, P: Plugin> RpcHandler for Dispatcher<'a, P> {
         let _t = trace_block_payload("Dispatcher::idle", &["plugin"],
                                      format!("token: {}", token));
         let view_id: ViewId = token.into();
-        let v = bail!(self.views.get_mut(&view_id), "idle", self.pid, view_id);
-        self.plugin.idle(v);
+        //let v = bail!(self.plugin_context.get_view(&view_id), "idle", self.pid, view_id);
```
jansol: No need to keep removed code around. We have version control for that.

rust/lsp-lib/src/lsp_plugin.rs
```diff
@@ -14,23 +14,34 @@
...
+/// Convenience for unwrapping a view, when handling RPC notifications.
+macro_rules! bail {
+    ($opt:expr, $view:expr) => ( match $opt {
+        Some(t) => t,
+        None => {
+            eprintln!("Path undefined for View {:?}", $view);
+            return
+        }
+    })
+}
```
jansol: Bail does not seem like an intuitive name for this but

> nit picking

so maybe it's just me

rust/plugin-lib/src/dispatch.rs
```diff
@@ -52,15 +53,55 @@ macro_rules! bail_err {
 /// to the plugin,
 pub struct Dispatcher<'a, P: 'a + Plugin> {
     //TODO: when we add multi-view, this should be an Arc+Mutex/Rc+RefCell
-    views: HashMap<ViewId, View<P::Cache>>,
+    //views: HashMap<ViewId, View<P::Cache>>,
```
jansol: Please don't leave in commented-out code unless you have a good reason to. If you meant to keep it due to the comment, might as well update the type signature to reflect the one named in the comment.

jansol:
Added a couple of notes. Feel free to change my mind if you disagree with them.

Superficially the PR looks okay but I don't know the details of LSP stuff.

rust/lsp-lib/src/utils.rs
```diff
@@ -205,4 +205,11 @@ pub fn start_new_server(
     }
 
     Ok(language_server_client)
+}
+
+/// Helper Method to get Url of the View path
+/// Panics: If view does not have a path or the given path
+/// can not be converted to Url
```
cmyr: This comment doesn't seem to be accurate? As written this function doesn't panic.

rust/plugin-lib/src/dispatch.rs
```diff
@@ -51,16 +52,55 @@ macro_rules! bail_err {
 /// Handles raw RPCs from core, updating state and forwarding calls
 /// to the plugin,
 pub struct Dispatcher<'a, P: 'a + Plugin> {
-    //TODO: when we add multi-view, this should be an Arc+Mutex/Rc+RefCell
-    views: HashMap<ViewId, View<P::Cache>>,
+    plugin_context: PluginContext<<P>::Cache>,
```
cmyr: In general, when I think of a 'context' as we've used it in this project, I tend to think of it as a _temporary collection of resources that are needed for some particular operation_.

So with that in mind, I would probably not have the plugin context owned by the dispatcher, like this, but rather have it created as needed, borrowing contents from the dispatcher:

```rust
/// Provides access to all active views.
struct PluginContext<'a, C: Cache> {
    views: &'a mut HashMap<ViewId, View<C>>,
}
```

Then it's possible that you would run into borrow check issues creating this, in which case you could destructure self (something like this, I haven't tried to compile):

```rust
fn idle(&mut self, _ctx: &RpcCtx, token: usize) {
        let view_id: ViewId = token.into();
        let Dispatcher { mut plugin, mut views, .. } = self;
        let ctx = PluginContext { views: &mut views };
        self.plugin.idle(view_id, &mut ctx);
     }
```

And this highlights another issue here, which is using the `view_id` as the idle token.

The purpose of this is so that the idle handler can know which view to call idle on in the future. This is nice for plugins that only care about a view at a time. With this change, it isn't necessary; the idle token could be anything.

One option, however, would be to keep the existing behaviour while adding a special 'global' idle token, say `usize::max_size()`; in this case we could keep the current behaviour if the idle token is a view id, but if it was this new token we could call a new plugin fn, `idle_global` (or something) which would be passed this new `PluginContext` struct.

This is just an idea, and there are tradeoffs, but it does have the advantage of keeping the current API pretty much entirely the same, which is nice for plugins that prefer it. 🤷‍♂️

rust/lsp-lib/src/conversion_utils.rs
```diff
@@ -116,3 +126,47 @@ pub(crate) fn core_hover_from_hover<C: Cache>(
...
+    let utf16_offset = line_utf16_offset + (position.character as usize);
+    rope.convert_metrics::<Utf16CodeUnitsMetric, BaseMetric>(utf16_offset)
+}
+
+pub(crate) fn core_location_from_location<C: Cache>(plugin_context: &mut PluginContext<C>, location: &Location) -> CoreLocation {
+    let path = location.uri.to_file_path().unwrap();
+    let view = plugin_context.get_view_for_path(&path);
+    
+    let contents = match view {
+        Some(view) => view.get_document().unwrap(),
```
cmyr: It seems unfortunate to always be doing this work when we have the view open; we should be able to just get the line we need and then do the utf16 count on it (and we could still convert to a rope in that case, which is convenient). I don't think this is _super_ important, though.

rust/lsp-lib/src/conversion_utils.rs
```diff
@@ -116,3 +126,47 @@ pub(crate) fn core_hover_from_hover<C: Cache>(
...
+        None => {
+            let mut f = File::open(&path).expect("can't find file");
```
cmyr: Is crashing the best option in this case?

rust/lsp-lib/src/conversion_utils.rs
```diff
@@ -116,3 +126,47 @@ pub(crate) fn core_hover_from_hover<C: Cache>(
...
+
+            let mut contents = String::new();
+            f.read_to_string(&mut contents).expect("can't read file");
```
cmyr: ditto about crashing here, I'm not sure it's necessary.

rust/lsp-lib/src/utils.rs
```diff
@@ -205,4 +205,11 @@ pub fn start_new_server(
...
+pub fn get_view_uri<C: Cache>(view: &View<C>) -> Option<Url> {
```
cmyr: If this is the only place that you're using the `bail` macro, and you want to have some fun, you could consider just writing a `uri_for_view!` macro?

nangtrongvuon:
Well, time to dust off that definition implementation! Overall the support for this in the front end is already somewhat there, as already mentioned. Just need a little bit of tweaking to bring it to parity with this.

One other thing that was mentioned is now client needs a way to open a view as requested by core. @cmyr Maybe we should look into that new view stuff again as well?

cmyr:
@nangtrongvuon I think we're going to go the simpler route; frontend will get a list of locations (often just one) in response to a goto request, and then frontend will have to manually open the new file.

nangtrongvuon:
Sounds good. I'll hack on it and have something ready soon.

rust/core-lib/src/client.rs
```diff
@@ -188,6 +188,16 @@ impl Client {
         ))
     }
 
+    pub fn handle_definition(&self, view_id: ViewId, request_id: usize, result: Vec<Location>) {
+        self.0.send_rpc_notification("show_hover", &json!(
```
nangtrongvuon: This is meant to be "show_definition" right?

betterclever: Thanks for pointing out. 👍  Fixed.

rust/lsp-lib/src/lsp_plugin.rs
jansol: My main issue is that this also does a match and only conditionally bails out.

jansol:
This needs rebasing on top of current master.

cmyr:
I let this get stale, and I expect the plugin architecture to change quite a bit again soon, so I'm going to close this for now.

PR #781, opened 2018-08-12
Fix crash on file reload (fix #780)

### LINKED ISSUE #780: Crash when reloading document if size has decreased, with selection at EOF
If a file is open in the editor, and has no unsaved changes, and the file is modified by another process in such a way that its size is decreased, we will panic when we try to reload the file.

- open a file in the editor.
- move the cursor to the end of the file.
- in another editor, delete the last line

We _do_ try to defend against this, but there's a small oversight in our implementation. (digging in now, will probably have a fix shortly)

cmyr:


jansol:
Dropping selections when the file is reloaded is quite expected I think (while it certainly would be cool to be able to retain them there might be some cases that would resolve in unexpected ways - I really should read through the rope science series to get a grip on this edit stuff).

The change itself LGTM so I'd merge this as-is and open a separate issue for retaining selections on a reload.

PR #783, opened 2018-08-16
Don't crash if logging setup fails

cmyr:
This isn't the ideal, solution, which would involve a) letting the client specify the location for the log file and b) still logging to stderr if a log file can't be written; however this will at least fix https://github.com/google/xi-mac/issues/248.


cmyr:
@betterclever yes, I do think we want the client to be able to specify if and where logs should be written, I'm happy for that to be separate work though.

PR #784, opened 2018-08-18
Make themes dir when making config dir

cmyr:
Users who already have a config directory will need to manually create the 'themes' directory, but this will create it for new installs.


virattara:
[This](https://github.com/google/xi-editor/blob/master/rust/core-lib/src/config.rs#L499) will make it?

cmyr:
Ah I'd missed that, thanks @virattara!

PR #785, opened 2018-08-21
Selection granularity for drag gestures

ryanbloom:
Like most text editors, Xi lets you select a word by double-clicking or a line by triple-clicking. However, if you then drag the cursor, it will extend the selection one character at a time. This PR assigns the appropriate granularity to drag gestures, depending on whether they started with a single, double, or triple click.

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -438,9 +438,9 @@ impl View {
             sel.add_region(SelRegion::new(last.start, offset));
             sel
         };
-        let min = sel.last().unwrap().start;
+        let range_start = sel.last().unwrap().start;
         self.set_selection(text, sel);
-        self.start_drag(offset, min, offset, SelectionGranularity::Point, false);
+        self.start_drag(range_start, range_start, range_start, SelectionGranularity::Point, false);
 }
 
     /// Selects the given region and supports multi selection.
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -264,7 +264,7 @@ impl View {
         match ty {
             GestureType::PointSelect => {
                 self.set_selection(text, SelRegion::caret(offset));
-                self.start_drag(offset, offset, offset, SelectionGranularity::Point);
+                self.start_drag(offset, offset, offset, SelectionGranularity::Point, false);
             },
             GestureType::RangeSelect => self.select_range(text, offset),
             GestureType::ToggleSel => self.toggle_sel(text, offset),
@@ -341,16 +341,10 @@ impl View {
                 return;
             }
         }
-        self.drag_state = Some(DragState {
-            base_sel: selection.clone(),
-            offset,
-            min: offset,
-            max: offset,
-            granularity: SelectionGranularity::Point,
-        });
         let region = SelRegion::caret(offset);
         selection.add_region(region);
         self.set_selection_raw(text, selection);
+        self.start_drag(offset, offset, offset, SelectionGranularity::Point, true)
     }
 
     /// Move the selection by the given movement. Return value is the offset of
@@ -446,7 +440,7 @@ impl View {
         };
         let min = sel.last().unwrap().start;
         self.set_selection(text, sel);
-        self.start_drag(offset, min, offset, SelectionGranularity::Point);
+        self.start_drag(offset, min, offset, SelectionGranularity::Point, false);
 }
 
     /// Selects the given region and supports multi selection.
@@ -468,7 +462,7 @@ impl View {
         };
 
         self.select_region(text, SelRegion::new(start, end), multi_select);
-        self.start_drag(offset, start, end, SelectionGranularity::Word);
+        self.start_drag(offset, start, end, SelectionGranularity::Word, multi_select);
     }
 
     /// Selects an entire line and supports multi selection.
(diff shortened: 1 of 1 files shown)


cmyr:
I'm a bit behind on xi stuff but I'll have some time to catch up next week. @ryanbloom very happy to see you digging into the Rust side of stuff! 

One thought I have from just skimming: If we're going to introduce a granularity enum, should we make sure that we're using it consistently? It seems like there's some duplication between this new enum and the whole 'gesture type' concept. Can they be unified somehow? @scholtzan you might have a good sense of this..

scholtzan:
Okay, that sounds reasonable.

Regarding unifying the gestures, it might be a good idea to have a general `Select` gesture or at least I don't see a reason against it. To prevent compatibility issues we could for now have both the new `Select` gesture type and leave the old gestures (maybe mark them as deprecated and then later remove them). Perhaps, we could even combine `Select` and `MultiSelect` by setting a flag for multiple lines.

ryanbloom:
Combining all the select gestures sounds good to me. If we go that route, there are only two options left in the `GestureType` enum—are you thinking `Select` and `RangeSelect` would become top-level `ViewEvent`s? The matching code might be cleaner that way.

scholtzan:
There is also the `ToggleSel` gesture, unless that can also be combined into the other gesture. In any case, I think it is better to keep them as `GestureType`s and consider them as gestures. It is quite likely that there will be more gestures in the future.

ryanbloom:
Okay, here's a tentative version of the unified-select-gestures refactor.

I decided to deprecate `ToggleSel` in favor of `Select {granularity: Point, multi: true}`, since from the frontend's perspective it's a single click with the multi-select modifier. The behavior isn't *quite* the same as the other multi-select gestures, but it's close enough in my opinion. If it were a separate gesture, there wouldn't be a defined behavior for `{granularity: Point, multi: true}`.

@cmyr If this looks like the right direction to you, I'll go ahead and update the protocol docs and the xi-mac implementation.

cmyr:
@ryanbloom Ok, I think that this approach is solid.

- I would add a log warning if people call the old API, pointing them to the new API, and warning that the old one will go away at some point.

- A piece of future work I've been imagining for a while (but which may not have an issue?) is replacing the various distinct movement rpcs with a single 'movement' rpc that takes a `MovementUnit` and a `modify_sel` argument. In this world, the `RangeSelect` feels awfully close to sending this movement RPC, with a `MovementUnit::Exact(usize)` argument. The even larger question then becomes, why do we treat keyboard and mouse input so differently? is there some generalization of the idea of changing cursor/selection state that would be helpful? (There might not be, i'm just thinking out loud)

ryanbloom:
You're right, both `Select` and `RangeSelect` sure do have a lot in common with `Move`. If we merged one of them in, you could get the other just by toggling `modify_sel`. Add in all the modifier options and we'd have something like this:

```rust
enum Movement {
	Up, Down, ...,
	Exact(usize, SelectionGranularity),
}
enum SelectionChange {
	Replace,   // click or arrow keys
	Add,       // command+click
	Range,     // shift+click or shift+arrows
	Rectangle, // option+shift+click
}
struct Select {
	movement: Movement,
	change: SelectionChange,
}

// RangeSelect becomes
Select {
	movement: Movement::Exact(123, SelectionGranularity::Point),
	change: SelectionChange::Range
}

// MoveWordLeft becomes
Select {
	movement: Movement::LeftWord,
	change: SelectionChange::Replace
}

// etc.
```
I could imagine two reasons we might go with this:
* We find a way to unify the code for handling these gestures
* It somehow makes it easier for frontends to use different input methods (touchscreens, Vim, etc.)?

On the other hand, this protocol would be a little less straightforward than what we have now: it introduces an extra level of nesting, and `Drag` is still a separate operation somewhere. So unless either of those benefits apply, I think it's fine to keep `Select` and `RangeSelect` separate from `Move`.

Anyway, we're getting into separate-PR territory here. It makes sense to at least unify the `Move` RPCs, so I'll do that on this branch.

cmyr:
@raphlinus thoughts on this? It feels like @ryanbloom is pretty close to coming up with a nice underlying abstraction for the general idea of 'manipulating selections from a frontend', but it isn't an entirely free lunch...

nangtrongvuon:
I think this is super cool! I'm with @cmyr on this one though, I do think the end goal should be to unify mouse and keyboard input into a single set of gestures. I really don't see the merit to having it be treated separately, since one of Xi's advertised features is that it works with any frontend, and many people will probably prefer a keyboard-oriented frontend (like vi/emacs and terminal setups!)

I might be off on this one, but I think the granularity themselves could be replaced with the movements? So instead of a "Word" selection granularity, we could try and see what direction the drag is in, and do a selection based on MoveWordLeft/Right. 

I don't see anything particularly wrong with the code, so it's fine by me!

ryanbloom:
It sounds like unifying the gestures is the favored approach. Before we go too far down the rabbit hole of figuring out how, I opened a separate issue (#826) for discussion.

I moved the protocol changes off of this branch, so it should be ready to merge.

cmyr:
Cool! the discussion issue looks good too, will take a look.

Haven't really gone through this yet, but one thing we will want is to add docs for the new API.

ryanbloom:
@cmyr Is there anything in particular you think needs documenting? I decided to hold off on adding client APIs here until we figure out how the protocol should work.

cmyr:
I would add short doc comments for `SelectionGranularity` and its members. you can add the `needs review` tag if you think this is good to go. :)

Issue #713, opened 2018-06-17
panicked at 'attempt to subtract with overflow' when I type space followed by backspace


bvinc:
To reproduce simply open up a new view, type a space, then a backspace.

Linux, Fedora 28
rustc 1.26.2 (594fb253c 2018-06-01)
rustup
gxi
git revision befef5cf

```
thread '<unnamed>' panicked at 'attempt to subtract with overflow', /home/brain/.cargo/git/checkouts/xi-editor-968440e5b4201956/befef5c/rust/core-lib/src/editor.rs:358:21
stack backtrace:
   0:     0x55e4e460cb4b - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::h30cc8d009960e7d3
                               at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1:     0x55e4e46131e0 - std::sys_common::backtrace::print::h0a61d49a4e79b1ca
                               at libstd/sys_common/backtrace.rs:71
                               at libstd/sys_common/backtrace.rs:59
   2:     0x55e4e46075ad - std::panicking::default_hook::{{closure}}::h6d4e5c7d0d288208
                               at libstd/panicking.rs:207
   3:     0x55e4e460730b - std::panicking::default_hook::h908bd919f2aa6c02
                               at libstd/panicking.rs:223
   4:     0x55e4e46079ff - std::panicking::rust_panic_with_hook::habacef8c65a3b0d4
                               at libstd/panicking.rs:402
   5:     0x55e4e4607822 - std::panicking::begin_panic_fmt::h20e14b91a98870b3
                               at libstd/panicking.rs:349
   6:     0x55e4e4607752 - rust_begin_unwind
                               at libstd/panicking.rs:325
   7:     0x55e4e4656b70 - core::panicking::panic_fmt::h8a30aa26a6a035a8
                               at libcore/panicking.rs:72
   8:     0x55e4e4656a96 - core::panicking::panic::hebbf35921bdc7e33
                               at libcore/panicking.rs:51
   9:     0x55e4e4062b7a - xi_core_lib::internal::editor::Editor::delete_backward::h4231060fbcb63561
                               at /home/brain/.cargo/git/checkouts/xi-editor-968440e5b4201956/befef5c/rust/core-lib/src/editor.rs:358
  10:     0x55e4e40662b5 - xi_core_lib::internal::editor::Editor::do_edit::h3bc39f7f43d4e6f2
[...]

PR #788, opened 2018-09-02
Backspace now deletes up to tab_size spaces

### LINKED ISSUE #716: backspace with translate_tabs_to_spaces should delete up to tab_width preceding spaces
This was previously implemented (in #460) but broke somewhere along the way. That PR contains a good discussion of the motivation and ultimate design. Care should be taken not to regress on #714/#713. I'd like this to also include a test case in `event_context.rs`; doing this correctly might involve modifying the `ContextHarness` used for testing to allow it to be passed a custom config, which wasn't possible when that was first written.

As a bonus it would be great to also have a regression test for #713.

nangtrongvuon:
This is an attempt to fix #716. 

Backspace should now have correct behavior, which is deleting up to config.tab_size at a time for each backspace press. Care was also taken not to regress #713/#714.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -345,7 +345,7 @@ impl Editor {
                 // backspace deletes max(1, tab_size) contiguous spaces
                 let (_, c) = view.offset_to_line_col(&self.text, region.start);
 
-                let tab_off = c & config.tab_size;
+                let tab_off = c % config.tab_size;
                 let tab_size = config.tab_size;
                 let tab_size = if tab_off == 0 { tab_size } else { tab_off };
                 let tab_start = region.start.saturating_sub(tab_size);


rust/core-lib/src/editor.rs
```diff
@@ -345,22 +345,31 @@ impl Editor {
                 // backspace deletes max(1, tab_size) contiguous spaces
                 let (_, c) = view.offset_to_line_col(&self.text, region.start);
 
-                let tab_off = c & config.tab_size;
+                // Offset of cursor position from nearest tab
+                let tab_off = c % config.tab_size;
```
cmyr: I think this change alone is the fix?? I have no idea how this was a `&` before, I think it must've been a typo. Changing this to a `%` should enough?

rust/core-lib/src/editor.rs
```diff
@@ -345,22 +345,31 @@ impl Editor {
...
                 let tab_size = config.tab_size;
-                let tab_size = if tab_off == 0 { tab_size } else { tab_off };
+
+                // Size of tab, counted from cursor to nearest tab
+                let tab_size = if tab_off == 0 { if c > tab_off { tab_size } else { c - tab_off } } else { tab_off };
                 let tab_start = region.start.saturating_sub(tab_size);
-                let preceded_by_spaces = region.start > 0 &&
-                    (tab_start..region.start).all(|i| self.text.byte_at(i) == b' ');
+                let preceded_by_spaces = region.start > 0 && tab_size > 0
+                && self.text.slice_to_string(tab_start, region.start).chars().all(|i| i.is_whitespace());
```
cmyr: I'm not sure about this change; among other things it now means that if tab_size is 4 this will delete three tabs and a space on backspace. In addition it requires us to allocate a string, instead of doing the comparison in the existing buffer.

nangtrongvuon:
You're right, I definitely came to the same original algorithm, which is checking where the tab starts. The "&" sign definitely threw me off.

I reduced the fix to just that, and I think it works as expected now!

PR #792, opened 2018-09-07
Add support for duplicating lines

scholtzan:
Duplicates lines that are selected

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -24,6 +24,7 @@ use xi_rope::delta::{self, Delta, Transformer};
 use xi_rope::engine::{Engine, RevId, RevToken};
 use xi_rope::spans::SpansBuilder;
 use xi_trace::trace_block;
+use xi_rope::tree::Cursor;
 
 use config::BufferItems;
 use event_context::MAX_SIZE_LIMIT;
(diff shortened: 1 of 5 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -687,6 +688,38 @@ impl Editor {
         }
     }
 
+    fn duplicate_line(&mut self, view: &View, config: &BufferItems) {
+        let mut builder = delta::Builder::new(self.text.len());
+        // get affected lines
+        let mut lines = BTreeSet::new();
```
scholtzan: This part is almost the same as in `modify_indent`. Perhaps there is a way to reuse the code...

cmyr: that could make sense at some point, feels like this is okay for now though.

scholtzan:
I added it under "Other movement and deletion commands". Not sure if that is the right section though.

nangtrongvuon:
I noticed Travis has been building this branch for a while. Maybe something got stuck?

cmyr:
hmm, good point. I checked and it said it passed, just this graphic hasn't updated?

eyelash:
After duplicating multiple lines the selection is somewhat unexpected:

before | after
:-:|:-:
![before][before] | ![after][after]

This makes it hard to run `duplicate_line` multiple times.

[before]: https://user-images.githubusercontent.com/2668288/45583828-44fa6d00-b8c9-11e8-8c53-d10822463256.png
[after]: https://user-images.githubusercontent.com/2668288/45583831-4e83d500-b8c9-11e8-871f-44a9b76a858b.png

cmyr:
hmm good point, I think I'd call that a bug. @scholtzan?

scholtzan:
Oh, I will look into it.

PR #798, opened 2018-09-15
Fix warning: function is never used: `split_as_leaves`

yvan-sraka:
```
   --> rope/src/rope.rs:699:1
    |
699 | fn split_as_leaves(mut s: &str) -> Vec<String> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(dead_code)] on by default
```

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -696,6 +696,7 @@ impl TreeBuilder<RopeInfo> {
     }
 }
 
+#[cfg(test)]
 fn split_as_leaves(mut s: &str) -> Vec<String> {
     let mut nodes = Vec::new();
     while !s.is_empty() {


cmyr:
Thanks for this, was just noticing that it in the console.

This fn _is_ used in tests, so we can't remove it; however if we annotate it with `#[cfg(test)]` it won't be compiled during a normal build, and we won't get this log message. I would be happy to merge that change!

cmyr:
Sorry, I didn't notice that this got updated; and meanwhile, another change (#800) has resolved this warning, so I'm going to close this. Thanks for the PR though, always good to do this little cleanup stuff.

If you'd like to take another shot, there's another warning right now originating in the syntect plugin..

PR #799, opened 2018-09-15
Make iterators respect codepoint boundaries

cmyr:
Previously, it was possible to specify ranges that were not codepoint
boundaries, which would cause crashes. This was also the case for
methods that use these methods under the hood, such as
Rope::slice_to_string.


cmyr:
After discussion am going to rethink this approach.

PR #801, opened 2018-09-15
Add comparison utilities to rope crate

cmyr:
This adds compare.rs, which includes some helpers for quickly comparing
regions of ropes.

These helpers will be used in a follow-on patch.

### DIFF
--- a/rust/rope/benches/diff.rs
+++ b/rust/rope/benches/diff.rs
@@ -1,0 +1,81 @@
+#![feature(test)]
+
+extern crate test;
+extern crate xi_rope;
+
+use test::Bencher;
+use xi_rope::compare;
+
+static EDITOR_STR: &str = include_str!("../../core-lib/src/editor.rs");
+static VIEW_STR: &str = include_str!("../../core-lib/src/view.rs");
+
+static INTERVAL_STR: &str = include_str!("../src/interval.rs");
+static BREAKS_STR: &str = include_str!("../src/breaks.rs");
+
+#[bench]
+fn ne_idx_sw(b: &mut Bencher) {
+    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].iter()
+        .fold(String::new(), |mut acc, s| { acc.push_str(s); acc });
+    let mut two = one.clone();
+    unsafe {
+        let b = two.as_bytes_mut();
+        let idx = b.len() - 200;
+        b[idx] = 0x02;
+    }
+
+    b.iter(|| {
+        compare::ne_idx_fallback(one.as_bytes(), one.as_bytes());
+        compare::ne_idx_fallback(one.as_bytes(), two.as_bytes());
+    })
+}
+
+#[bench]
+fn ne_idx_hw(b: &mut Bencher) {
+    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].iter()
+        .fold(String::new(), |mut acc, s| { acc.push_str(s); acc });
+    let mut two = one.clone();
+    unsafe {
+        let b = two.as_bytes_mut();
+        let idx = b.len() - 200;
+        b[idx] = 0x02;
+    }
+
+    b.iter(|| {
+        compare::ne_idx(one.as_bytes(), one.as_bytes());
+        compare::ne_idx(one.as_bytes(), two.as_bytes());
+    })
+}
+
+#[bench]
+fn ne_idx_sw_rev(b: &mut Bencher) {
+    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].iter()
+        .fold(String::new(), |mut acc, s| { acc.push_str(s); acc });
+    let mut two = one.clone();
+    unsafe {
+        let b = two.as_bytes_mut();
+        let idx = 200;
+        b[idx] = 0x02;
+    }
+
+    b.iter(|| {
+        compare::ne_idx_rev_fallback(one.as_bytes(), one.as_bytes());
+        compare::ne_idx_rev_fallback(one.as_bytes(), two.as_bytes());
+    })
+}
+
+#[bench]
+fn ne_idx_hw_rev(b: &mut Bencher) {
+    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].iter()
+        .fold(String::new(), |mut acc, s| { acc.push_str(s); acc });
+    let mut two = one.clone();
+    unsafe {
+        let b = two.as_bytes_mut();
+        let idx = 200;
+        b[idx] = 0x02;
+    }
+
+    b.iter(|| {
+        compare::ne_idx_rev(one.as_bytes(), one.as_bytes());
+        compare::ne_idx_rev(one.as_bytes(), two.as_bytes());
+    })
+}
--- a/rust/rope/benches/diff.rs
+++ b/rust/rope/benches/diff.rs
@@ -14,8 +14,7 @@ static BREAKS_STR: &str = include_str!("../src/breaks.rs");
 
 #[bench]
 fn ne_idx_sw(b: &mut Bencher) {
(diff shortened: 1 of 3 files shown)


rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+    let two = &two[two.len() - min_len..];
+    debug_assert_eq!(one.len(), two.len());
+    let mut idx = min_len;
+    loop {
+        let mask: i32;
+        if idx < SSE_STRIDE {
+            let mut one_buf: [u8; 16] = [0; 16];
+            let mut two_buf: [u8; 16] = [0; 16];
+            let mut temp_idx = SSE_STRIDE - idx;
+            for i in 0..idx {
```
raphlinus: I think this might be better expressed something like `one_buf[temp_idx .. SSE_STRIDE] = one[..idx]` (not sure about exact syntax).

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+pub fn ne_idx_rev(one: &[u8], two: &[u8]) -> Option<usize> {
+    let min_len = one.len().min(two.len());
+    let one = &one[one.len() - min_len..];
+    let two = &two[two.len() - min_len..];
+    debug_assert_eq!(one.len(), two.len());
+    let mut idx = min_len;
+    loop {
+        let mask: i32;
+        if idx < SSE_STRIDE {
+            let mut one_buf: [u8; 16] = [0; 16];
```
raphlinus: Maybe use SSE_STRIDE instead of hardcoded 16? Seems inconsistent.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub fn ne_idx(one: &[u8], two: &[u8]) -> Option<usize> {
+    ne_idx_fallback(one, two)
+}
+
+#[inline]
+#[allow(dead_code)]
+#[doc(hidden)]
+pub fn ne_idx_fallback(one: &[u8], two: &[u8]) -> Option<usize> {
+    for i in 0..one.len().min(two.len()) {
```
raphlinus: This can be written something like `one.iter().zip(two.iter()).position(|a, b| a != b)`.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+
+#[cfg(not(any(target_arch = "x86", target_arch = "x86_64")))]
+pub fn ne_idx_rev(one: &[u8], two: &[u8]) -> Option<usize> {
+    ne_idx_rev_fallback(one, two)
+}
+
+#[inline]
+#[allow(dead_code)]
+#[doc(hidden)]
+pub fn ne_idx_rev_fallback(one: &[u8], two: &[u8]) -> Option<usize> {
```
raphlinus: Similarly with rposition.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+#[inline(always)]
+#[doc(hidden)]
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub fn fast_cmpestr_mask(one: &[u8], two: &[u8]) -> i32 {
+    debug_assert_eq!(one.len(), 16);
+    debug_assert_eq!(two.len(), 16);
+    unsafe {
+        let onev = _mm_loadu_si128(one.as_ptr() as *const _);
+        let twov = _mm_loadu_si128(two.as_ptr() as *const _);
+        let mask = _mm_cmpestrm(onev, 16, twov, 16, SSID_OPTS);
```
raphlinus: I think using _SIDD_BIT_MASK instead of _SIDD_UNIT_MASK would let you get rid of the following _mm_movemask_epi8.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+
+        debug_assert!(self.target.is_boundary::<BaseMetric>());
+        debug_assert!(self.base.is_boundary::<BaseMetric>());
+        debug_assert!(base_leaf.is_char_boundary(base_leaf_off));
+        debug_assert!(target_leaf.is_char_boundary(target_leaf_off));
+
+        self.base_chunk = &base_leaf[..base_leaf_off];
+        self.target_chunk = &target_leaf[..target_leaf_off];
+
+        loop {
```
raphlinus: Another way to write this kind of loop is to have a zip-like iterator over two ropes; that lets you separate out the chunk iteration logic (and reduces the amount of repeated code). But I think this is ok.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+/// ```
+/// use xi_rope::compare::fast_cmpestr_mask;
+/// let one = "aaaaaaaaaaaaaaaa";
+/// let two = "aa3aaaaa9aaaEaaa";
+/// let exp = "0001000100000100";
+/// let result = format!("{:016b}", fast_cmpestr_mask(one.as_bytes(), two.as_bytes()));
+/// assert_eq!(result.as_str(), exp);
+/// ```
+///
+#[inline(always)]
```
raphlinus: This should _probably_ be protected by a `#[target_feature(enable = "sse4.2")]`. See [memchr-rs](https://github.com/BurntSushi/rust-memchr/blob/master/src/x86/sse42.rs#L24) source for guidance.

This function is also unsafe, because it is UB if the CPU doesn't support SSE4.2.

rust/rope/src/compare.rs
```diff
@@ -0,0 +1,550 @@
...
+        let twov = _mm_loadu_si128(two.as_ptr() as *const _);
+        let mask = _mm_cmpestrm(onev, 16, twov, 16, SSID_OPTS);
+        _mm_movemask_epi8(mask)
+    }
+}
+
+/// Returns the lowest `i` for which `one[i] != two[i]`, if one exists.
+#[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+pub fn ne_idx(one: &[u8], two: &[u8]) -> Option<usize> {
+    let min_len = one.len().min(two.len());
```
raphlinus: This should do runtime SSE feature detection. The [standard](https://doc.rust-lang.org/std/macro.is_x86_feature_detected.html) way to do that is:

```
if is_x86_feature_detected!("sse4.2") {
    ...
} else {
    ne_idx_fallback(one, two)
}
```

rust/rope/src/compare.rs
cmyr: interestingly I had that originally and it performed significantly worse on benchmarks, which I was surprised by. It's possible I was doing something wrong..

rust/rope/benches/diff.rs
```diff
@@ -0,0 +1,81 @@
...
+
+static EDITOR_STR: &str = include_str!("../../core-lib/src/editor.rs");
+static VIEW_STR: &str = include_str!("../../core-lib/src/view.rs");
+
+static INTERVAL_STR: &str = include_str!("../src/interval.rs");
+static BREAKS_STR: &str = include_str!("../src/breaks.rs");
+
+#[bench]
+fn ne_idx_sw(b: &mut Bencher) {
+    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].iter()
```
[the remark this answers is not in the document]
cmyr: I agree in the longer term (if for instance we start adding performance regression testing to CI) we'll want to be more cautious about reproducibility; maybe at that point we'll want a separate crate that handles all benchmarks, and includes a bunch of test data in-tree.

rust/rope/src/compare.rs
cmyr: If this is the case, I couldn't figure it out. `_mm_cmpestrm` always returns a `__m128i`, and I need to extract some int type from that somehow. Open to suggestions!

raphlinus: Have you tried `_mm_cvtsi128_si32`?

cmyr: okay cool that worked 👍

PR #802, opened 2018-09-15
 Compute deltas between arbitrary ropes

cmyr:
This includes the commit from #801; that PR should be reviewed first.

This adds a preliminary algorithm for computing an efficient delta
between two ropes.

~This algorithm produces poor deltas in a few pathological cases, such as
when a document is compared against itself with its lines shuffled, or
when a pair of documents contains few line breaks and have numerous small
edits; however it has consistently quick running time.~

### DIFF
--- a/rust/rope/src/diff.rs
+++ b/rust/rope/src/diff.rs
@@ -19,7 +19,7 @@ use std::collections::HashMap;
 use compare::RopeScanner;
 use delta::{Delta, DeltaElement};
 use interval::Interval;
-use rope::{Rope, RopeInfo, RopeDelta};
+use rope::{LinesMetric, Rope, RopeInfo, RopeDelta};
 use tree::{Node, NodeInfo};
 
 use memchr::memchr;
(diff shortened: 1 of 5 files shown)


rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+        // for each matching region, we extend it forwards and backwards.
+        // we keep track of how far forward we extend it each time, to avoid
+        // having a subsequent scan extend backwards over the same region.
+        let mut prev_end = start_offset;
+
+        for (targ_off, base_off) in longest_subseq {
+            if targ_off <= prev_end { continue; }
+            let (left_dist, mut right_dist) = expand_match(base, target, base_off,
+                                                           targ_off, prev_end);
+            if targ_off + right_dist > target_end {
```
raphlinus: Can you write this:

    right_dist = right_dist.min(target_end - targ_off)

I _think_ that subexpression is guaranteed nonnegative (thus no integer panic), but haven't verified completely.

rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+            let base_start = base_off - left_dist;
+            let len = left_dist + right_dist;
+            prev_end = targ_start + len;
+
+            builder.copy(base_start, targ_start, len);
+        }
+
+        if diff_end > 0 {
+            builder.copy(base.len() - diff_end,
+            target.len() - diff_end, diff_end);
```
raphlinus: I think this line should be indented.

rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+        }
+        self.ops.push(DiffOp { target_idx: target, base_idx: base, len: len })
+    }
+
+    fn to_delta(self, base: &Rope, target: &Rope) -> RopeDelta {
+        let mut els = Vec::with_capacity(self.ops.len() * 2);
+        let mut targ_pos = 0;
+        for DiffOp { base_idx, target_idx, len } in self.ops {
+            if target_idx > targ_pos {
+                let iv = Interval::new_closed_open(targ_pos, target_idx);
```
raphlinus: This should probably be changed to `Interval::new` now, but will likely compile.

rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+                els.push(DeltaElement::Insert(target.subseq(iv)));
+            }
+            els.push(DeltaElement::Copy(base_idx, base_idx + len));
+            targ_pos = target_idx + len;
+        }
+
+        if targ_pos < target.len() {
+            let iv = Interval::new_closed_open(targ_pos, target.len());
```
raphlinus: ditto

rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+/// Fast iterator over lines in a string, not removing newline characters.
+struct LineScanner<'a> {
+    inner: &'a str,
+    idx: usize,
+}
+
+impl<'a> Iterator for LineScanner<'a> {
```
raphlinus: This doesn't return the last partial line, no? Maybe want to document that, if for no other reason than the fact that its behavior is different than lines_raw.

rust/rope/src/diff.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        if start_offset > 0 {
+            builder.copy(0, 0, start_offset);
+        }
+
+        // if our preliminary scan finds no differences we're done
+        if start_offset == base.len() && target.len() == base.len() {
+            return builder.to_delta(base, target);
+        }
+
```
raphlinus: As discussed in irc, here's my proposal for the way forward, which should be a follow-on issue.

Create a RopeSlice struct, which is a pair of (&Rope, Range). Custom implement Eq and Hash traits, to internally use iter_chunks to compute both. Don't use lines_raw iterator, use cursor with LinesMetric. I think this gives you maximum efficiency, as everything is borrowed and there's no allocation or copying, even in the case of very long lines.

I did verify that you can do DefaultHasher::write() with different slice boundaries and get the same state, though I don't believe that's written down explicitly.

rust/rope/src/diff.rs
cmyr: ah yea that's much cleaner 👍

rust/rope/src/diff.rs
cmyr: Nope that's a bug 👍

PR #804, opened 2018-09-16
Bugfix: Duplicate line will duplicate selected regions instead of lines

scholtzan:
This is a fix for duplicating lines to get the same behaviour as in Sublime: If <kbd>cmd</kbd><kbd>shift</kbd><kbd>D</kbd> is pressed, then all lines that have a caret get duplicated. In case of selections, only the selection will get duplicated (not the entire line).

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -706,14 +706,12 @@ impl Editor {
 
         for (start, end) in to_duplicate {
             // insert duplicates
-            let duplicated_text = self.text.slice_to_string(start..end);
             let iv = Interval::new_closed_open(start, start);
+            builder.replace(iv, self.text.slice(start..end));
 
             // last line does not have new line character so it needs to be manually added
             if end == self.text.len() {
-                builder.replace(iv, Rope::from(duplicated_text + &config.line_ending))
-            } else {
-                builder.replace(iv, Rope::from(duplicated_text))
+                builder.replace(iv, Rope::from(&config.line_ending))
             }
         }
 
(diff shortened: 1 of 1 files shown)


cmyr:
I haven't really used either feature, but they both seem valid/interesting, and quite different. I'm not sure I'd like to come down on either side, but I do think going with `duplicate_selection` would at least leave us open to `duplicate_lines` in the future?

scholtzan:
@eyelash Thanks for pointing me to Atom! The behaviour there feels a bit more intuitive. I made some changes so that the behaviour is similar to Atom's and actually duplicates lines also for selections.

scholtzan:
I feel like the correct behaviour is having them all in one undo group. Otherwise, if we duplicate a lot of lines accidentally, wouldn't it then be necessary to do <kbd>cmd</kbd><kbd>Z</kbd> several times?

I don't have write access to this repo so I cannot merge it myself.

cmyr:
oops sorry, I had the wrong permissions set on the @xi-editor/xi-core team. Should be fixed now!

cmyr:
also my thinking on undo is that it's more likely that you accidentally do one extra duplication than that you accidentally do many, so that having undo act on only the last would be more likely to be helpful? Just thinking out loud..

scholtzan:
Oh I see, that makes probably more sense. 
Now I excited and already merged it :( But yeah it sounds more like a bug than a feature.

cmyr:
yea my understanding is that events that are tagged with `EditType::Other` should always be assigned to their own undo group, so i'm not sure why that isn't happening here.

scholtzan:
I am currently using `EditType::InsertChars` here.

cmyr:
oh huh I didn't notice that, but it explains the behaviour. I'd leave it up to you, happy to merge a follow-up if you think that's the right choice. 👍

PR #811, opened 2018-09-18
Implement display for Rope

zettaittenani:
Fix https://github.com/xi-editor/xi-editor/issues/772

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -708,6 +708,16 @@ impl fmt::Debug for Rope {
     }
 }
 
+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if f.alternate() {
+            write!(f, "{}", String::from(self))
+        } else {
+            write!(f, "Rope({})", String::from(self))
+        }
+    }
+}
+
 impl Add<Rope> for Rope {
     type Output = Rope;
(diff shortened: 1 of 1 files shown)


rust/rope/src/rope.rs
```diff
@@ -708,6 +708,16 @@ impl fmt::Debug for Rope {
     }
 }
 
+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if f.alternate() {
```
cmyr: Thinking about this, I think that the _default_ representation should just print the contents of the rope, similar to how the default representation of `String` just prints the contents of the String. In general, we would like to behave more or less the same as the built in string types do.

rust/rope/src/rope.rs
```diff
@@ -708,6 +708,16 @@ impl fmt::Debug for Rope {
...
+            write!(f, "{}", String::from(self))
```
cmyr: Although this works, and is probably reasonable for most uses, I will observe that it isn't _ideal_; specifically it is unnecessarily allocating a new string each time. A more efficient approach would be to iterate through the nodes of the rope with `self.iter_chunks_all()`, and then `write!()` each chunk directly.

cmyr:
This has been stolen away by #830 😢. Thank you for the PR, much appreciated!

Issue #810, opened 2018-09-17
Range select command doesn't shrink the selection


portstrom:
I would expect pointing the range select command at a point within the selected range to move the caret there, shrinking the selected range. Now it does nothing.

cmyr:
I agree this is not the expected behaviour of this feature. 👍

cmyr:
fixed in #817

PR #817, opened 2018-09-19
Fix RangeSelect behaviour 

haroldmrt:
This correct #810 and allow to drag a selection after a RangeSelect.

Before the fix, dragging after doing a RangeSelect would result in a new selection beginning at the end of the previous one. Now it correctly extend the selection. This is not perfect as the selection can only be shrinked to the original RangeSelect.


cmyr:
Looks good, thanks! and thanks for the review @scholtzan 😎

PR #814, opened 2018-09-18
Update contributor guidelines

cmyr:
This is intended to be a fairly comprehensive contributors guide. Feedback very welcome.

If you've been given commit rights to a repo, please check out the 'review process' section, and if anything is unclear let me know; this represents a change to our current process. In particular, if you have been regularly contributing to one of xi-editor/xi-mac, you will now be expected to merge changes on your own, after they have been approved; and you will be able to seek approval from a larger number of other contributors.

If this gets merged, I will update the xi-mac CONTRIBUTORS file to link to this one.

[Rendered version](https://github.com/xi-editor/xi-editor/blob/3433601c91796244aba0a6eba4aa75858a249ca3/CONTRIBUTING.md)

cc @raphlinus @scholtzan @betterclever @jansol @nangtrongvuon @mqzry @lukakerr @ryanbloom @LiHRaM

PR #818, opened 2018-09-19
Update and rename test runner script

cmyr:
This replaces the previous `run_ci` script with a new `run_all_checks` script that runs tests, checks for warnings, and runs benchmarks.

If merged, I'll update #814 to suggest people run this script before opening a PR.

Note: I'm bad at bash, lots of stackoverflow on this one. 😕


rust/run_all_checks
```diff
@@ -0,0 +1,27 @@
...
+
+if ! (cargo test --all); then
+    exit $?
+fi
+
+printf 'Tests passed!\n'
+printf 'Running benchmarks ("rustup run nightly cargo bench --all"):\n'
+
+if rustup run nightly cargo --version; then
+    rustup run nightly cargo bench --all
```
raphlinus: The returned status is not being checked?

PR #828, opened 2018-09-22
Some basic rope benches

mqzry:
A fork of https://github.com/xi-editor/xi-editor/pull/644.
Updated the benches to to use the new `edit_str` signature.

As @vlovich said, these benches are very basic and it should be expanded.
Either 
i)we merge this and come up with more interesting benches in a independent PR
ii) we Keep this PR open and I fill up this branch with more bench code


cmyr:
@mqzry Do you have a preference between those two approaches?

One thing I notice is that there seems to be some duplicated code here (a quick `rg build_triangle` shows that we currently have... three versions of this function in various test files?) so I wonder if we can consolidate that somehow.

Issue #772, opened 2018-07-31
Implement Display for Rope


cmyr:
Just noticed while looking at #771 that ~these aren't~ this isn't implemented.

zettaittenani:
I really want to implement this. How about the code below?
Testing passed.

```diff
diff --git a/rust/rope/src/rope.rs b/rust/rope/src/rope.rs
index 21d6410..3978dd2 100644
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -708,6 +708,16 @@ impl fmt::Debug for Rope {
     }
 }

+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if f.alternate() {
+            write!(f, "{}", String::from(self))
+        } else {
+            write!(f, "Rope({})", String::from(self))
+        }
+    }
+}
+
 impl Add<Rope> for Rope {
     type Output = Rope;
     fn add(self, rhs: Rope) -> Rope {
@@ -1174,7 +1184,7 @@ mod tests {

         let rope_with_emoji = Rope::from("hi\ni'm\n😀 four\nlines");
         let utf16_units = rope_with_emoji.measure::<Utf16CodeUnitsMetric>();
-
+
         assert_eq!(utf16_units, 20);

         // position after 'f' in four
@@ -1193,4 +1203,14 @@ mod tests {
         let utf8_offset = rope_with_emoji.convert_metrics::<Utf16CodeUnitsMetric, BaseMetric>(utf16_units);
         assert_eq!(utf8_offset, 19);
     }
+
+    #[test]
+    fn display() {
+        let rope = Rope::from("hi\ni'm\nfour\nlines");
+        let display_fmt_alt = "hi\ni'm\nfour\nlines";
+        assert_eq!(&format!("{:#}", rope), display_fmt_alt);
+
+        let display_fmt = "Rope(hi\ni'm\nfour\nlines)";
+        assert_eq!(&format!("{}", rope), display_fmt);
+    }
 }
```

cmyr:
@zettaittenani Okay, I think that code looks good!

If you wanted to get fancy, here are some extra ideas:

- instead of `String::from(..)`, you could iterate through the contents of the rope with `Rope::iter_chunks`, and then write these out. This would save having to reallocate the whole contents.

[...]

PR #830, opened 2018-09-22
Display for Rope.

sjoshid:
It looks like there has been no real update for #772. @cmyr is this what you were looking for?
I can write few test cases too. But first wanted to know if this is acceptable.


rust/rope/src/rope.rs
```diff
@@ -730,6 +730,16 @@ impl<'a> From<&'a Rope> for String {
     }
 }
 
+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut ci = self.iter_chunks_all();
+        for s in ci.next() {
+            write!(f, "{}", s).is_ok();
```
mqzry: You should branch on `f.alternate()` like the impl of `fmt::Debug` and the other pull request.

rust/rope/src/rope.rs
```diff
@@ -730,6 +730,16 @@ impl<'a> From<&'a Rope> for String {
     }
 }
 
+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut ci = self.iter_chunks_all();
```
mqzry: The function `iter_chunks_all` does not exist. Use `iter_chunks(0..self.len())` instead to get a iterator over the whole rope.

cmyr: Yea, iter_chunks_all was removed a few days ago, the replacement is `iter_chunks(..)`. (we can now specify ranges in these functions, which is convenient).

rust/rope/src/rope.rs
cmyr: In my opinion an alternate representation isn't strictly necessary, that's more a fancy bonus feature.

One issue here though is that we want to propagate any error that occur in the `write!` macro; the `is_ok()` call basically just ignores them, we should be able to just do `write!(...)?;` and the error will be handled correctly.

rust/rope/src/rope.rs
sjoshid: I thought I had my fork synced with remote but it wasnt. I have made changes in my latest commit.

rust/rope/src/rope.rs
sjoshid: I read in @cmyr comment somewhere it isn't super important. Hence I ignored it.

rust/rope/src/rope.rs
```diff
@@ -756,6 +756,16 @@ impl<'a> From<&'a Rope> for String {
     }
 }
 
+impl fmt::Display for Rope {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut ci = self.iter_chunks(..);
+        for s in ci.next() {
```
cmyr: my only other observation here is that in rust you would normally just write this as,

```rust
for s in self.iter_chunks(..) {
// do something
}
```

That is to say, you don't need the extra step of assigning the iterator to a variable, you can just use it directly.

sjoshid: Great. Will make that change.

sjoshid:
I agree.

cmyr:
Thanks for this!

Issue #655, opened 2018-05-12
Refine backspace behavior


raphlinus:
The recent #520 changed backspace behavior from deleting individual codepoints (which is too small) to deleting entire grapheme clusters (which is too aggressive). The correct behavior is complex. It should:

* Remove individual accents from script base+modifier sequences
* Remove individual indic characters from grapheme clusters
* Delete emoji as a block
* Delete both CR and LF (see google/xi-win#34)
* Delete both the base character and variation selector

In some cases (CRLF, emoji, variation selectors), this logic aligns with grapheme clusters, in others it does not.

A good model for the behavior is Android's[ `getOffsetForBackspaceKey`](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/BaseKeyListener.java#111) in BaseKeyListener.java.

YangKeao:
I am working on this Issue. 

Where should I write functions to get the codePoint's unicode property (should I write them into xi-unicode?) to judge whether it's a combining character or ...? or there is already a library to do this?

And I have noticed that the test for delete backward doesn't involve any of these things, even codepoint and grapheme is both OK for them. So should I add (or modify) some tests about it?

cmyr:
@YangKeao There is a rust unicode crate for looking up unicode properties, but I'm not sure if we want to bring that in as a dependency or not. I'm tagging @raphlinus, he'll have a clearer sense of what fits best.

Re: tests, yes, some proper test cases would definitely make sense for this.

YangKeao:
@raphlinus Do you have any suggestion on whether we should include another rust Unicode crate to look up Unicode properties?

raphlinus:
I haven't researched this in detail, but here are some thoughts. I'm not aware of any crate that provides the unicode properties we need, but it's possible some have been developed since I last looked in detail. If we do take a dependency on a crate, I want to be careful about the weight, probably best measured by binary size impact. Since I've needed various unicode properties for line breaking, I implemented them myself in the xi-unicode subcrate in the xi repo, as suggested above. I'd be fine with that in any case.

We'll definitely want tests, and I think those can be adapted readily from the Android code base, which tests the various backspace cases extensively.

YangKeao:
I am working on this issue now. And my work is nearly a translation of [getOffsetForBackspaceKey](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/BaseKeyListener.java#111) in BaseKeyListener.java .

[...]

PR #837, opened 2018-09-25
Refine backspace behavior

YangKeao:
It's about issue #655.

Have passed all [tests from Android](https://android.googlesource.com/platform/frameworks/base/+/master/core/tests/coretests/src/android/text/method/BackspaceTest.java). Most logic is the same as [code in Android](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/BaseKeyListener.java#111).

The only difference is in `editor.rs:474` because I haven't thought up a implementation of `UCharacter.getCombiningClass(codePoint) == 0`.

Some functions such as `is_emoji` are defined and implemented in `xi-unicode`. I don't know whether the performance of them can be better ( through a better algorithm  )


cmyr:
@YangKeao Thanks for this, very exciting! This will need a careful review so it might take a few more days to get around to, but the overall approach looks good and I'm very happy that you've ported all of the tests, they look fairly comprehensive.

rust/core-lib/src/editor.rs
```diff
@@ -357,8 +359,174 @@ impl Editor {
                     && config.use_tab_stops {
                     tab_start
                 } else {
-                    self.text.prev_grapheme_offset(region.end)
-                        .unwrap_or(region.end)
+
```
cmyr: This fn is definitely too long now. I might consider moving this to `movement.rs`, or I might even consider adding a `backspace.rs` file?

rust/core-lib/src/editor.rs
```diff
@@ -357,8 +359,174 @@ impl Editor {
...
+                    #[derive(PartialEq)]
+                    enum State {
+                        Start,
+                        LF,
```
cmyr: Nit: Rust idiom is to only capitalize the first letter of acronyms/initialisms, so this would be `Lf`, and below would be `BeforeZwj`, etc.

rust/core-lib/src/editor.rs
```diff
@@ -357,8 +359,174 @@ impl Editor {
...
+                                    delete_code_point_count += last_seen_vs_code_point_count + 1;
+                                }
+                                state = State::Finished;
+                            }
+                            State::BeforeVS => {
+                                if is_emoji(code_point) {
+                                    delete_code_point_count += 1;
+                                    state = State::BeforeEmoji;
+                                } else {
+                                    if !is_variation_selector(code_point) { //TODO: UCharacter.getCombiningClass(codePoint) == 0
```
cmyr: Is this TODO for future work or for now?

rust/unicode/src/lib.rs
```diff
@@ -190,6 +190,51 @@ impl LineBreakLeafIter {
...
+pub fn is_emoji_modifier_base(c: char) -> bool {
+    is_in_asc_list(c, &EMOJI_MODIFIER_BASE_TABLE, 0, EMOJI_MODIFIER_BASE_TABLE.len() - 1)
+}
+
+pub fn is_tag_spec_char(c: char) -> bool { '\u{E0020}' <= c && c <= '\u{E007E}' }
+
+pub fn is_emoji_cancel_tag(c: char) -> bool { c == '\u{E007F}' }
+
+pub fn is_zwj(c: char) -> bool { c == '\u{200D}' }
+
```
cmyr: An alternative to this would be an `EmojiExt` trait + `impl EmojiExt for char`. I'm not sure if that is any better. (Just a thought)

rust/unicode/src/tables.rs
```diff
@@ -1498,3 +1498,6 @@ pub const LINEBREAK_STATE_MACHINE: [u8; 3827] = [
     16, 17, 146, 147, 148, 149, 150, 151, 152, 153, 26, 27, 28, 29, 30, 159,
     160, 161, 162, 163, 36, 165, 166, 167, 168, 169, 170,
 ];
+
```
cmyr: we'll want to add `#[cfg_attr(rustfmt, rustfmt_skip)]` to these tables, so they're ignored when #846 lands.

rust/unicode/src/tables.rs
```diff
@@ -1498,3 +1498,6 @@ pub const LINEBREAK_STATE_MACHINE: [u8; 3827] = [
...
...}', '\u{1F984}', '\u{1F985}', '\u{1F986}', '\u{1F987}', '\u{1F988}', '\u{1F989}', '\u{1F98A}', '\u{1F98B}', '\u{1F98C}', '\u{1F98D}', '\u{1F98E}', '\u{1F98F}', '\u{1F990}', '\u{1F991}', '\u{1F992}', '\u{1F993}', '\u{1F994}', '\u{1F995}', '\u{1F996}', '\u{1F997}', '\u{1F998}', '\u{1F999}', '\u{1F99A}', '\u{1F99B}', '\u{1F99C}', '\u{1F99D}', '\u{1F99E}', '\u{1F99F}', '\u{1F9A0}', '\u{1F9A1}', '\u{1F9A2}', '\u{1F9B0}', '\u{1F9B1}', '\u{1F9B2}', '\u{1F9B3}', '\u{1F9B4}', '\u{1F9B5}', '\u{1F9B6}', '\u{1F9B7}', '\u{1F9B8}', '\u{1F9B9}', '\u{1F9C0}', '\u{1F9C1}', '\u{1F9C2}', '\u{1F9D0}', '\u{1F9D1}', '\u{1F9D2}', '\u{1F9D3}', '\u{1F9D4}', '\u{1F9D5}', '\u{1F9D6}', '\u{1F9D7}', '\u{1F9D8}', '\u{1F9D9}', '\u{1F9DA}', '\u{1F9DB}', '\u{1F9DC}', '\u{1F9DD}', '\u{1F9DE}', '\u{1F9DF}', '\u{1F9E0}', '\u{1F9E1}', '\u{1F9E2}', '\u{1F9E3}', '\u{1F9E4}', '\u{1F9E5}', '\u{1F9E6}', '\u{1F9E7}', '\u{1F9E8}', '\u{1F9E9}', '\u{1F9EA}', '\u{1F9EB}', '\u{1F9EC}', '\u{1F9ED}', '\u{1F9EE}', '\u{1F9EF}', '\u{1F9F0}', '\u{1F9F1}', '\u{1F9F2}', '\u{1F9F3}', '\u{1F9F4}', '\u{1F9F5}', '\u{1F9F6}', '\u{1F9F7}', '\u{1F9F8}', '\u{1F9F9}', '\u{1F9FA}', '\u{1F9FB}', '\u{1F9FC}', '\u{1F9FD}', '\u{1F9FE}', '\u{1F9FF}'];
```
raphlinus: How much of this is contiguous ranges, so binary search on range (as used to be done in the Rust stdlib) would be more space-efficient? But I'm fine with this as is, in IRC we're talking about bloat measured in megabytes, and this would be offset by code expansion for the more sophisticated binary search.

cmyr:
@YangKeao hello! wanted to check in on this, since I think it's very close to ready to merge, and it might be harder to merge the longer we wait. Do you have an idea of when you might be able to find a few minutes to finish up?

YangKeao:
@cmyr Sorry for waiting :cry: . I may work on this the day after tomorrow.

rust/core-lib/src/editor.rs
YangKeao: That's from Android's [getOffsetForBackspaceKey:249](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/BaseKeyListener.java#249). I need to write a function to get combining class (maybe in future work?) . However, without this, I passed all tests provided by Android :laughing: .

rust/core-lib/src/editor.rs
YangKeao: That's from Android's [getOffsetForBackspaceKey:249](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/method/BaseKeyListener.java#249). I need to write a function to get combining class (maybe in future work?) . However, without this, I passed all tests provided by Android :laughing: .

rust/unicode/src/lib.rs
YangKeao: I agree it is better!

rust/unicode/src/tables.rs
YangKeao: OK! I have added.

rust/core-lib/src/editor.rs
YangKeao: I don't think `movement.rs`  is a good place for it. Should I move the whole `delete_backward` function to a new file or just some parts of it? 

If break up it into multiple functions, I cannot even think up proper function names of them :cry: .

cmyr: So yes, I could imagine moving the whole function elsewhere, or at least everything that replaces `prev_grapheme_offset`. You might call this, `fn offset_for_delete_backwards(offset: usize, text: &Rope) -> usize`.

I think this probably belongs in core, somewhere. A new file would be fine, `backspace.rs`?

`State` gets a `fn execute()` or `fn run()` or fn

YangKeao: OK! I have moved it.

rust/core-lib/src/lib.rs
```diff
@@ -124,6 +125,7 @@ use internal::line_cache_shadow;
 use internal::width_cache;
 #[cfg(feature = "ledger")]
 use internal::fuchsia;
+use internal::backspace;
```
cmyr: this 'internal' stuff has been deleted, so you don't need to make this change; maybe rebase to `master`?

rust/core-lib/src/backspace.rs
```diff
@@ -0,0 +1,213 @@
+// Copyright 2017 The xi-editor Authors.
```
cmyr: 2018!

YangKeao: :laughing: Sorry!

rust/core-lib/src/lib.rs
YangKeao: OK. I have rebased!

PR #838, opened 2018-09-25
remove `From` string impls from ViewId

### LINKED ISSUE #832: Deserializing ViewId can panic
`ViewId` has a custom deserializer implementation that [panics when receiving a string that doesn't match the expected format](https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/tabs.rs#L856). Deserializers should never panic, but return an error. **Suggested solution:** Remove custom serializer and deseralizer, as they are pointless.

euclio:
The deserializer should fail with an error, not panic, if the view id
does not have the expected format.

Fixes #832.


euclio:
Hm. Looks like the doctests rely on the old `Into` impls.

cmyr:
Thanks for this, it looks good! We implement `From<usize>` for `ViewId` for other purposes, you should be able to use that in the doctests?

PR #839, opened 2018-09-26
Fix injection of view_id

dsp:
Injecting view_id into the params section of the EditCommand "cmd" seems to be incorrect. I discovered this by attempting to use xi_core_lib::rpc as an external crate to generate requests for xi-core. This failed with "missing field: view_id". This pull requests fixes this and adds a test checking the invariant.

I am honestly not sure what our policy of adding tests is, because there seem to be very little in xi-core.


dsp:
Rebased.

As discussed in IRC, I am attempting to use it as the basis of a xi_protocol crate and use it to create a benchmark program to benchmark the RPC interface.

PR #840, opened 2018-09-26
Add missing length check in Delta::is_simple_delete

cmyr:
Hit this playing with something else, all credit goes to liberal use of `debug_assert!` 🙏

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -101,7 +101,7 @@ impl<N: NodeInfo> Delta<N> {
                     end < self.base_len
                 } else if let DeltaElement::Copy(b1, e1) = self.els[1] {
                     // Deletion in middle
-                    end < b1 && e1 == self.base_len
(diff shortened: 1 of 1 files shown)


rust/rope/src/delta.rs
```diff
@@ -101,7 +101,7 @@ impl<N: NodeInfo> Delta<N> {
                     end < self.base_len
                 } else if let DeltaElement::Copy(b1, e1) = self.els[1] {
                     // Deletion in middle
-                    end < b1 && e1 == self.base_len
+                   self.els.len() == 2 && end < b1 && e1 == self.base_len
```
scholtzan: Just some general observation: wouldn't it be more expressive to use `elements` instead of `els`?

cmyr: I agree, and if we really want we could rename it to 'els' on the wire, if that's a concern? @raphlinus?

PR #841, opened 2018-09-26
Make line endings stable during tests

dsp:
When running tests on Windows, we default to Windows line endings.
The tests however expect UNIX-style line endings. Let's override the
config during tests to ensure we always get unix style line endings.


dsp:
I changed this, so that we never load windows specifics during tests.

Issue #673, opened 2018-05-25
Adopt rustfmt


cmyr:
At some point we will want to start using rustfmt, and include a rustfmt pass in CI.

Todos:
- [ ] Update contributing.md to mention rustfmt (also clippy, which was added a while back)
- [ ] Add [`rustfmt.toml`](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md)
- [ ] Add rustfmt to [CI](https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server)
- [ ] Format all existing crates
- [ ] add some little `pre-merge.sh` script that runs rustfmt and clippy? (And mention this in contributing.md?)

Formatting options bikeshed:
In general I'm happy to just stick with the defaults. I would prefer [`fn_args_density`](https://github.com/rust-lang-nursery/rustfmt/blob/master/Configurations.md#fn_args_density) to be 'compressed', but that option isn't stable so can do without.

I don't love `max_width = 100`. I like aiming for 80, but making that a hard limit might be annoying?

If anyone wants to take a stab at this you're very welcome. 😎

Lapz:
I'll work on this

cmyr:
@Lapz cool, thanks!

Lapz:
I been working on this and have decided to use a pre git commit hook for the pre-merge.sh script.Would that be ok?
I've also encountered a [bug](https://github.com/rust-lang-nursery/rustfmt/issues/2760) with rustfmt that I've reported upstream.

cmyr:
@Lapz yea I think that pre-commit would make sense. Maybe adding that can be a separate PR? It can also include running clippy.

We do also have a `ci.local` script in the project root that is intended to run all the tests, it hasn't been used much though.

PR #846, opened 2018-09-29
Another stab at #673

Lapz:
Another stab at #673 but this time much simpler

* Added rustfmt to the CI.
* Ran rustfmt on the whole codebase

**Note**: Running cargo fmt will error on stable but a fix is present in nightly


cmyr:
Can you clarify what you mean about this erroring on stable? it seems to work for me?

rust/core-lib/src/client.rs
```diff
@@ -42,150 +42,176 @@ impl Client {
...
 
-    pub fn update_cmds(&self, view_id: ViewId,
-                       plugin: &str, cmds: &[Command]) {
-        self.0.send_rpc_notification("update_cmds",
-                                     &json!({
+    pub fn update_cmds(&self, view_id: ViewId, plugin: &str, cmds: &[Command]) {
+        self.0.send_rpc_notification(
+            "update_cmds",
+            &json!({
                                          "view_id": view_id,
```
cmyr: it looks like maybe rustfmt is leaving the contents of the macro alone or something? This alignment seems screwy

cmyr:
1. We'll also want to update the `rust/run_all_checks` script to run `rustfmt`.

2. I think maybe this could be good motivation for me to finish up some outstanding reviews, so that we can get any larger PRs merged before running this, because merging after or rebasing might be hairy.

In any case will try and get this resolved quickly.

Thanks @Lapz!

rust/core-lib/src/client.rs
Lapz: Rust fmt dosen't touch macros so it will have to be done manually

Lapz:
I've you changes in #850 in as well and made the change to run_all_checks script

cmyr:
Okay, thanks for grabbing the stuff from 850. We'll also want to rerun `cargo fmt` now that we're ignoring those files.

I've pushed #851, which I'm going to merge shortly, and which will make everything work on stable for now.

My suggestion:

- once #851 is merged, fetch master
- `git rebase -i master`, dropping 957ae88
- rerun `cargo fmt`. You might need to run it twice
- add that commit here
- if we get through CI, we can merge.

One other note is that rustfmt _might_ produce different results on 1.28 (which we use in CI) and 1.29 (current stable). You can install 1.28 through rustup with `rustup toolchain add 1.28.0`, and then run `rustup component add rustfmt-preview`.

rustfmt.toml
```diff
@@ -0,0 +1,5 @@
+force_explicit_abi = true
+max_width = 120
```
cmyr: Can we get rid of this line? I think `100` is fine, and our original motivation for 120 (do what fuchsia does) is less relevant now.

cmyr:
@Lapz So I think this probably requires us to move a bit more quickly than your schedule might allow, so with your approval I would like to:

- cherry-pick your commits here that involve updating the travis config, adding rustfmt.toml, and the docs
- commit these changes, without any rustfmt changes


This will then let us commit the actual reformatted files separately, in chunks, taking care not to create any big merge conflicts for active PRs.

In any case I appreciate the work you've put into this!

cmyr:
@Lapz I think this is too big a change to try and do all at once, and I think it was probably unfair of me to ask you to do it; it was always going to be hard to get this merged. I'm going to take this on myself, and do it in stages, trying to avoid interfering with other work. Thank you for the effort, and I'd be happy to find another better starter issue for you if you like!

Issue #805, opened 2018-09-16
End of line behavior for command to transpose characters


portstrom:
Xi has a command called `transpose`. It works similarly to the command `transpose-chars` in Emacs, but doesn't get the end of the line behavior right. From the [Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/emacs/Transpose.html):

> Normally, `C-t` transposes the two characters on either side of point. When given at the end of a line, rather than transposing the last character of the line with the newline, which would be useless, `C-t` transposes the last two characters on the line. So, if you catch your transposition error right away, you can fix it with just a `C-t`.

cmyr:
We have in general chosen ST3 as a model when nitting various little behaviours, and I'm happy with that choice.

In this case, both approaches address the fundamental problem that transposing a newline is probably not helpful; and if we pretend that newlines don't exist, transposing with the next non-newline character makes sense to me. If there's a specific utility-based argument for the other approach I'm happy to hear it.

portstrom:
The quote from the Emacs manual makes it clear why that behavior is preferred and why the other behavior is even considered useless. In general I think Emacs and Vi are going to have every text related feature more well-designed than any editor with a shorter history. I assume the feature in Sublime works differently not because it was a conscious design choice, but simply because the author of Sublime never used Emacs, and doesn't know how useful this feature can be when implemented properly.

If you insist on having the Sublime feature, you can simply make two different commands for the two features, and any frontend can choose which one to use, or let the user configure the choice. The Sublime feature will however, as the Emacs manual says, be useless.

raphlinus:
The existing behavior was a deliberate choice (mostly because I felt Sublime was a good model to follow), but if people have strong preferences I'm willing to change it.

sidsrvstv:
Are there any issues I can help out with?

PR #847, opened 2018-09-29
Emacs style transpose

psylinse:
This addresses #805.

This is the first rust code I've written so please don't be shy with feedback. This was done more to understand xi and rust. If it's way off the mark no worries if you just close it.


cmyr:
Okay, the code looks good. There are tests for various edit actions in `event_context.rs`; I'm not sure if transpose is covered there, but you are very welcome to add some tests.

It's possible at some point in the future that this behaviour will branch on some config option; perhaps for now you could keep a comment in place, similar to the one there already?

cmyr:
@psylinse bump?

psylinse:
I will get to this on Saturday.

psylinse:
@cmyr Okay, I have rebased and added a comment. Unfortunately, I don't have time to dig into tests or anything so I hope this is sufficient!

PR #855, opened 2018-10-01
Fix incorrect behavior for transpose at end-of-line

### LINKED ISSUE #805: End of line behavior for command to transpose characters

revolter:
Fixes #805.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -606,20 +606,29 @@ impl Editor {
         (as_interval, interval_rope)
     }
 
-    fn do_transpose(&mut self, view: &View) {
+    fn do_transpose(&mut self, view: &View, config: &BufferItems) {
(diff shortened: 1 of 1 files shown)


revolter:
Oh, I'm too tired and didn't notice. Feel free to close.

cmyr:
@revolter no worries, thanks for the PR in any case; will close if/when the other is merged.

PR #861, opened 2018-10-02
Make`slice_to_string` return `Cow<str>` instead of `String`

martin-fink:
Make `slice_to_string` return `Cow<str>` instead of `String`. This avoids making a copy of the string on every call, even if the string is never modified. 

This is my first contribution to a Rust open source project, I hope my code is acceptable how it is

If not, I'm open for suggestions to improve or fix it.

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -398,7 +398,7 @@ impl Editor {
         if save {
             let saved = self.extract_sel_regions(&deletions)
                 .unwrap_or_default();
-            *kill_ring = saved.into();
+            *kill_ring = saved.as_ref().into();
         }
         self.delete_sel_regions(&deletions);
     }
@@ -420,7 +420,7 @@ impl Editor {
 
     /// Extracts non-caret selection regions into a string,
     /// joining multiple regions with newlines.
-    fn extract_sel_regions(&self, sel_regions: &[SelRegion]) -> Option<String> {
+    fn extract_sel_regions(&self, sel_regions: &[SelRegion]) -> Option<Cow<str>> {
         let mut saved = None;
         for region in sel_regions {
             if !region.is_caret() {
@@ -428,8 +428,8 @@ impl Editor {
                 match saved {
                     None => saved = Some(val),
                     Some(ref mut s) => {
-                        s.push('\n');
-                        s.push_str(&val);
+                        s.to_mut().push('\n');
+                        s.to_mut().push_str(&val);
                     }
                 }
             }
@@ -571,7 +571,7 @@ impl Editor {
 
     pub(crate) fn do_copy(&self, view: &View) -> Value {
         if let Some(val) = self.extract_sel_regions(view.sel_regions()) {
-            Value::String(val)
+            Value::String(val.as_ref().to_owned())
         } else {
             Value::Null
         }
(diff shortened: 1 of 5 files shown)


rust/rope/src/rope.rs
```diff
@@ -634,18 +634,27 @@ impl Rope {
         leaf.as_bytes()[pos]
     }
 
-    // TODO: this should be a Cow
-    // TODO: a case can be made to hang this on Cursor instead
-    pub fn slice_to_string<T>(&self, range: T) -> String 
+    pub fn slice_to_string<T>(&self, range: T) -> Cow<str>
```
cmyr: I think we should also rename this function, since it no longer returns a `String`. I could imagine calling it `slice_to_cow` or `get_slice` or `slice_cow` or `cow_slice`; even maybe `slice_str`?

rust/core-lib/src/editor.rs
```diff
@@ -398,7 +398,7 @@ impl Editor {
         if save {
             let saved = self.extract_sel_regions(&deletions)
                 .unwrap_or_default();
-            *kill_ring = saved.into();
+            *kill_ring = saved.as_ref().into();
```
cmyr: I _think_ all these `as_ref().into()` can be replaced with `into_owned()`.

rust/core-lib/src/editor.rs
```diff
@@ -622,8 +622,9 @@ impl Editor {
                 if let Some(end) = self.text.next_grapheme_offset(middle) {
                     if start >= last {
                         let interval = Interval::new_closed_open(start, end);
-                        let swapped = self.text.slice_to_string(middle..end) +
-                                      &self.text.slice_to_string(start..middle);
+                        let swapped: String = format!("{}{}",
```
cmyr: A perhaps slightly cleaner way to write this would be,
```rust
let before =  self.text.slice_to_string(start..middle);
let after = self.text.slice_to_string(middle..end);
let swapped: String = [after, before].concat();
```

(or some variant on that; but the basic pattern of `concat`ing an array.)

rust/plugin-lib/src/base_cache.rs
```diff
@@ -433,7 +433,7 @@ mod tests {
             if offset > self.0.len() {
                 Err(Error::Other("offset too big".into()))
             } else {
-                let chunk = self.0.slice_to_string(offset..end_off);
+                let chunk = self.0.slice_to_cow(offset..end_off).as_ref().to_owned();
                 Ok(GetDataResponse { chunk, offset, first_line, first_line_offset })
             }
```
cmyr: one last `into_owned()`

rust/rope/src/rope.rs
```diff
@@ -752,7 +761,7 @@ impl From<Rope> for String {
 
 impl<'a> From<&'a Rope> for String {
     fn from(r: &Rope) -> String {
-        r.slice_to_string(..)
+        r.slice_to_cow(..).as_ref().to_owned()
```
cmyr: 😉

rust/rope/src/rope.rs
```diff
@@ -1265,4 +1274,35 @@ mod tests {
...
+
+        let rope = Rope::from(&long_text);
+
+        let cow = rope.slice_to_cow(..);
+
+        assert!(long_text.len() > 1024);
+        assert_eq!(
+            cow,
+            Cow::Owned(long_text) as Cow<str>
+        );
```
cmyr: Can we try another check here that a smaller slice of this rope (at the beginning of the rope, because even a small slice that happens to be on a leaf boundary will require `Owned`) is also `Borrowed`?

martin-fink:
Thank you for your patience and for your help!

PR #864, opened 2018-10-02
Support capitalizing text

scholtzan:
This adds support for capitalizing selected text. The behaviour is similar to the behaviour in XCode.


rust/core-lib/src/editor.rs
```diff
@@ -685,6 +686,48 @@ impl Editor {
...
+                // capitalize each word in the current selection
+                let (start, end) = word_cursor.select_word();
+
+                if start < end {
+                    let interval = Interval::new_closed_open(start, end);
+                    let word = self.text.slice_to_string(start..end);
+
+                    // first letter is uppercase, remaining letters are lowercase
+                    let (first_char, rest) = word.split_at(1);
+                    let mut capitalized_text = String::from("");
```
cmyr: note/non-blocking: you can get an empty string with `String::new()`; and if we know approximately our length (ignoring, say, ß) we might save ourselves an allocation here with `String::with_capacity(word.len())`.

rust/core-lib/src/editor.rs
```diff
@@ -685,6 +686,48 @@ impl Editor {
...
+                    capitalized_text.push_str(&first_char.to_uppercase());
+                    capitalized_text.push_str(&rest.to_lowercase());
+                    builder.replace(interval, Rope::from(capitalized_text));
```
cmyr: I wonder if we can get away with only replacing the first char with the results of `to_uppercase`? it seems like the second part of this is a noop?

jansol: The comment above specifically says that everything except the first char is lowercased. (they might not be lowercase at first)

rust/core-lib/src/editor.rs
raphlinus: I think [a, b].concat() is the preferred idiom here.

rust/core-lib/src/editor.rs
scholtzan: Yeah, I thought it might be good to handle cases like: thIS iS wEiRd → This Is Weird

rust/core-lib/src/editor.rs
```diff
@@ -685,6 +686,48 @@ impl Editor {
...
+
+            loop {
+                // capitalize each word in the current selection
+                let (start, end) = word_cursor.select_word();
+
+                if start < end {
+                    let interval = Interval::new_closed_open(start, end);
+                    let word = self.text.slice_to_string(start..end);
+
+                    // first letter is uppercase, remaining letters are lowercase
```
raphlinus: I'm trying to think through the Unicode consequences of this. Certainly it seems to be making assumptions that the text is alphabetic, but since this is opt-in, I think maybe that's reasonable. Certainly there are cases this might break a grapheme cluster boundary (like emoji), but in those cases I expect the case-transforms will be no-ops, thus the word will be a no-op as well.

PR #870, opened 2018-10-03
Cleanup

dsp:
While reading the rope code I stumbled across some old, commented out code. I am proposing to delete it since it has not been updated or used since 2 years. Feel free to ignore and reject this pull request if we think there is value in keeping it around visible in the repo.


cmyr:
@dsp can we get a rebase?

dsp:
@cmyr rebased :)

PR #875, opened 2018-10-03
put plugin_rpc back

dzhou121:
Can we re-enable plugin_rpc to how it was previously?

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -702,7 +702,7 @@ impl Editor {
 
                 if start < end {
                     let interval = Interval::new_closed_open(start, end);
-                    let word = self.text.slice_to_string(start..end);
+                    let word = self.text.slice_to_cow(start..end);
 
                     // first letter is uppercase, remaining letters are lowercase
                     let (first_char, rest) = word.split_at(1);
--- a/rust/core-lib/src/plugins/mod.rs
+++ b/rust/core-lib/src/plugins/mod.rs
@@ -134,6 +134,13 @@ impl Plugin {
                                                 "request_id": request_id,
                                                 "position": position}))
     }
+
+    pub fn dispatch_command(&self, view_id: ViewId, method: &str, params: &Value) {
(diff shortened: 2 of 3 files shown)


cmyr:
Does this work? Are you using this? (Where are you using this??)

dzhou121:
Yes it works. I'm using this for my front-end to plugin communication. I'm writing a vim key plugin for things I can't emulate in the front-end directly, and I will send lots of custom commands from front-end to plugin.

cmyr:
@dzhou121 I've been thinking a lot about how vim mode is going to work long-term, and I'm also thinking about unifying the frontend and plugin commands, which I think came up in an earlier discussion we had? If you're interested in trying to work on vim-mode in core, I would be happy to share my current thoughts..

cmyr:
also this PR will need to be rebased to master in order to pass CI.

dzhou121:
Vim-mode is kind of a complicated key binding and I feel it is more suitable to be a plugin.

I do agree with unifying the frontend and plugin commands, and I've asked for this a while ago 
https://github.com/crane-editor/crane/issues/1

Thinking about a vim-mode plugin, it needs the ability to change cursor position.

dzhou121:
Also, in additional to unifying the rpc commands, plugin should be able to send notifications to front-end as well. 

Take the lsp plugin for example, it needs to send to the front-end with diagnostics, alerts etc. I'm aware that you can add these to the core and some of them have, it's just more flexible for front-ends and plugins to be able to talk with each other with modifying the core.

cmyr:
I agree that vim-mode should be _something_ like a plugin, but I think it needs to be handled differently than the current plugin system; it needs deeper integration into core, and it needs to handle raw keyboard events from the frontend.

I think diagnostics/alerts etc can all be mediated by the core. For instance, core will decide whether to invalidate a given diagnostic in response to an edit (if the line that contains the diagnostic is deleted, we will clear the diagnostic; the frontend will know have enough information to do this).

Issue #807, opened 2018-09-16
Commands to increment and decrement numbers


portstrom:
Vim has commands to [increment](https://vimhelp.appspot.com/usr_26.txt.html#26.2) (`Ctrl-A`) and decrement (`Ctrl-X`) numbers. These would be good to have in Xi as well.

Some details to pay attention to:
* If the caret is not pointing at a number, the next number after the caret is selected, but only within the same line.
* After the command, the caret is placed at the end of the number.
* Negative numbers are meaningfully handled.

cmyr:
I'd put this firmly in 'one day'/'nice to have' territory? I do use this in vim sometimes, but it still feels like a slightly odd feature, that is mostly justified by it's use in macros? maybe?

Does it handle hex/octal/binary?

In any case, this does feel like a fun fairly self-contained project, so I would certainly be open to having someone work on it if they were interested.

portstrom:
I don't use Vim often, and I don't even now how to make macros in Vim, but when I do use Vim, I often use this feature. When I use Emacs I miss it.

Octal numbers are handled in the confusing way you can see following the link. I think it would be better not to handle octal numbers, or have interpretation of numbers depend on the language being edited.

terhechte:
I'd like to give this a try. If I manage to implement this, then I'll also add support to Xi-Mac

terhechte:
I've added an implementation here: https://github.com/xi-editor/xi-editor/pull/877

PR #877, opened 2018-10-03
[807] Add number operations

terhechte:
#807  

This includes two new operations: "IncreaseNumber" and "DecreaseNumber". 
The behaviour is as follows:

If there is a number next to or on the beginning of the region, then
this number will be increased / decreased and the cursor will be placed at the end of the number.

If there is a number somewhere in the region (but not on the beginning of the region) nothing will happen.

Some Examples with a increment `transform_function`:

"|1234" -> "1235|"
"12|34" -> "1235|"
"-|12" -> "-11|"
"1[23 and other contents|]" -> "124| and other contents"
"[text and 42 a number|]" -> Nothing happens
"[test and 4|] a number" -> Nothing happens

Multiple regions are supported.

Xi-Mac support can be found here:
https://github.com/xi-editor/xi-mac/pull/288/


cmyr:
This looks good, someone will get around to a closer review shortly. One thing this does need though is to be added to the frontend docs (`docs/docs/frontend-protocol.md`).

terhechte:
I've added it to the frontend docs

cmyr:
@terhechte this will need to get rebased to origin/master to pass CI.

terhechte:
@cmyr I've merged master in and changed to `slice_to_cow`. Everything should build fine now.

portstrom:
Pay attention to what happens when the number overflows. Your implementation works with `i32` and doesn't do any overflow check. Better change that to `i64` and use [`checked_add`](https://doc.rust-lang.org/std/primitive.i64.html#method.checked_add) and [`checked_sub`](https://doc.rust-lang.org/std/primitive.i64.html#method.checked_sub).

terhechte:
@portstrom good catch. I've moved to i128 to also support really huge numbers and `wrapping_add` which feels like a good-enough solution.

rust/core-lib/src/event_context.rs
```diff
@@ -839,4 +839,145 @@ mod tests {
...
+        assert_eq!(harness.debug_render(),"this is a 999| text example");
+
+        // Number Wrapping
+        ctx.do_edit(EditNotification::MoveToEndOfDocument);
+        ctx.do_edit(EditNotification::DeleteToBeginningOfLine);
+        ctx.do_edit(EditNotification::Insert { chars: format!("{}", i128::max_value()) });
+        ctx.do_edit(EditNotification::Gesture { line: 0, col: 11, ty: PointSelect });
+        ctx.do_edit(EditNotification::MoveRightAndModifySelection);
+        ctx.do_edit(EditNotification::IncreaseNumber);
+        assert_eq!(harness.debug_render(), format!("{}|", i128::min_value()));
```
cmyr: with the assert macros you don't need the `format!` call; the asserts treat all arguments after the statement as a format argument, so you can just write, say, `assert!(false, "this is a {} example", "pretty bad");`

portstrom:
Why do `wrapping_add` when you can simply do `if let Some(number) = line[begin..end].parse::<i128>().and_then(transform_function).ok() {`, which lets you use `checked_add`? Also notice I removed the `&` and `*` which are not needed.

terhechte:
@cmyr @portstrom Thanks for the detailed feedback. I'll have a go at it today. I'll look into the WordCursor. I found the various ways of accessing the text (rope, view, cursors) a bit overwhelming, I have to admit (given that the documentation is also sparse) and went with a solution that seemed manageable.

terhechte:
@portstrom Good catch, I've updated it. thanks.

rust/core-lib/src/event_context.rs
```diff
@@ -954,14 +953,29 @@ mod tests {
...
         ctx.do_edit(EditNotification::DeleteToBeginningOfLine);
-        ctx.do_edit(EditNotification::Insert { chars: format!("{}", i128::max_value()) });
-        ctx.do_edit(EditNotification::Gesture { line: 0, col: 11, ty: PointSelect });
-        ctx.do_edit(EditNotification::MoveRightAndModifySelection);
+        ctx.do_edit(EditNotification::Insert { chars: "10_000".into() });
+        ctx.do_edit(EditNotification::MoveToEndOfDocument);
+        ctx.do_edit(EditNotification::IncreaseNumber);
+        assert_eq!(harness.debug_render(), "10_000|");
+
+        // decimals are kinda accounted for (i.e. 4.55 becomes 4.56 (good), but 4.99 becomes 4.100 (bad)
```
cmyr: this is okay, we're never going to handle floating point 'correctly', and just consistently treating anything we handle as an integer feels okay to me.

rust/core-lib/src/editor.rs
```diff
@@ -686,6 +686,44 @@ impl Editor {
...
+    fn change_number<F: Fn(i128) -> Option<i128>>(&mut self, view: &View,
+                                        transform_function: F) {
+        let mut builder = delta::Builder::new(self.text.len());
+        for region in view.sel_regions() {
+
+            let mut cursor = WordCursor::new(&self.text, region.end);
+            let (mut start, end) = cursor.select_word();
+
+            // see if the word begins with a -
+            if start > 0 && self.text.slice_to_cow(start - 1..start) == "-" {
```
cmyr: So I don't think this is a problem now, but at some point we will likely want to implement some approximation of [UAX29](http://unicode.org/reports/tr29/) which means it's likely there are characters that will be word boundaries that will have multi-byte representations in utf8; if that was the case this `slice` would panic, because `start - 1` would not be a word boundary. 

What we probably want to do is to explicitly grab the preceding codepoint boundary, and check that character. This is all handled by the generic `Cursor` type in the rope crate, but there's a convenience method on `Rope` itself; you can just do `rope.prev_codepoint_offset(start)`. **edit** I'll leave that there for informative purposes, but there's a better way to do this: you can just do `rope.byte_at(start - 1)`; this won't panic if we're not on a char boundary, and you can just compare the bytes directly.

terhechte: Thanks for the feedback! I'll give this a try later!

terhechte: Added it, works great.

portstrom:
@cmyr That's something [Clippy](https://github.com/rust-lang-nursery/rust-clippy) detects. You should really use clippy before merging anything at all.

terhechte:
@portstrom There seems to be a ticket for rustfmt and clippy integration: https://github.com/xi-editor/xi-editor/issues/673

cmyr:
@portstrom we would happily accept a PR including clippy fixes and/or adding a clippy check to CI!

Issue #863, opened 2018-10-02
core should send client a list of available languages on startup


cmyr:
This will be the first step in letting clients manually specify or override the language for a file.

This involves adding a new core->client RPC, which will need to go in the documentation.

Implementation will be very similar to `available_themes`; indeed the language list is available in the `finish_init` fn in `tabs.rs`. 

This should be sent as a list of 'LanguageId' `String`s.

daubaris:
Would love to work on this :) any tips?

cmyr:
not much more than are in the original issue; take a look at how the `available_themes` RPC works, and maybe use that as a model? If you have any specific questions, you can ask them here or at irc.mozilla.org #xi. 😎

daubaris:
Hi again, I've made some progress with this issue, but I've found myself cornered with language_list. 

As you've mentioned implementation should be very similar to `available_themes` fn.
So basically it should look like this:

```
pub fn available_languages(&self, languages: Vec<String>) {
    self.0.send_rpc_notification("available_languages",
                                 &json!({"languages": languages}))
}
```

Or should I pass a `Vec<LanguageId>`?

Also while looking into tabs.rs `finish_init` function I cant seem to find mentioned language list, because it's executed in the `finalize_new_views` fn in tabs.rs with a config.

Could you explain a bit further about this language list?

cmyr:
yea, so you can send `LanguageId`s, which are just strings. You may need to add `#[derive(Serialize)]` to `LanguageId`. 

The fn in `tabs.rs` is actually called `finish_setup`.

PR #878, opened 2018-10-03
#863 core should send client a list of available languages on startup

daubaris:
#863 

Added the implementation of available languages.

Things to note:
Not quite sure if line 197 in tabs.rs is a good solution, would love to receive some feedback about it.

### DIFF
--- a/rust/core-lib/src/syntax.rs
+++ b/rust/core-lib/src/syntax.rs
@@ -41,7 +41,7 @@ pub struct LanguageDefinition {
 /// A repository of all loaded `LanguageDefinition`s.
 #[derive(Debug, Default)]
 pub struct Languages {
-    named: HashMap<LanguageId, Arc<LanguageDefinition>>,
+    pub named: HashMap<LanguageId, Arc<LanguageDefinition>>,
     extensions: HashMap<String, Arc<LanguageDefinition>>,
 }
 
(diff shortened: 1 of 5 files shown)


daubaris:
Going to fix rpc tests

rust/core-lib/src/syntax.rs
```diff
@@ -41,7 +41,7 @@ pub struct LanguageDefinition {
 /// A repository of all loaded `LanguageDefinition`s.
 #[derive(Debug, Default)]
 pub struct Languages {
-    named: HashMap<LanguageId, Arc<LanguageDefinition>>,
+    pub named: HashMap<LanguageId, Arc<LanguageDefinition>>,
```
cmyr: this is unnecessary; `Languages` has an `iter` fn, which returns an iterator over the `LanguageDefinition` objects.

rust/core-lib/src/tabs.rs
```diff
@@ -194,6 +194,8 @@ impl CoreState {
         let plugin_paths = self.config_manager.get_plugin_paths();
         self.plugins.reload_from_paths(&plugin_paths);
         let languages = self.plugins.make_languages_map();
+        let languages_ids = languages.named.keys().cloned().collect::<Vec<_>>();
```
cmyr: as mentioned above, this can just be `languages.iter().cloned().collect::<Vec<_>>();`.

cmyr: er, the above is slightly wrong. I would do this as,
```rust
langauges.iter().map(|l| l.name.clone()).collect::<Vec<_>>();
```

daubaris: Great! I’ll fix it as soon as I get back from work!

PR #879, opened 2018-10-03
Try RPC deserialization first with params

scholtzan:
This is a first attempt to fix https://github.com/xi-editor/xi-editor/issues/833.
It works when I send an empty dictionary as param: `document.sendRpcAsync("find_next", params: [:])`

### DIFF
--- a/rust/core-lib/src/rpc.rs
+++ b/rust/core-lib/src/rpc.rs
@@ -534,18 +534,16 @@ impl<'de, T: Deserialize<'de>> Deserialize<'de> for EditCommand<T>
         let helper = InnerId::deserialize(&v).map_err(de::Error::custom)?;
         let InnerId { view_id } = helper;
 
-        if T::deserialize(v.clone()).is_err() {
-            // if params are empty, remove them
-            let remove_params = match v.get("params") {
-                Some(&Value::Object(ref obj)) => obj.is_empty(),
-                Some(&Value::Array(ref arr)) => arr.is_empty(),
-                Some(_) => return Err(de::Error::custom("'params' field, if present, must be object or array.")),
-                None => false,
-            };
+        // if params are empty, remove them
+        let remove_params = match v.get("params") {
+            Some(&Value::Object(ref obj)) => obj.is_empty() && T::deserialize(v.clone()).is_err(),
+            Some(&Value::Array(ref arr)) => arr.is_empty() && T::deserialize(v.clone()).is_err(),
+            Some(_) => return Err(de::Error::custom("'params' field, if present, must be object or array.")),
+            None => false,
+        };
 
-            if remove_params {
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/rpc.rs
```diff
@@ -513,16 +513,19 @@ impl<'de, T: Deserialize<'de>> Deserialize<'de> for EditCommand<T>
...
-        let remove_params = match v.get("params") {
-            Some(&Value::Object(ref obj)) => obj.is_empty(),
-            Some(&Value::Array(ref arr)) => arr.is_empty(),
-            Some(_) => return Err(de::Error::custom("'params' field, if present, must be object or array.")),
-            None => false,
-        };
 
-        if remove_params {
-            v.as_object_mut().map(|v| v.remove("params"));
+        if T::deserialize(v.clone()).is_err() {
```
cmyr: This is a little concerning, because it means we're always cloning the contents of each RPC we receive, which I suspect has measurable performance implications.

I think it would be okay to take this approach _only_ when the params are empty; in that case the clone will be cheap. In any case it's annoying that we can't somehow limit the clone to situations where deserialize fails; or alternately that we can't recover the original value in the failure case. 😞

scholtzan: Yeah :(

rust/core-lib/src/rpc.rs
```diff
@@ -533,10 +533,11 @@ impl<'de, T: Deserialize<'de>> Deserialize<'de> for EditCommand<T>
         let mut v = Value::deserialize(deserializer)?;
         let helper = InnerId::deserialize(&v).map_err(de::Error::custom)?;
         let InnerId { view_id } = helper;
+
         // if params are empty, remove them
         let remove_params = match v.get("params") {
-            Some(&Value::Object(ref obj)) => obj.is_empty(),
-            Some(&Value::Array(ref arr)) => arr.is_empty(),
+            Some(&Value::Object(ref obj)) => obj.is_empty() && T::deserialize(v.clone()).is_err(),
+            Some(&Value::Array(ref arr)) => arr.is_empty() && T::deserialize(v.clone()).is_err(),
```
cmyr: I think this is an improvement, but it still means we're going to be trying and failing to deserialize an awful lot of RPCs, and I _am_ curious about the performance implications.

scholtzan: Good question, I'll look into this later today.

scholtzan: Interesting. I measured the deserialization performance for `find_next` with and without parameters. There is basically no difference. 
I measured the time using `xi_trace` from line 537 until after line 549. With and without parameters it takes ~ 0.050ms.

cmyr: Okay, thanks for measuring!

Issue #866, opened 2018-10-02
allow client to override a buffer's language


cmyr:
This requires a new top-level RPC, which we might call `set_language`, which passes a `ViewId` and a `LanguageId`. When handling in `tabs.rs`, we'll need to call `config_manager.override_language(..)`.

I'm not sure if this is the approach we'll want to take long term, but I think it's a reasonable at this point and it isn't hard to change later.

Won't be useful until #863 is also finished, and should probably be implemented alongside a `xi-mac` PR for easy testing.

mrandri19:
I've started working on it

PR #881, opened 2018-10-03
#866 allow client to override a buffer's language

mrandri19:
I've added `set_language`, but haven't added tests yet since I haven't figured out the correct way to do it and it doesn't seem that all of the rpc commands are tested.

My editor removed some trailing whitespace, if that makes the diff noisy I can fix it.

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -73,13 +73,13 @@ note for `new_view`. Errors are not currently reported.
 
 `set_theme {"theme_name": "InspiredGitHub"}`
 
-Requests that core changes the theme. If the change succeeds the client
+Asks core to change the theme. If the change succeeds the client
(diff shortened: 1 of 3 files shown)


docs/docs/frontend-protocol.md
```diff
@@ -73,9 +73,14 @@ note for `new_view`. Errors are not currently reported.
 
 `set_theme {"theme_name": "InspiredGitHub"}`
 
-Requests that core change the theme. If the change succeeds the client
```
cmyr: This is twice now someone has made this change, and twice I've complained that, actually, this is totally acceptable English, as in "we ask that they remove their shoes." Although it seems like this is confusing people, so maybe we could just go with "Asks core to change the theme"?

mrandri19: No problem, will change :)

Issue #836, opened 2018-09-24
log file path should be specified by client


cmyr:
logging setup currently fails on macOS for release builds because the application cannot write new files inside its app bundle. 

In any case, the expected location for logs is system dependent, and we should respect that. There may be a rust crate for finding log file paths, but we probably want the client to have ultimate control.

An alternative approach would be to build logging into the RPC process, such that all logs end up getting sent to the client, which can do with them what it wants; there are problems with this, though, such as how we would log errors in the RPC system itself. 😕

06kellyjac:
Looking into it 🙂

cmyr:
@06kellyjac great! I _think_ the way that we will want to do this is by passing an argument to xi-core at exec time; we need to know the log path in `rust/src/main.rs`, so it can't happen later like with config paths.

We'd want this argument to be optional, so that clients that don't specify a log path still work fine. I'd also suggest just handling arg parsing manually for now, it's probably not worth bringing in a dependency yet. 😎

06kellyjac:
@cmyr I did a bit of digging before seeing your message so I went on a different route 😅. I found a crate that gets OS standard paths that can be the defaults (and I played around with it a bit); in this case for logs we'd want `XDG_DATA_DIRS` and other OS equivalents.

Technically there are two I looked at:

- <https://crates.io/crates/dirs>
- <https://crates.io/crates/directories>

They're both primarily by the same person and both up to date. I went with `dirs` as it looked quite nice to use and was more "low-level".

I'll get some of the code up soon for you to pick apart. I'm a bit new to Rust so any and all nitpicking is 100% welcome.

---

How do you want the argument to be passed in?

- `xi-editor --log-dir="SOMEPATH"`
- `LOG_DIR=log-dir="SOMEPATH" xi-editor`
- both?

Also for do you want `xi-core.log` & `xi-lsp-plugin.log` to go in the same directory?

I was thinking default to what's spat out of `dirs` & put both in there, then have `--log-dir` override that
(& then if you really wanted have a `--log-dir-lsp` to override the `xi-lsp-plugin.log` but maybe keep it to just `--log-dir` for now?)

Thanks 😄

cmyr:
@06kellyjac I think we just one a single directory, and we'll put all log files there. `--log-dir` works. the `dirs` crate looks nice and light weight, so I'd go with that.

So: sounds good, basically!

[...]

PR #886, opened 2018-10-04
Added functionality to get OS standard directory to put the logs.

06kellyjac:
Deals with part of issue #836 

Acceptance Criteria:

- ~Get OS standard directories by default~
  - [X] Done for `xi-core.log`
   - ~Done for `xi-lsp-plugin.log`~
- ~Accept argument override `--log-dir` for front ends such as `xi-mac`~
  - ~Done for `xi-core.log`~
  - ~Done for `xi-lsp-plugin.log`~

Just the OS directories for `xi-core.log` for now. Argument override will come in another PR.
After that I'll look at sharing both of those changes for `xi-core.log` and `xi-lsp-plugin.log`.

### DIFF
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -64,7 +64,7 @@ fn setup_logging() -> Result<(), fern::InitError> {
         Err(_) => log::LevelFilter::Info,
     };
 
-    let fern_dispatch = fern::Dispatch::new()
+    let mut fern_dispatch = fern::Dispatch::new()
         .format(|out, message, record| {
             out.finish(format_args!(
                 "{}[{}][{}] {}",
@@ -76,18 +76,21 @@ fn setup_logging() -> Result<(), fern::InitError> {
         }).level(level_filter)
         .chain(io::stderr());
 
-    match path_for_log_file(XI_LOG_FILE) {
-        Ok(logging_file_path) => fern_dispatch
-            .chain(fern::log_file(logging_file_path)?)
-            .apply()?,
-        Err(err) => {
-            fern_dispatch.apply()?;
-            warn!(
-                "There was an issue getting the path for the log file: {}, falling back to stderr.",
-                err,
-            );
-        }
+    let path_result = path_for_log_file(XI_LOG_FILE);
+    // If the logging_file_path returned successfully, add the logfile capability to fern
+    if let Ok(logging_file_path) = &path_result {
+        fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
     }
+
+    // Start fern
+    fern_dispatch.apply()?;
+
+    // If the logging_file_path returned an error, print it with fern/logs
+    if let Err(e) = &path_result {
+        let message = "There was an issue getting the path for the log file";
+        warn!("{}: {:?}, falling back to stderr.", message, e);
+    }
+
     info!("Logging with fern is setup");
     Ok(())
 }
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -24,7 +24,7 @@ extern crate xi_rpc;
 
 use std::fs;
 use std::io;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use xi_core_lib::XiCore;
 use xi_rpc::RpcLoop;
@@ -45,7 +45,7 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
     }
 }
 
-fn path_for_log_file<P: AsRef<str>>(filename: P) -> Result<PathBuf, io::Error> {
+fn path_for_log_file<P: AsRef<Path>>(filename: P) -> Result<PathBuf, io::Error> {
     let mut logging_directory = get_logging_directory()?;
     // Create the logging directory
     fs::create_dir_all(&logging_directory)?;
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -12,17 +12,33 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 use std::io;
+use std::fs;
+use std::path::PathBuf;
 
 #[macro_use]
 extern crate log;
 extern crate chrono;
 extern crate fern;
 
+extern crate dirs;
+use dirs::data_local_dir;
+
 extern crate xi_core_lib;
 extern crate xi_rpc;
 
 use xi_core_lib::XiCore;
 use xi_rpc::RpcLoop;
+
+fn get_logging_directory() -> Result<PathBuf, fern::InitError> {
+    let xi_directory = "xi-core/";
+    if let Some(mut log_dir) = data_local_dir() {
+        log_dir.push(xi_directory);
+        Ok(log_dir)
+    } else {
+        let dir_error = io::Error::new(io::ErrorKind::Other, "dir was not able to find a directory");
+        Err(fern::InitError::Io(dir_error))
+    }
+}
 
 fn setup_logging() -> Result<(), fern::InitError> {
     let level_filter = match std::env::var("XI_LOG") {
@@ -32,9 +48,16 @@ fn setup_logging() -> Result<(), fern::InitError> {
             _ => log::LevelFilter::Info,
         },
         // Default to info
-        Err(_) => log::LevelFilter::Info
+        Err(_) => log::LevelFilter::Info,
     };
 
+    let logging_directory = get_logging_directory()?;
+    // Create the logging directory
+    fs::create_dir_all(&logging_directory)?;
+
+    let xi_log_file = "xi-core.log";
+    let mut logging_file_path = logging_directory;
+    logging_file_path.push(xi_log_file);
     fern::Dispatch::new()
         .format(|out, message, record| {
             out.finish(format_args!(
@@ -44,10 +67,9 @@ fn setup_logging() -> Result<(), fern::InitError> {
                 record.level(),
                 message
             ))
-        })
-        .level(level_filter)
-        .chain(std::io::stderr())
-        .chain(fern::log_file("xi-core.log")?)
+        }).level(level_filter)
+        .chain(io::stderr())
+        .chain(fern::log_file(logging_file_path)?)
         .apply()?;
     Ok(())
 }
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -22,15 +22,15 @@ extern crate dirs;
 extern crate xi_core_lib;
 extern crate xi_rpc;
 
-use std::io;
 use std::fs;
+use std::io;
 use std::path::PathBuf;
 
 use xi_core_lib::XiCore;
 use xi_rpc::RpcLoop;
 
-const XI_DIRECTORY: &'static str = "xi-core";
-const XI_LOG_FILE: &'static str = "xi-core.log";
+const XI_DIRECTORY: &str = "xi-core";
+const XI_LOG_FILE: &str = "xi-core.log";
 
 fn get_logging_directory() -> Option<PathBuf> {
     if let Some(mut log_dir) = dirs::data_local_dir() {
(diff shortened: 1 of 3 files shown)


rust/src/main.rs
```diff
@@ -12,18 +12,34 @@
...
+use std::fs;
+use std::path::PathBuf;
 
 #[macro_use]
 extern crate log;
 extern crate chrono;
 extern crate fern;
 
+extern crate dirs;
+use dirs::data_local_dir;
```
cmyr: rust style is to put `use` statements below `extern crate` statements.

In this case, my personal preference might be to just skip the `use` statement entirely, since we only just this symbol once, in one place; it's not too much bother to just type `dir::data_local_dir()` at the call site.

rust/src/main.rs
```diff
@@ -32,9 +48,16 @@ fn setup_logging() -> Result<(), fern::InitError> {
             _ => log::LevelFilter::Info,
         },
         // Default to info
-        Err(_) => log::LevelFilter::Info
+        Err(_) => log::LevelFilter::Info,
     };
 
+    let logging_directory = get_logging_directory()?;
```
cmyr: I might call this function `path_for_log_file(filename: &str)` (or, better,` path_for_log_file<P: AsRef<Path>>(filename: P)`), and pass it in the name of the file; and then I would move in the `create_dir_all` and the path appending to that function. This would make it easier to eventually share this with another binary.

rust/src/main.rs
```diff
@@ -32,9 +48,16 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+    // Create the logging directory
+    fs::create_dir_all(&logging_directory)?;
```
cmyr: Failure to create the log directory (as with failure to find a log directory) is not reason enough for us to exit; so instead of the `?` here, we should check the error, print a warning with `eprintln`, and then continue to run.

rust/src/main.rs
```diff
@@ -11,19 +11,48 @@
...
 extern crate xi_rpc;
 
+use std::io;
+use std::fs;
+use std::path::PathBuf;
+
 use xi_core_lib::XiCore;
 use xi_rpc::RpcLoop;
 
+const XI_DIRECTORY: &'static str = "xi-core";
```
cmyr: I believe that clippy will warn that the `&'static` is redundant here. You used to need it, but this changed sometime this year.

rust/src/main.rs
```diff
@@ -11,19 +11,48 @@
...
+        log_dir.push(XI_DIRECTORY);
+        Some(log_dir)
+    } else {
+        eprintln!("[WARNING] The dir library was not able to find a directory for this platform");
+        None
+    }
+}
+
+fn path_for_log_file<P: AsRef<str>>(filename: P) -> Option<PathBuf> {
+    if let Some(mut logging_directory) = get_logging_directory() {
```
cmyr: just an observation: you could write this `let mut logging_directory = get_logging_directory()?;`, and the `?` will just return `None` if the result is `None`, and you can continue without the `{ .. }` block.

rust/src/main.rs
```diff
@@ -11,19 +11,48 @@
...
+    if let Some(mut log_dir) = dirs::data_local_dir() {
+        log_dir.push(XI_DIRECTORY);
+        Some(log_dir)
+    } else {
+        eprintln!("[WARNING] The dir library was not able to find a directory for this platform");
+        None
+    }
+}
+
+fn path_for_log_file<P: AsRef<str>>(filename: P) -> Option<PathBuf> {
```
cmyr: although it really shouldn't matter, you should be able to have `AsRef<Path>` instead of `AsRef<str>`. `String` and `&str` both implement `AsRef<Path>`, so you can still call this with string types, but you can also call it with a `Path` or `PathBuf`, and the intent is clearer.

rust/src/main.rs
```diff
@@ -11,19 +11,48 @@
...
+    if let Some(mut logging_directory) = get_logging_directory() {
+        // Create the logging directory
+        match fs::create_dir_all(&logging_directory) {
+            Ok(_) => (), Err(why) => eprintln!("[WARNING] {:?}", why)
+        }
+        logging_directory.push(filename.as_ref());
+        Some(logging_directory)
+    } else { None }
```
cmyr: although I think this is fine, rustfmt will make us put `None` on a newline.

rust/src/main.rs
```diff
@@ -32,23 +61,33 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+        .chain(io::stderr());
+
+    if let Some(logging_file_path) = path_for_log_file(XI_LOG_FILE) {
+        fern_dispatch
+            .chain(fern::log_file(logging_file_path)?)
+            .apply()?;
+    } else {
+        eprintln!("[WARNING] There was an issue getting the path for the log file, using stdout");
+        fern_dispatch
+            .chain(io::stdout())
```
cmyr: we can't chain to `stdout`; it is being piped to the client, which will try and interpret it as RPCs. 😱

if `fern_dispatch` was declared `mut`, you could do this with just,
```rust
// {
// earlier stuff
    if let Some(logging_file_path) = path_for_log_file(XI_LOG_FILE) {
        fern_dispatch = fern_dispatch.chain(..);
    }
    info!("Logging with fern setup");
    fern_dispatch.apply()
}
```

or something similar.

06kellyjac: Aaaaaaahhhhhhh I see. Fixed 😅

rust/src/main.rs
06kellyjac: I've ran `rustfmt` on `main.rs` now.

rust/src/main.rs
06kellyjac: I still need the `&str` as `const` must be annotated with a type but the `'static` lifetime is indeed redundant.
I've ran the fixed version by clippy and didn't have any complaints this time.
`const XI_DIRECTORY: &str = "xi-core";`

rust/src/main.rs
```diff
@@ -79,12 +76,15 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
-            "[WARNING] There was an issue getting the path for the log file, just using stderr"
-        );
+    match path_for_log_file(XI_LOG_FILE) {
+        Ok(logging_file_path) => {
+            fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?)
+        }
+        Err(e) => {
+            let error_message =
+                "[WARNING] There was an issue getting the path for the log file, falling back to stderr.";
+            eprintln!("{}\nError:\n{:?}", error_message, e)
```
jansol: This could probably be reordered a bit so it finalizes the (stderr-only) logger and *then* prints the error message using the appropriate macro.

rust/src/main.rs
06kellyjac: I've given this a go & swapped the Options back to Results as I wanted it to respond with an io Not Found error.
I've tried forcing `get_logging_directory` to always respond with the error; it was caught fine (by the match on line 79) & didn't cause any exit or crash but still gives some valuable info.

rust/src/main.rs
06kellyjac: I looked into using `PathBuf` instead of `str` but:
> the trait `std::convert::AsRef<std::path::PathBuf>` is not implemented for `str`

Using `Path` worked fine though.

rust/src/main.rs
06kellyjac: Yeah I was half and half on whether to do the error logging before or after starting fern.

I've given it a go. Please tell me if there's a cleaner way to do the `apply()`s and then log the error. 🙂

jansol: You could get creative with if let:
```
let p = path_for_log_file(XI_LOG_FILE);
if let Ok(logging_file_path) = p {
    fern_dispatch.chain(fern::log_file(logging_file_path)?);
}
fern_dispatch.apply()?;
if let Err(e) = p {
    warn!("There was an issue getting the path for the log file: {}, falling back to stderr.", e);
}
```

I think that's about as linear as you can get.

06kellyjac: Mmmmm. I wrote it out but idk if it looks cleaner:

Old:

```rust
let fern_dispatch = fern::Dispatch::new().etc.etc.....

match path_for_log_file(XI_LOG_FILE) {
	Ok(logging_file_path) => fern_dispatch
		.chain(fern::log_file(logging_file_path)?)
		.apply()?,
	Err(err) => {
		fern_dispatch.apply()?;
		warn!(
			"There was an issue getting the path for the log file: {}, falling back to stderr.",
			err,
		);
	}
}
info!("Logging with fern is setup");
Ok(())
```

New:

```rust
let mut fern_dispatch = fern::Dispatch::new().etc.etc.....

let p = path_for_log_file(XI_LOG_FILE);
if let Ok(logging_file_path) = &p {
	fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
}
fern_dispatch.apply()?;
if let Err(e) = &p {
	warn!(
		"There was an issue getting the path for the log file: {:?}, falling back to stderr.",
		e,
	);
}
info!("Logging with fern is setup");
Ok(())
```
🤷‍♂️ 

I think personally I lean towards the match statement. Which do you prefer? 😃

06kellyjac: Yeah I get you.

I tried adding some comments and spacing and it looks much more readable than it was before & is DRY.

```rust
let mut fern_dispatch = fern::Dispatch::new().etc.etc.....

let path_result = path_for_log_file(XI_LOG_FILE);

// If the logging_file_path returned successfully, add the logfile capability to fern
if let Ok(logging_file_path) = &path_result {
	fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
}

// Start fern
fern_dispatch.apply()?;

// If the logging_file_path returned an error, print it with fern/logs
if let Err(e) = &path_result {
	warn!(
		"There was an issue getting the path for the log file: {:?}, falling back to stderr.",
		e,
	);
}

info!("Logging with fern is setup");
Ok(())
```

Good balance of both worlds?

06kellyjac: I've pushed this up (with a slight change to `warn!()` to make it a bit cleaner IMO).

rust/src/main.rs
```diff
@@ -11,19 +11,48 @@
...
+
+fn get_logging_directory() -> Result<PathBuf, io::Error> {
+    match dirs::data_local_dir() {
+        Some(mut log_dir) => {
+            log_dir.push(XI_DIRECTORY);
+            Ok(log_dir)
+        }
+        None => Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "The dir library was not able to find a directory for this platform",
```
jansol: Nit: the wording here is quite vague, something like "No standard logging directory known for this platform" would be more descriptive I think? Mentioning the library is not necessary either.

That aside, nice job!

06kellyjac: You're right. I'd thought I'd at least added "logging directory"; good spot.
Also yea I guess knowing the library is a bit useless. Very easy to check where the log came from & see the library from there.
Changes pushed.

jansol:
> as I can tell we don't actually know what path we're using for the logging. Maybe we would want to log this as well?

I'm not sure that is useful considering that you need to know it in order to see the log to begin with

cmyr:
Yea, but at least you would also see it in stderr, which would tell you where it is in the future? Like I just literally do not know what directory macOS is going to return for this (`/var/log`? `~/Library/Logs`)? 

In any case this isn't a blocker, once we have the `--log-dir` flag this becomes a non-issue. 👍

06kellyjac:
Would this be better added in as just Documentation?

I can add some in-code documentation with `///`.
I can also add some details in the docs folder of this project but I'm not sure which file to put it in...

jansol:
More documentation is almost always good. If you have ideas for improving it, feel free to apply them in your next PR. (or a separate one if that feels more appropriate)

PR #888, opened 2018-10-05
Adding functionality to change the default logging directory.

06kellyjac:
## Summary
Add a way to override the default locations for the log file. This can be useful for frontend's that use the `xi-core` and want all the logs in one place.  
This also allows for users to use multiple frontends that keep their `xi-core` logging separate.  
For example: If I was to use `xi-mac` and `xi-tui` with the `xi-core` (on a Mac) as it is now they would both be logging to the same file (`/Users/MyUserName/Library/Application Support/xi-core/xi-core.log`) which can be a pain when debugging to name just one issue.

This PR deals strictly with adding the ability to change the log file name and/or directory. A future PR will deal with extending this to the `xi-lsp-plugin.log` file which `xi-core` also writes to.

## Related Issues

Relates to #836

## Checklist

- [X] Can choose a different **directory name** for the log file
  - Default name: `xi-core`
  - Example:
    - [X] `xi-core --log-dir cool`
    - [X] `XI_LOG_DIR="cool" xi-core`
    - `XI_LOG_DIR="uncool" xi-core --log-dir cool`
  - Resulting location:
    - Linux/Redox: `/home/username123/.local/share/cool/xi-core.log`
    - Mac: `/User/username123/Application Support/cool/xi-core.log`
    - Windows:`C:\Users\username123\AppData\local\cool\xi-core.log`
- [X] Can choose a different **file name** for the log file
  - Default name: `xi-core.log`
  - Example:
    - [X] `xi-core --log-file amazing.log`
    - [X] `XI_LOG_FILE="amazing.log" xi-core`
    - `XI_LOG_FILE="unamazing.log" xi-core --log-file amazing.log`
  - Resulting location:
    - Linux/Redox: `/home/username123/.local/share/xi-core/amazing.log`
    - Mac: `/User/username123/Application Support/xi-core/amazing.log`
    - Windows:`C:\Users\username123\AppData\local\xi-core\amazing.log`
- See comment below.    ~*Optional* - Can completely override the path (maybe with the flag `--log-dir` or a new flag called `--log-path`)~
  - ~Example: `xi-core --log-path /User/username123/some_dir/other_dir`~
  - ~Resulting location: `/User/username123/some_dir/other_dir/xi-core.log`~

## Review Checklist

### DIFF
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -105,6 +105,12 @@ fn generate_logging_path(logfile_config: LogfileConfig) -> Result<PathBuf, io::E
         Some(file_name) => file_name,
         None => PathBuf::from(XI_LOG_FILE),
     };
+    if logfile_file_name.eq(Path::new("")) {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "A blank file name was supplied",
+        ));
+    };
     // Use the directory name set in logfile_config or fallback to the default
     let logfile_directory_name = match logfile_config.directory {
         Some(dir) => dir,
@@ -208,7 +214,7 @@ fn main() {
         );
     }
     if let Some(e) = logging_error {
-        warn!("Unable to successfully generate the logging path: {:?}", e)
+        warn!("Unable to successfully generate the logging path: {}", e)
     }
 
     match rpc_looper.mainloop(|| stdin.lock(), &mut state) {
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -39,12 +39,10 @@ fn get_logging_directory_path<P: AsRef<Path>>(directory: P) -> Result<PathBuf, i
             log_dir.push(directory);
             Ok(log_dir)
         }
-        None => Err(
-            io::Error::new(
-                io::ErrorKind::NotFound,
-                "No standard logging directory known for this platform",
-            )
-        ),
+        None => Err(io::Error::new(
+            io::ErrorKind::NotFound,
+            "No standard logging directory known for this platform",
+        )),
     }
 }
 
@@ -73,14 +71,13 @@ fn setup_logging(logging_path_result: Result<PathBuf, io::Error>) -> Result<(), 
 
     if let Ok(logging_file_path) = &logging_path_result {
         // Ensure the logging directory is created
-        let parent_path = logging_file_path.parent().ok_or(
-            io::Error::new(
-                io::ErrorKind::InvalidInput,
-                format!(
-                    "Unable to get the parent of the following Path: {}", logging_file_path.display(),
-                )
-            )
-        )?;
+        let parent_path = logging_file_path.parent().ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "Unable to get the parent of the following Path: {}",
+                logging_file_path.display(),
+            ),
+        ))?;
         fs::create_dir_all(parent_path)?;
         // Attach it to fern
         fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
@@ -98,7 +95,10 @@ fn setup_logging(logging_path_result: Result<PathBuf, io::Error>) -> Result<(), 
             warn!("{}: {:?}, falling back to stderr.", message, e);
         }
         Ok(logging_file_path) => {
-            info!("Logging to the following file: {}", logging_file_path.display());
+            info!(
+                "Logging to the following file: {}",
+                logging_file_path.display()
+            );
         }
     }
     Ok(())
@@ -139,7 +139,9 @@ fn get_flags() -> HashMap<String, Option<String>> {
 
             // Check the next argument doesn't start with the flag prefix
             // map_or accounts for peek returning an Option
-            let next_arg_not_a_flag: bool = args_itterator.peek().map_or(false, |val| !val.starts_with(flag_prefix));
+            let next_arg_not_a_flag: bool = args_itterator
+                .peek()
+                .map_or(false, |val| !val.starts_with(flag_prefix));
             if next_arg_not_a_flag {
                 flags.insert(key, args_itterator.next());
             }
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -62,19 +62,19 @@ fn get_logging_directory_path<P: AsRef<Path>>(directory: P) -> Result<PathBuf, i
 /// use std::path::PathBuf;
 /// let path_with_file = PathBuf::from("/some/directory/then/file");
 /// assert_eq!(Some(OsStr::new("file")), path_with_file.file_name());
-/// assert_eq!(create_parent_directory(path_with_file).is_ok(), true);
+/// assert_eq!(create_log_directory(path_with_file).is_ok(), true);
 ///
 /// let path_with_other_file = PathBuf::from("/other_file");
 /// assert_eq!(Some(OsStr::new("other_file")), path_with_other_file.file_name());
-/// assert_eq!(create_parent_directory(path_with_file).is_ok(), true);
+/// assert_eq!(create_log_directory(path_with_file).is_ok(), true);
 ///
 /// // Path that is just the root or prefix:
 /// let path_without_file = PathBuf::from("/");
 /// assert_eq!(None, path_with_file.file_name());
-/// assert_eq!(create_parent_directory(path_with_file).is_ok(), false);
+/// assert_eq!(create_log_directory(path_with_file).is_ok(), false);
 /// ```
-pub fn create_parent_directory(path_with_file: &PathBuf) -> Result<(), io::Error> {
-    let parent_path = path_with_file.parent().ok_or(io::Error::new(
+fn create_log_directory(path_with_file: &PathBuf) -> Result<(), io::Error> {
+    let log_dir = path_with_file.parent().ok_or(io::Error::new(
         io::ErrorKind::InvalidInput,
         format!(
             "Unable to get the parent of the following Path: {}",
@@ -82,7 +82,7 @@ pub fn create_parent_directory(path_with_file: &PathBuf) -> Result<(), io::Error
         ),
     ))?;
     // Try to create the directory.
-    fs::create_dir_all(parent_path)?;
+    fs::create_dir_all(log_dir)?;
     Ok(())
 }
 
@@ -111,7 +111,7 @@ fn setup_logging(logging_path_result: Option<PathBuf>) -> Result<(), fern::InitE
 
     if let Some(logging_file_path) = &logging_path_result {
         // Try to create the parent directories for the logging file in the logging file path
-        create_parent_directory(&logging_file_path)?;
+        create_log_directory(&logging_file_path)?;
 
         // Attach it to fern
         fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -81,7 +81,6 @@ fn create_log_directory(path_with_file: &PathBuf) -> Result<(), io::Error> {
             path_with_file.display(),
         ),
     ))?;
-    // Try to create the directory.
     fs::create_dir_all(log_dir)?;
     Ok(())
 }
@@ -110,10 +109,8 @@ fn setup_logging(logging_path: Option<PathBuf>) -> Result<(), fern::InitError> {
         .chain(io::stderr());
 
     if let Some(logging_file_path) = &logging_path {
-        // Try to create the parent directories for the logging file in the logging file path
         create_log_directory(&logging_file_path)?;
 
-        // Attach it to fern
         fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
     };
 
@@ -226,11 +223,11 @@ fn main() {
     let stdout = io::stdout();
     let mut rpc_looper = RpcLoop::new(stdout);
 
-    let flags: HashMap<String, Option<String>> = get_flags();
+    let flags = get_flags();
 
     let logfile_config = generate_logfile_config(&flags);
 
-    let logging_path_result: Result<PathBuf, io::Error> = generate_logging_path(logfile_config);
+    let logging_path_result = generate_logging_path(logfile_config);
 
     let mut logging_path: Option<PathBuf> = None;
     let mut logging_error: Option<io::Error> = None;
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -102,6 +102,30 @@ fn setup_logging() -> Result<(), fern::InitError> {
     Ok(())
 }
 
+fn prepare_logging_path(logfile_config: LogfileConfig) -> Result<PathBuf, io::Error> {
+    // Use the file name set in logfile_config or fallback to the default
+    let logfile_file_name = match logfile_config.file {
+        Some(file_name) => file_name,
+        None => XI_LOG_FILE.to_string(),
+    };
+    // Use the directory name set in logfile_config or fallback to the default
+    let logfile_directory_name = match logfile_config.directory {
+        Some(dir) => dir,
+        None => XI_LOG_DIR.to_string(),
+    };
+
+    let mut logging_directory_path = get_logging_directory_path(logfile_directory_name)?;
+
+    // Add the file name & return the full path
+    logging_directory_path.push(logfile_file_name);
+    Ok(logging_directory_path)
+}
+
+struct LogfileConfig {
+    directory: Option<String>,
+    file: Option<String>,
+}
+
 fn get_flags() -> HashMap<String, Option<String>> {
     let mut flags: HashMap<String, Option<String>> = HashMap::new();
 
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -152,8 +152,13 @@ fn get_flags() -> HashMap<String, Option<String>> {
 
 fn generate_logfile_config(flags: HashMap<String, Option<String>>) -> LogfileConfig {
     // If the key is set, get the Option within
-    let log_dir_flag_option = flags.get("log-dir").cloned().unwrap_or(None);
-    let log_file_flag_option = flags.get("log-file").cloned().unwrap_or(None);
+    let log_dir_flag_option = std::env::var("XI_LOG_DIR")
+        .ok()
+        .or(flags.get("log-dir").cloned().unwrap_or(None));
+
+    let log_file_flag_option = std::env::var("XI_LOG_FILE")
+        .ok()
+        .or(flags.get("log-file").cloned().unwrap_or(None));
 
     LogfileConfig {
         directory: log_dir_flag_option,
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -22,6 +22,7 @@ extern crate dirs;
 extern crate xi_core_lib;
 extern crate xi_rpc;
 
+use std::collections::HashMap;
 use std::fs;
 use std::io;
 use std::path::{Path, PathBuf};
@@ -111,6 +112,26 @@ fn setup_logging() -> Result<(), fern::InitError> {
 
     info!("Logging with fern is setup");
     Ok(())
+}
+
+fn get_flags() -> HashMap<String, Option<String>> {
+    let mut flags: HashMap<String, Option<String>> = HashMap::new();
+
+    let flag_prefix = "-";
+    let mut args_itterator = std::env::args().peekable();
+    while let Some(arg) = args_itterator.next() {
+        if arg.starts_with(flag_prefix) {
+            let key = arg.trim_left_matches(flag_prefix).to_string();
+
+            // Check the next argument doesn't start with the flag prefix
+            // map_or accounts for peek returning an Option
+            let next_arg_not_a_flag: bool = args_itterator.peek().map_or(false, |val| !val.starts_with(flag_prefix));
+            if next_arg_not_a_flag {
+                flags.insert(key, args_itterator.next());
+            }
+        }
+    }
+    flags
 }
 
 fn main() {
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -90,15 +90,15 @@ fn setup_logging(logging_path_result: Result<PathBuf, io::Error>) -> Result<(), 
     // Log details of the logging_file_path result using fern/log
     // Either logging the path fern is outputting to or the error from obtaining the path
     match &logging_path_result {
-        Err(e) => {
-            let message = "There was an issue getting the path for the log file";
-            warn!("{}: {:?}, falling back to stderr.", message, e);
-        }
         Ok(logging_file_path) => {
             info!(
                 "Logging to the following file: {}",
                 logging_file_path.display()
             );
+        }
+        Err(e) => {
+            let message = "There was an issue getting the path for the log file";
+            warn!("{}: {:?}, falling back to stderr.", message, e);
         }
     }
     Ok(())
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -152,6 +152,17 @@ fn main() {
     let stdout = io::stdout();
     let mut rpc_looper = RpcLoop::new(stdout);
 
+    let flags: HashMap<String, Option<String>> = get_flags();
+
+    // If the key is set, get the Option within
+    let log_dir_flag_option = flags.get("log-dir").cloned().unwrap_or(None);
+    let log_file_flag_option = flags.get("log-file").cloned().unwrap_or(None);
+
+    let logfile_config = LogfileConfig {
+        directory: log_dir_flag_option,
+        file: log_file_flag_option,
+    };
+
     if let Err(e) = setup_logging() {
         eprintln!(
             "[ERROR] setup_logging returned error, logging disabled: {:?}",
--- a/rust/src/main.rs
+++ b/rust/src/main.rs
@@ -57,7 +57,7 @@ fn path_for_log_file<P: AsRef<Path>>(filename: P) -> Result<PathBuf, io::Error> 
     Ok(logging_directory)
 }
 
-fn setup_logging() -> Result<(), fern::InitError> {
+fn setup_logging(logging_path_result: Result<PathBuf, io::Error>) -> Result<(), fern::InitError> {
     let level_filter = match std::env::var("XI_LOG") {
         Ok(level) => match level.to_lowercase().as_ref() {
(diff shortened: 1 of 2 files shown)


rust/src/main.rs
```diff
@@ -29,13 +29,27 @@ use std::path::{Path, PathBuf};
 use xi_core_lib::XiCore;
 use xi_rpc::RpcLoop;
 
-const XI_DIRECTORY: &str = "xi-core";
+const XI_LOG_DIR: &str = "xi-core";
 const XI_LOG_FILE: &str = "xi-core.log";
 
-fn get_logging_directory() -> Result<PathBuf, io::Error> {
+fn get_logfile_directory_name() -> String {
+    match std::env::var("XI_LOG_DIR") {
```
cmyr: so we've moved away from using env vars for stuff like this; for instance we used to do this for config files (and might still support it) but now we set those directories through the RPC system.

We can't do that here, but I think it makes more sense to just pass args to the executable directly, unless you have a particular reason for preferring env vars?

rust/src/main.rs
```diff
@@ -29,13 +29,27 @@ use std::path::{Path, PathBuf};
...
+        Ok(name) => name,
+        Err(_) => String::from(XI_LOG_DIR),
+    }
+}
+
+fn get_logfile_file_name() -> String {
```
cmyr: both this and the previous fn could return `PathBuf` instead of `String`. These fns also mean we're always allocating, even when we're just using the statics, but we probably don't need to sweat a few tens of ns here just yet, and this is nice and clear.

06kellyjac:
Both envs and args supported. If both are supplied envs take priority.

I'm thinking the full path override can be a different Issue/PR if it's *actually needed* at any point as it would probably take some tweaking and we'd have to decide in the (unlikely) event of `XI_LOG_DIR xi-core --log-path /Some/path/here` if the `XI_LOG_DIR` env would take priority over the flag ... etc ...

---

Please be super nitpick-ie as knowing everything I've done wrong will help my Rust learning.
😃

06kellyjac:
TODO: Can probably swap some Strings to PathBufs in places.

rust/src/main.rs
```diff
@@ -45,15 +46,7 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
...
-fn path_for_log_file<P: AsRef<Path>>(filename: P) -> Result<PathBuf, io::Error> {
-    let mut logging_directory = get_logging_directory()?;
-    // Create the logging directory
-    fs::create_dir_all(&logging_directory)?;
-    logging_directory.push(filename.as_ref());
-    Ok(logging_directory)
-}
-
-fn setup_logging() -> Result<(), fern::InitError> {
+fn setup_logging(logging_path_result: Result<PathBuf, io::Error>) -> Result<(), fern::InitError> {
```
jansol: I'm pretty sure the path should be completely known by this point, which means that this fn should take an Option<&Path> instead of a Result

rust/src/main.rs
```diff
@@ -76,32 +69,130 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+    if let Ok(logging_file_path) = &logging_path_result {
+        // Ensure the logging directory is created
+        let parent_path = logging_file_path.parent().ok_or(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            format!(
+                "Unable to get the parent of the following Path: {}",
+                logging_file_path.display(),
+            ),
+        ))?;
+        fs::create_dir_all(parent_path)?;
```
jansol: This recursively creates *all* missing directories so erroring if the parent is missing just above this seems confusing. I guess it would make sense on windows where you might be logging to an external drive that goes missing, but in the more common case `path/with/nonexisting/directories` missing intermediate directories should not be an error IMHO.

I'd like at least some clarification to comments and the error message. In particular a mention that the path here already contains the filename, because otherwise this becomes very confusing (erroring if someone wants to use `/` as their log dir is just weird).

rust/src/main.rs
06kellyjac: The reason I send through the PathBuf or the io:Error is the Error can be logged cleanly with fern on line 99 but I can remove it & do a normal eprintln if that's preferred.

rust/src/main.rs
```diff
@@ -76,32 +69,130 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+
+struct EnvFlagConfig {
+    env_name: &'static str,
+    flag_name: &'static str,
+}
+
+fn extract_env_or_flag(flags: &HashMap<String, Option<String>>, conf: EnvFlagConfig) -> Option<String> {
+    std::env::var(conf.env_name)
+        .ok()
+        .or(flags.get(conf.flag_name).cloned().unwrap_or(None))
```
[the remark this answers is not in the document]
06kellyjac: I can swap the precedence but if we do we probably want to document that behavior very clearly as I can't think of any programs off the top of my head that take cli args over env variables.

rust/src/main.rs
06kellyjac: ### My thought process:

Well I moved the `fs::create_dir_all()` to right next to the `fern::log_file()` bit as it ensures that we've tried to make sure  there's a directory there before trying to write a file inside.
As it was before any old PathBuf could have been passed in and there was no guarantee that we'd tried to make the required directories.

As we're sending in a PathBuf that looks something like the following: `/home/username123/.local/share/xi-core/xi-core.log` and `fs::create_dir_all()` just wants the directory bit.
```
/home/username123/.local/share/xi-core/xi-core.log
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

So 'PathBuf`s `.parent()` will get that bit **but** will return None if the PathBuf is just `/` or a "prefix".
I can't give `fs::create_dir_all()` a `None`.
I turn the option into a `Result` so I can use `?`/`try!`

**Note:** Someone can still have it log out to `/` they just can't try log *directly* to `/`. Fern wants a location for the log *file* to go and as a result the `PathBuf` *must* be something like `/xi-core.log` to log to the `/` directory

I could do something like the following:
```
match logging_file_path.parent() {
    Some(parent_path) => {
            fs::create_dir_all(parent_path)?;
            // Attach it to fern
            fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
    }
        None => ()
}
```
Or
```
match logging_file_path.parent() {
    Some(parent_path) => {
            fs::create_dir_all(parent_path)?;
    }
        None => ()
}
// Attach it to fern
fern_dispatch = fern_dispatch.chain(fern::log_file(logging_file_path)?);
```
But then I don't have an Err any more and it will allow people to put a PathBuf of `/` into `fern::logfile` which will *always* fail.

So I could have just done what I did with the conversion to a Result & then using `?`/`try!`.


---

I do agree that I should add some comments to explain what's going on.
If you have any questions or suggestions for the comments I'm also on irc.

rust/src/main.rs
06kellyjac: Ok, I'll go swap it to an `eprintln` & `Option`

rust/src/main.rs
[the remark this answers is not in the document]
06kellyjac: Done

rust/src/main.rs
06kellyjac: Done

jansol: I was thinking that the caller could hold on to the Err and log it *after* calling this function.

06kellyjac: Ahh that would be better.

06kellyjac: Sorted

rust/src/main.rs
```diff
@@ -76,37 +69,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
-    info!("Logging with fern is setup");
     Ok(())
 }
 
+fn generate_logging_path(logfile_config: LogfileConfig) -> Result<PathBuf, io::Error> {
+    // Use the file name set in logfile_config or fallback to the default
+    let logfile_file_name = match logfile_config.file {
+        Some(file_name) => file_name,
+        None => PathBuf::from(XI_LOG_FILE),
+    };
```
jansol: This could have a fallback for the case of an empty filename. Or disable file logging in that case? I guess that would be more useful.

rust/src/main.rs
```diff
@@ -76,37 +69,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+    // Use the directory name set in logfile_config or fallback to the default
+    let logfile_directory_name = match logfile_config.directory {
+        Some(dir) => dir,
+        None => PathBuf::from(XI_LOG_DIR),
+    };
```
jansol: Explicitly specifying an empty string would currently result in logging directly to `~/.local/` I think? Not sure that is the most intuitive thing to do... But I'm not sure what would be better either. Maybe default to the current working directory in that case?

rust/src/main.rs
```diff
@@ -76,37 +69,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+
+    let flag_prefix = "-";
+    let mut args_itterator = std::env::args().peekable();
+    while let Some(arg) = args_itterator.next() {
+        if arg.starts_with(flag_prefix) {
+            let key = arg.trim_left_matches(flag_prefix).to_string();
+
+            // Check the next argument doesn't start with the flag prefix
+            // map_or accounts for peek returning an Option
+            let next_arg_not_a_flag: bool = args_itterator
```
jansol: This has ttoo many tts. ;)

rust/src/main.rs
```diff
@@ -76,37 +69,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+
+    let mut logging_path: Option<PathBuf> = None;
+    let mut logging_error: Option<io::Error> = None;
+    match logging_path_result {
+        Ok(val) => logging_path = Some(val),
+        Err(e) => logging_error = Some(e),
+    };
+    if let Err(e) = setup_logging(logging_path) {
         eprintln!(
             "[ERROR] setup_logging returned error, logging disabled: {:?}",
```
jansol: "not enabled" would be more accurate.

rust/src/main.rs
```diff
@@ -76,37 +69,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+    // Use the directory name set in logfile_config or fallback to the default
+    let logfile_directory_name = match logfile_config.directory {
+        Some(dir) => dir,
+        None => PathBuf::from(XI_LOG_DIR),
+    };
+
+    let mut logging_directory_path = get_logging_directory_path(logfile_directory_name)?;
+
+    // Add the file name & return the full path
+    logging_directory_path.push(logfile_file_name);
```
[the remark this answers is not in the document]
jansol: Nevermind, `push` might be better after all -- otherwise we might end up accidentally replacing the last directory.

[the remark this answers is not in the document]
06kellyjac: Yea, that would be nice semantically but sadly it doesn't really look very good as I'd still have to do a `.push()`.

Without push:

```
let mut directory = PathBuf::from("/a_dir/another_dir");
assert_eq!(directory.to_str(), Some("/a_dir/another_dir"));
directory.set_file_name("filename");
assert_eq!(directory.to_str(), Some("/a_dir/filename"));
```

With push:

```
let mut directory = PathBuf::from("/a_dir/another_dir");
assert_eq!(directory.to_str(), Some("/a_dir/another_dir"));
directory.push("temporary");
assert_eq!(directory.to_str(), Some("/a_dir/another_dir/temporary"));
directory.set_file_name("filename");
assert_eq!(directory.to_str(), Some("/a_dir/another_dir/filename"));
```

rust/src/main.rs
06kellyjac: I ran it a couple ways and the current behavior is (for XDG style systems like Linux & Redox):
- `xi-core --log-dir` -> `/home/username/.local/share/xi-core/xi-core.log`
- `xi-core --log-dir --other-flag` -> `/home/username/.local/share/xi-core/xi-core.log`
- `xi-core --log-dir  --other-flag` -> `/home/username/.local/share/xi-core/xi-core.log` (2 spaces between the flags, doesn't seem to want to show up)
- `xi-core --log-dir ""` -> `/home/username/.local/share/xi-core.log`
- `xi-core --log-dir " "` -> `/home/username/.local/share/ /xi-core.log`
- `xi-core --log-dir something` -> `/home/username/.local/share/something/xi-core.log`
- `xi-core --log-dir "something"` -> `/home/username/.local/share/something/xi-core.log`

In my opinion those are as I'd expect them to be.

As for defaulting to the current directory, I was under the impression falling back to the current directory was part of the reason for this work as writing to within the `XiEditor.app/Contents/MacOS` folder was causing issues.
Also it's pretty common behavior for programs to go ahead and write logs off in.

Also assuming `/some/directory` was the working dir it's still easy to do with `$PWD` or an absolute path:

- `xi-core --log-dir .` -> `/home/username/.local/share/./xi-core.log`
- `xi-core --log-dir "."` -> `/home/username/.local/share/./xi-core.log`
- `xi-core --log-dir $PWD` -> `/some/directory/xi-core.log`
- `xi-core --log-dir /some/directory` -> `/some/directory/xi-core.log`
- `xi-core --log-dir "/some/directory"` -> `/some/directory/xi-core.log`
- `xi-core --log-dir ~/directory` -> `/home/username/directory/xi-core.log`
- `xi-core --log-dir "~/directory"` -> `/home/username/.local/share/~/directory/xi-core.log`

The only behavior that's a bit odd/annoying to me is:
Taking `"~/directory"` (in quotes) literally is quite normal cli behavior but `"/some/directory"` (in quotes) isn't taken in a similar way and results in `/some/directory/xi-core.log` instead of `/home/username/.local/share//some/directory/xi-core.log` (which would be simplified to `/home/username/.local/share/some/directory/xi-core.log`) as it starts with `/` and `PathBuf` sees that as a reason to replace the `PathBuf` with an absolute path.
I'd rather that be one way or the other.

rust/src/main.rs
06kellyjac: Done 👍

rust/src/main.rs
06kellyjac: Done 👍

rust/src/main.rs
06kellyjac: I'd probably lean towards the fallback & then have `--log-disable` or something as a flag for disabling logging.
But yea I do need to account for `" "`.

rust/src/main.rs
06kellyjac: I swear every hour I swap between flavoring `if let` over `match` when that'd probably be better and vice versa >.<

As the code is now it `setup_logging` will never recive a `PathBuf` of just `/` but if the code is changed `setup_logging` could recive a `PathBuf` of just `/` so that code is to handle that.

I split it into a separate function wrote some decent docs for it but I can't find them now 😭

rust/src/main.rs
06kellyjac: I tried `echo "~"` on `zsh` and `bash` on my machine and it didn't expand. `echo "$HOME"` and `echo ~` expand though.

06kellyjac: I think as it is now it all behaves in a predictable manner.
We can document it's behavior nice and clearly somewhere and the logging to stderr that says where the logfile is can be handy.


We could also add at some point a nice way for Frontends (& plugins) to see where the logfile is & show the user.
I can imagine bringing up a command pallet in a frontend and running a command called `Show logfile` or `Open logfile` or `Open logfile in file browser` or something.

rust/src/main.rs
```diff
@@ -45,15 +46,47 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
...
+/// // Path that is just the root or prefix:
+/// let path_without_file = PathBuf::from("/");
+/// assert_eq!(None, path_with_file.file_name());
+/// assert_eq!(create_log_directory(path_with_file).is_ok(), false);
+/// ```
+fn create_log_directory(path_with_file: &PathBuf) -> Result<(), io::Error> {
+    let log_dir = path_with_file.parent().ok_or(io::Error::new(
+        io::ErrorKind::InvalidInput,
+        format!(
+            "Unable to get the parent of the following Path: {}",
```
jansol: I'd mention that the path should contain a filename.

rust/src/main.rs
```diff
@@ -45,15 +46,47 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
...
+            path_with_file.display(),
+        ),
+    ))?;
+    // Try to create the directory.
+    fs::create_dir_all(log_dir)?;
+    Ok(())
 }
 
-fn setup_logging() -> Result<(), fern::InitError> {
+fn setup_logging(logging_path_result: Option<PathBuf>) -> Result<(), fern::InitError> {
```
jansol: Just `logfile_path` is fine now that it's an Option.

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
 
-    // If the logging_file_path returned an error, print it with fern/logs
-    if let Err(e) = &path_result {
-        let message = "There was an issue getting the path for the log file";
-        warn!("{}: {:?}, falling back to stderr.", message, e);
+    // Log details of the logging_file_path result using fern/log
+    // Either logging the path fern is outputting to or the error from obtaining the path
+    match &logging_path_result {
+        Some(logging_file_path) => info!(
+            "Logging to the following file: {}",
```
jansol: I'd prefer something like "Writing logs to {}". Sounds less stiff and is shorter.

06kellyjac: Done 👍

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+            logging_file_path.display()
+        ),
+        None => warn!("There was no path supplied for the log file, falling back to stderr."),
```
jansol: "No path supplied for the log file, ..." or "No path was supplied for the log file, ..." sounds better IMHO.
"Log file without a name requested; only logging to stderr" might be more descriptive though.

rust/src/main.rs
06kellyjac: Good spot. Done.

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+    };
+    if let Err(e) = setup_logging(logging_path) {
         eprintln!(
-            "[ERROR] setup_logging returned error, logging disabled: {:?}",
+            "[ERROR] setup_logging returned error, logging not enabled: {:?}",
             e
         );
     }
+    if let Some(e) = logging_error {
+        warn!("Unable to successfully generate the logging path: {}", e)
```
jansol: "successfully" is redundant.

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
 
-    // If the logging_file_path returned an error, print it with fern/logs
-    if let Err(e) = &path_result {
-        let message = "There was an issue getting the path for the log file";
-        warn!("{}: {:?}, falling back to stderr.", message, e);
+    // Log details of the logging_file_path result using fern/log
+    // Either logging the path fern is outputting to or the error from obtaining the path
+    match &logging_path {
+        Some(logging_file_path) => info!("Writing logs to: {}", logging_file_path.display()),
+        None => warn!("No path was supplied for the log file, not saving logs to disk, falling back to just stderr"),
```
jansol: This would be easier to read as two sentences I think. "No path was supplied for the log file. Not saving logs to disk, only writing to stderr."

06kellyjac: Done :)

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+}
+
 fn main() {
     let mut state = XiCore::new();
     let stdin = io::stdin();
     let stdout = io::stdout();
     let mut rpc_looper = RpcLoop::new(stdout);
 
-    if let Err(e) = setup_logging() {
+    let flags: HashMap<String, Option<String>> = get_flags();
```
cmyr: you can get away without all of these type declarations I suspect.

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
...
+
+    let logfile_config = generate_logfile_config(&flags);
+
+    let logging_path_result: Result<PathBuf, io::Error> = generate_logging_path(logfile_config);
+
+    let mut logging_path: Option<PathBuf> = None;
```
cmyr: this is confusing to me; it seems like either only one of these can be `None`, so why can't we keep this a `Result<PathBuf, io::Error>`? I suspect there's a reason but it sure looks odd.

If we _do_ need these to be a pair of optionals, it might be cleaner to match directly on `generate_logging_path(..)`

*some time has passed*

okay _still_ trying to figure out what's going on here. It sure looks like,

```rust
let log_path = generate_logging_path(logfile_config);
if let Err(err) = log_path.as_ref() {
    warn!("Failed to get log file path: '{}'", err);
}
if let Err(err) = setup_logging(log_path.ok()) {
    eprintln!("[Error] logging setup failed: {:?}, err);
}
```

Will do what you want, and is quite a bit (to me) clearer?

rust/src/main.rs
```diff
@@ -45,15 +46,47 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
...
+/// ```
+fn create_log_directory(path_with_file: &PathBuf) -> Result<(), io::Error> {
+    let log_dir = path_with_file.parent().ok_or(io::Error::new(
+        io::ErrorKind::InvalidInput,
+        format!(
+            "Unable to get the parent of the following Path: {}, Your path should contain a file name",
+            path_with_file.display(),
+        ),
+    ))?;
+    // Try to create the directory.
```
cmyr: At the risk of being too nitty, an observation I have in a few places in this PR is that it might be slightly _over_ commented. I appreciate that you're trying to write 'good code', and that's an excellent thing to aspire to, but it's easy to get a little overzealous sometimes.

I think comments are important, but the one case where I like to discourage them is when reading the comment is actually _slower_ than reading the code. In this instance, the name of the function tells us all we need to know, and the `?` tells us this is failable.

In any case, I would normally not mention this but you've been pretty receptive to the review process so far, and I think this advice might be helpful. This is not a blocking problem by any means, I'm not asking for a change, just explaining what I see. 😎

rust/src/main.rs
```diff
@@ -45,15 +46,47 @@ fn get_logging_directory() -> Result<PathBuf, io::Error> {
     }
 }
 
-fn path_for_log_file<P: AsRef<Path>>(filename: P) -> Result<PathBuf, io::Error> {
-    let mut logging_directory = get_logging_directory()?;
-    // Create the logging directory
-    fs::create_dir_all(&logging_directory)?;
-    logging_directory.push(filename.as_ref());
-    Ok(logging_directory)
+/// This function tries to create the parent directories for a file
```
cmyr: Nice, I appreciate doc comments.

rust/src/main.rs
```diff
@@ -76,37 +109,147 @@ fn setup_logging() -> Result<(), fern::InitError> {
         }).level(level_filter)
         .chain(io::stderr());
 
-    let path_result = path_for_log_file(XI_LOG_FILE);
-    // If the logging_file_path returned successfully, add the logfile capability to fern
-    if let Ok(logging_file_path) = &path_result {
+    if let Some(logging_file_path) = &logging_path {
+        // Try to create the parent directories for the logging file in the logging file path
```
cmyr: ditto my earlier comment, also for the next comment on 116; in both of these cases I think the code speaks for itself.

(I think this is true of a few other places as well, but it's all subjective, I'm not going to try and point them all out.)

rust/src/main.rs
06kellyjac: I'm happy to remove them if you prefer. It was mostly for while I was tweaking things.

rust/src/main.rs
06kellyjac: I usually under comment so I've probably put a few too many here and there.
Happy to remove them as I agree in these cases the code is already self explanatory.

Keep up the nitpicking!

rust/src/main.rs
06kellyjac: See other comment for 116

rust/src/main.rs
06kellyjac: It used to be that the `Result<PathBuf, io::Error>` was passed into `setup_logging` but it was probably cleaner to just give it an `Option<PathBuf>`.
The reason for the separate values is so the `warn!` could go after the `setup_logging` part.

06kellyjac: I just tried this code & the `warn!` bit didn't show up 😢 
```
let logging_path_result = Err(io::Error::new(
    io::ErrorKind::InvalidInput,
    "A blank file name was supplied",
));

if let Err(e) = logging_path_result.as_ref() {
    warn!(
        "Unable to generate the logging path to pass to set up: {}",
        e
    )
};
if let Err(e) = setup_logging(logging_path_result.ok()) {
    eprintln!(
        "[ERROR] setup_logging returned error, logging not enabled: {:?}",
        e
    );
};
```

cmyr: Ah right I understand. We could do the reverse just as easily, though,

```rust
let log_path = generate_logging_path(logfile_config);
if let Err(err) = setup_logging(log_path.as_ref().map(|p| p.as_path()).ok()) {
    eprintln!("[Error] logging setup failed: {:?}, err);
}
if let Err(err) = log_path {
    warn!("Failed to get log file path: '{}'", err);
}
```

in this version `setup_logging` takes an `Option<&Path>`.

06kellyjac: Works well.

```
let logging_path_result = Err(io::Error::new(io::ErrorKind::InvalidInput, "A blank file name was supplied"));

let logging_path = logging_path_result
    .as_ref()
    .map(|p: &PathBuf| -> &Path { p.as_path() })
    .ok();

if let Err(e) = setup_logging(logging_path) {
    eprintln!(
        "[ERROR] setup_logging returned error, logging not enabled: {:?}",
        e
    );
}
if let Err(e) = logging_path_result.as_ref() {
    warn!(
        "Unable to generate the logging path to pass to set up: {}",
        e
    )
};
```

I also fixed some stuff to work with the new `Option<&Path>`.

06kellyjac: I also like that `create_log_directory` is just using a `&Path` now.

06kellyjac:
Thanks for all the feedback both of you.

PR #896, opened 2018-10-08
Use language from core in plugins

### LINKED ISSUE #891: Have plugins use core-determined language identifiers
Our use of the syntect plugin for syntax highlighting predates the existence of language identifiers in core. The syntect plugin current determines the syntax to use based on the path of the file, and updates it when the file has changed.

This is now redundant. Instead, the language identifier that is assigned in core (which may be guesed from the filename or the file's first line, or may be overridden by a user) should be the sole/canonical source of truth.

This means that:
- [ ] when a new document is opened, the `LanguageId` should be included in the set of info we send the plugin.
- [ ] when the `LanguageId` changes in core, we should update the plugin (and the client) similarly to how we do currently when a config changes. This _probably_ means adding a new core->plugin RPC; I could one day imagining combining these into a per-buffer 'metadata' RPC, but for now separate is fine.
- [ ] the syntect plugin should no longer guess the `LanguageId` on its own; it should use what is provided by the core.

There are still some kind of weird things related to languages and configs. For instance, there should be a config setting that allows the user to specify which language to assign to files, based on their extension; and so this config, when changed, could potentially change the language of an open file, which would cause it to update its config.

It also isn't clear how, long-term, we're going to manage plugin lifecycle, but we may end up doing something like #374, which might require changing how we tell plugins about language changes, but I think this change is reasonable for now.

akxcv:
## Summary

Currently, the language for syntax highlighting is determined in two places: in the core and in the syntect plugin. In order for the client to be able to change the language, we need to make the core the single source of truth (for languages, at least).

In this pull request:

- The syntect plugin is no longer guessing the language. Instead, it uses the language from the core.
- The core now notifies plugins and clients about language changes via `language_changed` RPC.

NOTE: I didn't write/run tests or add any docs yet. This is just a first preview.
NOTE 2: The only test I did is I compiled my local version of xi-mac with those changes, and everything worked as expected.
NOTE 3: You can see this working with this [version of xi-mac](https://github.com/xi-editor/xi-mac/pull/294).

## Related Issues

Related to [#282](https://github.com/xi-editor/xi-mac/issues/282)
Closes #891 

## Checklist

- [x] Make syntect use core-identified language on view init (fall back to plain text)
- [x] Send `language_changed` RPC to plugins
- [x] Handle `language_changed` RPC in syntect
- [x] Send `language_changed` RPC to client

## Review Checklist

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -308,12 +308,14 @@ impl ConfigManager {
     /// Sets a specific language for the given buffer. This is used if the
     /// user selects a specific language in the frontend, for instance.
     #[allow(dead_code)]
-    pub(crate) fn override_language(&mut self, id: BufferId, new_lang: LanguageId) {
+    pub(crate) fn override_language(&mut self, id: BufferId, new_lang: LanguageId) -> Option<Table> {
         let has_changed = self.buffer_tags.get_mut(&id)
             .map(|tag| tag.set_user(Some(new_lang)))
             .expect("buffer must exist");
         if has_changed {
-            self.update_buffer_config(id);
+            self.update_buffer_config(id)
+        } else {
+            None
         }
     }
 
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -307,7 +307,6 @@ impl ConfigManager {
 
     /// Sets a specific language for the given buffer. This is used if the
     /// user selects a specific language in the frontend, for instance.
-    #[allow(dead_code)]
     pub(crate) fn override_language(&mut self, id: BufferId, new_lang: LanguageId) -> Option<Table> {
         let has_changed = self.buffer_tags.get_mut(&id)
             .map(|tag| tag.set_user(Some(new_lang)))
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -343,6 +343,10 @@ impl<'a> EventContext<'a> {
         self.render()
     }
 
+    pub(crate) fn language_changed(&mut self, lang_id: &LanguageId) {
+        self.plugins.iter().for_each(|plug| plug.language_changed(self.view_id, lang_id));
+    }
+
     pub(crate) fn reload(&mut self, text: Rope) {
         //TODO: It would be nice if we could preserve the existing selections,
         //but to do that correctly we would need to compute a real delta between
@@ -439,7 +443,7 @@ impl<'a> EventContext<'a> {
             Err(err) => warn!("Hover Response from Client Error {:?}", err)
         }
     }
-     
+
     /// Gives the requested position in UTF-8 offset format to be sent to plugin
     /// If position is `None`, it tries to get the current Caret Position and use
     /// that instead
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -344,9 +344,15 @@ impl<'a> EventContext<'a> {
         self.render()
     }
 
-    pub(crate) fn language_changed(&mut self, lang_id: &LanguageId) {
-        self.client.language_changed(self.view_id, lang_id);
-        self.plugins.iter().for_each(|plug| plug.language_changed(self.view_id, lang_id));
+    pub(crate) fn language_changed(
+        &mut self,
+        old_language_id: &LanguageId,
+        new_language_id: &LanguageId
+    ) {
+        self.language = new_language_id.clone();
+        self.client.language_changed(self.view_id, new_language_id);
+        self.plugins.iter()
+            .for_each(|plug| plug.language_changed(self.view_id, old_language_id, new_language_id));
     }
 
     pub(crate) fn reload(&mut self, text: Rope) {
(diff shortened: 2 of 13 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -541,7 +541,11 @@ impl CoreState {
     fn do_set_language(&mut self, view_id: ViewId, language_id: LanguageId) {
         if let Some(view) = self.views.get(&view_id) {
             let buffer_id = view.borrow().get_buffer_id();
-            self.config_manager.override_language(buffer_id, language_id);
+            self.make_context(view_id).unwrap().language_changed(&language_id);
```
akxcv: I don't like the fact that `self.make_context(view_id).unwrap()` is repeated, but if I do `let context = self.make_context(view_id).unwrap()`, I cannot compile because I borrow `self` mutably and immutably at the same time...

rust/syntect-plugin/src/main.rs
```diff
@@ -192,9 +193,13 @@ impl<'a> Syntect<'a> {
     }
 
     /// Wipes any existing state and starts highlighting with `syntax`.
-    fn do_highlighting(&mut self, view: &mut MyView) {
+    fn do_highlighting(&mut self, view: &mut MyView, lang: Option<&LanguageId>) {
```
akxcv: not sure if this signature is clear enough

rust/core-lib/src/config.rs
```diff
@@ -308,12 +308,14 @@ impl ConfigManager {
     /// Sets a specific language for the given buffer. This is used if the
     /// user selects a specific language in the frontend, for instance.
     #[allow(dead_code)]
```
cmyr: we can remove this `#[allow(..)]` now; it was in place because I wrote this function at the same time as a bunch of other config stuff, and thought it was useful to keep it in place even though we weren't using it; now that we're using this flag (that silences a compiler warning) is unnecessary.

rust/plugin-lib/src/dispatch.rs
```diff
@@ -97,6 +97,16 @@ impl<'a, P: 'a + Plugin> Dispatcher<'a, P> {
         v.config = conf.unwrap();
     }
 
+    fn do_language_changed(&mut self, view_id: ViewId, new_lang: LanguageId) {
```
cmyr: my thinking is that we would in this case update the view itself, before calling the plugin, sort of like how we do above in `do_did_save`; I'm not sure if with language we will want to pass along what the previous language was, but I think it can't hurt?

we do the config one backwards because it's important to let the plugin know the _set_ of options that are updated, and that seemed like the easiest way to do that. It does feel a bit gross though. 😕

rust/syntect-plugin/src/main.rs
cmyr: I'm slightly confused by why we need to add this param; it seems to make things more complicated. By associating the syntax with the view object, we can make sure we're using the correct syntax for the correct view. I'll point out where I see this being a problem, below

rust/syntect-plugin/src/main.rs
```diff
@@ -192,9 +193,13 @@ impl<'a> Syntect<'a> {
...
         let initial_state = {
-            let syntax = self.guess_syntax(view.get_path());
+            let language_id = lang.unwrap_or(view.get_language_id());
```
cmyr: Do we ever modify the view's language id? If we don't, then this is going to be using the stale id if `None` is passed, such as after a save event.

akxcv: We do modify it every time, but yes, I think we should remove the `new_lang` parameter to avoid errors in the future and to make the signature simpler.

rust/core-lib/src/event_context.rs
```diff
@@ -344,9 +344,15 @@ impl<'a> EventContext<'a> {
         self.render()
     }
 
-    pub(crate) fn language_changed(&mut self, lang_id: &LanguageId) {
-        self.client.language_changed(self.view_id, lang_id);
-        self.plugins.iter().for_each(|plug| plug.language_changed(self.view_id, lang_id));
+    pub(crate) fn language_changed(
+        &mut self,
+        old_language_id: &LanguageId,
```
cmyr: I'm not sure we need to plumb through the old language? The plugin lib should already know what the previous language was, so if it needs that information it's available.

rust/plugin-lib/src/lib.rs
```diff
@@ -146,7 +146,12 @@ pub trait Plugin {
                       changes: &ConfigTable);
 
     #[allow(unused_variables)]
-    fn language_changed(&mut self, view: &mut View<Self::Cache>, new_lang: LanguageId) {}
+    fn language_changed(
```
cmyr: similarly in the pluging lib, assuming we've already updated the view to have the new language, passing it to the function directly might be unnecessary? I can understand passing the old language (because we've overwritten it in the view, and so the plugin has no other way of getting it) but I'm not sure we need the new language. This is similar to how we handle when the path changes; we update the view with the new path, and then pass along the old path in case it might be helpful.

rust/plugin-lib/src/dispatch.rs
akxcv: I think that in the current version I'm only passing what's appropriate. Please tell me if I'm wrong somewhere.

PR #904, opened 2018-10-10
Only check whitespace when opening an existing file

### LINKED ISSUE #900: Only check whitespace when opening an existing file
Currently, we check a file's whitespace when creating a new empty buffer, which will be a noop but causes spurious warnings in the  view. Instead we should only check whitespace when opening an existing buffer.

This should be a quick fix for someone; a good place to start is by searching for the text, "set_user_config failed to update config, no tables were returned" in `core-lib/src/tabs.rs`.

YangKeao:
## Summary
Fix #900. Add a tag `is_existing_file` to `pending_views` and detect_whitespace according to it.

## Related Issues
closes #900

### DIFF
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -112,7 +112,7 @@ pub struct CoreState {
     /// it can be passed to plugins.
     self_ref: Option<WeakXiCore>,
     /// Views which need to have setup finished.
-    pending_views: Vec<(ViewId, Table, bool)>,
+    pending_views: Vec<(ViewId, Table)>,
     peer: Client,
     id_counter: Counter,
     plugins: PluginCatalog,
@@ -354,10 +354,9 @@ impl CoreState {
         let view_id = self.next_view_id();
         let buffer_id = self.next_buffer_id();
 
-        let mut is_existing_file = false;
 
         let rope = match path.as_ref() {
-            Some(p) => {is_existing_file = true; self.file_manager.open(p, buffer_id)?},
+            Some(p) => self.file_manager.open(p, buffer_id)?,
             None => Rope::from(""),
         };
 
@@ -373,7 +372,7 @@ impl CoreState {
         //view_id before we can send any events to this view. We use mark the
         // viewa s pending and schedule the idle handler so that we can finish
         // setting up this view on the next runloop pass.
-        self.pending_views.push((view_id, config, is_existing_file));
+        self.pending_views.push((view_id, config));
         self.peer.schedule_idle(NEW_VIEW_IDLE_TOKEN);
 
         Ok(json!(view_id))
@@ -551,8 +550,12 @@ impl CoreState {
 
     fn finalize_new_views(&mut self) {
         let to_start = mem::replace(&mut self.pending_views, Vec::new());
-        to_start.iter().for_each(|(id, config, is_existing_file)| {
-            if *is_existing_file {
+        to_start.iter().for_each(|(id, config)| {
+            let buffer_id = self.views.get(&id).map(|v| v.borrow().get_buffer_id()).unwrap();
+            let has_contents = self.editors.get(&buffer_id)
+                .map(|e| !e.borrow().get_buffer().len() == 0)
+                .unwrap_or_default();
+            if has_contents {
                 let modified = self.detect_whitespace(*id, config);
                 let config = modified.as_ref().unwrap_or(config);
                 let mut edit_ctx = self.make_context(*id).unwrap();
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -566,11 +568,16 @@ impl CoreState {
 
     fn finalize_new_views(&mut self) {
         let to_start = mem::replace(&mut self.pending_views, Vec::new());
-        to_start.iter().for_each(|(id, config)| {
-            let modified = self.detect_whitespace(*id, config);
-            let config = modified.as_ref().unwrap_or(config);
-            let mut edit_ctx = self.make_context(*id).unwrap();
-            edit_ctx.finish_init(&config);
+        to_start.iter().for_each(|(id, config, is_existing_file)| {
+            if *is_existing_file {
```
cmyr: instead of passing through `is_existing_file`, we could just do,

```rust
let has_contents = self.editors.get(&buffer_id)
    .map(|e| !e.get_buffer().is_empty())
    .unwrap_or_default();
if has_contents {
    // ...
} else {
    // ...
}
```

This might be easier? It means less state to worry about.

rust/core-lib/src/tabs.rs
```diff
@@ -370,8 +370,10 @@ impl CoreState {
         let view_id = self.next_view_id();
         let buffer_id = self.next_buffer_id();
 
+        let mut is_existing_file = false;
+
         let rope = match path.as_ref() {
-            Some(p) => self.file_manager.open(p, buffer_id)?,
+            Some(p) => {is_existing_file = true; self.file_manager.open(p, buffer_id)?},
```
cmyr: oh also: `rustfmt` (which i'll turn on soon) is going to make us put this on separate lines.

rust/core-lib/src/tabs.rs
YangKeao: Yes. So that only when open an non-empty file whitespace will be checked.

rust/core-lib/src/tabs.rs
```diff
@@ -567,10 +568,19 @@ impl CoreState {
     fn finalize_new_views(&mut self) {
         let to_start = mem::replace(&mut self.pending_views, Vec::new());
         to_start.iter().for_each(|(id, config)| {
-            let modified = self.detect_whitespace(*id, config);
-            let config = modified.as_ref().unwrap_or(config);
-            let mut edit_ctx = self.make_context(*id).unwrap();
-            edit_ctx.finish_init(&config);
+            let buffer_id = self.views.get(&id).map(|v| v.borrow().get_buffer_id()).unwrap();
```
LiHRaM: The `detect_whitespace` method already fetches the `buffer_id` - if you make the check over there, you can just return `None` on false and keep `finalize_new_views` simpler?

cmyr: That's a good point, this becomes basically a one-liner in `detect_whitespace`: 
```rust
if editor.borrow().get_buffer().is_empty() { return None; }
```

YangKeao: Thanks for your advice!

rust/core-lib/src/tabs.rs
```diff
@@ -577,6 +578,14 @@ impl CoreState {
...
         let buffer_id = self.views.get(&id).map(|v| v.borrow().get_buffer_id())?;
+        if self.editors.get(&buffer_id)
+            .map(|e| {
+                e.borrow().get_buffer().len() == 0
+            })
+            .unwrap_or_default() {
+            return None;
+        }
+
         let editor = self.editors.get(&buffer_id).expect("existing buffer_id must have corresponding editor");
```
cmyr: we already get the editor here; if we move the check down below this line, it's just:

```rust
if editor.get_buffer().is_empty() {
    return None;
}
```

cmyr: maybe you were thinking that this handles the case when there is no editor better?

cmyr: @YangKeao thoughts on this?

cmyr: @YangKeao 😁

YangKeao:
Sorry for replying late :cry:.

YangKeao:
Sorry, there are some bugs. I will fix them.

cmyr:
No problem, thanks for the fix! It looks like you'll also need to run rustfmt on this, it's something we just added to CI.

If you're interested in continuing to work on xi stuff when you have time, I would be happy to add you to the xi-editor organization?

YangKeao:
Aha, I am very interested in rust and xi-editor and willing to work on it. Thank you :smile:

cmyr:
Okay, sent. I'm done for the day but will check in on this tomorrow. ☀️

Issue #328, opened 2017-05-29
`add_selection_above` and `add_selection_below` polish


eyelash:
- Running `add_selection_above` with a cursor somewhere on the first line should not add a cursor to the beginning of the document.
- Running `add_selection_below` with a cursor somewhere on the last line should not add a cursor to the very end of the document.
- When a selection spans multiple characters we should probably add a selection that spans the same range on the line above/below like Atom does (what I previously called Atom behaviour). Here we would need to discuss what to do if the line above/below is shorter than the current line. Atom skips the line if there would be no overlap but still adds a cursor if the selection is a caret.
- Atom skips empty lines if the selection is not a caret on the first column.

eyelash:
A few examples of Atom's behaviour:
Consider the following document:
```
abcd

ef
ghi
jklmn
```
If you add a cursor before `a` and call `add_selection_below`, a cursor on the empty line 2 will be added.
If you add a cursor between `a` and `b` and call `add_selection_below`, a selection between `e` and `f` will be added.
If you select `cd` and call `add_selection_below`, `i` will also be selected. If you call `add_selection_below` again, another selection will be added for `lm`.

ishanjain28:
HI, I would like to work on this issue, From your description of how this feature should work, It looks like you are talking about the multiple cursor support that is available in VSCode, Atom or vim.

The rpc calls `add_selection_below` and `add_selection_above` are already used in xi-editor code to select the text below and above the cursor when Shift+Down/Up keys are pressed. 

So, I'd suggest naming them `add_cursor_below` and `add_cursor_above`. 

cc @cmyr, I want to work on adding multiple cursor support. I checked the code in core-lib but I can use some help in implementing multiple cursors.

cmyr:
@ishanjain28 multiple cursor support is already implemented; this issue is about polishing two commands related to multi-cursor.

On xi-mac, this is mapped to <kbd>control</kbd>+<kbd>shift</kbd>+<kbd>↑</kdb> (or ↓). This adds a new cursor at the position of the current cursor, on the previous (or next) line.

If you play around with this, you can see the problem originally reported. Would be very happy to have you explore this, it shouldn't be too much work.

ishanjain28:
@cmyr I would love to work on this issue but I am on arch linux and I use `xi-gtk` as the frontend, And it doesn't has support for multiple cursors.

eyelash:
[...]

PR #906, opened 2018-10-10
add selection above and below will only move to lines with the same horizontal position

timpbrenner:
## Summary
This introduces two new types of Movement, UpWithHorizPos and DownWithHorizPos. These will move to the next line that is long enough to maintain the current cursor position. They are employed only the selection up and selection down events right now (control+shift+↑ or ↓).

My thought was that this behavior was different enough than up or down to make a new movement type.

## Related Issues
#328


cmyr:
@timpbrenner thanks for this! need to get my head around the original issue, but will get you a review soon. :)

cmyr:
@ryanbloom you might also be interested in this?

rust/core-lib/src/movement.rs
```diff
@@ -81,19 +85,41 @@ fn vertical_motion(r: SelRegion, view: &View, text: &Rope, line_delta: isize,
...
+
+        return (view.line_col_to_offset(text, line, col), Some(col));
     }
-    let new_offset = view.line_col_to_offset(text, line, col);
-    (new_offset, Some(col))
+
+    loop {
+        // If the line is longer than the current cursor position, break.
+        let line_length = view.offset_of_line(text, line.saturating_add(1)) - view.offset_of_line(text, line);
+        if line_length > col { break; }
```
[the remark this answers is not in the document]
timpbrenner: Empty lines have a length of 1, I am assuming its the newline. So >= would put a cursor on a line with 1 character shorter. If this is confusing I can put in a comment.

timpbrenner:
I will defer on the design decision. This implementation made more sense to me and seemed more useful. I can change it to only skip over blank lines, should be a simple change.

I am fine looking into selections into another PR, I will change the PR so it won't close the issue and will take a look after this is merged.

I see what you use case there and I do realize its strange but I am not sure what else should happen? Maybe its an argument to only skip blank lines? Again I will defer on the design decision.

Thank you for your feedback.

rust/core-lib/src/movement.rs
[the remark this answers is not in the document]
ryanbloom: Oops, my bad! The problems I was having must have been the related to the `horiz` issue. I don't think a comment is necessary, but it's up to you.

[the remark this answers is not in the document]
timpbrenner: I added a comment, I could definitely be confusing.

rust/core-lib/src/movement.rs
```diff
@@ -80,20 +84,49 @@ fn vertical_motion(r: SelRegion, view: &View, text: &Rope, line_delta: isize,
...
+    if !enforce_horiz_pos {
+        if line > n_lines {
+            return (text.len(), Some(col));
+        }
+
+        return (view.line_col_to_offset(text, line, col), Some(col));
+    }
+
+    // If the active columns is longer than the current line, use the current line length.
+    let col = if line_length < col { line_length } else { col };
```
timpbrenner: @ryanbloom I did this to ensure the column is never longer then the starting position. I wasn't sure of a better way to get the current cursor's horizontal position. This works though it may not be optimal.

cmyr:
@timpbrenner Sorry for the delay on this, I've had trouble getting my head back around this problem and I feel like it needs careful thought. Will try to block out some time later this week. In any case, thanks for your patience!

rust/core-lib/src/movement.rs
```diff
@@ -45,6 +45,10 @@ pub enum Movement {
     UpPage,
     /// Move down one viewport height.
     DownPage,
+    /// Move up to the next line that can preserve the cursor position.
+    UpEnforceHorizPos,
```
cmyr: Bikeshed: I might go with `UpExactPosition`?

rust/core-lib/src/movement.rs
```diff
@@ -62,7 +66,7 @@ pub enum Movement {
 /// Note: in non-exceptional cases, this function preserves the `horiz`
 /// field of the selection region.
 fn vertical_motion(r: SelRegion, view: &View, text: &Rope, line_delta: isize,
-    modify: bool) -> (usize, Option<HorizPos>)
+    modify: bool, enforce_horiz_pos: bool) -> (usize, Option<HorizPos>)
```
cmyr: This function is becoming _very_ hard to understand, but I'm not totally sure what the right solution is. I certainly wouldn't want to ever add _another_ bool, after this one, and I'm not sure about this one.

This particular movement feels like it maybe has enough specific logic that it would be worth putting in its own function? I'm not _sure_ this is best, but I'd encourage you to maybe play around with that and see if it simplifies anything?

rust/core-lib/src/movement.rs
```diff
@@ -80,20 +84,48 @@ fn vertical_motion(r: SelRegion, view: &View, text: &Rope, line_delta: isize,
     // This code is quite careful to avoid integer overflow.
     // TODO: write tests to verify
     let line = view.line_of_offset(text, active);
+    let mut line_length = view.offset_of_line(text, line.saturating_add(1)) - view.offset_of_line(text, line);
     if line_delta < 0 && (-line_delta as usize) > line {
-        return (0, Some(col));
+        if enforce_horiz_pos {
```
cmyr: thinking out loud: you could write this bit,

```rust
let row = if enforce_horiz_pos { active } else { 0 };
return (row, Some(col));
```

which _might_ be an improvement? Not a huge win though.

rust/core-lib/src/movement.rs
timpbrenner: Naming is never Bikeshedding, after working with the code I think position seems to be implied to be Horiz position. I think your name makes more sense, will change.

rust/core-lib/src/movement.rs
timpbrenner: I will play around with it. My problem when looking at it is that so much of what these two methods would have in common is setup, getting positions and lengths and that struck me as odd to put into its own function. Let me play around with it and see what I can do.

rust/core-lib/src/movement.rs
timpbrenner: If I don't split the methods I will make this change.

rust/core-lib/src/movement.rs
cmyr: I've also gone and done and run rustfmt on the whole repo while sleeping on this, so that will be another minor headache. I hope that if you run rustfmt on your branch separately (with the 1.29 compiler, not sure if there are any changes in rustfmt to 1.30) you'll still get a clean merge.

rust/core-lib/src/movement.rs
cmyr: Yea my main motivation is just how many different exit points this function has; it gets quite hard to follow. Another option might be to do all the setup in a shared function, and then branch afterwards? Just thinking out loud.

rust/core-lib/src/movement.rs
timpbrenner: I hadn't messed around with rustfmt so it just gave me an excuse to, didn't end up being a problem. Thanks for the heads up.

timpbrenner:
@cmyr I ended up splitting into 2 functions with one to return info about the cursor position. And by not having access to the current offset I was forced to both reduce and simplify the exits from the functions. I think this is an definite improvement, let me know what you think. Thank you.

rust/core-lib/src/event_context.rs
```diff
@@ -1727,4 +1727,45 @@ mod tests {
         four really nice|\n\
         lines to see." );
     }
+
+    #[test]
```
cmyr: I like this test, covers the general cases well and is easy to follow.

rust/core-lib/src/movement.rs
```diff
@@ -68,18 +72,10 @@ fn vertical_motion(
...
-    // The active point of the selection
-    let active = if modify {
-        r.end
-    } else if line_delta < 0 {
-        r.min()
-    } else {
-        r.max()
-    };
-    let col = if let Some(col) = r.horiz { col } else { view.offset_to_line_col(text, active).1 };
+    let (col, line, n_lines) = selection_position(r, view, text, line_delta, modify);
```
cmyr: No change required, but looking at this I think I would personally keep `n_lines` out of this function; it seems pretty separate and it should be a one-liner in both places it's used?

rust/core-lib/src/movement.rs
```diff
@@ -88,14 +84,84 @@ fn vertical_motion(
...
     (new_offset, Some(col))
 }
 
+/// Compute movement based on vertical motion by the given number of lines skipping
+/// any line that is shorter than the current cursor position.
+fn vertical_motion_exact_pos(
+    r: SelRegion,
+    view: &View,
+    text: &Rope,
+    line_delta: isize,
```
cmyr: I'm not sure that `line_delta` really makes sense as an argument in this case? we're moving up and down by a unit, and that unit is a function of the text/view and the current region, so this could just be an `is_down` bool or something? This saves us having to worry about handling negative numbers, integer conversion, etc.

rust/core-lib/src/movement.rs
```diff
@@ -88,14 +84,84 @@ fn vertical_motion(
...
+        init_line - (-line_delta as usize)
+    } else {
+        init_line.saturating_add(line_delta as usize)
+    };
+
+    // If the active columns is longer than the current line, use the current line length.
+    let col = if line_length < col { line_length - 1 } else { col };
+
+    loop {
+        line_length =
```
cmyr: although I think it's fine, I did stare at this for a while, thinking about the EOF case. I think the `saturating_add` is unnecessary here, because line + 1 for any valid line is <= the total number of bytes in the document; if this number would overflow we've already crashed.

rust/core-lib/src/movement.rs
```diff
@@ -88,14 +84,84 @@ fn vertical_motion(
...
+            view.offset_of_line(text, line.saturating_add(1)) - view.offset_of_line(text, line);
+
+        // If the line is longer than the current cursor position, break.
+        // We use > instead of >= because line_length includes newline.
+        if line_length > col {
```
cmyr: I also might consider combining these two statements into a single compound statement, eg, `if x || y || z { break; }`, although this is mostly a style/personal preference thing, so not a blocker.

rust/core-lib/src/movement.rs
timpbrenner: Fair, will change.

rust/core-lib/src/movement.rs
timpbrenner: My thought was that we could do a move where line_delta would be a minimum, but I'm not sure how much sense that makes. I will change it.

rust/core-lib/src/movement.rs
timpbrenner: I'll defer to your expertise here, will change.

rust/core-lib/src/movement.rs
timpbrenner: Well in the case of going past the line limit (first or last) I want to set the line to the init_line so I think this makes more sense as two if statements. Unless I am missing something I am going to leave as is.

timpbrenner:
@cmyr I understand the confusion, I thought number wise, you were talking visually. I changed it to move_up thinking that move will indicate more of the visual movement rather than line number change.

cmyr:
@timpbrenner great, will get this merged on green. Thanks again for all the work!

PR #917, opened 2018-10-12
MultiLineSelect doesn't need to move cursor

dzhou121:
## Summary
The current Gesture of MultiLineSelect has a bug that the cursor can't go to a position above the starting selection cursor position.  

Also, when you do a line selection, the cursor doesn't need to move to the end of the selection. 

## Review Checklist

### DIFF
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -639,6 +639,7 @@ mod tests {
     #[test]
     fn test_gestures() {
         use rpc::GestureType::*;
+        use rpc::MouseAction;
         let initial_text = "\
         this is a string\n\
         that has three\n\
@@ -733,11 +734,11 @@ mod tests {
         that has three\n\
         [lines.|]" );
 
-        ctx.do_edit(EditNotification::Gesture { line: 1, col: 2, ty: MultiLineSelect });
+        ctx.do_edit(EditNotification::Drag(MouseAction { line: 1, column: 2, flags: 0, click_count: Some(1) }));
         assert_eq!(harness.debug_render(),"\
         this is a string|\n\
-        [th|at has three\n\
-        lines.]" );
+        [that has three\n\
+        lines.|]" );
 
         ctx.do_edit(EditNotification::SelectAll);
         assert_eq!(harness.debug_render(),"\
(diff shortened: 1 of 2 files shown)


cmyr:
@ryanbloom you've been in this code a bit lately, want to take a look?

ryanbloom:
@dzhou121 I see there's something funny going on with `MultiLineSelect` on master, but it also seems to be happening on this branch. Could you describe how the behavior is supposed to be different?

(I'm using the xi-mac frontend, in case that has something to do with it.)

ryanbloom:
Huh, that sounds pretty similar to how xi-mac does it. Here's the bug that I'm seeing (here and on master):

lineselect bug

The mouse is dragged to the top of the selection, but the cursor stays at the bottom. Does this branch fix the problem for your frontend?

I took a look at your code, and the main difference seems to be that `scroll_to_cursor` isn't called after the selection changes. I don't *think* that should affect where the cursor actually ends up. Maybe the different behavior we're getting has to do with how frontends handle scroll events?

dzhou121:
I think this PR doesn't fix the bug for xi-mac, because if ```scroll_to_cursor``` isn't called, the cursor is left at the end of the selection. If ```scroll_to_cursor``` is called, there will be a explicit ```scroll_to``` which points the cursor to the end of selection. 

It fixed my front-end is because my front-end is emulating vim, so the ```scroll_to``` forces my cursor back to the end of the selection.

To fix the bug for xi-mac, I think we'll need to call ```scroll_to_cursor``` to the offset of the ```drag_to```

cmyr:
@dzhou121 I'm still a little confused. I  think the best option here is to add some tests to `event_context.rs` that demonstrate the expected behaviour?

cmyr:
@dzhou121 hi! Not sure if you're still interested in this patch, but if you can include a test case to show the behaviour you expect I would be happy to merge.

dzhou121:
I think I was confused about the position of the cursor and the ```scroll_to``` event. So the actual cursor is in the ```update``` event, but in my frontend I'm using the ```scroll_to``` for the cursor position. So in this PR, it changes the behaviour of ```scroll_to``` event not to point to the end of the selection, but to point to the position in the incoming ```gesture``` or ```drag```.

cmyr:
So this is all really on me, I never fully got my head around this PR. @dzhou121 if you're still interested in this I'd be happy to try again, otherwise I'm going to close. Sorry!

PR #919, opened 2018-10-13
Improve sse4.2 string compare performance

cmyr:
This version just uses bytewise compare, instead of the fancier (and
slower) string-specific operations. This shows a marked improvement, and
the new sse version is actually faster than the avx version, which is
slightly surprising; it may be that the avx loads are higher latency and
that's dominating? I've also only tested on one machine, a laptop with
turbo-boost disabled.

## Review Checklist

### DIFF
--- a/rust/rope/benches/diff.rs
+++ b/rust/rope/benches/diff.rs
@@ -36,86 +36,75 @@ static TARG_STR: &str = "This adds some function, I guess?, that has a size set 
 
 Currently my sense of smell (and the pain of implementing Write) might be too much, since bytes will probably always be fed one by one anyway. Otherwise crying might be needed (one by one is inefficient).";
 
+fn make_test_data() -> (Vec<u8>, Vec<u8>) {
+    let one = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].concat().into_bytes();
+    let mut two = one.clone();
+    let idx = one.len() / 2;
+    two[idx] = 0x02;
+    (one, two)
+}
+
 #[bench]
 fn ne_idx_sw(b: &mut Bencher) {
-    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].concat();
-    let mut two = one.clone();
-    unsafe {
-        let b = two.as_bytes_mut();
-        let idx = b.len() - 200;
-        b[idx] = 0x02;
-    }
+    let (one, two) = make_test_data();
 
     b.iter(|| {
-        compare::ne_idx_fallback(one.as_bytes(), one.as_bytes());
-        compare::ne_idx_fallback(one.as_bytes(), two.as_bytes());
+        compare::ne_idx_fallback(&one, &one);
+        compare::ne_idx_fallback(&one, &two);
     })
 }
 
 #[bench]
 fn ne_idx_sse(b: &mut Bencher) {
-    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].concat();
-    let mut two = one.clone();
-    unsafe {
-        let b = two.as_bytes_mut();
-        let idx = b.len() - 200;
-        b[idx] = 0x02;
-    }
+    let (one, two) = make_test_data();
 
     let mut x = 0;
     b.iter(|| {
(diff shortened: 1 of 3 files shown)


cmyr:
Okay, I've cleaned this up and it is behaving more reasonably.

rust/rope/benches/diff.rs
```diff
@@ -19,95 +19,95 @@ extern crate xi_rope;
...
-fn ne_idx_hw(b: &mut Bencher) {
-    let one: String = [EDITOR_STR, VIEW_STR, INTERVAL_STR, BREAKS_STR].concat();
-    let mut two = one.clone();
-    unsafe {
-        let b = two.as_bytes_mut();
-        let idx = b.len() - 200;
-        b[idx] = 0x02;
-    }
+fn ne_idx_sse(b: &mut Bencher) {
+    let (one, two) = make_test_data();
```
raphlinus: Do we want to `if is_x86_feature_detected("sse4.2")` around this so that it won't crash when benchmarking on ancient systems?

rust/rope/benches/diff.rs
```diff
@@ -19,95 +19,95 @@ extern crate xi_rope;
...
-        let idx = b.len() - 200;
-        b[idx] = 0x02;
-    }
+    let (one, two) = make_test_data();
 
     let mut dont_opt_me = 0;
     b.iter(|| {
-        dont_opt_me += compare::ne_idx_avx(one.as_bytes(), two.as_bytes()).unwrap_or_default();
-        dont_opt_me += compare::ne_idx_avx(one.as_bytes(), one.as_bytes()).unwrap_or_default();
+        dont_opt_me += unsafe { compare::ne_idx_avx(&one, &two).unwrap_or_default() };
```
raphlinus: Ditto `avx2`.

rust/rope/src/compare.rs
```diff
@@ -29,39 +29,38 @@ const SSE_STRIDE: usize = 16;
...
+    let mask = _mm_cmpeq_epi8(onev, twov);
+   (!_mm_movemask_epi8(mask)) ^ HIGH_HALF_MASK as i32
 }
 
 const AVX_STRIDE: usize = 32;
 
 /// Like above but with 32 byte slices
+#[cfg(target_arch = "x86_64")]
 #[target_feature(enable="avx2")]
 #[doc(hidden)]
```
raphlinus: The order of these attributes is not consistent, but that's very much a nit.

rust/rope/src/compare.rs
```diff
@@ -110,38 +113,27 @@ pub fn ne_idx_avx(one: &[u8], two: &[u8]) -> Option<usize> {
...
         }
-        let i = mask.trailing_zeros() as usize;
-        if i != 32 { return Some(idx + i); }
         idx += SSE_STRIDE;
-        if idx >= min_len { break; }
     }
     None
 }
 
 #[cfg(target_arch = "x86_64")]
```
raphlinus: This definitely needs to be marked unsafe, right now it's a safety hole. Also it should be `#[target_feature(enable)]`. My sense is that you made changes to the forward version and didn't make corresponding changes for rev?

rust/rope/src/compare.rs
```diff
@@ -405,9 +397,9 @@ mod tests {
         let two = "aaaa";
         let tre = "aaba";
         let fur = "";
-        assert!(ne_idx_sse42(one.as_bytes(), two.as_bytes()).is_none());
-        assert_eq!(ne_idx_sse42(one.as_bytes(), tre.as_bytes()), Some(2));
-        assert_eq!(ne_idx_sse42(one.as_bytes(), fur.as_bytes()), None);
+        assert!(ne_idx_sse(one.as_bytes(), two.as_bytes()).is_none());
```
raphlinus: So many of these tests fail if you specify `RUSTFLAGS='-C target-cpu=native'` for example. I think what you want to do is this:

* Don't `#[cfg(target_feature)]`.
* Do `#[cfg(target_arch)]`.
* Do runtime `is_x86_feature_detected` check.
* Wrap call in `unsafe`.

Apply consistently to all such tests; I see some closer but none fully correct.

Would be good to also test on aarch64, but I don't have a system super handy. I could probably get rust compiler set up on my rpi3 though.

cmyr:
@raphlinus this is ready for another look whenever

PR #920, opened 2018-10-13
Fix table of contents in CONTRIBUTING.md

ttaomae:
## Summary
This fixes a few small errors in the table of contents of the CONTRIBUTING.md file.

On a related note, I also noticed that docs/contribute.md seems to be out of date. Should that file be updated to match .github/CONTRIBUTING.md?

## Related Issues
None.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -11,13 +11,16 @@ follow the [Code of Conduct](https://github.com/xi-editor/xi-editor/blob/master/
     - [Improving and reviewing docs](#improving-and-reviewing-docs)
     - [Reviewing and testing changes](#reviewing-and-testing-changes)
 - [Proposing and making changes](#proposing-and-making-changes)
-	- [Finding something to work on](#finding-something-to-work-on)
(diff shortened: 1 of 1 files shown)


cmyr:
@ttaomae Nice catch! and yes, the other docs could be updated too; I'm not sure if it's best to copy this file over, or to just link to it? Two copies is always tricky. I wonder if there's a way to symlink it or something?

cmyr:
@jansol @LiHRaM seeing a spurious windows CI failure here, is there some option to increase the storage on the image or something?

ttaomae:
@cmyr I did a quick search and it looks like it is possible to use a symlink. https://github.com/s4y/gh-pages-symlink-test

cmyr:
@ttaomae cool, we'd certainly welcome a PR with that change 😄

ttaomae:
@cmyr Hmm, thinking about it a little more, I don't think it would work in this situation since it would be missing the Jekyll front matter.

cmyr:
mm, right. Copy paste works for me I guess?

PR #924, opened 2018-10-15
Fix freeze on `measure_width` request

mrandri19:


rust/rpc/src/lib.rs
```diff
@@ -271,6 +273,10 @@ impl<W: Write + Send> RpcLoop<W> {
                     let json = match self.reader.next(&mut stream) {
                         Ok(json) => json,
                         Err(err) => {
+                            if self.peer.0.is_blocked.load(Ordering::Relaxed) {
```
cmyr: Although it's unlikely to be a problem in practice, I think we should use `Ordering:SeqCst` here; I think as written there's still a theoretical risk of deadlock.

rust/rpc/src/lib.rs
```diff
@@ -271,6 +273,10 @@ impl<W: Write + Send> RpcLoop<W> {
                     let json = match self.reader.next(&mut stream) {
                         Ok(json) => json,
                         Err(err) => {
+                            if self.peer.0.is_blocked.load(Ordering::SeqCst) {
```
cmyr: Okay, so this actually made me do some reading on the various orderings, and I think we have another option: this load should just `Acquire`, and the store below should use `Release`. Sorry for the confusion. :)

mrandri19: Done

PR #926, opened 2018-10-15
Add documentation for `measure_widths` rpc method

### LINKED ISSUE #923: Add documentation for `measure_widths` rpc method

mrandri19:
Closes #923

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -554,6 +554,15 @@ interface Line {
 }
 ```
 
+#### measure_width
+
+```
+measure_width [{"id": number, "strings": string[]}] <- {"id":0, "result":[[28.0,8.0]]}
+```
+
+Asks the frontend to send, for each pair of theme id and string
+array, the typographic lenght of each string.
+
 ---
 
 #### theme_changed


docs/docs/frontend-protocol.md
```diff
@@ -554,6 +554,15 @@ interface Line {
...
 ```
 
+#### measure_width
+
+```
+measure_width [{"id": number, "strings": string[]}] <- {"id":0, "result":[[28.0,8.0]]}
+```
+
+Asks the frontend to send, for each pair of theme id and string
+array, the typographic lenght of each string.
```
cmyr: There's a typo in 'length', but I also think we could maybe make this a bit clearer, maybe:

"Asks the frontend to measure the typographic widths of a group of strings. The frontend should return an array of arrays, one for each item in the input array, containing the widths of each of that item's `strings` when rendered with the style indicated by that items `id` argument."

That still feels a bit too complicated, it is a tricky one.

mrandri19: It sounds good to me, maybe add some references to what is the typographic width (maybe linking to the implementation in xi-mac) and where it's used (the rope science posts about word wrapping)

cmyr: Oops sorry, missed that!

How about,

"Asks the frontend to measure the _display_ widths (the width when rendered and presented on screen) of a group of strings. The frontend should return an array of arrays, one for each item in the input array, containing the widths of each of that item's strings when rendered with the style indicated by that items id argument.

These widths are used to determine how to calculate line breaks and other attributes that depend on the behaviour of the client's text rendering system."

Feel free to add links if you wish?

eyelash:
Speaking of `measure_widths`, I always wondered why there is no `view_id` argument for this request. how does this work if two views use different font sizes?

PR #928, opened 2018-10-16
Positions of matches in find_status

scholtzan:
This is currently a WIP and requires some discussion about the protocol changes.

## Summary

I started looking into https://github.com/xi-editor/xi-editor/issues/834 to add the positions of the find matches to `find_status`. For now I added the line numbers of each match to `find_status`, but I was wondering if it makes sense to provide the exact position (start and end since regex could go over multiple lines) or would that be too much information getting sent which might negatively affect performance?

I was thinking to change `find_status` to something like the following: 

```
find_status {
    "view_id": "view-id-1", 
    "queries": [{
        "id": 1, 
        "chars": "a", 
        "case_sensitive": false, 
        "is_regex": false, 
        "whole_words": true, 
        "matches": 6,
        "positions": [{          // This part starting here would be new
            "start_line": 1,
            "start_col": 12,
            "end_line": 1,
            "end_col": 13
        },
        {
            "start_line": 2,
            "start_col": 2,
            "end_line": 2,
            "end_col": 3
        }]
    }]
}
```

Technically, if we send all the positions then `matches` would be redundant. Though, it might still be useful to have as a quick way to get the number of matches?

## Related Issues
https://github.com/xi-editor/xi-editor/issues/834

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -695,9 +695,9 @@ Removes a status item from the front end.
 
 Find supports multiple search queries.
 
-`find_status {"view_id": "view-id-1", "queries": [{"id": 1, "chars": "a", "case_sensitive": false, "is_regex": false, "whole_words": true, "matches": 6}]}`
+`find_status {"view_id": "view-id-1", "queries": [{"id": 1, "chars": "a", "case_sensitive": false, "is_regex": false, "whole_words": true, "matches": 6, "lines": [1, 3, 3, 6]}]}`
 
-Notifies the client about the current search queries and search options.
+Notifies the client about the current search queries and search options. `lines` indicates for each match its line number.
 
 #### replace_status
 
--- a/rust/core-lib/src/find.rs
+++ b/rust/core-lib/src/find.rs
@@ -112,7 +112,7 @@ impl Find {
                 is_regex: None,
                 whole_words: None,
                 matches: self.occurrences.len(),
-                lines: self.occurrences.iter().map(|o| view.offset_to_line_col(text, o.min()).0).collect(),
+                lines: self.occurrences.iter().map(|o| view.offset_to_line_col(text, o.min()).0 + 1).collect(),
             }
         } else {
             FindStatus {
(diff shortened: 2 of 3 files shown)


cmyr:
So in a sense we already send _some_ of these positions to the frontend, at least for visible regions, because we send them as highlights.

My longterm vision for selections, find highlights, and other annotations is to move them out of the style system, and send them as (probably) a separate field in the update protocol. In that case, this probably becomes redundant.

In any case, I think that the scenarios where a frontend wants the spans for a find highlight are probably limited to when that find is on screen, so I'm not sure we need to full spans for all results. My first inclination would be to just send the line numbers, as an array of ints; this should give the client the information they'd need to, say, display result indicators in a scroll wheel.

It may be that in the future we find that doing all searching up front is too slow; if this is the case we might need to break it up into batches, and then we'll have to make the update to the client be incremental/streaming. I _suspect_ we'll want to go this route eventually, but I'm okay with not doing it yet.

So: my vote is to just send the line number for each line that contains a find result, but if you have a particular case in mind for why we might want to send full spans I'm ready to change my mind. :)

scholtzan:
I agree that line numbers are probably sufficient for now.

I was thinking about two use cases where actual positions might be useful:
* If we have regular expressions with matches over several lines it might be nice to display indicators for all of the lines involved (similar to what happens in IntelliJ)
* In the _veeeery_ far future perhaps there will be plugins, like a minimap, that would require information about all match position to display them. However, I am sure many things will change until then, so it would not make sense to plan for that.

I should look into separating find highlights and selection from the style system. It keeps coming up quite often recently ;)

cmyr:
@scholtzan yea, my feeling is that we're ready to make the move on selections. The more time I've had to think about that the clearer it is to me that it's the correct decision.

I'm not familiar with intelliJ, so I'd be curious to see a screenshot.

For minimap: I suspect that will have to be pretty tightly integrated into core, although I guess there are a few ways to go about it; but yea I don't think we need to plan for that quite yet.

scholtzan:
IntelliJ has an indicator that covers the entire range for the match:
screen shot 2018-10-15 at 21 56 39

Other editors, such as Xcode, only show an indicator only for the start of the match.

scholtzan:
I updated `find_status` and added the field `lines` which contains for each match the line number.

portstrom:
I see the number of positions is unbounded. This would be very bad if you open a 1 GB file and start searching for something common. As the idea to represent matches as ranges rather than a single line number, I think that can also be a solution to limit the number of positions sent to the client. If the distance between two ranges is less than 0.5% of the total number of lines in the file, then merge those ranges into one.

scholtzan:
You are definitely right. Currently, sending the find positions in `find_status` is more of a workaround and will at some point be replaced by what we come up with in https://github.com/xi-editor/xi-editor/issues/643

PR #931, opened 2018-10-16
Allow CI failures on windows (temporary)

cmyr:


cmyr:
This issue is being discussed here, we can turn this back on when it gets resolved:
https://travis-ci.community/t/there-is-not-enough-space-on-the-disk/338/13

Issue #913, opened 2018-10-11
State restoration


cmyr:
Breaking this out from #865.

State restoration has two goals: preventing data loss and allowing the user to quit and reopen their client without losing their open documents and window positions, etc.

This is not a small feature! Off the top of my head it requires:

- [ ] keeping track of the _absolute_ delta between the current buffer state and the last saved state, for each open buffer; this is distinct from the undo stack (the undo stack can contain a lot of additonal information, but is also bounded in length; there is no guarantee that the undo stack provides enough information to recreate the document)
- [ ] keeping track of the view state (the selections and the viewport position) for all views, and making all of this data serializable
- [ ] keeping track of undo state; this is non trivial because it is currently tied up in the general CRDT state, which we _don't_ want to try and serialize/load on the fly; instead we want to serialize some summary of each undo op on the stack; this should probably be a follow-up project
- [ ] keep track of any user-overridden config settings;
- [ ] providing an API for the client to pass us arbitrary state that it will need in order to restore the session, such as the positions of windows and whether they're displaying a sidebar or anything else a particular client might provide
- [ ] persisting this on the fly, efficiently and performantly
- [ ] _restoring_ this state, _quickly_, when needed;
- [ ] probably some other stuff I'm forgetting?

This is closely related to https://github.com/xi-editor/xi-mac/issues/290; basically xi-mac currently relies on a bunch of Cocoa APIs for things like state restoration, but those make it harder to do other things, so we'd like to ditch them. In this case, xi-mac will rely on this work.

A first version of this should focus on just the data-protection angle; being able to track unsaved changes, but ignoring things like view or undo state.

akxcv:
It would be nice not to bother with session restoration if Xi is opened by double-clicking a file, for instance. In case the user just runs Xi itself, we probably want to try to restore the last session. Requesting last session data explicitly will also allow us to put this behind a menu, for example.

I imagine the following scenario for how the session is restored:

1. The client explicitly asks whether there is any data about the last session via a new RPC. This should _probably_ be synchronous.
2. If the answer is positive, the client might send a (synchronous?) RPC to restore the last session. (The client might not want to actually restore the session, e.x. we could just show the user that there is something to be restored, but the user might not want to)
3. Restoration happens in the core (views are opened, cursor and viewport positions are set, config overrides are applied).
4. After the restoration is over, the core sends an RPC including view ids and any other required data to the client.

cmyr:
[...]

PR #939, opened 2018-10-18
[RFC] Session restoration

akxcv:
This is the first draft for the design of session restoration (#913).

### [Rendered](https://github.com/akxcv/xi-editor/blob/rfc/session-restoration/rfcs/session-restoration.md)

### DIFF
--- a/rfcs/session-restoration.md
+++ b/rfcs/session-restoration.md
@@ -1,0 +1,65 @@
+## Motivation
+
+As per [October 2018 Roadmap](https://github.com/xi-editor/xi-editor/issues/937), the focus now
+is on making xi-editor viable for daily use.
+
+To be viable for daily use, an editor must be able to prevent data loss. This feature would also
+allow for better continuation between closing and opening the editor, as it would allow the user
+to reopen files just as they left them when closing the editor.
+
+See also: [#913](https://github.com/xi-editor/xi-editor/issues/913).
+
+## Proposal
+
+### Overview
+
+Upon initialization, the core creates a `Session` object which is responsible for storing [session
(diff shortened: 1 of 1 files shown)


cmyr:
broadly seems good, some very quick notes:

- it might be worth separating buffer info from view info; we will at some point support multiple views per buffer.

- json is reasonable at first, but we might want to consider [bincode](https://github.com/TyOverby/bincode); I'm not sure if this is already a dependency of ours (we might bring it in for theme caching?) but if it is then it might make sense to start there

- we might not want to write out the whole file each time there's a small update; we could have some approach where we write out changes as a log, and then periodically compact the log? We want to think about how to minimize the amount we're writing.

cmyr:
yea, that's more or less my thinking, but obviously I haven't done any measuring or anything, so it's just gut feeling. It might be worth implementing this with like a `StateSerializer` trait where we can easily swap in different implementations?

jansol:
If atomic saves rely on backup files e.g. like in vim, maybe store deltas at first against the original file, but after a certain amount of time (or # of edits) flush them to the backup and start storing deltas against that?

jansol:
If atomic saves rely on backup files e.g. like in vim, maybe store deltas at first against the original file, but after a certain amount of time (or # of edits) flush them to the backup and start storing deltas against that?

jansol:
Any news on this?

akxcv:
> Any news on this?

I've been vacationing, and now I'm about to start working on this again.

cmyr:
this particular attempt at this feature is on hold, and another attempt will provide its own design.

Issue #914, opened 2018-10-11
Remove Interval data structure


raphlinus:
The xi-editor codebase is complex. Some of the complexity is in service of important goals, for example the asynchronous design is in service of greater responsiveness and not blocking user input. Other sources of complexity are not pulling their weight. I think the best example of that is the `Interval` datatype. This issue contains a plan to remove it.

## Original motivation

The notion of "interval" is tied to metrics and, in particular, boundaries. It was also intended to capture subtle behavior about editing. For example, when inserting a character at the beginning or end of a span, does the span extend to the inserted character, or is the character inserted outside it? A large part of the idea of "interval" was to specify the various choices, and was directly inspired by the [SPAN_INCLUSIVE_EXCLUSIVE](https://stackoverflow.com/questions/9879233/explain-the-meaning-of-span-flags-like-span-exclusive-exclusive) mechanism in Android. The idea was that the desired behavior would naturally fall out of algebra between points and spans. That idea has not been realized.

Also, there was a vague inspiration (from the CRDT literature) that positions in a string as it's being edited correspond to real numbers. Thus, "after 2" and "before 3" might be different locations after an insert at 3, even though they're the same cursor position before. This idea also never really materialized, and instead we should use "priorities," which I think are still a good idea.

Lastly, when expressing intervals in terms of lines rather than bytes, I wanted to use an interval closed at the end to represent the partial line at the end.

I'll use the standard notation (same as [Wikipedia](https://en.wikipedia.org/wiki/Interval_(mathematics))) here. [0, 1) is a closed-open interval, (1, 2) is open-open, (2, 3] is open-closed, and [3, 4] is closed-closed.

## The plan

The plan, in brief, is to find all occurrences of intervals which are not closed-open, and reimplement the underlying logic to use closed-open intervals instead. Once all uses of intervals are closed-open, they can be replaced with a more idiomatic `Range` instead (see #576 for some of the existing work in using `Range`).

Doing a simple `git grep Interval::new | grep -v "closed_open"` search, I see a number of uses, though most seem to be tests. My gut feeling is that most can be changed straightforwardly, but breaks will require a re-think.

## Breaks

[...]

PR #941, opened 2018-10-24
Make all intervals closed_open

cmyr:
This is a step towards #914, removing the interval data type completely.
In order to simplify those changes, this changes the implementation of
Interval to only ever represent a closed-open range; this will make it easier
to reason about whatever changes are required to get wrapping and spans working
again.

## Review Checklist

### DIFF
--- a/rust/rope/Cargo.toml
+++ b/rust/rope/Cargo.toml
@@ -4,7 +4,7 @@ description = "A generic rope data structure built on top of B-Trees."
 license = "Apache-2.0"
 name = "xi-rope"
 repository = "https://github.com/google/xi-editor"
-version = "0.2.1"
+version = "0.3.0"
 
 [dependencies]
 bytecount = "0.3.1"
--- a/rust/rope/Cargo.toml
+++ b/rust/rope/Cargo.toml
@@ -4,7 +4,7 @@ description = "A generic rope data structure built on top of B-Trees."
 license = "Apache-2.0"
 name = "xi-rope"
 repository = "https://github.com/google/xi-editor"
-version = "0.2.0"
+version = "0.2.1"
 
 [dependencies]
 bytecount = "0.3.1"
--- a/rust/rope/src/breaks.rs
+++ b/rust/rope/src/breaks.rs
@@ -47,7 +47,7 @@ impl Leaf for BreaksLeaf {
     fn push_maybe_split(&mut self, other: &BreaksLeaf, iv: Interval) -> Option<BreaksLeaf> {
         //eprintln!("push_maybe_split {:?} {:?} {}", self, other, iv);
         let (start, end) = iv.start_end();
-        let start_test = if iv.is_start_closed() { start } else { start + 1 };
+        let start_test = start + 1;
         for &v in &other.data {
             if start_test <= v && v <= end {
                 self.data.push(v - start + self.len);
@@ -254,7 +254,7 @@ mod tests {
         }
         for _ in 0..n {
             let len = node.len();
-            let empty_interval_at_end = Interval::new_open_closed(len, len);
+            let empty_interval_at_end = Interval::new(len, len);
             node.edit(empty_interval_at_end, testnode.clone());
         }
         node
--- a/rust/rope/src/breaks.rs
+++ b/rust/rope/src/breaks.rs
@@ -47,9 +47,8 @@ impl Leaf for BreaksLeaf {
     fn push_maybe_split(&mut self, other: &BreaksLeaf, iv: Interval) -> Option<BreaksLeaf> {
         //eprintln!("push_maybe_split {:?} {:?} {}", self, other, iv);
         let (start, end) = iv.start_end();
-        let start_test = start + 1;
         for &v in &other.data {
-            if start_test <= v && v <= end {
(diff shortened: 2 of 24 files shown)


cmyr:
I've added a commit to this that makes the changes described in #914.

rust/rope/Cargo.toml
```diff
@@ -4,7 +4,7 @@ description = "A generic rope data structure built on top of B-Trees."
 license = "Apache-2.0"
 name = "xi-rope"
 repository = "https://github.com/google/xi-editor"
-version = "0.2.0"
+version = "0.2.1"
```
raphlinus: If we're bumping, probably to 0.3. I think this is not semver-compatible.

rust/rope/src/breaks.rs
```diff
@@ -47,7 +47,7 @@ impl Leaf for BreaksLeaf {
     fn push_maybe_split(&mut self, other: &BreaksLeaf, iv: Interval) -> Option<BreaksLeaf> {
         //eprintln!("push_maybe_split {:?} {:?} {}", self, other, iv);
         let (start, end) = iv.start_end();
-        let start_test = if iv.is_start_closed() { start } else { start + 1 };
+        let start_test = start + 1;
```
raphlinus: I think this would be better as `start < v` below. Of course, I expect this to change a lot, so doesn't matter much.

rust/rope/src/interval.rs
```diff
@@ -12,109 +12,83 @@
...
-    }
+    pub start: usize,
+    pub end: usize,
 }
 
 impl Interval {
-    pub fn new(start: usize, start_closed : bool, end: usize, end_closed: bool) -> Interval {
-        let start = (start as u64) * 2 + if start_closed { 0 } else { 1 };
-        let end = (end as u64) * 2 + if end_closed { 1 } else { 0 };
+    pub fn new(start: usize, end: usize) -> Interval {
```
raphlinus: This is the place to decide if there's an end >= start invariant. My feeling is yes; so it should be added as a docstring and a debug assert.

rust/rope/src/interval.rs
```diff
@@ -162,123 +136,122 @@ impl Interval {
...
-            end: self.end - amount2,
+            start: self.start - amount,
+            end: self.end - amount,
         }
     }
 
     // insensitive to open or closed ends, just the size of the interior
     pub fn size(&self) -> usize {
-        self.end() - self.start()
+        self.end().saturating_sub(self.start())
```
raphlinus: If there's an end >= start invariant, this can (and probably should) just stay `-`.

rust/rope/Cargo.toml
cmyr: My thinking here was just that I wanted to use the `#[deprecated]` annotation on the old Interval constructors, and that wants to know the version the deprecation occurred. I think with semver, pre 1.0, anything goes, although there may be some unofficial conventions we're violating.

raphlinus: No, it's not "anything goes," 0.x.y to 0.x.(y+1) is supposed to not break code, but 0.x to 0.(x+1) is allowed to (this is slightly different than the original semver, as I recall).

jansol: Cargo docs mention that official semver does not actually specify 0.x.y as a special case but that cargo among others considers any 0.x.y version bumps to be breaking changes.

cmyr: It's also unclear that this breaks any code; at least it doesn't break API? The only API breaking change here is that I made `push_subseq` `pub(crate)` from `pub` because I thought it probably shouldn't be public API, but that's really out of scope anyway.

In either case, happy to defer to the crowd. I was thinking of doing a 3.0 of the rope crate shortly (and publishing to crates.io) but I wanted to finish this current batch of cleanup first.

cmyr: oops no it does remove some public methods from `Interval`, so yea, breaking.

rust/rope/src/spans.rs
```diff
@@ -170,11 +171,10 @@ impl<T: Clone + Default> Spans<T> {
         let new_end = xform.transform(base_end, true);
         let mut builder = SpansBuilder::new(new_end - new_start);
         for (iv, data) in self.iter() {
-            let (start_closed, end_closed) = (iv.is_start_closed(), iv.is_end_closed());
-            let start = xform.transform(iv.start() + base_start, !start_closed) - new_start;
-            let end = xform.transform(iv.end() + base_start, end_closed) - new_start;
-            if start < end || (start_closed && end_closed) {
-                let iv = Interval::new(start, start_closed, end, end_closed);
+            let start = xform.transform(iv.start() + base_start, false) - new_start;
+            let end = xform.transform(iv.end() + base_start, true) - new_start;
```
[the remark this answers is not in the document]
cmyr: Oops yep that's an error.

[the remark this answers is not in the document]
raphlinus: Agree after understanding this. I think the doc comment could be improved, perhaps from "at the given offset" to "into the region defined by base_start and base_end". And that could probably be an Interval rather than a pair of offsets, but maybe one thing at a time.

PR #952, opened 2018-10-27
Only run CI rustfmt check on stable Rust

cmyr:
Rustfmt will always fail on nightly, because there are different rules.


cmyr:
Yea my patience levels for figuring out config languages is... low. 😞

cmyr:
Looking at this again I should make sure it's actually running at all, will test a bit more 👍

cmyr:
okay double checked, looks good.

PR #953, opened 2018-10-27
Add 'real world' editing benchmark

cmyr:
This benchmark attempts to mimic what actual writing looks like; a bunch
of one-character edits one after another, followed by deleting
something, and periodically moving to another location in the buffer.


cmyr:
Yea it's pretty simple, but i was playing around with an optimization to do small edits in-place where possible and none of the current benchmarks are very well suited to that.

PR #957, opened 2018-10-28
Atomic Save

### LINKED ISSUE #918: Atomic save
Inspired by #913, here's another 'don't lose user data' feature:

Currently, we save files in place. This is potentially dangerous; if there is a panic or a power failure while we're overwriting a file, we risk destroying user data.

What most editors do is to write the file first to a temporary path, and then move this file to the final path with `mv` or similar. You could also copy the original file to a backup location, and then overwrite the original. For more information, here's a SO question: [how does vim save files?](https://stackoverflow.com/questions/13563627/how-does-vim-save-files)

mbStavola:
## Summary
We don't want to accidentally corrupt user data while trying to _save_ their data, so we should be a little more defensive in that process.

I took the approach to save the new content to a swapfile before overwriting the original. Currently the swapfile is written to the same directory that the final final will be written to, but we can change that to a temp directory or some xi specific directory if we want.

## Related Issues
Resolves #918

## Checklist
- [x] Test
- [ ] Review

## Review Checklist

### DIFF
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -188,9 +188,10 @@ where
 fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result<()> {
     use std::fs;
 
-    let tmp_extension = path.extension()
+    let tmp_extension = path
+        .extension()
         .and_then(|extension| extension.to_str())
-        .map_or_else(|| "swp".to_string(),|extension| format!("{}.swp", extension));
+        .map_or_else(|| "swp".to_string(), |extension| format!("{}.swp", extension));
     let tmp_path = &path.with_extension(tmp_extension);
 
     let mut f = File::create(tmp_path)?;
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -191,7 +191,7 @@ fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result
     let tmp_extension = path
         .extension()
         .and_then(|extension| extension.to_str())
-        .map_or_else(|| "swp".to_string(), |extension| format!("{}.swp", extension));
+        .map_or_else(|| "swp".to_string(), |extension| extension.to_string() + ".swp");
     let tmp_path = &path.with_extension(tmp_extension);
 
     let mut f = File::create(tmp_path)?;
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -186,7 +186,14 @@ where
 }
 
 fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result<()> {
-    let mut f = File::create(path)?;
+    use std::fs;
+
+    let tmp_extension = path.extension()
+        .and_then(|extension| extension.to_str())
+        .map_or_else(|| "swp".to_string(),|extension| format!("{}.swp", extension));
+    let tmp_path = &path.with_extension(tmp_extension);
+
+    let mut f = File::create(tmp_path)?;
     match encoding {
         CharacterEncoding::Utf8WithBom => f.write_all(UTF8_BOM.as_bytes())?,
         CharacterEncoding::Utf8 => (),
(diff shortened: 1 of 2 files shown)


rust/core-lib/src/file.rs
```diff
@@ -186,7 +186,15 @@ where
 }
 
 fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result<()> {
-    let mut f = File::create(path)?;
+    use std::fs;
+
+    let tmp_extension = path
+        .extension()
+        .and_then(|extension| extension.to_str())
+        .map_or_else(|| "swp".to_string(), |extension| format!("{}.swp", extension));
```
jansol: You can use `extension.to_string() + ".swp"` which is faster than `format!()`. This also allows you to turn `extension` into a `String` earlier, which in turn simplifies this line a bit.

mbStavola: Made the first change, but I couldn't figure out a good way to do the conversion to `String` earlier-- since `extension` is an `OsStr`, it's a bit of a pain.

jansol: I think you can do something like
```rust
path
.extension()
.and_then(|extension| extension.to_str()?.to_string())
.map_or_else(|| "swp".to_string(), |extension| extension+".swp")
```

If the `to_str()` fails, `?` will just return `None` from the closure.

mbStavola: Unfortunately, `to_string()` returns `String` instead of `Option<String>` so we can't return like that. However, we _could_ add an `into()` call to coerce it into a `Some`, which is pretty neat!

rust/core-lib/src/file.rs
```diff
@@ -186,7 +186,15 @@ where
...
-    let mut f = File::create(path)?;
+    use std::fs;
+
+    let tmp_extension = path
+        .extension()
+        .and_then(|extension| extension.to_str()?.to_string().into())
+        .map_or_else(|| "swp".to_string(), |extension| extension + ".swp");
+    let tmp_path = &path.with_extension(tmp_extension);
+
+    let mut f = File::create(tmp_path)?;
```
cmyr: This will fail of the swap file already exists. We probably don't want that. What do we want? Is this a situation where we might consider alerting the user?

rust/core-lib/src/file.rs
```diff
@@ -186,7 +186,15 @@ where
 }
 
 fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result<()> {
-    let mut f = File::create(path)?;
+    use std::fs;
```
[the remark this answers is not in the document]
mbStavola: Ah! Forgot about this, I'm usually a little lazy and import to the function scope before moving it to the top of the file if need be. Will fix.

rust/core-lib/src/file.rs
```diff
@@ -186,7 +186,15 @@ where
...
+
+    let tmp_extension = path
+        .extension()
+        .and_then(|extension| extension.to_str()?.to_string().into())
```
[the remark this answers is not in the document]
mbStavola: Honestly, it's less effort to just keep it as an `OsString` and it works everything else. Why not?

rust/core-lib/src/file.rs
mbStavola: It doesn't fail if the swap file exists (according to [docs](https://doc.rust-lang.org/src/std/fs.rs.html#370) and my light testing with `xi-gtk`), but I agree that the behavior of ignoring and overwriting the swapfile is a bit weird.

I'm thinking maybe instead of saying "hey, you're going to overwrite the temp file" when trying to save, we could give the user the option to restore from the swapfile if it's still hanging around when they reopen the file? It seems to make more sense that way since we're adding this to prevent data-loss?

**However**, this seems to have a suspiciously high level of overlap with https://github.com/xi-editor/xi-editor/pull/939 I think. Maybe this Atomic Save is just a poor man's version of the `Session`object from the RFC? If that's the case, it's quite possible that this PR isn't really a super useful addition in the long run.

This is more of an abstract design question rather than an implementation detail; maybe it warrants more discussion to properly flesh out.

cmyr: Ah okay I'd misread the docs, this should overwrite the file.

I agree that there's really no great way for us to surface this information to the user, right now: ideally we would want to be able to open the temp file for them, and _really ideally_ we'd want to be able to show them a diff; but we don't really have either of these functions atm.

So I'll walk my suggestion back: I think that the overwrite is fine for now, and it's a nice theoretical win that if we screw up while writing we at least don't corrupt the existing file.

cmyr:
@mbStavola want to rebase this and I can make sure the issues I had are resolved?

cmyr:
Yep lgtm!

Issue #937, opened 2018-10-17
Roadmap (October 2018)


cmyr:
The principal focus of the next batch of work is turning xi-mac into a viable (albeit _minimal_) editor for daily use. To that end, the following is what I anticipate working on for the next 3-6 months:

### Width measurement and word wrap (overview/tracking at #565):
- [ ] remove Interval (#914): Not _directly_ related to width measurement, but it probably isn't worth doing word wrap based on the Interval code and _then_ removing it.
- [ ] Word wrap (#610, mostly)
- [ ] tracking longest line (#479): this needs a bit more thought, but it relates to word wrap in that it will share the width cache. 

### minimal language features
I've been avoiding this because doing it 'correctly' (in my estimation) involves confronting a bunch of fundamental problems about how languages are defined, the role of plugins, etc. But I'd like to be working in xi day-to-day, and the main things that bother me right now are comment toggling and manual indentation, so doing a 'good enough' implementation of this, for now, makes sense. In any case, there are two obvious approaches here; the first is to get this working through syntect metadata, and the other is to add something simple through the config system similar to what [vscode does](https://code.visualstudio.com/docs/extensionAPI/language-support#_smart-bracket-matching). 

- [ ] simple auto indent (#424)
- [ ] comment toggling (#795)
- [ ] _maybe_ bracket matching (#647)
    - [ ] _probably requires_ implementing edit priority (#933)

### misc other projects
There are a number of projects currently being worked on by other contributors:
- [ ] async loading of large files (#902, @Lacaranian)
- [ ] preliminary state restoration (#913, @akxcv)
- [ ] refactoring selection/movement commands (#826, @ryanbloom)
- [ ] _maybe_ treat selection / find spans differently from styles (_needs issue_, @scholtzan?)

As well as a number of projects related to the mac client, which are out of scope here.

## Getting involved

If you're interested in hacking on xi-editor, feel free to browse the issues. If you're interested in taking on something deeper, there are a number of open-ended questions and projects that could be interesting to explore:

#### Alternate syntax parsers (highlighters)

In xi-editor, syntax highlighting is provided by plugins. Currently there's only one such plugin, based on syntect. It would be fun to experiment with others. There are two good projects here: one is the currently broken plugin in /experimental/lang (#925), and the other would be writing a new plugin that uses tree-sitter (#753). Either of these projects would help expose any limitations of our current approach, and help us think about how best to implement full multi-plugin support.

[...]

PR #971, opened 2018-11-02
Simple syntect-based autoindent

### LINKED ISSUE #424: Auto-indentation

cmyr:
This adds support for basic autoindent, for languages supported by
syntect, through the syntect-plugin.

This implementation is not quite as fancy as Sublime's; we only use the
two most basic regex patterns, for 'increase next' and 'decrease
current' indentation level.

It is, however, actual honest-to-goodness syntax-aware auto-indent,
which is a step in the right direction.

Closes #424
Progress on #937

-----------

This patch has a sibling at https://github.com/trishume/syntect/pull/223; currently this is pointing to my syntect fork, so it shouldn't be merged before that does. In the meantime, anyone who is interested is encouraged to play around with this; I've only really tested it in rust any python, and not extensively in either case.

## Review Checklist

### DIFF
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -274,20 +274,14 @@ impl<'a> Syntect<'a> {
             return Ok(());
         }
         let prev_line = self.previous_nonblank_line(view, line)?;
-        let indent_level = self.indent_level_of_line(view, prev_line);
-        let increase = self.test_increase(view, prev_line + 1)?;
         let decrease = self.test_decrease(view, line)?;
-        let indent_level = match (increase, decrease) {
-            (true, false) => indent_level + tab_size,
-            (false, true) => indent_level.saturating_sub(tab_size),
-            _ => indent_level,
-        };
-
-        if indent_level != current_indent {
-            self.set_indent(view, line, indent_level)
-        } else {
-            Ok(())
+        if decrease {
+            let indent_level = self.indent_level_of_line(view, prev_line).saturating_sub(tab_size);
+            if indent_level != current_indent {
+                return self.set_indent(view, line, indent_level);
+            }
         }
+        Ok(())
     }
 
     fn set_indent(&self, view: &mut MyView, line: usize, level: usize) -> Result<(), Error> {
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -32,7 +32,9 @@ use xi_plugin_lib::{mainloop, Cache, Error, Plugin, StateCache, View};
 use xi_rope::{Interval, RopeDelta};
 use xi_trace::{trace, trace_block};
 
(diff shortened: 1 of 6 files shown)


scholtzan:
The behaviour when writing plain text is slightly weird. If I indent something then it gets automatically unindented. For example, here I am trying to indent a list item: 

out

cmyr:
@scholtzan Okay I've pushed a small change that should prevent us from so aggressively decreasing indent level.

scholtzan:
Cool! It feels much better now :)

scholtzan:
Another thing I noticed is that when text is pasted, it doesn't get indented:
out

rust/syntect-plugin/src/main.rs
```diff
@@ -210,82 +214,171 @@ impl<'a> Syntect<'a> {
...
+            //eprintln!("auto indenting {}, prev_level {}", line, base_indent);
+            self.set_indent(view, line, indent_level)
+        } else {
+            Ok(())
+        }
+    }
 
-        if is_newline {
-            let line_num = view.line_of_offset(start).unwrap();
+    /// Called when actviely editing a line; cheifly checks for whether or not
```
scholtzan: Nit: actively; chiefly

scholtzan: But this feels pretty good when writing code :)

cmyr:
Oups, missed that paste comment @scholtzan. I think that's a slightly different feature, and is out of scope for this PR; that and bulk-reindent are things that I'll open issues for afterwards.

rust/syntect-plugin/src/main.rs
```diff
@@ -210,82 +219,175 @@ impl<'a> Syntect<'a> {
...
+    fn autoindent_line(&mut self, view: &mut MyView, line: usize) -> Result<(), Error> {
+        let _t = trace_block("Syntect::autoindent", &["syntect"]);
+        debug_assert!(line > 0);
+        let tab_size = view.get_config().tab_size;
+        let current_indent = self.indent_level_of_line(view, line);
+        let base_indent = self
+            .previous_nonblank_line(view, line)?
+            .map(|l| self.indent_level_of_line(view, l))
+            .unwrap_or(0);
+        let increase_level = self.test_increase(view, line)?;
```
scholtzan: If I test here whether it should be decreased with `test_decrease` it returns an error (`error in autoindent PeerDisconnect`). The reason for this seems to be that at some point in `get_metadata` were are trying to get the `line` with `view.get(line)`. View seems to access some cache with the lines but cannot find `line` and returns an error. I am not sure if that is the expected behaviour?

PR #973, opened 2018-11-03
Fix off-by-one in plugin cache update logic

cmyr:
We were in correctly declining to apply a delta that was an append at
the end of file, which caused us to unnecessarily invalidate our cache.

## Review Checklist

### DIFF
--- a/rust/plugin-lib/src/base_cache.rs
+++ b/rust/plugin-lib/src/base_cache.rs
@@ -322,16 +322,15 @@ impl ChunkCache {
         let start = iv.start();
         let end = iv.end();
         // we only apply the delta if it is a simple edit, which
-        // begins in the interior of our chunk.
+        // begins inside or immediately following our chunk.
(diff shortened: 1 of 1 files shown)


rust/plugin-lib/src/base_cache.rs
```diff
@@ -331,7 +331,7 @@ impl ChunkCache {
         // The one 'complex edit' we should probably be handling is
         // the replacement of a single range. This could be a new
         // convenience method on `Delta`?
-        if start < self.offset || start >= self.offset + self.contents.len() {
+        if start < self.offset || start > self.offset + self.contents.len() {
```
scholtzan: I'm wondering, if it might be helpful to add a short explanation to the comment above? Otherwise, I could imagine that someone in the future might think `"after our chunk" >= self.offset + self.contents.len()` and changes it back accidentally.

cmyr: good point, updated!

PR #979, opened 2018-11-09
Update crossbeam, bump deps

cmyr:
This adds support for the latest version of crossbeam, and includes a `cargo update` run for good measure.

## Review Checklist


rust/rpc/src/lib.rs
```diff
@@ -339,7 +339,8 @@ impl<W: Write + Send> RpcLoop<W> {
                     }
                 }
             }
-        });
+        }).unwrap();
```
jansol: If I understand correctly this will panic on an Err? But the next thing in the file is an if that returns either Ok or Err so panicing here doesn't really make sense at a quick glance.

jansol: point

PR #986, opened 2018-11-12
Update protocol semantic information

scholtzan:
This is a WIP for https://github.com/xi-editor/xi-editor/issues/643
I also added a xi-mac implementation for testing: https://github.com/xi-editor/xi-mac/pull/345

### DIFF
--- a/rust/core-lib/src/annotations.rs
+++ b/rust/core-lib/src/annotations.rs
@@ -14,11 +14,13 @@
 
 //! Management of annotations.
 
-use std::collections::HashMap;
 use serde_json::Value;
+use std::collections::HashMap;
+use std::iter;
 
 use plugins::PluginId;
-use xi_rope::Interval;
+use view::View;
+use xi_rope::{Interval, Rope};
 use xi_rope::spans::Spans;
 
 pub type AnnotationType = String;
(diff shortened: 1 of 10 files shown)


rust/core-lib/src/event_context.rs
```diff
@@ -225,6 +225,9 @@ impl<'a> EventContext<'a> {
             UpdateSpans { start, len, spans, rev } => self.with_editor(|ed, view, _, _| {
                 ed.update_spans(view, plugin, start, len, spans, rev)
             }),
+            UpdateAnnotations { start, len, type_id, spans } => self.with_editor(|ed, view, _, _| {
+                ed.update_annotations(view, plugin, start, len, type_id, spans)
```
scholtzan: How do plugins update annotations? Is it right to have a separate RPC for this?

cmyr: There's no updating in place; to update the plugin just sends a new annotation.

rust/core-lib/src/editor.rs
```diff
@@ -861,6 +863,29 @@ impl Editor {
...
+        &mut self,
+        view: &mut View,
+        plugin: PluginId,
+        start: usize,
+        len: usize,
+        type_id: AnnotationType,
+        spans: Vec<AnnotationSpan>,
+    ) {
+        let _t = trace_block("Editor::update_annotations", &["core"]);
+        let start = start;
```
cmyr: 😏

rust/core-lib/src/plugins/rpc.rs
```diff
@@ -180,6 +189,7 @@ pub enum PluginRequest {
 pub enum PluginNotification {
     AddScopes { scopes: Vec<Vec<String>> },
     UpdateSpans { start: usize, len: usize, spans: Vec<ScopeSpan>, rev: u64 },
+    UpdateAnnotations { start: usize, len: usize, type_id: AnnotationType, spans: Vec<AnnotationSpan> },
```
cmyr: like with `UpdateSpans`, we're going to need to thread a `rev` through here.

rust/core-lib/src/view.rs
```diff
@@ -789,12 +815,24 @@ impl View {
                 }
             }
         }
+
+        let start_off = self.offset_of_line(text, self.first_line);
```
cmyr: maybe we can move all this off into a helper function?

rust/core-lib/src/annotations.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.items.edit(interval, items);
+    }
+}
+
+/// A region of an `Annotation`.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AnnotationSlice {
+    pub annotation_type: AnnotationType,
+    /// Annotation occurrences, guaranteed non-descending start order.
+    pub ranges: Vec<(usize, usize)>,
```
cmyr: as raph mentions in some comments on the RFC, we probably don't want to use byte indexes for this, because the frontend doesn't really know about them; to the frontend everything is in visual line/col space.

rust/core-lib/src/annotations.rs
```diff
@@ -0,0 +1,139 @@
...
+            } else {
+                annotations.push(Annotations {
+                    items: items,
+                    annotation_type: type_id
+                });
+            }
+
+            *e = annotations;
+
+        }).or_insert(vec![Annotations {
```
cmyr: this doesn't quite work, because the items that are sent could represent only a small region of the total document. Instead we need to create a new spans object of the correct size, and then update it with the items we've been sent.

rust/core-lib/src/editor.rs
scholtzan: 😏

rust/core-lib/src/annotations.rs
scholtzan: Currently, these ranges are translated into line/col when the annotations are converted to json

scholtzan:
There is some code for invalidating selections and find highlights which could be removed. However, since we are still supporting frontends using the old update protocol, this code can't get removed yet.

rust/core-lib/src/annotations.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub items: Spans<Value>,
+    pub annotation_type: AnnotationType,
+}
+
+/// A region of an `Annotation`.
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AnnotationSlice {
+    annotation_type: AnnotationType,
+    /// Annotation occurrences, guaranteed non-descending start order.
+    ranges: Vec<[usize; 4]>,
```
cmyr: I might give this a type, to make it a bit more self-documenting?

rust/core-lib/src/annotations.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use plugins::PluginId;
+use view::View;
+use xi_rope::spans::Spans;
+use xi_rope::{Interval, Rope};
+
+pub type AnnotationType = String;
+
+/// Annotation types used in core.
+pub enum CoreAnnotationType {
```
cmyr: Another option:

```rust
pub type AnnotationType {
    Selection,
    Find,
    Other(String),
}

impl AnnotationType {
    fn as_str(&self) -> &str {
        match self {
            AnnotationType::Find => "find",
            AnnotationType::Other(ref s) => s,
            // etc;
        }
    }
}
```

eyelash:
I'm assuming this changes the front-end protocol (I don't really have time to keep up with the core changes at the moment) so it would be nice if it could be documented in [the documentation](https://github.com/xi-editor/xi-editor/blob/master/docs/docs/frontend-protocol.md) and I will hopefully be able to update xi-gtk before the old protocol gets dropped.

scholtzan:
You are absolutely right! I'll update the documentation.

cmyr:
@eyelash fwiw this doesn't break anything; the new stuff is current added in addition to the old stuff so that we can experiment freely.

PR #998, opened 2018-11-19
Fix Lang Plugin

### LINKED ISSUE #925: Fix experimental/lang plugin
There's a broken plugin sitting in /experimental/lang that it would be nice to try and get working again; it's a proof of concept for a high-performance parsing plugin that can be used as an alternative to syntect.

This was moved into experimental because it was annoying always trying to keep it updated while doing a bunch of refactoring on the plugin library, but as that has momentarily stabilized it might be a good time to try and get it working again?

This would be a good exercise for someone who is curious about how syntax highlighting and plugins work, and is looking for a warm-up.

mbStavola:
## Summary
Gets the `experimental/lang` plugin working again by migrating the code to the new plugin API.

Thankfully, it was similar enough to the `syntect` plugin that I was able to use it as a reference. However, a shim needed to be added that converts the plugin's native `StateEl` into `syntect` friendly _scopes_.

## Related Issues
Closes #925

## Checklist
- [x] Get things _actually_ highlighting
- [x] Make sure the chosen scopes are actually _accurate_
- [x] Cleanup 

## Review Checklist


cmyr:
If you log the scopes as they're sent by the plugin, are they the same as you see them here? If not then it will be some issue with the mapping in core, otherwise it will be something in the plugin.

This plugin was always very much an experiment, and sent a pretty minimal set of scopes. It's probably possible that they're wrong as well.

mbStavola:
Looks like this is functional for the most part! Key part here was doing what the `syntect` plugin does and sending over whitespace as `["source.rust"]`. Had a bit of fun with the peg faux-DSL and was pleasantly surprised at how straight forward it was.

screen shot 2018-11-20 at 11 28 56 pm
screen shot 2018-11-20 at 11 34 25 pm
screen shot 2018-11-20 at 11 37 11 pm

I think we can queue this up for a review so I can clean up some of the code / fix any bugs.

mbStavola:
@cmyr I wasn't able to reproduce the exact issue you mentioned (on my end it was always sending spans), but I did find that there were still some issues with the scopes. I was able to _more_ closely model the `syntect`plugin by making the default parse case return the top-level scope. This gets us wayyyy more consistent behavior while highlighting.

Hopefully this also addresses your issue?

rust/experimental/lang/src/rust.rs
```diff
@@ -38,56 +42,108 @@ pub enum StateEl {
...
-        StateEl::Keyword => style.font = 1,
-        StateEl::Operator => style.fg_color = 0xFFAA2244,
-        StateEl::PrimType => style.fg_color = 0xFF44AAAA,
+impl StateEl {
+    /// Transform a [StateEl] into a [Vec] of Syntect Scope [String]s.
+    /// See [this](https://github.com/sublimehq/Packages/blob/master/Rust/Rust.sublime-syntax)
+    /// for reference.
+    pub fn as_scopes(&self) -> Scope {
+        let scope_strs = match self {
+            StateEl::Source => vec!["source.rust"],
```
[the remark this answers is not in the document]
mbStavola: I couldn't make these either `const` or `static` since `vec![]` and `Vec::new()` aren't `const` functions. I'm pretty unfamiliar with either concept in Rust, so I'm not sure if there is some finagling I could do to get it working. I did a quick read and it seemed not really possible.

What I ended up doing was bringing in `lazy_static` and pulled out the `Vecs`-- the thought  being that it might help future refactoring. With the `to_vec()` calls, I'm still allocating as much as I was before, but I'm hoping that we can migrate the `Scope` type from `Vec<String>` to `&[&str]` or something in the future.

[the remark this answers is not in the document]
cmyr: right, sorry, I should've been more clear.

What I was thinking is that instead of these being `Vec<String>`s, they can be `const &[&str]`s. In fact, given that we know all of our scopes up front, it might even make sense to just give them numerical IDs; we could send `def_styles` right after launch, and then exclusively use integer ids internally?

In any case this plugin is 100% a demo so I don't think we need to worry too much.

rust/experimental/lang/src/language/rust.rs
```diff
@@ -38,56 +54,107 @@ pub enum StateEl {
...
-        StateEl::StrQuote => style.fg_color = 0xFF998844,
-        StateEl::CharQuote => style.fg_color = 0xFF998844,
-        StateEl::Invalid => style.fg_color = 0xFFFF0000,
-        StateEl::NumericLiteral => style.fg_color = 0xFF6644EE,
-        StateEl::CharConst => style.fg_color = 0xFF8866EE,
-        StateEl::Keyword => style.font = 1,
-        StateEl::Operator => style.fg_color = 0xFFAA2244,
-        StateEl::PrimType => style.fg_color = 0xFF44AAAA,
+impl StateEl {
+    /// Transform a [StateEl] into a [Vec] of Syntect Scope [String]s.
```
cmyr: out of date?

rust/experimental/lang/src/main.rs
```diff
@@ -14,142 +14,277 @@
...
-mod colorize;
-mod statestack;
+mod language;
+mod parser;
 mod peg;
+mod statestack;
+
+const LINES_PER_RPC: usize = 50;
+
+type Scope = Vec<String>;
```
cmyr: this is less useful now that we're using integer id's. I don't think it's necessary now, but if we were polishing this i might suggest `ScopeId` typedef.

rust/experimental/lang/src/main.rs
```diff
@@ -14,142 +14,277 @@
...
+    type Cache = StateCache<State>;
+
+    fn update(
+        &mut self,
+        view: &mut View<Self::Cache>,
+        _delta: Option<&RopeDelta>,
+        _edit_type: String,
+        _author: String,
+    ) {
+        let _guard = trace_block("ExperimentalLang::update", &["experimental-lang"]);
```
cmyr: Each trace we record increases the overall size and possibly sample count (we only keep 100k events per source, maybe?) so while tracing is definitely good it's also good to be judicious. Since this call should always return immediately there's not much to learn here; it's more important to trace things that have variable runtimes or that we're otherwise suspicious of, performance wise.

rust/experimental/lang/src/statestack.rs
```diff
@@ -31,7 +30,8 @@ struct Entry<T> {
 }
 
 pub trait NewState<T> {
```
cmyr: if you have context for this, adding a doc comment would probably benefit future archaeologists (I have no idea what's going on here, for instance)

rust/experimental/lang/src/statestack.rs
```diff
@@ -101,3 +97,43 @@ impl<T: Clone + Hash + Eq, N: NewState<T>> Context<T, N> {
         result
     }
 }
+
+pub struct HolderNewState<T> {
```
cmyr: ditto my earlier doc comment... comment, a bit if info here would save me having to actually understand the _code_ 😉

rust/experimental/lang/src/language/rust.rs
mbStavola: Yeah, this can go, but I'm going to keep the scope reference link around-- probably attach it to `ALL_SCOPES`

rust/experimental/lang/src/main.rs
mbStavola: Yeah, good point. Chatty logs (traces?) were helpful while actually getting this working, but we can cut down on most of them now. I'm going to remove all of the ones in the lifecycle methods, but I'll probably stick one in `compute_syntax`.

rust/experimental/lang/src/statestack.rs
mbStavola: This used to be used to allow the `Context` to either store a `Style` or some debug info... I'm not sure it really _should_ be around anymore... I'm pretty sure I can just replace the `HolderNewState` usage with a top of stack query.

PR #1006, opened 2018-11-21
[WIP] Restore selections on undo/redo

### LINKED ISSUE #458: undo doesn't remember cursor position/selections
It seems pretty common that if you undo past a selection edit it restores the selection.  XiEditor doesn't seem to do that.

Steps to reproduce:
1. Select a block of text
2. Delete
3. Undo

Expected:
Deleted text re-appears selected

Actual:
Deleted text re-appears but not selected.  In fact cursor position doesn't change ever on undo.

cmyr:
This is a quick stab at #458; at the very least it needs some cleanup/docs/clone-removal. Mostly wanted to open an early PR so other people could play around with it. To whit:

@nangtrongvuon I'm curious if this solves the last problems with #849?

@lord So for this to work I basically have to thread the `EditType` through to the view. Maybe this means we could just use `EditType` instead of `InsertDrift`? In any case i have some larger thoughts about the fundamental problem with trying to get "selection adjustment after edit" right; will try to write that up soon.

(closes #458, eventually, maybe)

## Review Checklist


lord:
Sounds good to me! Probably actually even cleaner than having a hard coded map of which edits types have which drifts, but also maybe worth documenting at some point how each of the edit types work, since otherwise it could be confusing about what exactly is the difference between them? Unless somebody reads through the code closely.

cmyr:
Yea, The `EditType` (and undo grouping in general) has always seemed like a bit of a wart, not sure what the best fixup should be. Documentation certainly helps.

cmyr:
This is currently blocked on some correctness issues when handling async edits to an existing undo group; shouldn't be impossible to get right but my focus has been elsewhere.

PR #1009, opened 2018-11-23
Use previous line for auto-indent after newline

scholtzan:
## Summary
When hitting enter at the beginning of the line, the line gets indented. This can result in some weird behaviour, for example when moving closing brackets:

out

This happens because auto-indent is looking at the indentation of the previous non-blank line. I am not entirely sure if there is a good reason to not just choose the previous line. 

In this PR, auto-indent is changed to use the previous line to determine the indentation level which results in the expected behaviour. 

There are still a few usages of `previous_nonblank_line`. I am not entirely sure if they could also be replaced with `previous_line`, but when I did some testing, I didn't notice any weird behaviour. So maybe `previous_nonblank_line` can be removed altogether?

## Review Checklist

### DIFF
--- a/rust/syntect-plugin/src/main.rs
+++ b/rust/syntect-plugin/src/main.rs
@@ -277,11 +277,13 @@ impl<'a> Syntect<'a> {
         let tab_size = view.get_config().tab_size;
         let current_indent = self.indent_level_of_line(view, line);
         let base_indent = self
-            .previous_nonblank_line(view, line)?
+            .previous_line(view, line)?
             .map(|l| self.indent_level_of_line(view, l))
             .unwrap_or(0);
+
         let increase_level = self.test_increase(view, line)?;
         let indent_level = if increase_level { base_indent + tab_size } else { base_indent };
+
         if indent_level != current_indent {
             self.set_indent(view, line, indent_level)
         } else {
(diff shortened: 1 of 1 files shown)


cmyr:
The problem is that this won't work if you have blank lines with no whitespace, which is common.

The problem here, as I understand it: we have two checks that we do, one on 'newline', which figures out the subsequent indentation level, and one on normal edits which figures out if we should decrease the indentation level.

In your example, when we insert the newline we check what the indentation level would normally be for a new line in this situation, which is that it would be one 'tab'. However we don't check whether the text on the new line should have it's level _decreased_. If we did, these would cancel out, and the indentation level would be correct.

So it's a subtle one. I think we want to keep `previous_nonblank_line` around, though. This points out one of the more annoying problems with this implementation, which is that it's basically impossible to write tests for, as written. When we implement `reindent` that might let us do some better tests here? 😕🤞

scholtzan:
Okay, I see. 
I was playing around with what you are describing. When checking whether the level should be decreased, it ended up almost always decreasing.

rust/syntect-plugin/src/main.rs
```diff
@@ -277,25 +277,28 @@ impl<'a> Syntect<'a> {
...
         let current_indent = self.indent_level_of_line(view, line);
         let base_indent = self
-            .previous_nonblank_line(view, line)?
+            .previous_line(view, line)?
             .map(|l| self.indent_level_of_line(view, l))
             .unwrap_or(0);
+
         let increase_level = self.test_increase(view, line)?;
         let indent_level = if increase_level { base_indent + tab_size } else { base_indent };
+
```
scholtzan: If I test here whether it should be decreased with `test_decrease` it returns an error (`error in autoindent PeerDisconnect`). The reason for this seems to be that at some point in `get_metadata` were are trying to get the `line` with `view.get(line)`. View seems to access some cache with the lines but cannot find `line` and returns an error. I am not sure if that is the expected behaviour?

scholtzan:
I digged a bit deeper into the auto-indent stuff, specifically checking whether we should decrease the indentation after newline. When I was trying to do this I encountered some strange problem. It seems after entering newline, the new line is not yet cached in the plugin's view and therefore no metadata for this line can be retrieved. I am not sure if this is actually the expected behaviour?

PR #1010, opened 2018-11-23
[RFC] Annotations (update protocol changes)

cmyr:
This is a proposal for _annotations_, a new core feature that would allow things like inline error display, version control status, and any other sort of region-based metadata.

cc @scholtzan, who has already started work on this, as well as @mmatoszko and @rakaramos, who have expressed interest in frontend work for some of the related features.

### [rendered](https://github.com/xi-editor/xi-editor/blob/46fe750f3156317f459a0d8073eae3283b7679e8/rfcs/2018-11-23-annotations.md)


rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+frontend, without using the text styling API.
+
+This API will be the basis for a wide variety of future features,
+including things like:
+
+- compiler warnings / lints / diagnostic messages
+- displaying other user's cursors during multi-user editing
+- clickable external links
+- version control status
+- breakpoints
```
raphlinus: Are annotations the mechanism for conveying autocompletion suggestions to the front-end? I think so. In either case, it's good to be explicit. Obviously the annotation type is more complex than, say, a warning, so fleshing that out will be future work.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+{
+    "ops": [Op]?,
+    "annotations": [AnnotationSet]?,
+    // other fields
+}
+
+// An AnnotationSet:
+{
+    "type": String,
+    "ranges": [[Number, Number]],
```
raphlinus: My understanding is that we don't use byte ranges in the update protocol. I believe when we discussed this, we were considering making the endpoints of a range be (visual) line number and offset-within-line, where the latter could reasonably be utf-8 offset, utf-16, or possibly both.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+    "payloads": [Any]?,
+    "metadata": Any? // maybe?
+}
```
raphlinus: Another thing we discussed, which I think would be useful to include here, is a total count of the annotations of a given type (whether or not they intersect the visible region).

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+struct Annotations {
+    items: Spans<Value>,
+    type_id: AnnotationType,
+}
+
+/// A region of an `Annotations`; this is the type we put on the wire.
+struct AnnotationSlice {
+    type: AnnotationType,
+    /// guaranteed non-descending start order
+    ranges: Vec<(usize, usize)>,
```
raphlinus: FWIW, here I do believe byte ranges are appropriate.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+corresponding schemas as part of the xi-editor project, and more may be
+added in the future; but new types could also be defined by a plugin,
+and as long as they're supported by a frontend that's fine.
+
+Some examples of possible types and their possible payloads:
+
+**'`selection`'**: The payload for a selection is simply the location in
+the selection of the cursor. This might be represented by a number in
+`[0, 1, 2]`, corresponding to 'start', 'end/upstream', or
+'end/downstream'.
```
raphlinus: The upstream/downstream distinction is important for both start and end positions. Interestingly, it could be subsumed by using line/offset instead of byte range - "upstream" is the end of line n, while "downstream" is offset 0 of line n+1.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+variety_
+
+### plugin API
+
+Plugins add, update, and remove annotations with a single protocol
+method:
+
+```javascript
+{
+    "method": "update_annotations",
```
raphlinus: This update needs to be based on a rev, as annotations will be subject to OT if the document has been edited concurrently with the annotation being in flight.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+to the frontend to display as desired.
+
+## Questions / considerations
+
+A number of things have come up while drafting this proposal that are
+worth noting.
+
+#### metadata handling
+
+We probably want (and have assumed having,elsewhere in this proposal)
```
raphlinus: typo nit: space

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+some way to associate metadata with a _set_ of annotations;
+for instance 'find' annotations might include the total
+number of results and their approximate locations. This does add some
```
raphlinus: Right, here's the discussion of total count. My personal sense is that it makes sense for it to be part of the generic update protocol and not type-specific metadata, but I can be talked out of it.

rfcs/2018-11-23-annotations.md
```diff
@@ -0,0 +1,406 @@
...
+want to either destroy modified annotations or to expand them, like
+selections; but we may want to have an 'update strategy' that is
+determined by the annotation's type.
+
+#### Annotations for inter-plugin communication
+A far future concern: but we may want to explore whether annotations
+also make sense as a mechanism for plugins to communicate metadata
+to one another.
+
+#### Sharing annotations between views
```
raphlinus: I'm now curious how other editors handle this. Some annotations are clearly per-view (selection, find, autocompletion suggestions). Others I can imagine being per-document (eg breakpoints). I'm curious how other editors handle this; if you have a document open and do "new view", what stuff shows in the new view?

rfcs/2018-11-23-annotations.md
cmyr: I think that autocomplete is probably too stateful to really fit into this system, as I've been imagining it. Once sent, annotations should be valid until they are invalidated by an edit or by the provider; autocomplete wants to be invalidated by cursor movements, and are just in general far more transient; I think it needs its own API.

jansol: Do they really have to be invalidated by cursor movement? The client could just hide them when they aren't relevant. And if there are edits they should indeed be invalidated anyway so the position of the cursor doesn't matter at that point.

rfcs/2018-11-23-annotations.md
cmyr: Okay, I don't fully grok the upstream/downstream logic, so this might be a poor example.

rfcs/2018-11-23-annotations.md
cmyr: Even if we include 'total count' as a general parameter, we would still likely want generic metadata for other purposes; for instance find may want to indicate which of the `n` total results is currently selected. I don't feel like I have a particularly clear answer here, right now; the metadata question will probably need to get hashed out further as we explore concrete use cases.

rfcs/2018-11-23-annotations.md
cmyr: All I mean here is that if an autocomplete window is up and the user clicks elsewhere, autocomplete goes away; 'hiding' and 'invalidating' are equivalent under this model, since the client just draws what we send it.

cmyr:
I'm going to merge this; I had hoped to do a bit more revising but I think it's more important that we get this in, being as it is already being implemented.

PR #1011, opened 2018-11-25
tail toggle?

sjoshid:
PR for #922


rust/Cargo.toml
```diff
@@ -36,4 +36,4 @@ members = [
 ]
 
 [profile.bench]
-lto = true
+lto = true
```
jansol: stray edit

rust/core-lib/src/tabs.rs
```diff
@@ -689,7 +689,7 @@ impl CoreState {
                     .find(|v| v.borrow().get_buffer_id() == buffer_id)
                     .map(|v| v.borrow().get_view_id())
                     .unwrap();
-                self.make_context(view_id).unwrap().reload(text);
+                self.make_context(view_id).unwrap().reload_tail(text);
```
jansol: We want to keep the old reload behavior for files that are not being tailed, no?

sjoshid: Good point. But the method containing this `handle_open_file_fs_event` is tagged as notify. Besides, we monitor four notify events: NoticeWrite, Write, Chmod and Create. `reload_tail` will handle all four correctly unless Im missing a test case.

cmyr:
What's the status on this? @sjoshid are you waiting for review? Apologies if so, sort of lost track here. 😅

sjoshid:
@cmyr 
At this point we need that review you talked about. Let me know if you see issues.

cmyr:
@sjoshid okay cool, will get to this shortly. :)

rust/core-lib/src/file.rs
```diff
@@ -42,6 +44,8 @@ pub struct FileManager {
     /// A monitor of filesystem events, for things like reloading changed files.
     #[cfg(feature = "notify")]
     watcher: FileWatcher,
+    #[cfg(feature = "notify")]
```
cmyr: I don't think this needs to be behind the feature flag; if the feature doesn't exist this will still compile fine, it just won't be used (which is fine).

Edit:

Also, I'm not sure if this makes sense here? If this is on it would mean that it's on for all files?

rust/core-lib/src/file.rs
```diff
@@ -136,12 +178,23 @@ impl FileManager {
     fn save_new(&mut self, path: &Path, text: &Rope, id: BufferId) -> Result<(), FileError> {
         try_save(path, text, CharacterEncoding::Utf8)
             .map_err(|e| FileError::Io(e, path.to_owned()))?;
+
+        #[cfg(feature = "notify")]
+        let new_tail_details = TailDetails {
```
cmyr: This looks like you could `derive(Default)` for this type, and then just do `TailDetails::default()`.

rust/core-lib/src/file.rs
```diff
@@ -160,11 +213,74 @@ impl FileManager {
...
+    P: AsRef<Path>,
+{
+    let mut f =
+        File::open(path.as_ref()).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+    let end_position =
+        f.seek(SeekFrom::End(0)).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+
+    let diff = end_position - current_position;
+    let mut buf = vec![0; diff as usize];
+    f.seek(SeekFrom::Current(-(buf.len() as i64))).unwrap();
```
cmyr: this could fail; we should handle the error.

rust/core-lib/src/rpc.rs
```diff
@@ -175,12 +175,19 @@ pub enum CoreNotification {
     /// ```
     Plugin(PluginNotification),
     /// Tells `xi-core` to close the specified view.
-    CloseView { view_id: ViewId },
+    CloseView {
```
cmyr: I think these changes are unintended?

rust/core-lib/src/tabs.rs
```diff
@@ -319,6 +319,7 @@ impl CoreState {
             // handled at the top level
             ClientStarted { .. } => (),
             SetLanguage { view_id, language_id } => self.do_set_language(view_id, language_id),
+            ToggleTail { enabled } => self.file_manager.tail_toggle = enabled,
```
cmyr: I think we want this to be a per-view option; so it should include a `ViewIdentifier`, and should apply to that particular view, not to all files.

rust/core-lib/src/tabs.rs
cmyr: I think there's a misunderstanding, here. I'm going to explain the 'notify' tag in my top level comment; but in this function what we should be doing, I think, is checking if the file is marked as being in 'tail mode', and if it is we call `reload_tail()` instead of `reload()`. We don't want to change anything about the existing reload behaviour, which should still continue to work.

rust/core-lib/src/tabs.rs
```diff
@@ -678,8 +679,9 @@ impl CoreState {
         //TODO: currently we only use the file's modification time when
         // determining if a file has been changed by another process.
         // A more robust solution would also hash the file's contents.
+        let tail_toggle = self.file_manager.tail_toggle;
 
-        if has_changes && is_pristine {
+        if has_changes && is_pristine && tail_toggle {
```
cmyr: This should be called even if `tail_toggle` is false (and `tail_toggle` should probably be a per-file setting).

rust/core-lib/src/file.rs
```diff
@@ -136,12 +178,23 @@ impl FileManager {
     fn save_new(&mut self, path: &Path, text: &Rope, id: BufferId) -> Result<(), FileError> {
         try_save(path, text, CharacterEncoding::Utf8)
             .map_err(|e| FileError::Io(e, path.to_owned()))?;
+
+        #[cfg(feature = "notify")]
+        let new_tail_details = TailDetails {
+            current_position_in_tail: 0,
```
cmyr: I don't think we should need to track this explicitly? Unless I'm misunderstanding, this should always be the same as the current length of the text we have loaded?

rust/core-lib/src/file.rs
```diff
@@ -136,12 +178,23 @@ impl FileManager {
...
+            is_tail_mode_on: false,
+            is_at_bottom_of_file: false,
```
cmyr: this might also be something we can determine from the state of the view; if `view.first_line + view.height` == the total number of lines, we know the last line is visible.

rust/core-lib/src/file.rs
```diff
@@ -160,11 +213,74 @@ impl FileManager {
             return Err(FileError::HasChanged(path.to_owned()));
         } else {
             let encoding = self.file_info[&id].encoding;
+            #[cfg(feature = "notify")]
+            self.watcher.unwatch(&path, OPEN_FILE_EVENT_TOKEN);
```
cmyr: This unwatch/watch shouldn't be necessary; this is why we stash the mod time of our own writes, and ignore the corresponding events on load.

rust/core-lib/src/editor.rs
```diff
@@ -179,6 +179,14 @@ impl Editor {
         self.set_pristine();
     }
 
+    pub fn tail_append(&mut self, tail: Rope) {
```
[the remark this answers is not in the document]
sjoshid: Done.

rust/core-lib/src/file.rs
```diff
@@ -84,6 +95,14 @@ impl FileManager {
         self.file_info.get(&id)
     }
 
+    #[cfg(feature = "notify")]
+    pub fn get_current_position_in_tail(&self, id: &BufferId) -> u64 {
+        match self.file_info.get(id) {
+            Some(v) => v.tail_details.current_position_in_tail,
+            None => 0,
```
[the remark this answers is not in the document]
sjoshid: This isn't the only case where a key lookup is made. Im going to lookup elsewhere in code to check how None (when key is not present in map) is handled in Xi.

rust/core-lib/src/file.rs
sjoshid: Since we don't want to have a global tail flag, Im going to remove this. Moving it to `FileInfo`.

rust/core-lib/src/file.rs
sjoshid: Done.

rust/core-lib/src/file.rs
sjoshid: Agree.

rust/core-lib/src/tabs.rs
sjoshid: This is something I wasnt sure about when I implemented. I was going to implement per view tail but then thought if a file isn't active, no events should be triggered, including `Write` event. I need to handle a case where an inactive tab becomes active.

rust/core-lib/src/file.rs
```diff
@@ -28,6 +28,8 @@ use xi_rpc::RemoteError;
 
 use tabs::BufferId;
 
+use std::io::Seek;
+use std::io::SeekFrom;
```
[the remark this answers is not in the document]
sjoshid: Yup.

rust/core-lib/src/tabs.rs
sjoshid: This is clear to me now. I will bring back `reload`.

rust/core-lib/src/file.rs
sjoshid: Correct. It should be length of text.

rust/core-lib/src/file.rs
sjoshid: This is much better than current implementation. Thanks for suggestion. :)

rust/core-lib/src/file.rs
sjoshid: > ignore the corresponding events on load.
Just to confirm this, we dont want to handle events after I modify and save a tailed file? I tried this in Vscode and it does support the behavior.

sjoshid:
@cmyr 
Thanks for detailed review and your patience. Im going to make necessary changes. You talked about other changes you haven't mentioned. Let me know what they are so I include them as well.

rust/core-lib/src/file.rs
cmyr: I'm not totally sure I understand; we want to still receive events after modifying the file, but that's the current behaviour.

rust/core-lib/src/file.rs
cmyr: thinking about this more, getting the total number of visual lines in the document might be slightly complicated right now; if you get stuck let me know.

rust/core-lib/src/tabs.rs
cmyr: we receive events regardless of whether we are focused on a view. 👍

cmyr:
I'd like to focus more on just the overall approach for now; I suspect there are some fairly big changes to make, so I'll do another round of review then. :)

rust/core-lib/src/file.rs
sjoshid: This is reverted. My test case was wrong.

rust/core-lib/src/file.rs
```diff
@@ -106,11 +125,27 @@ impl FileManager {
             let _ = File::create(path).map_err(|e| FileError::Io(e, path.to_owned()))?;
         }
 
+        #[cfg(feature = "notify")]
```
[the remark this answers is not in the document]
sjoshid: I have reverted the entire `open` method. Moving tailing login in `try_load_file` which check if file is being tailed or not. Is this what is expected?

rust/core-lib/src/tabs.rs
sjoshid: Implemented per-view option. Client side changes are pending.

rust/core-lib/src/file.rs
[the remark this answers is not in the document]
sjoshid: Removed this method as part of refactoring.

rust/core-lib/src/file.rs
```diff
@@ -254,38 +254,46 @@ fn try_load_file<P>(file_manager: &FileManager, buffer_id: BufferId, path: P) ->
...
-                f.read_exact(&mut bytes).unwrap();
-
-                new_tail_details = TailDetails {
-                    current_position_in_tail: end_position,
-                    is_tail_enabled: true,
-                    is_at_bottom_of_file: true,
-                };
-            } else {
-                debug!("Tail is false, So loading entire file.");
+    #[cfg(feature = "notify")]
```
sjoshid: `try_load_file` is common for fuchsia and non-fuchsia. Is this correct way to handle this?

rust/core-lib/src/tabs.rs
```diff
@@ -319,6 +319,7 @@ impl CoreState {
             // handled at the top level
             ClientStarted { .. } => (),
             SetLanguage { view_id, language_id } => self.do_set_language(view_id, language_id),
+            #[cfg(feature = "notify")]
```
sjoshid: Making a check here since fuchsia does not have notify.

sjoshid:
Is there any plan to use a different third party crate instead of notify for fuchsia? I ask because if that is the case I can remove compiler checks as the underlying logic remains the same.

rust/core-lib/src/tabs.rs
cmyr: So I think the best way to handle this is to remove the `cfg` here and above, and instead have two versions of `do_toggle_tail`; one of them is exactly what you have right now, with `#[cfg(feature = "notify")]`, and the other one looks like,

```rust
 #[cfg(not(feature = "notify"))]
fn do_toggle_tail(&mut self, _view_id: ViewId, _enabled: bool) {
    warn!("do_toggle_tail called without notify feature enabled.");
}
```

You can see another example of this pattern (without the `warn!` macro) here: https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/tabs.rs#L671

rust/core-lib/src/file.rs
cmyr: Well how I'd think about this is, what in the following block works or doesn't work without the `notify` feature? From just skimming the code, it looks like there's no use of `notify` at all; We're just referencing some types that are themselves hiding behind `cfg(feature = "notify")`. Is this necessary? For instance, we could _always_ have `tail_details` exist, even if notify was disabled; it just wouldn't do anything, and I think that's probably fine?

sjoshid:
@cmyr 
Apologies for the crickets here. Resumed work on finding the bottom of file for tailing. Im planning to use rope's line metrics to get the max line and check if that is <= (height + first) aka last line of view. `do_edit` of view does get a rope which can be used to find max line. 
You did mention in one of your previous comments this might get complicated. So I might be missing something here.

cmyr:
The complication is that the lines in a view are not necessarily all 'hard' lines, e.g. end in a newline; some of them could be wrapped lines.

To know if we're currently at the bottom of the file, we want to know if the visual line containing EOF is in the visual region.

So we want to figure out the last visible line, with like `view.line_of_offset(&text, text.len())`, and then compare this to height + first_line.

It's important that we go through the view, and not just the rope, because the rope only knows about newlines but the view knows about line breaks.

sjoshid:
@cmyr 
There is one concern I would like to highlight here: What is the expected behavior when
tail mode is enabled and EOF is NOT visible? We discussed this in the issue thread but it
was too soon to discuss then. 

**Option 1: Scroll position**
This is easier. Keep tailing and building up the rope as usual but keep the 
scroll position constant. When EOF becomes visible keep changing the scroll position
to end. This will handle the case where events might have stopped coming when EOF was
not visible. 

**Option 2: Maintain a rope buffer**
When EOF is not visible instead of reloading we keep appending rope delta to this 
buffer. (Or I could not change the cursor position stored in tail details but using it
 means opening a file and reading from it). When EOF is visible append the buffer to 
existing rope. 

These ideas are at a very high level. But I wanted your opinion before I move further. 
Let me know if you have something else in mind. 
I might be going on a tangent. ¯\_(ツ)_/¯

cmyr:
I think when EOF is not visible we should behave as we would if not tailing; we reload the buffer as normal on change, but don't change the viewport. This shouldn't be too hard, since it's how things work currently?

sjoshid:
> we reload the buffer as normal on change, but don't change the viewport

Got it. 

> This shouldn't be too hard, since it's how things work currently?

Currently the viewport always follows the cursor. It doesn't honor my scroll. It's the `render_if_dirty` method that scrolls the viewport to cursor. I played around with the `render` method (that calls `render_if_dirty)` but no matter what it is always called. 
However, I was able to control the viewport scroll in `render_if_dirty` by checking if EOF is visible. But since `render` is used widely this change will have huge impact I guess. Im thinking is there a less impactful method for what Im trying to do. 
Any leads?

sjoshid:
I remember trying that option. It didnt work because `render` is called from many places. Even if I have my own `render_tail` it still won't stop chasing the cursor. 
I did come up with a hybrid approach which so far is working correctly. I created `render_tail` as you said along with a `tail_enabled` check in `render`. This check will control if the view needs to scroll to cursor. Im making `tail_enabled` false by default which should not impact the existing behavior.

cmyr:
Okay, that sounds reasonable. The other option this maybe suggests is to just check if tail is enabled, in render, and then pick a code path based on that? But no worries, just an idea that occurred to me.

Do you think this is ready for another look?

sjoshid:
That was actually my first attempt. The reason it didnt work is the same: `render` method being called from other places. I had something on the line of:
```
if tail_enabled
    if eof_visible { 
          Scroll to end
     }
else {
    //scroll to cursor just like today
}
```
Now the problem is when I scroll, `render` is called immediately even before view gets a chance to set `eof_visible` to false. This makes scrolling impossible because `eof_visible` is always true and that means view always scroll to the eof. I maybe wrong but it took sometime for me to figure the weird scroll behavior. 

> Do you think this is ready for another look?

Not yet. It will be by this weekend for sure.

rust/core-lib/src/view.rs
```diff
@@ -297,11 +307,13 @@ impl View {
         self.size = size;
     }
 
-    pub fn set_scroll(&mut self, first: i64, last: i64) {
+    pub fn set_scroll(&mut self, text: &Rope, first: i64, last: i64) {
         let first = max(first, 0) as usize;
         let last = max(last, 0) as usize;
         self.first_line = first;
         self.height = last - first;
+        let eof_line = self.line_of_offset(text, text.len()) + 1;
```
sjoshid: Is +1 needed here?

sjoshid:
I think it is ready for review.

rust/core-lib/src/file.rs
```diff
@@ -259,8 +259,8 @@ where
 
                 new_tail_details = TailDetails {
                     current_position_in_tail: end_position,
-                    is_tail_enabled: true,
-                    is_at_bottom_of_file: true,
+                    is_tail_enabled: v.tail_details.is_tail_enabled,
+                    is_at_bottom_of_file: v.tail_details.is_at_bottom_of_file,
```
sjoshid: Would it be better to add a new fn in FileManager to update tail position instead of creating a new one each time?

rust/core-lib/src/view.rs
cmyr: It is if this value is being used as either the "number of lines" or as the end bound of a half open range. That is to say: if a document has 'one' line, `line_of_offset(text.len()) == 0`.

sjoshid: Ok. Then +1 here looks correct.

sjoshid:
@cmyr 
Did you get a chance to review this? I would be happy to make more changes if required. 
xi-mac issue is still open. Im planning to get that ready after this.

cmyr:
Oops sorry no, I haven't taken a look yet. Will do in the next couple days!

rust/Cargo.toml
cmyr: @sjoshid this is unresolved

rust/core-lib/src/file.rs
```diff
@@ -161,29 +172,30 @@ impl FileManager {
...
-        f.metadata().map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?.modified().ok();
-    let mut bytes = Vec::new();
-    f.read_to_end(&mut bytes).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
 
-    let encoding = CharacterEncoding::guess(&bytes);
-    let rope = try_decode(bytes, encoding, path.as_ref())?;
-    let info = FileInfo { encoding, mod_time, path: path.as_ref().to_owned(), has_changed: false };
-    Ok((rope, info))
+    #[cfg(feature = "notify")]
+    pub fn toggle_tail(&mut self, id: BufferId, enabled: bool) -> Result<(), FileError> {
```
cmyr: As a public function, this should have at least a brief bit of documentation.

rust/core-lib/src/file.rs
cmyr: Can you explain more what you mean?

rust/core-lib/src/rpc.rs
cmyr: @sjoshid this is still unresolved

rust/core-lib/src/file.rs
sjoshid: The only difference between `new_tail_details` and old `new_tail_details` is the `current_position_in_tail`. Instead of creating a new `TailDetails` one each time, I was thinking maybe I can just update `current_position_in_tail` of existing one.

rust/core-lib/src/rpc.rs
sjoshid: I see similar changes for other enum types. I think `rustfmt`  did this but if not, I can revert these changes.

rust/Cargo.toml
sjoshid: I can revert this.

rust/core-lib/src/file.rs
sjoshid: I agree. I will add comments.

rust/core-lib/src/rpc.rs
sjoshid: I just confirmed. rustfmt did this change.

rust/core-lib/src/view.rs
```diff
@@ -92,6 +92,10 @@ pub struct View {
 
     /// Annotations provided by plugins.
     annotations: AnnotationStore,
+
+    is_tail_enabled: bool,
+
+    eof_visible: bool,
```
sjoshid: Added comments these two new items too.

sjoshid:
Some changes went in as part of https://github.com/xi-editor/xi-mac/pull/373.

sjoshid:
@cmyr 
How is this looking? Also did you get a chance to see the xi-mac changes for debug config?

sjoshid:
@cmyr 
I have been testing this whenever I get a chance but Im not seeing expected behavior. It is due to the debounced mode of notify. When I bombard the file being tailed, `Write` event is never received. There are couple of ideas suggested at https://github.com/passcod/notify/issues/146 and I think we need those before I can make this a solid feature. 
Can we keep this open till then?

cmyr:
Ah okay, that makes sense, especially on mac; events will be aggregated, and we'll get one at most every 30s.

Happy to leave this open; ping me here or on zulip to discuss!

sjoshid:
@cmyr 
Im happy to report that https://github.com/passcod/notify/pull/183 has been merged in notify. We need to modify our side of notify to handle the new event. 
I feel Im going to get a lot of conflicts when I start work on this again so can we close this PR and open a fresh one? I'll port over all changes I have made so far.

jansol:
Sounds like a sensible course of action.

PR #1014, opened 2018-11-26
[WIP] Session restoration

### LINKED ISSUE #913: State restoration

akxcv:
## Summary

See #939. Closes #913.

This is a _very basic_ prototype of the session persistence part. For now, the only thing saved is the pristine delta.

Delta is saved to a file (`~/.xi-session` for now) every 1 second after any edit.

## Checklist

This checklist will be expanded as we go.

- [ ] Settle on session persistence mechanism
- [ ] Implement session persistence
- [ ] Settle on session restoration mechanism
- [ ] Implement restoration
- [ ] Test with [xi-mac](https://github.com/xi-editor/xi-mac)

## Review Checklist


rust/core-lib/src/session.rs
```diff
@@ -0,0 +1,54 @@
...
+use tabs::ViewId;
+use serde_json::{self, Value};
+
+use std::time::{Duration, Instant};
+use std::collections::HashMap;
+use std::fs::OpenOptions;
+use std::io::Write;
+
+// NOTE: increase delay to notice the effect
+pub(crate) const SESSION_SAVE_DELAY: Duration = Duration::from_millis(1000);
```
cmyr: maybe `Duration::from_secs(1)` is clearer?

rust/core-lib/src/session.rs
```diff
@@ -0,0 +1,54 @@
...
+
+// TODO: is this OK?
+pub(crate) const SESSION_SAVE_MASK: usize = 1 << 26;
```
cmyr: maybe we should declare these all in one place to make it easier to avoid collisions?

rust/core-lib/src/session.rs
```diff
@@ -0,0 +1,54 @@
+use xi_rope::rope::RopeInfo;
```
cmyr: needs a header

rust/core-lib/src/session.rs
```diff
@@ -0,0 +1,54 @@
...
+use xi_rope::delta::Delta;
```
cmyr: you can just import `xi_rope::RopeDelta`, which is an alias for `Delta<RopeInfo>`.

rust/core-lib/src/session.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    pub fn set_view_delta(&mut self, view_id: ViewId, delta: Delta<RopeInfo>) {
+        self.data.views.insert(view_id, BufferData { pristine_delta: delta });
+    }
+
+    pub fn save(&mut self) {
+        // TODO: save somewhere appropriate
+        let mut file = OpenOptions::new().create(true).write(true).open(
+            format!("{}/{}", env!("HOME"), ".xi-session")
+        ).unwrap();
```
cmyr: I think we should skip these unwraps (the json one is fine) but there may be platforms where this doesn't work, or we don't have write access, or there there's no $HOME set.

jansol:
This also needs a rebase at this point.

cmyr:
Closing this; it's still a desired feature, but will come in a different PR.

PR #1018, opened 2018-11-28
Add iter<M: Metric<N>>() method to Cursor

cmyr:
This is an ergonomics improvement; it allows a cursor be used as an Iterator,
which fits naturally into existing API and useage.

## Review Checklist


rust/rope/src/tree.rs
```diff
@@ -790,6 +808,27 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
     }
 }
 
+/// An iterator generated by a [`Cursor`], for some [`Metric`].
+pub struct CursorIter<'c, 'a: 'c, N: 'a + NodeInfo, M: 'a + Metric<N>> {
+    cursor: &'c mut Cursor<'a, N>,
+    __metric: PhantomData<&'a M>,
```
raphlinus: Why two underlines? I think one is standard. I could be wrong though.

cmyr: No better reason than, "I've seen this elsewhere". Looking into this I guess I've only seen that on things that have to be `pub` for some reason? will fix.

cmyr:
I looked into `DoubleEndedIterator` but my understanding is that that doesn't mean that we can go backwards; it means that we can access items from either end, converging on a middle. Could be wrong?

raphlinus:
I think that's right. Well, we can cross that again if we ever need a reverse iterator (maybe just a new `rev_iter`).

Issue #610, opened 2018-04-10
Incremental/async line breaking


raphlinus:
This is a design discussion for changes around line breaking. It's closely related to #565 (width measurement) and #58 (word wrap). I wanted to split out the discussion regarding generation and rendering of soft line breaks from width measurement, as the latter involves the front-end and the former is almost entirely a core consideration.

## Current state

The current prototype of line breaking is (mostly) incremental, but synchronous. Basically, that means that the recomputation of breaks might be inexpensive, as the "small delta" property is preserved, but the computation is done on every edit, before rendering. Incremental computation is not always possible, for example on document load and when changing the wrap width. In the current prototype, wrapping is pretty fast, but incorporating width measurement will slow things down.

To summarize very briefly, there's an optional `Breaks` structure in view. When not present, hard breaks are extracted from the text rope. When present, it overrides those breaks. The `after_edit` method in `View` updates the breaks, keeping them synchronized. The Breaks structure is an instance of rope, which is conceptually equivalent to a sequence of offsets for the breaks.

On edit, operational transformation preserves the breaks outside the edited range, deleting all the breaks inside. Then the `rewrap` method recomputes the breaks inside that range. It's a fairly clever algorithm that tries to touch only the lines that are absolutely needed, ie in the case of a very long paragraph it does _not_ rewrap the entire paragraph.

See [rope science 6](http://google.github.io/xi-editor/docs/rope_science_06.html) for a little discussion of the thinking that went into this algorithm, as well as plans for async.

## Going async

Opening or rewrapping a large document will be a poor experience, as the core will block for potentially very large amounts of time. Our approach is to go async, as is done for syntax highlighting.

In an async model, rewrapping is computed in chunks. For pending chunks, the breaks are not necessarily accurate. The algorithm is eventually consistent, so at some point the set of pending chunks becomes empty. In these intermediate states, the document is still rendered, and interaction is still possible.

Deleting _all_ the breaks for newly inserted text would be a very bad experience, as very long lines would result. Therefore, one change I propose is to separate hard and soft breaks. Hard breaks would be derived from the text rope, as is done in the no-breaks case, and the Breaks data structure would only contain soft breaks. The two sources of breaks would be merged on rendering (and for related computations such as converting between offsets and visual line numbers). One consequence is that there would no longer be two code paths (corresponding to the fact that Breaks is an Option); the non wrapped case would be represented by Breaks simply being empty.

[...]

PR #1021, opened 2018-11-30
Unify width and column wrapping

cmyr:
This reimplements column-based wrapping using the existing width-based
system, which reduces the number of code paths for both creating and updating
breaks by up to 50%!

This is currently _slightly_ hacky; we have to clear the width cache when
switching between word and column wrapping, because column wrapping now
uses the width cache but stores codepoint counts instead of actual
measured widths.

This is temporary; assuming we keep column-based wrapping around at all,
it will make sense in the future to get the 'unit width' for a given
style from the frontend, and use that + UAX#11 to calculate actual widths.

--------

this is an initial step towards #610; I'd like to avoid maintaining two distinct code paths/algorithms when implementing incremental wrapping.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/width_cache.rs
+++ b/rust/core-lib/src/width_cache.rs
@@ -27,7 +27,7 @@ use client::Client;
 /// Internally, it is implemented as an index into the `widths` array.
 pub type Token = usize;
 
-/// A measured width, in points.
+/// A measured width, in px units.
 type Width = f64;
 
 type StyleId = usize;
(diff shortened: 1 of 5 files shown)


jansol:
It would certainly make sense to implement column wrapping in terms of width. For monospaced text there shouldn't be any difference (except correct handling of full-width characters as a nice side effect) and for proportional text column wrapping hardly makes any sense anyway.

rust/core-lib/src/width_cache.rs
```diff
@@ -19,44 +19,93 @@ use std::collections::{BTreeMap, HashMap};
...
-use client::{Client, WidthReq};
+use client::Client;
 
 /// A token which can be used to retrieve an actual width value when the
 /// batch request is submitted.
 ///
 /// Internally, it is implemented as an index into the `widths` array.
 pub type Token = usize;
 
+/// A measured width, in points.
```
raphlinus: I prefer "px units", "points" is macOS specific and misleading (it can easily be confused with 1/72 inch).

rust/core-lib/src/width_cache.rs
```diff
@@ -19,44 +19,93 @@ use std::collections::{BTreeMap, HashMap};
...
+    pub strings: Vec<String>,
+}
+
+/// The response for a batch of [`WidthReq`]s.
+pub type WidthResponse = Vec<Vec<Width>>;
+
+/// A trait for types that provide width measurement. In the general case this
+/// will be provided by the frontend, but alternative implementations might
+/// be provided for faster measurement of 'fixed-width' fonts, or for testing.
+pub trait Measure {
```
raphlinus: Odd for me to prefer more verbose names, but WidthMeasure feels better here. I don't feel strongly, though.

rust/core-lib/src/width_cache.rs
```diff
@@ -19,44 +19,93 @@ use std::collections::{BTreeMap, HashMap};
...
+}
+
+impl Measure for Client {
+    fn measure_width(&self, request: &[WidthReq]) -> Result<WidthResponse, xi_rpc::Error> {
+        Client::measure_width(self, request)
+    }
+}
+
+/// temporary; a stand-in type for testing fixed-width measurement.
+/// This treats each codepoint as having width 1; at the very least we should
```
raphlinus: I wouldn't be so apologetic; this is still a useful measure. I would probably write this as "a measure taking into account unicode width might also be useful."

PR #1032, opened 2018-12-10
Adopt Rust 2018

scholtzan:
Some `cargo fix` magic to adopt Rust 2018.

I did the following steps (in case someone wants to reproduce):
1. Add `#![feature(rust_2018_preview)]` to `lib.rs`
2. Run `cargo +nightly fix --allow-dirty --edition `
3. Remove `#![feature(rust_2018_preview)]`
4. Add `edition = '2018'` to `Cargo.toml`

## Related Issues
https://github.com/xi-editor/xi-editor/issues/1030

## Review Checklist


russelltg:
You should not need the feature flag as rust 2018 is stable now. See https://rust-lang-nursery.github.io/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

jansol:
@russelltg Yes but we generally try to stay compatible with a couple of versions older than the latest stable rust.

raphlinus:
This bumps the Rust requirement to 1.31. With 1.30.1 I get `error: Edition 2018 is unstable and only available for nightly builds of rustc.`

That's not a "no," that's just pointing out that it varies from our usual policy. The `#![feature(rust_2018_preview)]` should be deleted and the README should be updated (it currently says 1.29).

jansol:
CI also needs to be updated in that case.

scholtzan:
Right, `#![feature(rust_2018_preview)]` would have not been necessary. I updated the version in `README`

cmyr:
@scholtzan we specify the rust version in a bunch of places, I would just `grep '1.29.2' from the directory root (there will be at least the two CI scripts, maybe the `run_all_checks` script, etc)

jansol:
Looks like some more `cargo fix` passes are in order, preferrably with the stable 1.31 toolchain.

rust/core-lib/src/view.rs
```diff
@@ -11,25 +11,26 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#![allow(clippy::range_plus_one)]
```
cmyr: I would prefer to just fix whatever this was pointing to?

rust/lsp-lib/src/types.rs
```diff
@@ -37,9 +36,6 @@ pub trait Callable: Send {
     );
 }
 
-//FIXME: this can be removed when this change makes it to stable:
-// https://github.com/rust-lang-nursery/rust-clippy/pull/3321
-#[cfg_attr(feature = "cargo-clippy", allow(boxed_local))]
```
cmyr: 🥂

rust/rope/src/compare.rs
```diff
@@ -13,9 +13,10 @@
 // limitations under the License.
 
 //! Fast comparison of rope regions, principally for diffing.
+#![allow(clippy::cast_ptr_alignment, clippy::unreadable_literal)]
```
cmyr: ditto keeping these attached to specific functions if possible; these lints are useful, and we only want to ignore them intentionally, case by case.

rust/rope/src/multiset.rs
```diff
@@ -356,7 +356,7 @@ impl fmt::Debug for Subset {
                     '+'
                 };
                 for _ in 0..s.len {
-                    try!(write!(f, "{}", chr));
+                    r#try!(write!(f, "{}", chr));
```
cmyr: we can just replace this with a `?` now I think

rust/core-lib/src/view.rs
scholtzan: Yeah, there was one case where changing this would return `RangeInclusive` while `Range` is expected.

cmyr: okay, cool.

scholtzan: I'll open an issue for this once this PR gets merged. I think it's a beginner-friendly refactoring issue

PR #1033, opened 2018-12-11
Word wrap checkpoint 3 (separate hard and soft breaks)

cmyr:
This is hopefully the last checkpoint PR before actually adding incremental word wrap.

This is based off of #1027, which should go in first.

Principally, as of this PR we no longer store hard breaks explicitly, depending instead on the fact that they exist as a function of the rope itself.

The main _code_ in this PR involves how we traverse the rope and the breaks structure in tandem, to correctly compute `offset_of_line` and `line_of_offset`; the latter is simple, and the former comes down to a binary search.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/linewrap.rs
+++ b/rust/core-lib/src/linewrap.rs
@@ -70,10 +70,11 @@ impl Lines {
     }
 
     pub(crate) fn visual_line_of_offset(&self, text: &Rope, offset: usize) -> usize {
-        match self.wrap {
-            WrapWidth::None => text.line_of_offset(offset),
-            _ => self.breaks.convert_metrics::<BreaksBaseMetric, BreaksMetric>(offset),
+        let mut line = text.line_of_offset(offset);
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/linewrap.rs
```diff
@@ -405,63 +407,146 @@ impl<'a> Iterator for VisualLines<'a> {
...
+    cur_line: usize,
+    total_lines: usize,
 }
 
 impl<'a> Iterator for MergedBreaks<'a> {
     type Item = usize;
 
     fn next(&mut self) -> Option<usize> {
-        // TODO: remove when soft and hard are separate
+        if self.text.pos() == self.offset && self.offset != self.text.total_len() {
```
jansol: Looks like `self.offset != self.text.total_len()` is used elsewhere too, maybe turn that into an `at_end()`?

cmyr: yea, good call.

PR #1041, opened 2018-12-15
Incremental word wrap initial impl

### LINKED ISSUE #610: Incremental/async line breaking

cmyr:
Edit: this includes commits from #1033, but maybe that should just get closed.

This is less ambitious than I'd initially planned; I really ended up getting stuck getting the partial invalidation correct, so this leaves the current behaviour in place where we invalidate everything on edit when wrap is turned on. I hope to revisit that soon, but this seemed like a good checkpoint.

There is clear follow-up here: I'd like to still do longest line tracking and logical vs visual lines.

There's also a major grossness where wrapping in a big document will cause the viewport to move after wrapping appears to be finished, because we're wrapping above the viewport and it changes offsets. But this is tractable.

In any case I think this is a good foundation, and likely worth merging as-is, but I expect to test it more over the next couple of days.

closes #610 
progress on #937 

## Review Checklist

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -893,7 +893,6 @@ impl View {
         let visible = self.first_line..self.first_line + self.height;
         let inval = self.lines.rewrap_chunk(text, width_cache, client, spans, visible);
         if let Some(InvalLines { start_line, inval_count, new_count }) = inval {
-            debug!("replaced {} breaks with {} at {}", inval_count, new_count, start_line);
             self.lc_shadow.edit(start_line, start_line + inval_count, new_count);
         }
     }
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -23,7 +23,7 @@ use crate::client::Client;
 use crate::edit_types::ViewEvent;
 use crate::find::{Find, FindStatus};
 use crate::line_cache_shadow::{self, LineCacheShadow, RenderPlan, RenderTactic};
-use crate::linewrap::{Lines, VisualLine, WrapWidth};
+use crate::linewrap::{InvalLines, Lines, VisualLine, WrapWidth};
 use crate::movement::{region_movement, selection_movement, Movement};
 use crate::rpc::{FindQuery, GestureType, MouseAction, SelectionModifier};
 use crate::selection::{Affinity, InsertDrift, SelRegion, Selection};
@@ -188,13 +188,17 @@ impl View {
         self.pending_render
     }
 
-    pub(crate) fn update_wrap_settings(&mut self, wrap_cols: usize, word_wrap: bool) {
+    pub(crate) fn update_wrap_settings(&mut self, text: &Rope, wrap_cols: usize, word_wrap: bool) {
         let wrap_width = match (word_wrap, wrap_cols) {
             (true, _) => WrapWidth::Width(self.size.width),
             (false, 0) => WrapWidth::None,
             (false, cols) => WrapWidth::Bytes(cols),
         };
-        self.lines.set_wrap_width(wrap_width);
+        self.lines.set_wrap_width(text, wrap_width);
+    }
+
+    pub(crate) fn needs_more_wrap(&self) -> bool {
+        !self.lines.is_converged()
     }
 
     pub(crate) fn do_edit(&mut self, text: &Rope, cmd: ViewEvent) {
@@ -615,6 +619,7 @@ impl View {
         style_spans: &Spans<Style>,
     ) -> Vec<isize> {
         let mut rendered_styles = Vec::new();
+        assert!(start <= end, "{} {}", start, end);
         let style_spans = style_spans.subseq(Interval::new(start, end));
 
         let mut ix = 0;
@@ -885,7 +890,12 @@ impl View {
         spans: &Spans<Style>,
     ) {
         let _t = trace_block("View::rewrap", &["core"]);
-        self.lines.rewrap_chunk(text, width_cache, client, spans);
+        let visible = self.first_line..self.first_line + self.height;
+        let inval = self.lines.rewrap_chunk(text, width_cache, client, spans, visible);
+        if let Some(InvalLines { start_line, inval_count, new_count }) = inval {
+            debug!("replaced {} breaks with {} at {}", inval_count, new_count, start_line);
+            self.lc_shadow.edit(start_line, start_line + inval_count, new_count);
+        }
     }
 
     /// Updates the view after the text has been modified by the given `delta`.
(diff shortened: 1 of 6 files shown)


rust/core-lib/src/view.rs
```diff
@@ -900,16 +909,12 @@ impl View {
...
-        if !self.lines.has_soft_breaks() {
-            let (iv, new_len) = delta.summary();
-            let start = self.line_of_offset(last_text, iv.start());
-            let end = self.line_of_offset(last_text, iv.end()) + 1;
-            let new_end = self.line_of_offset(text, iv.start() + new_len) + 1;
-            self.lc_shadow.edit(start, end, new_end - start);
-        } else {
-            //TODO (really) actual minimal invalidation, based on what wrapped.
-            self.set_dirty(text);
+        let visible = self.first_line..self.first_line + self.height;
```
raphlinus: Personally I would find spaces around the `..` clearer, as it binds more loosely than the `+`.

jansol: Looks like there's another case like this a few lines earlier. This sounds like a rustfmt change though, so fixing it would likely affect other parts of the code. Adding parentheses around the `+` term might help here though.

cmyr: yea my impression as well was that rustfmt wouldn't let me, bu it seems to be fine. I don't have much of a preference either way?

jansol: Reformatting this would definitely make it a good bit easier to scan. Other than that I'd say this is good to go.

jansol: Yeah we'll want to make sure the topmost visible line stays where it is in the view. That should be quite doable once we get line numbers to updates? Additionally I think rewrapping should prioritize the visible chunk a lot more, in sqlite.c I see quite a lot of nothing when I resize and scroll.

But yes, it's a great improvement and I was assuming this to be another checkpoint like the previous stuff :)

cmyr: it's a bit tricker than just line numbers, because the line cache operates on visible lines; when a wrap happens earlier in the document we need to essentially send the client the delta in the number of earlier lines, so that it can manually adjust the current position.

cmyr: And can you clarify what you mean by "quite a lot of nothing"? are you being left in an empty region?

jansol: Yes, basically everything except the old visible chunk is empty when resizing & scrolling in quick succession.

Hmm do you mean the cache operates on logical (wrapped) lines? That would indeed make it a bit tricky.

cmyr: In my usage logical lines = 'real' lines; delineated by 0, EOF, or newline boundaries. Visual lines = 'wrapped' lines, which may not be delineated in this way.

And yes, the cache works on visual (wrapped) lines; otherwise the client would need to know where to insert the breaks.

"Empty" means that it is not rendered? Is this on a debug build? (That shouldn't happen)

jansol: Ah, so it does show me the region where it *thinks* I should be right after resizing but that moves around the document while staying "static" in the view, hence empty regions when scrolling. Also it seems like I can't really interact with the text field while it rewraps? (select text etc) Hard to say for sure since it keeps warping around though.

cmyr: so what it's doing is staying on the current visual line number, which changes as the wrap happens. I think interaction should be possible, you would have to artificially slow down the wrapping to figure that out. When we're not jumping around that should be doable.

cmyr: update: rustfmt _does_ complain about spaces around `..`

PR #1042, opened 2018-12-15
Change delta_rev_head to try_delta_rev_head

### LINKED ISSUE #455: crash: 'base revision not found'
Had trouble reproducing in a smaller file, but reproducing consistently in jquery.js:

- open document
- select all
- delete
- undo
- immediately start typing

```
thread '<unnamed>' panicked at 'base revision not found', src/libcore/option.rs:819:4
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::option::expect_failed
   9: <core::option::Option<T>>::expect
  10: xi_rope::engine::Engine::delta_rev_head
  11: xi_core_lib::internal::editor::Editor::plugin_update_spans
  12: xi_core_lib::internal::plugins::manager::PluginManager::handle_plugin_notification
  13: <xi_core_lib::internal::plugins::PluginRef as xi_rpc::Handler>::handle_notification
  14: <xi_rpc::RpcLoop<W>>::mainloop::{{closure}}
  15: crossbeam::scoped::scope
  16: <xi_rpc::RpcLoop<W>>::mainloop
  17: xi_core_lib::internal::plugins::start_plugin_process::{{closure}}
```

### LINKED ISSUE #1028: replace Engine::delta_rev_head with try_delta_rev_head return a Result
This function can crash trivially, and is exposed publicly. It should be treated like how we treat `try_edit_rev` as of #993, returning a result that can be handled as appropriate by the caller.

Closing this issues closes #455.

camirmas:
## Summary

Fixes a bug which causes the application to crash upon receiving a nonexistent revision. The replacing function, `try_delta_rev_head`, now returns `Error::MissingRevision(rev)` instead of panicking.

Opted to swallow the error when the function is used in `editor.rs`, but am open to other error handling suggestions.

## Related Issues

resolves #1028 
resolves #455 

## Review Checklist

### DIFF
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -305,8 +305,11 @@ impl Editor {
         Some((delta, last_text, drift))
     }
 
-    pub(crate) fn delta_rev_head(&self, target_rev_id: RevToken) -> RopeDelta {
-        self.engine.try_delta_rev_head(target_rev_id).unwrap()
+    /// Attempts to find the delta from head for the given `RevToken`. Returns
+    /// `None` if the revision is not found, so this result should be checked if
+    /// the revision is coming from a plugin.
+    pub(crate) fn delta_rev_head(&self, target_rev_id: RevToken) -> Option<RopeDelta> {
+        self.engine.try_delta_rev_head(target_rev_id).ok()
     }
 
     #[cfg(not(target_os = "fuchsia"))]
@@ -887,6 +890,8 @@ impl Editor {
                 }
                 start = new_start;
                 end_offset = transformer.transform(end_offset, true);
+            } else {
+                error!("Revision {} not found", rev);
             }
         }
         let iv = Interval::new(start, end_offset);
--- a/rust/core-lib/src/editor.rs
+++ b/rust/core-lib/src/editor.rs
@@ -286,7 +286,7 @@ impl Editor {
         }
 
         let last_token = self.last_rev_id.token();
-        let delta = self.engine.delta_rev_head(last_token);
+        let delta = self.engine.try_delta_rev_head(last_token).expect("last_rev not found");
         // TODO (performance): it's probably quicker to stash last_text
         // rather than resynthesize it.
         let last_text = self.engine.get_rev(last_token).expect("last_rev not found");
@@ -306,7 +306,7 @@ impl Editor {
     }
 
     pub(crate) fn delta_rev_head(&self, target_rev_id: RevToken) -> RopeDelta {
-        self.engine.delta_rev_head(target_rev_id)
+        self.engine.try_delta_rev_head(target_rev_id).unwrap()
     }
 
     #[cfg(not(target_os = "fuchsia"))]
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/editor.rs
```diff
@@ -879,14 +879,15 @@ impl Editor {
...
-            if !transformer.interval_untouched(Interval::new(start, end_offset)) {
-                spans = spans.transform(start, end_offset, &mut transformer);
+            if let Ok(delta) = self.engine.try_delta_rev_head(rev) {
+                let mut transformer = Transformer::new(&delta);
+                let new_start = transformer.transform(start, false);
+                if !transformer.interval_untouched(Interval::new(start, end_offset)) {
+                    spans = spans.transform(start, end_offset, &mut transformer);
+                }
+                start = new_start;
+                end_offset = transformer.transform(end_offset, true);
```
cmyr: I think it's worth logging the error, since it's still definitely unexpected behaviour and probably indicates programmer or user error _somewhere_.

rust/core-lib/src/editor.rs
```diff
@@ -306,7 +306,7 @@ impl Editor {
     }
 
     pub(crate) fn delta_rev_head(&self, target_rev_id: RevToken) -> RopeDelta {
-        self.engine.delta_rev_head(target_rev_id)
+        self.engine.try_delta_rev_head(target_rev_id).unwrap()
```
cmyr: I _think_ it's probably best if we propagate this error up to the caller in this case. It's close, and I'm on the fence, but I think it's likely that future work will involve people using this function more, and I think that it should be communicated that there are cases that could fail.

camirmas: Makes sense. Also I noticed that there aren't any other `Result` types returned from `Editor` functions, is passing along an `Engine::Error` fine in this case or would we prefer to start supplying an `Editor::Error` to convert to?

cmyr: Good question. I agree it doesn't feel great to expose the engine error, and it also doesn't feel great to add a new error type for this single little case.

Another option would be to make this return an `Option<RopeDelta>`, and add a doc comment to this function explaining that it will return `None` if the revision is not found, and results should be checked if the revision is coming from a plugin?

camirmas: Cool so I made that an `Option` and added a doc, if there's some greater need for a new error type it will be easy to change in the future. Also added an error log where it's used in `event_context.rs`, but I'm unfamiliar with `SpecialEvent` handling so not sure if that's the right move

rust/core-lib/src/event_context.rs
```diff
@@ -195,8 +195,15 @@ impl<'a> EventContext<'a> {
                 // The action that follows the block must belong to a separate undo group
                 self.editor.borrow_mut().update_edit_type();
 
-                let delta = self.editor.borrow_mut().delta_rev_head(starting_revision);
-                self.update_plugins(&mut self.editor.borrow_mut(), delta, "core")
+                let res = self.editor.borrow_mut().delta_rev_head(starting_revision);
```
cmyr: I think unwrap here is actually fine; this revision comes from core (we stash it earlier in this fn) so if we can't find it something is seriously wrong.

rust/core-lib/src/editor.rs
cmyr: Okay cool! I think we can actually skip the error handling in `SpecialEvent`, but the other change looks good; should be ready to merge after that little change (I'll do another quick pass now and double check, but I think this is gtg)

Issue #1053, opened 2018-12-17
incremental line breaking follow-up


cmyr:
I'm going to merge #1041 shortly, but there are definitely some rough edges. This is a list of things I'd like to look into improving about the current behaviour:

- [ ] Figure out how not to make things jitter during rewrap. The main problem, here, is that when we rewrap above the visible region we're changing the total number of lines in the linecache, without changing what the frontend believes its current line is. The solution is to tell the frontend to silently adjust the current visual line's number by some +- value. This will require a cooresponding change to the update protocol.

- [ ] performance: we need to tune various constants to keep our RPC duration within an acceptable window. More ambitiously, I'd like to consider making these magic numbers (the number of items in a batch of width requests, the number of lines to wrap per batch, etc) not be constant, but to be tuned based on actual performance; basically we would start with some default value, time each operation, and update the size of the next batch accordingly. This would be helpful because we likely want very different values for these numbers depending on how warm the width cache is.

- [ ] make it so that if the viewport changes while wrap is in progress, we wrap the visible region immediately; the user should see no evidence that incremental wrapping happens, outside maybe of the scroller position changing sometimes.

- [ ] do incremental invalidation when breaks are present; we punted on this but it should be relatively easy to do.

- [ ] _maybe_: 'perfect' invalidation; currently we consider every line in a wrapped region to have been invalidated, even if that particular line might still be valid. For instance: a bunch of short lines that are not wrapped will still be invalidated after batch rewrapping, because we invalidate the entire region for which new breaks were calculated. Doing this correctly will require carefully tracking breaks, but is fairly straightforward, and would be a small win; in most cases we'd cause fewer lines to be invalided.

- [ ] _maybe_: Converge on soft breaks where possible: currently we always rewrap entire paragraphs, because the logic is easier. We should be able to converge on soft breaks where appropriate. (This feels fairly low priority; in the general case rewrapping to paragraph boundaries should be plenty quick, although it will fall down on the single-really-long-line case.)

PR #1054, opened 2018-12-18
Do minimal invalidation with soft breaks enabled

cmyr:
This is a little complicated, because we need different information
after an edit than we do when we're just rewrapping. Most importantly,
when an edit occurs it's possible that the edit deletes existing
breaks; this means we need to count the breaks in the pre-edit
rope and breaks structures in order to handle invalidation correctly.

----

progress on #1053

## Review Checklist

### DIFF
--- a/rust/core-lib/src/linewrap.rs
+++ b/rust/core-lib/src/linewrap.rs
@@ -82,6 +82,21 @@ pub(crate) struct InvalLines {
     pub(crate) new_count: usize,
 }
 
+/// Detailed information about changes to linebreaks, used to generate
+/// invalidation information. The logic behind invalidation is different
+/// depending on whether we're updating breaks after an edit, or continuing
+/// a bulk rewrapping task. This is shared between the two cases, and contains
+/// the information relevant to both of them.
+struct WrapSummary {
+    start_line: usize,
+    /// Total number of invalidated lines; this is meaningless in the after_edit case.
+    inval_count: usize,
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/linewrap.rs
```diff
@@ -82,6 +82,21 @@ pub(crate) struct InvalLines {
...
+/// the information relevant to both of them.
+struct WrapSummary {
+    start_line: usize,
+    /// Total number of invalidated lines; this is meaningless in the after_edit case.
+    inval_count: usize,
+    /// The total number of new (hard + soft) breaks in the wrapped region.
+    new_count: usize,
+    /// The number of new soft breaks.
+    new_soft: usize,
+}
```
jansol: This looks like a C-ism. Is any of the other fields unique to one case? If so, I'd probably turn this into a struct enum just to err on the safe side. If not, maybe consider making `invalid_count` a `Option<usize>`?

cmyr: The problem is  the function we're calling doesn't know which of the two callers it's dealing with. I've taken the approach of just making sure I point out the tricky bits and hopefully explain the situation. Doesn't feel great or anything but felt like the best option. 😞

jansol: Hmm yeah, reading the code it makes a bit more sense to have it be this way.

jansol:
Feels like nice speedup in sqlite3.c already.

PR #1056, opened 2018-12-18
Send real line numbers with updates

### LINKED ISSUE #184: Include line numbers when sending updates from core

cmyr:
With this PR, we will send (and update) logical line numbers when updating the client's line cache.

I wanted to get this finished before trying to fix the rewrap jitter, because this would've caused a related problem: even if the viewport didn't move during rewrap above the visible area, the line numbers would still have changed.

This _kind of_ depends on #1054; it works fine without it, but only because we're constantly invalidating the cache and so don't have to deal with stale updates.

closes #184 
progress on #937 

## Review Checklist

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -498,6 +498,7 @@ interface Op {
   op: "copy" | "skip" | "invalidate" | "update" | "ins"
   n: number  // number of lines affected
   lines?: Line[]  // only present when op is "update" or "ins"
+  ln?: number // the logical number for this line; null if this line is a soft break
 }
 ```
 
@@ -516,7 +517,11 @@ their count; and the editing operations may be more efficiently done in-place
 than by copying from the old state to the new].
 
 The "copy" op appends the `n` lines `[old_ix .. old_ix + n]` to the new lines
-array, and increments `old_ix` by `n`.
+array, and increments `old_ix` by `n`. Additionally, "copy" includes the `ln`
+field; this represents the new logical line number (that is, the 'real' line
+number, ignoring word wrap) of the first line to be copied. **Note:** if the
+first line to be copied is itself a wrapped line, the `ln` number will need to
+be incremented in order to be correct for the first 'real' line.
 
 The "skip" op increments `old_ix` by `n`.
 
@@ -537,6 +542,7 @@ present in the old state is copied at most once to the new state).
 ```
 interface Line {
   text?: string  // present when op is "update"
+  ln?: number // the logical/'real' line number for this line.
   cursor?: number[]  // utf-8 code point offsets, in increasing order
   styles?: number[]  // length is a multiple of 3, see below
 }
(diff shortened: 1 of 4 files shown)


rust/core-lib/src/linewrap.rs
```diff
@@ -681,6 +694,10 @@ impl<'a> MergedBreaks<'a> {
         }
     }
 
+    fn is_hard(&self) -> bool {
```
jansol: This could mayyybe have a slightly more descriptive name.

docs/docs/frontend-protocol.md
```diff
@@ -498,6 +498,7 @@ interface Op {
   op: "copy" | "skip" | "invalidate" | "update" | "ins"
   n: number  // number of lines affected
   lines?: Line[]  // only present when op is "update" or "ins"
+  ln?: number // the logical number for this line; null if this line is a soft break
```
eyelash: What does `this line` refer to here?

cmyr: Right, yea. This is definitely unclear.

This is the logical line number of the first line, if this is a "copy" op. (In theory this would also be the case for "update", but update doesn't exist and with the upcoming annotations work I think update will be removed.)

Something to note: if the first line in this "copy" is itself a continuation/wrapped line, `ln` will still refer to that line's logical number; this means that you'll need to increment `ln` to get the correct value for the first line that has a line number.

eyelash: Generally I'm not a fan of the complexity of the update protocol and I wonder if it is actually necessary. Did anyone ever benchmark the update protocol versus simply rerendering all visible lines for every update? I wouldn't be surprised if it wouldn't make any measurable difference in practice.

cmyr: as in not keeping a cache, and sending lines as needed?

I agree that this is an annoying amount of complexity. I think a longer-term solution might be to implement this all in rust as a library that clients can include.

This is guaranteed present if the op is a `copy`.

I hadn't thought of the situation where copy includes invalid lines; in this case I suspect this will just work, but I'd have to dig a bit.

eyelash: > as in not keeping a cache, and sending lines as needed?

I mean something similar to [what xray does](https://github.com/atom/xray/blob/7df278e3730eb66ee6ee73f39ff52249b326ef01/xray_core/src/buffer_view.rs#L992-L1058).

cmyr: That certainly does have an appealing simplicity. I've tended to defer to @raphlinus  on these types of architectural decisions, but I do think we're starting to bump up against some of the difficulties of our current approach, mostly that it involves a lot of ornate and finicky state tracking that's hard to maintain.

Issue #1059, opened 2018-12-18
Crash in rewrap


jansol:
I was playing around and decided to load the example file linked in the OP of xi-editor/xi-mac#20 and turns out this reliably crashes xi-core:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:355:21
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::panicking::panic
   9: <xi_core_lib::linewrap::MergedBreaks<'a> as core::iter::iterator::Iterator>::next
  10: xi_core_lib::linewrap::Lines::do_wrap_task
  11: xi_core_lib::view::View::rewrap
  12: xi_core_lib::event_context::EventContext::rewrap
  13: xi_core_lib::event_context::EventContext::update_wrap_settings
  14: xi_core_lib::event_context::EventContext::finish_init
  15: xi_core_lib::tabs::CoreState::handle_idle
  16: <xi_core_lib::core::XiCore as xi_rpc::Handler>::idle
  17: xi_rpc::do_idle
  18: xi_rpc::next_read
  19: <std::panic::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  20: std::panicking::try::do_call
  21: __rust_maybe_catch_panic
  22: crossbeam_utils::thread::scope
  23: <xi_rpc::RpcLoop<W>>::mainloop
  24: xi_core::main
  25: std::rt::lang_start::{{closure}}
  26: std::panicking::try::do_call
  27: __rust_maybe_catch_panic
  28: std::rt::lang_start_internal
  29: main
[2018-12-19][03:16:38][xi_rpc][ERROR] panic guard hit, closing runloop
```

cmyr:
closed by #1064

PR #1064, opened 2018-12-21
Fix logic error in Cursor::next

### LINKED ISSUE #1062: Cursor::next with pos == 0 returns None unexpectedly
The current behaviour for Cursor::next is to always return `Some(len)` (where `len` is the length of the tree) when reaching the end of the tree. Whether or not this is the best behaviour is up for debate, but in any case the current behaviour can fail unexpectedly in some cases. The following test case is an illustration,  using as input the file linked in https://github.com/xi-editor/xi-mac/issues/20:

```rust
#[test]
    fn wtf() {
        let s = include_str!("/Users/rofls/Downloads/Jobe_142274871.txt");
        let text = Rope::from(s);
        //text.iter_chunks(..).for_each(|s| eprint!("{}, ", s.len()));
        let x = text.iter_chunks(..).last().unwrap();
        let mut cursor = Cursor::new(&text, 0);
        assert_eq!(cursor.at_or_prev::<LinesMetric>(), Some(0));
        assert!(cursor.get_leaf().is_some());
        assert_eq!(cursor.next::<LinesMetric>(), None); // this should fail, but it passes
        assert_eq!(cursor.pos(), text.len());
        assert!(cursor.get_leaf().is_some()); // this fails, but it should pass?
    }
```

In any case, this is the cause of the crash in #1059.

raphlinus:
It was missing the end of the rope in cases where the tree was large.

Fixes #1062

## Review Checklist

(I'll do some doc updates as a separate PR)

### DIFF
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -668,6 +668,9 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
             // At this point, we know that (1) the next boundary is not not in
             // the cached subtree, (2) self.position corresponds to the begining
             // of the first leaf after the cached subtree.
+            if self.position == self.root.len() {
+                return Some(self.position);
+            }
             self.descend();
             return self.next::<M>();
         } else {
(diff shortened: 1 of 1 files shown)


cmyr:
This also closes #1059.

PR #1065, opened 2018-12-21
Ensure line breaks are up to date after movement

cmyr:
With this PR, we will verify that the visible region has up to date line breaks, creating them if necessary, before rendering the view.

Progress on  #1053

## Review Checklist

- [ ] I have responded to reviews and made changes where appropriate.
- [x] I have tested the code with `cargo test --all` / `./rust/run_all_checks`.
- [x] I have updated comments / documentation related to the changes I made.
- [x] I have rebased my PR branch onto xi-editor/master.

PR #1066, opened 2018-12-21
Keep viewport anchored during incremental rewrap

cmyr:
This introduces the concept of anchors and drift. Well, shift. I was
calling it drift in my head but it's shift in the code, which is less
poetic but more clearly suggests intentionality.

To state the problem: when we're calculating new line breaks, we do it
incrementally; that is, we calculate the breaks for the visible region,
render that immediately, and then incrementally calculate the breaks for
the rest of the document. The problem here comes when the incremental
work is happening above the visible region; this can change the number
of visual lines in the document, which means that the line the client
thinks is line 50 is now line 68; this means the view will jump
unexpectedly as these updates come in.

The solution is to pick an anchor point during rewrapping; in this
commit the anchor point is the offset of the newline character preceding
the first visible line.

We wrap the lines between this anchor point and the end of the view
immediately; all subsequent batches of wrapping happen either above or
below this point. If it happens above, we keep a running tally of the
number of lines added or removed; during the next update, we take this
number (a signed integer) and send it to the frontend as the "shift"
field of the "update" RPC. the frontend then reorients itself by
translating what it thinks its current position is by the "shift" value.
This way the view stays fixed, even as relative line numbers change.

-------

cc @eyelash 
progress on #1053 

This is meant to be used with #1065; it will be a bit jankier without that patch.

There will probably be some fiddling with this around when and where we set the anchor, and how we calculate batch size, but this is the last major bit of the incremental wrap work.

## Review Checklist

### DIFF
--- a/docs/docs/frontend-protocol.md
+++ b/docs/docs/frontend-protocol.md
@@ -493,6 +493,7 @@ update
   ops: Op[]
   view-id: string
   pristine: bool
+  shift?: number
 
 interface Op {
   op: "copy" | "skip" | "invalidate" | "update" | "ins"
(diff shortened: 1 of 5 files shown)


cmyr:
I'm going to close this for now; some further investigation has revealed a major oversight in our approach, which is going to require some rethinking.

Issue #826, opened 2018-09-22
Unify selection commands


ryanbloom:
Originally discussed in #785.

It would be nice to refactor the selection commands so they're not tied to specific input methods. This would be a better way to separate concerns, and potentially make xi more adaptable to new modes of input. Here's another half-baked proposal, based on @nangtrongvuon's comment and some earlier discussion in #159:

```rust
enum Movement {
    Previous, Next, Beginning, End,
    Exact(usize),
}
enum Granularity {
    Point,
    Word,
    Line,
    Paragraph,
    File,
}
enum SelectionChange {
    Replace,   // click or arrow keys
    Add,       // command+click
    Range,     // shift+click or shift+arrows
    Rectangle, // option+shift+click
}
struct Select {
    movement: Movement,
    granularity: Granularity,
    change: SelectionChange,
    drag: bool,
}

// RangeSelect becomes
Select {
    movement: Movement::Exact(123),
    granularity: Granularity::Point,
    change: SelectionChange::Range,
    drag: false
}

// MoveWordLeft becomes
Select {
    movement: Movement::Backward,
    granularity: Granularity::Word,
    change: SelectionChange::Replace,
    drag: false
}

// SelectAll becomes
Select {
    movement: Movement::Exact(0),
    granularity: Granularity::File,
    change: SelectionChange::Replace,
    drag: false
}

// RectangleDrag becomes
Select {
    movement: Movement::Exact(123),
    granularity: Granularity::Point,
    change: SelectionChange::Rectangle,
    drag: true
}
```

I see two main downsides to this sort of approach:

* **Protocol complexity:** This version results in longer commands being sent over the wire. Probably not much of a performance hit, but messages could take more code to construct and be less human-readable.

* **Useless combinations:** The more options we add, the more of these get introduced. As mentioned in #159, "beginning of character" doesn't make much sense. Also, "beginning of word" and "previous word" both correspond to option+left. Drag operations never add a new selection. You can't really make a rectangle with paragraph granularity. Et cetera.

Anyway, there are a lot of cases to consider here and I'm not completely sure that there's an elegant way to handle all of them. Suggestions are very welcome.

cmyr:
The complexity issue is worth thinking about, but I'm not worried about it from a performance standpoint; these are still relatively small messages, and the added cost is going to be in the are of, at worst, handfuls of microseconds.

[...]

PR #1072, opened 2019-01-01
Refactor gestures

ryanbloom:
## Summary
This is part 1 of some refactoring related to the selection commands. I'm focusing on gestures first; hopefully by the time this is merged I'll have a clearer idea of what to do about movements.

### Behavior changes
This is mostly about cleaning up internals, but a few changes will be visible from the frontend.
* Previously, xi would keep existing selections when the you pressed shift+click (unlike TextEdit), but they disappeared as soon as you started dragging. This pull request preserves them after you start dragging.
* RangeSelect now takes a granularity argument (e.g. for shift + double-click).
* Dragging with a granularity won't reverse the selection until you move a whole unit backwards. This one's a little hard to explain, but double-click in the middle of a long word and move the mouse a couple pixels to the left. The current version will move the cursor to the start of the SelRegion; this PR won't. Most other editors seem to do it this way, and it simplifies the code a bit.

## Related Issues
Discussed in #826.

## To do
- [x] Submit pull request to xi-mac (xi-editor/xi-mac#408)
- [x] Support old version of the protocol
- [x] Update documentation
- [ ] Test

## Review Checklist

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -276,34 +276,8 @@ impl View {
             }
             GestureType::Drag => self.do_drag(text, offset, Affinity::default()),
 
-            // Deprecated gestures, kept for compatibility
-            GestureType::PointSelect => {
-                warn!("The point_select gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Point, false)
-            }
-            GestureType::RangeSelect => {
-                warn!("The range_select gesture is deprecated; use select_extend instead");
-                self.extend_selection(text, offset, SelectionGranularity::Point)
-            }
-            GestureType::ToggleSel => {
-                warn!("The toggle_sel gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Point, true)
-            }
-            GestureType::LineSelect => {
-                warn!("The line_select gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Line, false)
-            }
-            GestureType::WordSelect => {
-                warn!("The word_select gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Word, false)
-            }
-            GestureType::MultiLineSelect => {
-                warn!("The multi_line_select gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Line, true)
-            }
-            GestureType::MultiWordSelect => {
-                warn!("The multi_word_select gesture is deprecated; use select instead");
-                self.select(text, offset, SelectionGranularity::Word, true)
+            _ => {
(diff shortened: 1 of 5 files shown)


cmyr:
Hey @ryanbloom sorry to let this hang, have been taking a bit of a break this past week. If I don't get to this over the weekend It'll be early next week. :)

rust/core-lib/src/view.rs
```diff
@@ -400,54 +402,82 @@ impl View {
...
+
+        self.drag_state = Some(DragState { base_sel, min, max, granularity });
     }
 
-    /// Selects an entire word and supports multi selection.
-    pub fn select_word(&mut self, text: &Rope, offset: usize, multi_select: bool) {
-        let (start, end) = {
-            let mut word_cursor = WordCursor::new(text, offset);
-            word_cursor.select_word()
+    /// Selects a specific range (eg. when the user performs SHIFT + click).
```
cmyr: I might say, 'expands an existing selection', for the first bit, instead? I find this RPC name confusing, generally, because double clicking a word is actually "selecting a range"; the thing that makes this distinct is that the result depends on the existing state.

rust/core-lib/src/view.rs
```diff
@@ -400,54 +402,82 @@ impl View {
...
+    pub fn select_range(&mut self, text: &Rope, offset: usize, granularity: SelectionGranularity) {
+        let (base_sel, last) = {
+            let mut base = Selection::new();
+            let (last, rest) = self.sel_regions().split_last().unwrap();
```
cmyr: i'm scratching my chin here wondering if this unwrap is safe?

ryanbloom:
Backwards compatibility is working with the latest commit.

I called the new RPC `SelectExtend` (as opposed to `SelectExpand`) in case we want to use "expand" for other operations like "select enclosing line." I think "extend" is a slightly better fit for this operation, but feel free to nitpick further if you'd like. 🙂

ryanbloom:
I've updated the documentation. Still a little unclear on that comment about converting gesture RPCs, though. Could you clarify?

rust/core-lib/src/view.rs
```diff
@@ -260,16 +253,28 @@ impl View {
...
         match ty {
+            GestureType::Select { granularity, multi } => {
+                self.select(text, offset, granularity, multi)
+            }
+            GestureType::RangeSelect { granularity } => {
+                self.select_range(text, offset, granularity)
+            }
+            GestureType::Drag => self.do_drag(text, offset, Affinity::default()),
+
+            // Deprecated gestures, kept for compatibility
```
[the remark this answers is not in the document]
cmyr: Yea, can we do it in `edit_types`?

Issue #816, opened 2018-09-18
Add a simple instruction how to download or install xi-editor


rubo77:
It is not obvious fo a new visitor, how to install or download a pre compiled package for Windows, Linux or OSX

cmyr:
This is a fair point. We don't currently offer a pre-compiled binary because the editor is not very polished, and we do not want to provide the impression that it is ready to daily use. Making people jump through the hoops of getting it built serves as a minor but maybe useful barrier to entry. :)

rubo77:
so we leave this open until tehere is the first polished version?

would be handy to be informed this way :)

cmyr:
@rubo77 that's a good question. There are various previous issues asking the same question (#7, e.g) , and in any case this is sort of a matter for the various front-ends, so your best bet would be to 'watch' the repo fro the frontend you're interested in. (the frontends are listed in the readme)

So I think I'd be inclined to close this issue?

lolgear:
@cmyr Please, do not close this issue. Better solution for this is header "This project in early stage. We do not offer simple installation instructions right now. Be patient and have fun with Rust :)" in Readme.

I think even first line would be great with Header1 style. ( # )

jansol:
Does #872 address this sufficiently?

lolgear:
@jansol Yes and no. It doesn't contain "This project in early stage. We do not offer simple installation instructions right now. Be patient and have fun with Rust".
These words are necessary in the top of Readme.
Maybe in top of Readme of both core and front-end repositories.

matkoniecz:
Yes, it would be nice to have explicit "It is in early stage and this editor is not ready for daily use as the main editor".

Readme has many hints leading in this direction that I now see, but making it clear would be a good idea.

seagle0128:
Can you please provide the beta pre-compiled binary for the users?

nangtrongvuon:
@seagle0128 The idea is that we don't support releases yet, so we also don't provide a pre-compiled binary. Although it doesn't seem too difficult to just have CI spit out a release build here and there? cc @cmyr ?

cmyr:
If someone wanted to manually package a binary and add it as a github release that would be fine by me.

PR #1073, opened 2019-01-04
explicitly describe as not ready for daily use

matkoniecz:
## Summary
Make more clear that while promising this project is not useful for people looking for a new editor - at this moment it is useful for people interested in developing a new editor

## Related Issues
related to #816

## Review Checklist

- [x] I have responded to reviews and made changes where appropriate (not yet applicabble).
- [x] I have tested the code with `cargo test --all` / `./rust/run_all_checks` (not applicable).
- [x] I have updated comments / documentation related to the changes I made (not applicable).
- [x] I have rebased my PR branch onto xi-editor/master (not yet applicable).

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ using modern software engineering techniques. It is initially built for
 macOS, using Cocoa for the user interface. There are also frontends for
 other operating systems available from third-party developers.
 
+It is in an early stage and this editor is not ready for daily use as the main one.
+
 Goals include:
 
 * ***Incredibly high performance***. All editing operations should commit and paint


jansol:
Thanks for the PR! Unfortunately I'm inclined to **close** this:

First off, this is the repo for xi-core which is not usable by itself to begin with, as is stated in the README multiple times already, starting from the very top of the file.

Additionally I use xi-mac as my primary code editor already, and I am fairly sure other developers have mentioned doing so as well so the suggested text as written is not strictly speaking true. It also comes of as quite self-deprecating and does not address the actual issue mentioned in the OP, namely the lack of downloadable binaries.

Speaking of xi-mac, the README in that repo actually does mention the lack of binaries as well as the plan for providing them in the future right at the beginning. Of course some people won't bother reading any of that and ask for binaries anyway. I don't think there is anything we can do about those.

matkoniecz:
Feel free to close if that would be misleading.

So with `xi-core` and `xi-mac` combination is ready? Maybe https://github.com/xi-editor/xi-editor#frontends should have one more category "ready for daily use" to differentiate https://github.com/xi-editor/xi-mac and https://github.com/sw5cc/xi-qt that are currently in the same section what seems to not make sense.

cmyr:
@matkoniecz I would be hesitant about 'ready for daily use', but xi-mac is certainly a useable, light-weight editor.

I appreciate the sentiment of this PR, but I agree with @jansol; I think it's clear enough right now that this is a work in progress. :)

jansol:
I'm absolutely on board with a progress indicator for the various known frontends. The main problem is that someone would have to evaluate them all and keep the indicators up to date. It's also not entirely clear to me how progress should be measured. If you have ideas, let's hear them.

matkoniecz:
I started writing abouit my idea but decided that it will be easier to propose it as a PR - see #1079

PR #1075, opened 2019-01-08
fix: ensure exec_path appends .exe on windows

### LINKED ISSUE #1074: xi-syntect-plugin: incorrect `exec_path` on windows
When building `xi-core` and `xi-syntect-plugin` on Windows, `xi-core` cannot find the `xi-syntect-plugin` and generates a File Not Found error.

I believe the cause is because of this generated manifest: https://github.com/xi-editor/xi-editor/blob/c8dbc7bde3354b2432918d643dc1ba70e57abc0b/rust/syntect-plugin/manifest.toml#L4

This fixes it and everything works:

```diff
diff --git a/rust/syntect-plugin/manifest.toml b/rust/syntect-plugin/manifest.toml
index a72f065..414fc1e 100644
--- a/rust/syntect-plugin/manifest.toml
+++ b/rust/syntect-plugin/manifest.toml
@@ -1,7 +1,7 @@
 name = "xi-syntect-plugin"
 version = "0.0.0"
 scope = "global"
-exec_path = "./bin/xi-syntect-plugin"
+exec_path = "./bin/xi-syntect-plugin.exe"
 activations = ["autorun"]
 commands = []
```

---

Could we perhaps generate the manifest each time the plugin is built and use something like this to ensure the `exec_path` is correct per platform?

```diff
diff --git a/rust/syntect-plugin/examples/make_manifest.rs b/rust/syntect-plugin/examples/make_manifest.rs
index 45a6332..1c3746f 100644
--- a/rust/syntect-plugin/examples/make_manifest.rs
+++ b/rust/syntect-plugin/examples/make_manifest.rs
@@ -50,7 +50,11 @@ fn main() -> Result<(), io::Error> {
         .collect::<Vec<_>>();

     let (name, version) = parse_name_and_version()?;
-    let exec_path = PathBuf::from(format!("./bin/{}", &name));
+    let exec_path = if cfg!(windows) {
+        PathBuf::from(format!("./bin/{}.exe", &name))
+    } else {
+        PathBuf::from(format!("./bin/{}", &name))
+    };

     let mani = PluginDescription {
         name,
```

Right now the `make_manifest.rs` file isn't called at all (I believe the current setup is to run it locally and stage the result - see this commit: https://github.com/xi-editor/xi-editor/commit/8c5c636fdaab0e53e98a358bc37b598cce347d6f). Perhaps moving it into a `build.rs` might be beneficial.

acheronfail:
Fixes #1074.

This PR ensures that when each plugin's `manifest.toml` is deserialised so that the `exec_path` has the `.exe` extension on windows platforms.

### DIFF
--- a/rust/core-lib/src/plugins/manifest.rs
+++ b/rust/core-lib/src/plugins/manifest.rs
@@ -16,7 +16,7 @@
 
 use std::path::PathBuf;
 
-use serde::Serialize;
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_json::{self, Value};
 
 use crate::syntax::{LanguageDefinition, LanguageId};
@@ -33,6 +33,7 @@ pub struct PluginDescription {
     pub scope: PluginScope,
     // more metadata ...
     /// path to plugin executable
+    #[serde(deserialize_with = "platform_exec_path")]
     pub exec_path: PathBuf,
     /// Events that cause this plugin to run
     #[serde(default)]
@@ -41,6 +42,15 @@ pub struct PluginDescription {
     pub commands: Vec<Command>,
     #[serde(default)]
     pub languages: Vec<LanguageDefinition>,
+}
+
+fn platform_exec_path<'de, D: Deserializer<'de>>(deserializer: D) -> Result<PathBuf, D::Error> {
+    let exec_path = PathBuf::deserialize(deserializer)?;
+    if cfg!(windows) {
+        Ok(exec_path.with_extension("txt"))
+    } else {
+        Ok(exec_path)
+    }
 }
 
 /// `PluginActivation`s represent events that trigger running a plugin.
--- a/rust/core-lib/src/plugins/manifest.rs
+++ b/rust/core-lib/src/plugins/manifest.rs
@@ -47,7 +47,7 @@ pub struct PluginDescription {
 fn platform_exec_path<'de, D: Deserializer<'de>>(deserializer: D) -> Result<PathBuf, D::Error> {
     let exec_path = PathBuf::deserialize(deserializer)?;
     if cfg!(windows) {
-        Ok(exec_path.with_extension("txt"))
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/plugins/manifest.rs
```diff
@@ -43,6 +44,15 @@ pub struct PluginDescription {
     pub languages: Vec<LanguageDefinition>,
 }
 
+fn platform_exec_path<'de, D: Deserializer<'de>>(deserializer: D) -> Result<PathBuf, D::Error> {
+    let exec_path = PathBuf::deserialize(deserializer)?;
+    if cfg!(windows) {
+        Ok(exec_path.with_extension("txt"))
```
cmyr: exe?

acheronfail: 🤦‍♂️ Wow, that is embarrassing. My fingers and my brain obviously did not agree!

acheronfail:
Also just found a path issue with the test on the travis build, will fix:

```
---- plugins::manifest::tests::platform_exec_path stdout ----
thread 'plugins::manifest::tests::platform_exec_path' panicked at 'assertion failed: `(left == right)`
  left: `"path/to\\binary.txt"`,
 right: `"path/to/binary.exe"`', core-lib\src\plugins\manifest.rs:257:13
```

cmyr:
Just for my understanding, what was the issue with the wrong path separator?

acheronfail:
> Just for my understanding, what was the issue with the wrong path separator?

It was my test that was the issue, previously it checked:

```rust
if cfg!(windows) {
    assert_eq!(exec_path, "path/to/binary.exe");
} else {
    assert_eq!(exec_path, "path/to/binary");
}
```

... which used Unix-style paths rather than Windows-style.
I changed the test just to check the end of the path instead.

cmyr:
Ah gotcha, cool. 👍

PR #1082, opened 2019-01-10
fix: move to end of paragraph was not working on last line without newline

### LINKED ISSUE #1081: move_to_end_of_line not working on last line, without newline
This is a regression related to the changes in cursor behaviour we've made (I suspect).

Basically: EOF used to be considered a boundary, and it isn't anymore. We probably need to explicitly use EOF if there's no next line boundary.

reproduction (in xi-mac):

- open a new file, enter some text but no newline. 
- <kbd>control</kbd><kbd>a</kbd> to move to the start of line. 
- <kbd>control</kbd><kbd>e</kbd> to move to the end of line. 

We expect the cursor to be at the end of line the line, but it will stay at the beginning.

Any patch that includes this should add a test in `event_context.rs`.

allaud:
## Summary
Fixes bug when goto end of line was not working for the last line

## Related Issues
closes #1081 

## Checklist

- [x] Add regression test for described case
- [x] Add actual fix for the issue

## Review Checklist

### DIFF
--- a/rust/core-lib/src/movement.rs
+++ b/rust/core-lib/src/movement.rs
@@ -248,8 +248,11 @@ pub fn region_movement(
                 } else if cursor.pos() == text.len() {
                     offset = text.len();
                 }
+                (offset, None)
+            } else {
+                //in this case we are already on a last line so just moving to EOL
+                (text.len(), None)
             }
-            (offset, None)
         }
         Movement::EndOfParagraphKill => {
             // Note: TextEdit would start at modify ? r.end : r.max()
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -742,6 +742,14 @@ mod tests {
         let harness = ContextHarness::new(initial_text);
         let mut ctx = harness.make_context();
 
+        ctx.do_edit(EditNotification::MoveDown);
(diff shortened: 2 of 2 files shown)


rust/core-lib/src/movement.rs
```diff
@@ -248,8 +248,11 @@ pub fn region_movement(
                 } else if cursor.pos() == text.len() {
                     offset = text.len();
                 }
+                (offset, None)
+            } else {
```
cmyr: In this case, we can just do `offset = text.len()`. Not sure if we can clean up the logic any, so that we don't have two branches with the same result?

allaud: Looks like the easiest way to keep code clean is just to use `(text.len(), None)`

fkorotkov:
@cmyr seems like Brew auto-update failed. Brew always tries to update first before installing. To have more reporoducible builds you might want to disable auto-updates via an environment variable.

```yaml
env:
  HOMEBREW_NO_AUTO_UPDATE: 1
```

Maybe it's a good time to disable it by default since it's not the first time I hear about the issue. 🤔

fkorotkov:
@cmyr working on a change to disable automatic auto-updates so people will need to run `brew update` explicitly. More determinism!

cmyr:
@fkorotkov awesome, thx!

Issue #2, opened 2016-04-27
Rpc over Stdin/stdout for plugins


natefinch:
I saw you mention plugins that communicate over pipes... Not sure exactly what you mean by pipes, but I wanted to float another suggestion - using json RPC over stdin/stdout for a plugin process. It obviates the need for any system resources, and anyone can write code that talks to stdin/stdout.  

I made a go package specifically for this here: HTTPS://github.com/natefinch/pie. Obviously that's go, not rust, but it may help explain the idea.. And if you do go in that direction, people who wanted to make a plugin in go could easily use that package.

raphlinus:
Yes, I meant stdin/stdout specifically, for the reasons you state: to make it super easy and low friction.

I'll look at the pie stuff - at first glance, it looks pretty similar to what I was thinking about. What I'd love to do is have a protocol so simple you can write a single-task plugin with 30 lines of code or so and no external libraries. But then there probably _will_ be a library that handles async behavior in a more sophisticated manner, like providing caches and managing invalidation for you. Go will be a high priority for such a library because of its combination of high performance and low developer friction.

developit:
FWIW, I've used a plugin architecture nearly identical to that PIE example a number of times, and found it scales well (in terms of complexity).  Particularly if you simply standardize on JSON-RPC, since libraries for that are near-universally available.

natefinch:
Ahh, awesome.  Yeah, if you want to write a plugin in Go, you don't even really need pie... it's only a few lines of code.. it's the pluggable application code that is really more complicated, and the plugin side is just there for symmetry.

This is a complete go program that will create a json rpc service over stdin/stdout, with service name "Plugin" and a method "Plugin.Hello".

```go
package main

import (
	"io"
	"net/rpc"
	"net/rpc/jsonrpc"
	"os"
)

func main() {
	s := rpc.NewServer()
	s.Register(Plugin{})
	s.ServeCodec(jsonrpc.NewServerCodec(rwCloser{os.Stdin, os.Stdout}))
}

// Plugin's methods become RPC calls.
type Plugin struct{}

func (Plugin) Hello() string { return "hello world!" }

// rwCloser just merges a ReadCloser and a WriteCloser into a ReadWriteCloser.
type rwCloser struct {
	io.ReadCloser
	io.WriteCloser
}

func (rw rwCloser) Close() error {
	err := rw.ReadCloser.Close()
	if err := rw.WriteCloser.Close(); err != nil {
		return err
	}
	return err
}
[...]

PR #1084, opened 2019-01-11
Fix performance issue with Cursor

### LINKED ISSUE #1061: Fix Cursor::prev calling compute_info unnecessarily
In certain pathological cases (a large number of nodes and a small number of boundaries in a given metric) `Cursor::prev` can exhibit very poor performance because it calls `compute_info` on each node. This came up when I was looking into #1059; unrelated to the crash, wrapping really long lines was visibly slow, and this came up in the profiler.

There's a TODO in the code, so this is mostly just a matter of figuring out what the right solution actually is. https://github.com/xi-editor/xi-editor/blob/master/rust/rope/src/tree.rs#L608

This isn't a super easy issue, but it is very contained; it's just a matter of figuring out the datastructure / algorithm and determining where we should actually be getting the info from.

raphlinus:
This patch fixes a number of performance issues with cursors, including
the immediate issue named in #1061, specifically that the `prev` method
was calling `compute_info`.

A deeper underlying problem is that these methods weren't O(log n) in
the worst case. The `next` method was partly fixed by #279, but had a
recursion and so could overflow the stack if the
`next_zero_measure_large` test was increased to `0..30`. The `prev`
method was never improved and had significantly worse performance.

The new strategy for both `prev` and `next` is as follows. First, they
try to find the next boundary in the current leaf. If not found, they
go to the next leaf and try to find the boundary there.

If the boundary is not found in those two leaves, then it is likely that
it is far away. At this point, the strategy shifts to a O(log n)
traversal of the tree to find the measure at the cursor, then another
traversal similar to `descend` to find the leaf containing the desired
boundary.

The logic is simplified a bit compared to #279 because it gives up after
two leaves rather than hunting in the cursor cache for a positive
measure subtree. Doing that would add complexity and improve performance
in cases where boundaries are much sparser than leaves (~1k) but denser
than the extent of the cache (~64k), but degrade performance where they
are even sparser than that. This seemed like a good tradeoff.

Closes #1061

## Summary
Example: `Add a pull request template. This will make it easier for newcomers to contribute to the project with confidence.`

## Related Issues
Example: `Related to #1`

Example: `closes #2`

## Checklist
Example:

- [ ] Example check-list item
- [x] Create a pull request template
- [ ] make xi perfect

## Review Checklist

### DIFF
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -228,23 +228,6 @@ impl<N: NodeInfo> Node<N> {
         }
     }
 
-    /// Returns the first child with a positive measure, starting from the
-    /// `j`th.  Also, returns the length in base units of the children we have
-    /// skipped. Note that if it returns `None` in the first component, we skip
-    /// all the children.
-    fn next_positive_measure_child<M: Metric<N>>(&self, j: usize) -> (Option<usize>, usize) {
-        let children = self.get_children();
-        let mut offset = 0;
-        for i in j..children.len() {
-            if children[i].measure::<M>() > 0 {
-                return (Some(i), offset);
-            } else {
-                offset += children[i].len();
-            }
-        }
-        (None, offset)
-    }
-
     fn get_leaf(&self) -> &N::L {
         if let NodeVal::Leaf(ref l) = self.0.val {
             l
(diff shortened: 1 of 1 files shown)


rust/rope/src/tree.rs
```diff
@@ -763,25 +713,35 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
         CursorIter { cursor: self, _metric: PhantomData }
     }
 
+    /// Tries to find the last boundary in the leaf the cursor is currently in.
+    ///
+    /// If the last boundary is at the end of the leaf, it is only counted if
+    /// it is less than `orig_pos`.
```
cmyr: to clarify: If we're at the 0 offset in leaf2, and we call this function, we want to avoid returning `leaf1.len()`, which would be the same as our starting offset?

Is that the only case that we're dealing with, here?

rust/rope/src/tree.rs
```diff
@@ -763,25 +713,35 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
+    #[inline]
+    fn last_inside_leaf<M: Metric<N>>(&mut self, orig_pos: usize) -> Option<usize> {
+        let l = self.leaf.expect("inconsistent, shouldn't get here");
+        let len = l.len();
+        if self.offset_of_leaf + len < orig_pos && M::is_boundary(l, len) {
+            let _ = self.next_leaf();
```
cmyr: I'm slightly confused, here. `self.next_leaf()` modifies `self.position`; we want to be returning `self.offset_of_leaf + len`, I think; I can't see that this is equivalent?

rust/rope/src/tree.rs
```diff
@@ -763,25 +713,35 @@ impl<'a, N: NodeInfo> Cursor<'a, N> {
...
-                } else {
-                    self.position = self.offset_of_leaf + offset_in_leaf;
-                }
-                return Some(self.position);
-            }
+        let l = self.leaf.expect("inconsistent, shouldn't get here");
+        let offset_in_leaf = self.position - self.offset_of_leaf;
+        let offset_in_leaf = M::next(l, offset_in_leaf)?;
+        if offset_in_leaf == l.len() && self.offset_of_leaf + offset_in_leaf != self.root.len() {
+            let _ = self.next_leaf();
```
cmyr: okay, so maybe this answers my question: if we're at a boundary at the end of a leaf we always want to advance to the next leaf?

rust/rope/src/tree.rs
raphlinus: Yes, that's the motivation. The `prev` method always yields a position strictly less than the starting position.

rust/rope/src/tree.rs
raphlinus: This handles specifically the case where the boundary is at the end of the current leaf. To satisfy the invariants, we advance to the start of the next leaf.

rust/rope/src/tree.rs
raphlinus: Yes, the logic for this specific case is parallel.

Issue #1068, opened 2018-12-22
Add custom syntax support


eyelash:
Similar to #637 but for syntaxes instead of themes.

I also noticed that syntax highlighting is missing for *.toml and *.xiconfig files which should probably be working out of the box.

cmyr:
Yea, I think we're about ready for this. It would also be nice to ship with our own set of default syntaxes; in particular I'd like syntaxes for git stuff (commit messages in particular) and toml is a clear omission.

If someone is interested in working on this, I'm happy to mentor; the main thing to do is go look at how syntect handles loading syntax files.

kanishkarj:
@cmyr I would like to check this out.

kanishkarj:
Hey, I have figured out how to load custom syntaxes using syntect. What should I do next @eyelash @cmyr ?

cmyr:
So there are at least two things I think we'll want to do. The first one is to have our own default syntax set that we support, so that we aren't relying on whatever is bundled with syntect. The second one is to have some way that the user can add additional syntaxes that will be loaded by syntect.

The first of these is easier; I think we will want to create a new repository where we keep syntax definitions, and keep a special syntect packfile of these, which we'll include in the syntect plugin binary, and load instead of just loading the defaults.

kanishkarj:
Shouldn't we close this issue and create a new issue : dynamic loading of syntaxes

jansol:
Sure, go ahead.

PR #1085, opened 2019-01-11
Custom syntax support

kanishkarj:
Hello all! 
This PR covers the issue : #1068 support for custom syntax. It implements the following features :
- [x] The syntaxes should be saved to the folder ~/.config/xi/syntaxes , if the folder is found to not exist or empty the plugin will load the default syntaxes provided by the plugin.
- [x] The syntaxes are loaded from path defined in default.toml.
- [x] the syntaxes are loaded from files only when the manifest.toml is updated. Else we will be using a local dump file located at : ~/.config/xi/dumps/syntax.dump .
- [x] The dump file will be generated when make_manifest.rs is run.
- [ ] Dynamic loading of syntaxes. 

To test this, check the README.md file that I have put in xi-editor/rust/syntect-plugin/ .


rust/core-lib/assets/defaults.toml
```diff
@@ -27,6 +27,10 @@ word_wrap = false
 
 autodetect_whitespace = true
 
+syntax_files = "~/.config/xi/syntaxes"
+
+syntaxes_dump_path = "~/.config/xi/dumps/syntax.dump"
```
cmyr: we can't assume this path is in use; the frontend can provide an arbitrary config path (for instance on xi-mac the config path is `~/Library/Application Support/XiEditor`.

You should be able to copy how we handle the themes directory; if the user has provided a config directory, we just append the path to that.

rust/core-lib/src/config.rs
```diff
@@ -16,6 +16,7 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
 use std::fs;
+use std::io::ErrorKind;
```
cmyr: you can combine this import into the statement before, so we only import from `std::io` once. :)

rust/syntect-plugin/README.md
```diff
@@ -0,0 +1,12 @@
+# Xi-Syntect
+
+First run the following commands on your command line :
+
+```shell
+mkdir ~/.config/xi/dumps
+mkdir ~/.config/xi/syntaxes
+```
```
cmyr: if we need these setup steps they should be part of the make file. As mentioned earlier, though, these paths are not hard coded.

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -21,12 +21,14 @@ use std::fs::{self, File};
...
 
+use crate::xi_core::config::ConfigManager;
 use crate::xi_core::plugin_manifest::*;
 use crate::xi_core::LanguageDefinition;
+use syntect::dumps::dump_to_file;
 use syntect::parsing::{SyntaxReference, SyntaxSet};
 use toml::Value;
 
-const OUT_FILE_NAME: &str = "generated_manifest.toml";
+const OUT_FILE_NAME: &str = "manifest.toml";
```
cmyr: hmm, I remember having _some_ reason why I wanted to use a different file name, but I forget what it was. 😕

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -41,7 +43,28 @@ fn parse_name_and_version() -> Result<(String, String), io::Error> {
...
 fn main() -> Result<(), io::Error> {
-    let syntax_set = SyntaxSet::load_defaults_newlines();
+    let config_values = ConfigManager::new(None, None);
+    let syntax_set = match &config_values.get_syntaxes_path() {
+        Ok(res) => {
+            let synt_set = SyntaxSet::load_from_folder(res).unwrap();
+            match &config_values.get_syntaxes_dump_path() {
+                Ok(path) => {
+                    match dump_to_file(&synt_set, path) {
+                        Ok(_) => {}
```
cmyr: I think it's idiomatic to use the unit type, `()`, here: so this would be `Ok(_) => (),`.

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -41,7 +43,28 @@ fn parse_name_and_version() -> Result<(String, String), io::Error> {
...
+                        Err(_) => {
+                            // Error Handling
```
cmyr: let's at least log these errors.

rust/syntect-plugin/manifest.toml
```diff
@@ -336,3 +423,69 @@ name = "YAML"
...
+extensions = ["tsx"]
+scope = "source.tsx"
+
+[[languages]]
+name = "VimL"
+extensions = ["vim"]
+scope = "source.viml"
+
+[[languages]]
+name = "TOML"
```
cmyr: hmm, i wonder why we have `TOML` twice?

rust/syntect-plugin/manifest_original.toml
```diff
@@ -0,0 +1,338 @@
+name = "xi-syntect-plugin"
```
cmyr: This file is maybe included by accident?

kanishkarj:
Hey, I am not getting why did the MacOS ones fail. Can some one help me out with that ?

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -41,7 +43,28 @@ fn parse_name_and_version() -> Result<(String, String), io::Error> {
 }
 
 fn main() -> Result<(), io::Error> {
-    let syntax_set = SyntaxSet::load_defaults_newlines();
+    let config_values = ConfigManager::new(None, None);
+    let syntax_set = match &config_values.get_syntaxes_path() {
```
cmyr: ah, I see what you're doing.

I don't think this file should load the config; it should be self contained.

This works fine for allowing us to create a custom default syntax set, but not for loading syntaxes dynamically.

I think we should treat these features independently; doing the custom defaults is a good first step, and it will give you a better understanding for how we should go about loading them dynamically (we might need to rethink how manifests work, for that).

rust/core-lib/assets/defaults.toml
kanishkarj: Okay, shall look into that.

kanishkarj:
So, Should I shall redo these things commit by commit. and we'll proceed with this as 2 different issues.

cmyr:
The mac failure is spurious, it's being looked into. We can ignore it, we also run mac on travis so if that passes we're okay. :)

rust/syntect-plugin/examples/make_manifest.rs
kanishkarj: The xi-editor ignores other files. I will have to write my changes on to `manifest.toml` else the changes aren't processed by the editor.

rust/syntect-plugin/Makefile
```diff
@@ -10,7 +10,8 @@ out/syntect: $(lang_configs) xi-syntect-plugin
 	mkdir -p out/syntect/bin
 	cp $(lang_configs) out/syntect
 	cp ../target/release/xi-syntect-plugin out/syntect/bin
-	cp manifest.toml out/syntect/manifest.toml
+	cp manifest.toml out/syntect/
+	cp syntaxes.packfile out/syntect/
```
cmyr: I think actually making this packfile should be a target in the makefile? There's an example of what this might look like in [the syntect repo](https://github.com/trishume/syntect/blob/master/Makefile)

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -65,6 +66,16 @@ fn main() -> Result<(), io::Error> {
     let toml_str = toml::to_string(&mani).unwrap();
     let file_path = Path::new(OUT_FILE_NAME);
     let mut f = File::create(file_path)?;
+
+    // Create dump file
+    match dump_to_file(&syntax_set, "syntaxes.packfile") {
+        Ok(_) => (),
+        Err(err) => {
+            // Error Handling
+            eprintln!("{}", err);
```
cmyr: we should also just exit with an error code, if this happens we have a problem that needs attention. I think `panic!` might even be appropriate, here?

rust/syntect-plugin/manifest.toml
```diff
@@ -6,9 +6,67 @@ activations = ["autorun"]
...
+scope = "source.assembly"
+
+[[languages]]
+name = "Crystal"
+extensions = ["cr"]
+first_line_match = "^#!/.*\\bcrystal"
+scope = "source.crystal"
+
+[[languages]]
+name = "Dart"
```
cmyr: are these languages included in our new repo's packages, or is this from when you were using that other source?

rust/syntect-plugin/original_manifest.toml
```diff
@@ -0,0 +1,338 @@
+name = "xi-syntect-plugin"
```
cmyr: do we need to commit this file?

rust/syntect-plugin/src/main.rs
```diff
@@ -731,7 +733,27 @@ impl<'a> Plugin for Syntect<'a> {
 }
 
 fn main() {
-    let syntax_set = SyntaxSet::load_defaults_newlines();
+    // get the Home dir (works with Windows / Linux / Mac)
+    let syntax_set = match dirs::home_dir() {
+        Some(path) => {
+            let mut path = String::from(path.to_str().unwrap());
+            // Set the path to the packfile.
+            path.push_str("/.config/xi/plugins/syntect/syntaxes.packfile");
```
cmyr: This shouldn't be necessary; that is, we don't want to be reading this from a file at runtime. Instead we want to use `include_bytes!` to include these syntaxes in the binary itself. To see how this works, you can go look at the `dumps.rs` file in syntect, which is where `load_defaults` is implemented.

rust/syntect-plugin/Cargo.toml
```diff
@@ -19,7 +20,7 @@ toml = "0.4"
 git = "https://github.com/trishume/syntect.git"
 branch = "master"
 default-features = false
-features = ["parsing", "assets","dump-load-rs", "metadata"]
+features = ["parsing", "yaml-load", "assets","dump-load-rs","dump-create", "metadata"]
```
cmyr: I'd double check that yaml-load is necessary, after you've made some other changes

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "rust/syntect-plugin/zola"]
+	path = rust/syntect-plugin/zola
+	url = https://github.com/getzola/zola
```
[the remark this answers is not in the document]
kanishkarj: Hey! That was a part of the previous commit. I have removed it in the updated one.

rust/syntect-plugin/Cargo.toml
kanishkarj: `SyntaxSet::load_from_folder` is dependent on it actually.

rust/syntect-plugin/manifest.toml
kanishkarj: Uhm, I think there has been an issue here. I have updated the manifest.toml in the upcoming commit.

rust/syntect-plugin/examples/make_manifest.rs
kanishkarj: Done.

kanishkarj:
Also, there is some issue with the go.sublime-syntax file. I have made a change to it and made a PR check it out.

rust/syntect-plugin/src/main.rs
kanishkarj: Check out my next commit.

rust/syntect-plugin/Makefile
```diff
@@ -20,6 +20,9 @@ install: out/syntect
 	mkdir -p $(XI_PLUGIN_DIR)
 	cp -r out/syntect $(XI_PLUGIN_DIR)
 
+make-assets:
```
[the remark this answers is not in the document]
kanishkarj: I am not clearly getting this, I shall look into this after reading that book on makefiles. I guess I misunderstood what you meant by targets.

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -65,6 +66,16 @@ fn main() -> Result<(), io::Error> {
     let toml_str = toml::to_string(&mani).unwrap();
     let file_path = Path::new(OUT_FILE_NAME);
     let mut f = File::create(file_path)?;
+
+    // Create dump file
+    match dump_to_file(&syntax_set, "syntaxes.packfile") {
+        Ok(_) => (),
+        Err(err) => {
+            // Error Handling
```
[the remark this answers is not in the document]
kanishkarj: updated

kanishkarj:
I guess I have covered everything. Check this out!

rust/syntect-plugin/examples/make_manifest.rs
cmyr: Yea, I think I did it the other way so that I could diff the two files, and then I would manually replace `manifest.toml` when I was satisfied? This doesn't seem necessary, though.

rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -41,7 +42,7 @@ fn parse_name_and_version() -> Result<(String, String), io::Error> {
 }
 
 fn main() -> Result<(), io::Error> {
-    let syntax_set = SyntaxSet::load_defaults_newlines();
+    let syntax_set = SyntaxSet::load_from_folder("syntect-resources/Packages").unwrap();
```
cmyr: So as discussed, we also need to `add_plain_text_syntax()` here. I'm not sure if the order we do this in matters, so I would just do exactly what `gendata.rs` does?

kanishkarj:
Here I am using the forks I made rather than the original repo as I have made some changes that have to be pushed

rust/syntect-plugin/Makefile
```diff
@@ -6,22 +6,27 @@ XI_PLUGIN_DIR ?= $(XI_CONFIG_DIR)/plugins
 
 lang_configs = assets/*.toml
 
-out/syntect: $(lang_configs) xi-syntect-plugin
+out/syntect: assets $(lang_configs) xi-syntect-plugin
```
cmyr: I think we don't actually want this here; we don't regenerate the assets each time somebody installs the plugin, but only during development.

PR #1098, opened 2019-01-23
Only check for x86 features on x86_64

### LINKED ISSUE #1097: i686 build broken
Hello,

I've just updated xi-editor for my application and it seems like the build on i686 is broken:

https://travis-ci.org/void-linux/void-packages/jobs/483135656

Run on VoidLinux i686.

cmyr:
This fixes an issue where when compiled for x86 (i686) we might
try to call a function that had not actually been compiled.

This brings up an obvious question, which is whether we should be
using at avx2/sse4.2 on i686? My understanding was that no i686
processors existed with avx2 support, and only a couple with sse4.2,
and so this didn't really feel worth it at the time.

- fixes #1097

@Cogitri I don't imagine you have any way to test this?

## Review Checklist

### DIFF
--- a/rust/rope/src/compare.rs
+++ b/rust/rope/src/compare.rs
@@ -16,6 +16,7 @@
 use crate::rope::{BaseMetric, Rope, RopeInfo};
 use crate::tree::Cursor;
 
+#[allow(dead_code)]
 const SSE_STRIDE: usize = 16;
 
 /// Given two 16-byte slices, returns a bitmask where the 1 bits indicate
@@ -56,6 +57,7 @@ pub unsafe fn sse_compare_mask(one: &[u8], two: &[u8]) -> i32 {
     (!_mm_movemask_epi8(mask)) ^ HIGH_HALF_MASK as i32
 }
 
+#[allow(dead_code)]
 const AVX_STRIDE: usize = 32;
 
 /// Like above but with 32 byte slices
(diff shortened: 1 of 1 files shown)


Cogitri:
I can test this in a few minutes 👍

raphlinus:
In answer to the questions: there definitely are 32 bit chips that support SSE 4.2, I think the Atom family is likely one of the more popular. For AVX2, probably few or none, but keep in mind it is possible to run code in 32 bit mode even on modern processors.

So it is reasonable to support these combinations. From my perspective, the cost is in testing and validation. I would consider 32-bit x86 (aka 686) a fairly fringe architecture at this point, so reasonable to just fall back to scalar code.

Cogitri:
+1 for less maintainance cost, i686 is (finally) slowly dying

cmyr:
@raphlinus If you were approaching this work, would make a different tradeoff from the one I've made, here?

raphlinus:
Within the goals of xi-editor, I would probably make the same tradeoff.

Cogitri:
Seems to work, throws two warnings though: https://travis-ci.com/Cogitri/void-packages/jobs/172506781#L1065

cmyr:
okay, updated to suppress those warnings.

Cogitri:
Alright, new build is running: https://travis-ci.com/Cogitri/void-packages/builds/98370741

Cogitri:
Alright, works for me.

PR #1099, opened 2019-01-24
Bugfix: return None when detected tab_size in file is 0

scholtzan:
This fixes https://github.com/xi-editor/xi-editor/issues/1091

The problem was that for each line the indentation level is determined and if at least one line has some indentation then the indentation level is returned. The returned indentation level is the one most frequently used in the file. So if there is, for example, one line with indentation level >0 and all other lines don't have indentation then it returns 0 instead of `None`.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/whitespace.rs
+++ b/rust/core-lib/src/whitespace.rs
@@ -59,7 +59,7 @@ impl Indentation {
                 } else {
                     Ok(None)
                 }
-            },
+            }
             _ => Ok(None),
         }
     }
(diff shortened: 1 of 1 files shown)


rust/core-lib/src/whitespace.rs
```diff
@@ -52,7 +52,14 @@ impl Indentation {
...
             (true, false) => Ok(Some(Indentation::Tabs)),
-            (false, true) => Ok(Some(Indentation::Spaces(extract_count(spaces)))),
+            (false, true) => {
+                let tab_size = extract_count(spaces);
+                if tab_size > 0 {
+                    Ok(Some(Indentation::Spaces(tab_size)))
+                } else {
+                    Ok(None)
+                }
+            },
```
LiHRaM: The formatter is complaining about the comma there
```suggestion
            }
```

scholtzan:
Thanks! I added the regression test.

Issue #1096, opened 2019-01-22
Syntax detection fails for git commit messages


cmyr:
As of #1085 we now have syntax support for git commit messages; however we do not correctly guess the syntax of commits.

This is a simple fix: these files have names like `COMMIT_EDITMSG`, with no extension. When guessing a syntax for a file, we check for a file extension, and compare that to a list of known file extensions. We should change this so that if a file name does not have an extension, we just use the file name itself.

This should be a one or two line change in `core-lib/src/syntax.rs`. We should also be able to add a test to that file as well.

kanishkarj:
I shall look into it.

PR #1100, opened 2019-01-25
Syntax detection fails for git commit messages #1096

kanishkarj:
Hey here's something I have come up with to issue #1096. 
- [x] Syntax detection fails for files without extensions.
- [ ] Tests

Can someone help me with the tests for this one?

### DIFF
--- a/rust/core-lib/src/syntax.rs
+++ b/rust/core-lib/src/syntax.rs
@@ -127,9 +127,6 @@ mod tests {
     use super::*;
 
     #[test]
-    pub fn language_for_name() {}
-
-    #[test]
     pub fn language_for_path() {
         let ld_rust = LanguageDefinition {
             name: LanguageId::from("Rust"),
--- a/rust/core-lib/src/syntax.rs
+++ b/rust/core-lib/src/syntax.rs
@@ -122,43 +122,50 @@ impl LanguageDefinition {
     }
 }
 
-// for testing
 #[cfg(test)]
-pub fn language_for_name() {}
+mod tests {
+    use super::*;
 
-#[cfg(test)]
-pub fn language_for_path() {
-    let ld_rust = LanguageDefinition {
-        name: LanguageId::from("Rust"),
-        extensions: vec![String::from("rs")],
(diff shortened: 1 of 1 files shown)


kanishkarj:
I have come up with the tests.

rust/core-lib/src/syntax.rs
```diff
@@ -120,3 +121,44 @@ impl LanguageDefinition {
         }
     }
 }
+
+// for testing
+#[cfg(test)]
+pub fn language_for_name() {}
```
cmyr: accident?

rust/core-lib/src/syntax.rs
```diff
@@ -120,3 +121,44 @@ impl LanguageDefinition {
...
+
+#[cfg(test)]
```
cmyr: the general way to write tests is to add a 'test' module; and then each test case is marked with `#[test]`. So this would look like,

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn my_function() {
        assert_eq!(String::from("one"), "one".to_string());
    }
}
```

`#[cfg(test)]` indicates conditional compliation; this section of code will only be compiled if we're compiling for testing. `#[test]` is a special procedural macro; it says that this function should be turned into a test case, and run when tests are run.

kanishkarj: Ooh.. that'll be done.

kanishkarj:
I think the code does well, and any suggestions on the structure?

rust/core-lib/src/syntax.rs
```diff
@@ -120,3 +121,51 @@ impl LanguageDefinition {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    pub fn language_for_name() {}
```
cmyr: I think we can just remove this for now

kanishkarj:
Ohk.. Shall remove it

kanishkarj:
Check this one out

cmyr:
Awesome, thanks!

Issue #1104, opened 2019-01-29
Add support for TOML syntax, use for .xiconfig files


cmyr:
We do not currently ship with toml syntax, but this is something we can now address. This will involve a pr to https://github.com/xi-editor/syntect-resources to add a TOML package; then we will need to regenerate the data files in the syntect plugin.

To make this work with `.xiconfig` files, we will need to manually add `xiconfig` to the list of file extensions associated with the toml syntax when we generate the manifest. This will require a special case in `generate_manifest.rs`; while iterating through the languages, if the language is toml, we need to manually add the new file extension before writing the manifest file.

jansol:
Closed by #1105

PR #1105, opened 2019-02-01
Add support for TOML syntax, use for .xiconfig files

kanishkarj:
addressing issue  #1104


rust/syntect-plugin/examples/make_manifest.rs
```diff
@@ -83,9 +84,16 @@ fn main() -> Result<(), io::Error> {
 }
 
 fn lang_from_syn<'a>(src: &'a SyntaxReference) -> LanguageDefinition {
+    let mut extensions = src.file_extensions.clone();
+
+    // add support for .xiconfig
+    if extensions.contains(&String::from("toml")) {
```
cmyr: Can we not write this as `.contains("toml")`?

kanishkarj: I tried that. It gives an error `expected struct `std::string::String`, found str`. It is same with the `String::from("xiconfig")`.

kanishkarj:
isn't `defaults_meta.packdump` related to indentation configurations like that? I think that's why it wasn't updated.

cmyr:
It is; however we need to update it anytime we're adding a new syntax, in order for comment toggling and indentation to work correctly.

rust/syntect-plugin/examples/make_manifest.rs
cmyr: ah that makes sense, for some reason i was thinking it was a set instead of a vector, where this would work.

Here's an alternative solution that is faster (not checking every extension) but I don't think this is very important in a tool that is only run once in a while, during development.

```rust
fn lang_from_syn<'a>(src: &'a SyntaxReference) -> LanguageDefinition {
    let mut def = LanguageDefinition {
        name: src.name.as_str().into(),
        extensions: src.file_extensions.clone(),
        first_line_match: src.first_line_match.clone(),
        scope: src.scope.to_string(),
        default_config: None,
    };
    if def.name.as_ref() == "TOML" {
        def.extensions.push(String::from("xiconfig"));
    }
    def
}
```

kanishkarj:
I ran it make_manifest again, but for some reason, it is not updating `defaults_meta.packdump`.

kanishkarj:
`defaults_meta.packdump` updated!

PR #1106, opened 2019-02-06
Automatically delete and load new/modified plugins

scholtzan:
I started working on some plugin stuff and noticed that new plugins added to the plugin directory and modified plugins don't get loaded automatically. So this PR adds this functionality. Also deleted plugins get removed from the editor.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -546,6 +546,22 @@ impl ConfigManager {
         }
         None
     }
+
+    /// Path to plugins sub directory inside config directory.
+    /// Creates one if not present.
+    pub(crate) fn get_plugins_dir(&self) -> Option<PathBuf> {
+        let plugins_dir = self.config_dir.as_ref().map(|p| p.join("plugins"));
+
+        if let Some(p) = plugins_dir {
+            if p.exists() {
+                return Some(p);
+            }
+            if fs::DirBuilder::new().create(&p).is_ok() {
+                return Some(p);
+            }
+        }
+        None
+    }
 }
 
 impl TableStack {
--- a/rust/core-lib/src/plugins/catalog.rs
+++ b/rust/core-lib/src/plugins/catalog.rs
@@ -48,6 +48,11 @@ impl<'a> PluginCatalog {
     pub fn reload_from_paths(&mut self, paths: &[PathBuf]) {
         self.items.clear();
         self.locations.clear();
+        self.load_from_paths(paths);
+    }
+
+    /// Loads plugins from paths and adds them to existing plugins.
+    pub fn load_from_paths(&mut self, paths: &[PathBuf]) {
         let all_manifests = find_all_manifests(paths);
         for manifest_path in &all_manifests {
             match load_manifest(manifest_path) {
@@ -79,9 +84,21 @@ impl<'a> PluginCatalog {
         self.items.keys()
     }
 
+    /// Returns the plugin located at the provided file path.
+    pub fn get_from_path(&self, path: &PathBuf) -> Option<Arc<PluginDescription>> {
+        self.items.values().find(|&v| {
+            v.exec_path.to_str().unwrap().contains(path.to_str().unwrap())
+        }).map(|plugin| plugin.clone())
+    }
+
     /// Returns a reference to the named plugin if it exists in the catalog.
     pub fn get_named(&self, plugin_name: &str) -> Option<Arc<PluginDescription>> {
         self.items.get(plugin_name).map(Arc::clone)
+    }
+
+    /// Removes the named plugin.
+    pub fn remove_named(&mut self, plugin_name: &str) {
+        self.items.remove(plugin_name);
     }
 }
 
--- a/rust/core-lib/src/plugins/catalog.rs
+++ b/rust/core-lib/src/plugins/catalog.rs
@@ -85,12 +85,11 @@ impl<'a> PluginCatalog {
     }
 
     /// Returns the plugin located at the provided file path.
-    #[allow(clippy::map_clone)]
     pub fn get_from_path(&self, path: &PathBuf) -> Option<Arc<PluginDescription>> {
         self.items
             .values()
             .find(|&v| v.exec_path.to_str().unwrap().contains(path.to_str().unwrap()))
-            .map(|plugin| plugin.clone())
+            .cloned()
     }
 
     /// Returns a reference to the named plugin if it exists in the catalog.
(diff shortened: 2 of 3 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -95,6 +96,9 @@ pub const OPEN_FILE_EVENT_TOKEN: WatchToken = WatchToken(2);
 #[cfg(feature = "notify")]
 const THEME_FILE_EVENT_TOKEN: WatchToken = WatchToken(3);
 
+#[cfg(feature = "notify")]
+const PLUGIN_EVENT_TOKEN: WatchToken = WatchToken(4);
```
scholtzan: Just something I was wondering, I'm not sure if it's ideal to manually set the "token id"? It might be better to generate the ID automatically when a new watch token is created and make sure it doesn't already exist. Well. I guess it's fine for now.

rust/core-lib/src/plugins/catalog.rs
```diff
@@ -79,10 +84,24 @@ impl<'a> PluginCatalog {
         self.items.keys()
     }
 
+    /// Returns the plugin located at the provided file path.
+    #[allow(clippy::map_clone)]
+    pub fn get_from_path(&self, path: &PathBuf) -> Option<Arc<PluginDescription>> {
+        self.items
+            .values()
+            .find(|&v| v.exec_path.to_str().unwrap().contains(path.to_str().unwrap()))
+            .map(|plugin| plugin.clone())
```
scholtzan: Maybe there is a more elegant way to do this since clippy was complaining?

jansol: From the [Great List of Clippy Lints](https://rust-lang.github.io/rust-clippy/master/index.html): Checks for usage of iterator.map(|x| x.clone()) and suggests iterator.cloned() instead.

However here we aren't dealing with an Iterator anymore, since `find()` only returns a n `Option<T>`, in this case `Option<&Arc<PluginDescription>>` because `values()` always yields references. Seems like replacing the `map()` just `.cloned()` would do the trick?

jansol: Also that clippy annotation can then be removed.

scholtzan: Thanks, fixed it :)

rust/core-lib/src/tabs.rs
cmyr: I think the longer-term solution will be to have idle work sort of like the RPC system, where we have an enum of idle events that we can schedule and then that get passed in the callback. This implementation is really just an old C idiom, and yea it isn't perfect.

rust/core-lib/src/plugins/catalog.rs
```diff
@@ -48,6 +48,11 @@ impl<'a> PluginCatalog {
     pub fn reload_from_paths(&mut self, paths: &[PathBuf]) {
         self.items.clear();
         self.locations.clear();
+        self.load_from_paths(paths);
+    }
+
+    /// Loads plugins from paths and adds them to existing plugins.
+    pub fn load_from_paths(&mut self, paths: &[PathBuf]) {
```
cmyr: I prefer `pub(crate)` for everything where possible. A lot of this code predates `pub(crate)` existing, so can be a bad example.

The main reason `pub(crate)` is better is that it lets the compiler give us more unused-code warnings. If something is marked `pub`, it doesn't warn us when it's unused because it might be used by another crate.

rust/core-lib/src/plugins/catalog.rs
```diff
@@ -79,10 +84,23 @@ impl<'a> PluginCatalog {
         self.items.keys()
     }
 
+    /// Returns the plugin located at the provided file path.
+    pub fn get_from_path(&self, path: &PathBuf) -> Option<Arc<PluginDescription>> {
```
cmyr: Instead of `&PathBuf`, prefer `&Path` or `P: AsRef<Path>`.

rust/core-lib/src/tabs.rs
```diff
@@ -42,6 +42,7 @@ use crate::event_context::EventContext;
 use crate::file::FileManager;
 use crate::line_ending::LineEnding;
 use crate::plugin_rpc::{PluginNotification, PluginRequest};
+use crate::plugins::rpc::ClientPluginInfo;
```
cmyr: I believe the alias `plugin_rpc` above refers to this file.

rust/core-lib/src/tabs.rs
```diff
@@ -732,6 +743,55 @@ impl CoreState {
...
+        use self::DebouncedEvent::*;
+        match event {
+            Create(ref path) | Write(ref path) => {
+                self.plugins.load_from_paths(&[path.clone()]);
+                if let Some(plugin) = self.plugins.get_from_path(path) {
+                    self.do_start_plugin(ViewId(0), &plugin.name);
+                }
+            }
+            // the way FSEvents on macOS work, we want to verify that this path
+            // has actually be removed before we do anything.
```
cmyr: s/be/been

rust/core-lib/src/tabs.rs
```diff
@@ -732,6 +743,55 @@ impl CoreState {
...
+                }
+            }
+            Chmod(ref path) | Remove(ref path) => {
+                if let Some(plugin) = self.plugins.get_from_path(path) {
+                    self.do_stop_plugin(ViewId(0), &plugin.name);
+                    self.do_start_plugin(ViewId(0), &plugin.name);
+                }
+            }
+            _ => (),
+        }
```
cmyr: looking back at this big switch statement I do wonder if there's room for simplification, maybe by adding a few helper functions?

I also wonder if there might be a simplification possible: On any event, find the plugin (by folder) that has changed. If it's been removed, stop it; if it's been added, start it, and in all other cases restart it?

Just thinking out loud, this approach works it just feels a bit more fragile / easier to screw up. I'm not sure my idea is any good, though.

rust/core-lib/src/config.rs
```diff
@@ -546,6 +546,22 @@ impl ConfigManager {
...
+        if let Some(p) = plugins_dir {
+            if p.exists() {
+                return Some(p);
+            }
+            if fs::DirBuilder::new().create(&p).is_ok() {
+                return Some(p);
+            }
+        }
+        None
+    }
```
cmyr: 🤔 

code golf:
```rust
let plugins_dir = self.config_dir.as_ref().map(|p| p.join("plugins"))?;
if !plugins_dir.exists() {
    let _ = fs::DirBuilder::new().create(&plugins_dir).ok()?;
}
Some(plugins_dir)
```

scholtzan:
Thanks! I can still/will create a new PR with the changes

PR #1107, opened 2019-02-06
Add newline to end of file if nonexistent

### LINKED ISSUE #1093: Add option to terminate file with a newline
Not 100% sure if this belongs into xi-editor, but it'd be nice if xi-editor offered an option to terminate files with a `line_ending` (newline). `git diff` without a newline at the end of a file isn't pretty (as it includes info that the file ending has been moved.

Cogitri:
## Summary
Right now xi-editor doesn't add a newline at the end of files. This makes for rather ugly diffs 
(since the final line, that doesn't have a newline at its end also is recognized as changed). This PR changes this behavior and adds a newline if the file doesn't end in one already.

## Related Issues
fixes #1093

What this still needs:

- [x] Fix getting the last line

### DIFF
--- a/rust/core-lib/assets/client_example.toml
+++ b/rust/core-lib/assets/client_example.toml
@@ -39,3 +39,6 @@ word_wrap = false
 
 # Detect tab and newline settings on file open
 autodetect_whitespace = true
+
+# Ensure file ends in a newline when saving
+save_with_newline = true
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -22,7 +22,7 @@ use std::time::{Duration, Instant};
 
 use serde_json::{self, Value};
 
-use xi_rope::{Interval, LinesMetric, Rope, RopeDelta};
+use xi_rope::{Cursor, Interval, LinesMetric, Rope, RopeDelta};
 use xi_rpc::{Error as RpcError, RemoteError};
 use xi_trace::trace_block;
 
@@ -492,6 +492,32 @@ impl<'a> EventContext<'a> {
             Err(err) => error!("plugin shutdown, do something {:?}", err),
         }
         self.editor.borrow_mut().dec_revs_in_flight();
+    }
+
+    /// Returns the text to be saved, appending a newline if necessary.
+    pub(crate) fn text_for_save(&mut self) -> Rope {
+        let editor = self.editor.borrow();
+        let mut rope = editor.get_buffer().clone();
+        let rope_len = rope.len();
+
+        if rope_len < 1 || !self.config.save_with_newline {
+            return rope;
+        }
+
+        let cursor = Cursor::new(&rope, rope.len());
+        let has_newline_at_eof = match cursor.get_leaf() {
+            Some((last_chunk, _)) => last_chunk.ends_with(&self.config.line_ending),
+            // The rope can't be empty, since we would have returned earlier if it was
+            None => unreachable!(),
+        };
+
+        if !has_newline_at_eof {
+            let line_ending = &self.config.line_ending;
+            rope.edit(rope_len.., line_ending);
+            rope
+        } else {
+            rope
+        }
     }
 
     /// Called after anything changes that effects word wrap, such as the size of
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -390,9 +390,10 @@ impl CoreState {
             None => return,
         };
 
-        let ed = &self.editors[&buffer_id];
+        let mut save_ctx = self.make_context(view_id).unwrap();
+        let fin_text = save_ctx.text_for_save();
 
-        if let Err(e) = self.file_manager.save(path, ed.borrow().get_buffer(), buffer_id) {
+        if let Err(e) = self.file_manager.save(path, &fin_text, buffer_id) {
             let error_message = e.to_string();
             error!("File error: {:?}", error_message);
             self.peer.alert(error_message);
--- a/rust/core-lib/assets/defaults.toml
+++ b/rust/core-lib/assets/defaults.toml
@@ -33,3 +33,5 @@ surrounding_pairs = [
   ["{", "}"],
   ["[", "]"],
 ]
+
+save_with_newline = true
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -186,6 +186,7 @@ pub struct BufferItems {
     pub word_wrap: bool,
     pub autodetect_whitespace: bool,
     pub surrounding_pairs: Vec<(String, String)>,
+    pub save_with_newline: bool,
 }
 
 pub type BufferConfig = Config<BufferItems>;


rust/core-lib/src/editor.rs
```diff
@@ -941,6 +942,17 @@ impl Editor {
 
         Some(GetDataResponse { chunk, offset, first_line, first_line_offset })
     }
+
+    pub(crate) fn pre_save(&mut self, view: &mut View, config: &BufferItems,) {
+        if let Some(last_line) = self.text.lines_raw(Interval::new(0,2)).last() {
```
Cogitri: Not quite sure how to get the last line here to check its line ending

rust/core-lib/src/editor.rs
```diff
@@ -941,6 +942,17 @@ impl Editor {
...
+            eprintln!("Last line: {}", last_line);
+            if let Ok(line_ending) = LineEnding::parse_chunk(&last_line) {
+                if line_ending.is_none() {
+                    self.insert_newline(view, config)
```
Cogitri: For this to work I need to get view&config passed into pre_save, which isn't very pretty. Also, I guess this inserts the newline at the current cursor position?

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +495,20 @@ impl<'a> EventContext<'a> {
...
 
+    pub(crate) fn pre_save(&mut self) {
+        let editor = self.editor.borrow();
+        let ed_text = editor.get_buffer();
+        let view = self.view.borrow();
+        let offset = view.line_of_offset(ed_text, ed_text.len()) + 1;
+        if let Some(last_char) = ed_text.lines_raw(Interval::new(offset, view.offset_of_line(ed_text, offset))).last() {
+            if let Ok(line_ending) = LineEnding::parse_chunk(&last_char) {
+                if line_ending.is_none() {
+                    eprintln!("No newline!");
```
Cogitri: I'm not quite sure what's the most elegant way to add a newline to the Rope

rust/core-lib/assets/client_example.toml
```diff
@@ -39,3 +39,7 @@ word_wrap = false
 
 # Detect tab and newline settings on file open
 autodetect_whitespace = true
+
+# If the file doesn't end with a newline this option adds a newline to it upon saving.
```
cmyr: I would keep this short; something to the effect of, "Ensure file ends in a newline when saving".

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +495,25 @@ impl<'a> EventContext<'a> {
         self.editor.borrow_mut().dec_revs_in_flight();
     }
 
+    /// Checks a newline needs to be added to the end of the file
+    pub(crate) fn pre_save_newline(&mut self) -> bool {
+        let editor = self.editor.borrow();
+        let ed_text = editor.get_buffer();
+        let view = self.view.borrow();
+        let offset = view.line_of_offset(ed_text, ed_text.len()) + 1;
```
cmyr: this name is slightly confusing; we use `offset` to refer to byte offsets, and this is a line number.

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +495,25 @@ impl<'a> EventContext<'a> {
...
+        if let Some(last_char) = ed_text
+            .lines_raw(Interval::new(offset - 1, view.offset_of_line(ed_text, offset)))
```
cmyr: I don't think this is going to work as expected. From the docs for `Rope::lines_raw`:

> An iterator over the raw lines. The lines, except the last, include the
> terminating newline.

So assuming the docs are correct, this will never return the trailing newline. Since all we really need to do is check the last byte, we should be able to write,

```rust
let rope_len = rope.len();
if rope_len > 0 && rope.byte_at(rope_len - 1) == b'\n' {
// .. logic
}
```

rust/core-lib/src/rpc.rs
```diff
@@ -569,7 +569,7 @@ impl<'de, T: Deserialize<'de>> Deserialize<'de> for EditCommand<T> {
             Some(_) => {
                 return Err(de::Error::custom(
                     "'params' field, if present, must be object or array.",
-                ))
+                ));
```
cmyr: was this a clippy lint or something?

rust/core-lib/src/tabs.rs
```diff
@@ -390,9 +390,23 @@ impl CoreState {
             None => return,
         };
 
-        let ed = &self.editors[&buffer_id];
+        let fin_text = if let Some(mut save_ctx) = self.make_context(view_id) {
+            let editor = save_ctx.editor.borrow();
+            let rope = editor.get_buffer();
+            if save_ctx.pre_save_newline() {
+                let mut text = rope.to_string();
+                text.push('\n');
```
cmyr: hmm: we really need to be using the platform specific newline character, here. We can get that from the config manager.

What I might actually do here is add a helper method to `EventContext`, like `append_newline`, that returns a new rope that we can write to disk.

rust/core-lib/src/tabs.rs
```diff
@@ -390,9 +390,23 @@ impl CoreState {
             None => return,
         };
 
-        let ed = &self.editors[&buffer_id];
+        let fin_text = if let Some(mut save_ctx) = self.make_context(view_id) {
+            let editor = save_ctx.editor.borrow();
+            let rope = editor.get_buffer();
+            if save_ctx.pre_save_newline() {
+                let mut text = rope.to_string();
```
cmyr: We should be able to avoid needing to roundtrip to the string, which will be expensive for big documents. Idiomatically this could look like,

```rust
let mut rope = editor.get_buffer().clone();
let len = rope.len();
rope.edit(len.., "\n");
// .. save as normal
```

rust/core-lib/assets/client_example.toml
Cogitri: Ah, that does sound way better.

rust/core-lib/src/event_context.rs
Cogitri: > An iterator over the raw lines. The lines, except the last, include the
terminating newline.

Ah, it does return the last line with a newline, it just assumes that that naturally never happens :)

Your suggestion is way prettier though

rust/core-lib/src/rpc.rs
Cogitri: Ah, rustfmt

rust/core-lib/src/tabs.rs
Cogitri: > hmm: we really need to be using the platform specific newline character, here. We can get that from the config manager.

Oops, right, was a little in my UNIX world here :)

rust/core-lib/src/tabs.rs
Cogitri: Ah, 'edit' was the function I was looking for :)

Cogitri:
Alright, mind looking at this again? :)

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +494,20 @@ impl<'a> EventContext<'a> {
         self.editor.borrow_mut().dec_revs_in_flight();
     }
 
+    /// Checks a newline needs to be added to the end of the file
+    pub(crate) fn pre_save_newline(&mut self) -> Rope {
+        let editor = self.editor.borrow();
+        let mut rope = editor.get_buffer().clone();
+        let rope_len = rope.len();
+        if rope_len > 0 && (rope.byte_at(rope_len - 1) == b'\n' || rope.byte_at(rope_len - 1) ==   b'\r') {
```
scholtzan: Tiny nit: there are two spaces before `b'\r'`

Cogitri: Fixed.

Cogitri:
I've removed the config option, IMHO not saving without a newline shouldn't ever be done (and I don't know of any editor which allows that). On the other hand, we shouldn't assume what the user wants I guess, but please do tell me your opinion on this :)

ryanbloom:
I was a little confused when I first saw Atom adding newlines to my files. I don't think I'm the only one; see discussion in atom/atom#4741. It's a good feature to have, and for UNIX at least it makes sense as the default, but I think it should be *possible* to disable this behavior in the configuration.

Cogitri:
Alright, I'll optionalise it then. Thanks for your opinion :)

rust/core-lib/src/tabs.rs
```diff
@@ -390,9 +390,15 @@ impl CoreState {
             None => return,
         };
 
-        let ed = &self.editors[&buffer_id];
+        let fin_text = if let Some(mut save_ctx) = self.make_context(view_id) {
+            save_ctx.pre_save_newline()
+        } else {
+            let editor = &self.editors[&buffer_id].borrow();
```
cmyr: If we can't make a context, the only reasonable explanation is that the view or buffer is missing, so this branch might be redundant? It's very hard to imagine a situation where we get here.

In fact: assuming we needed a `ViewId` in order to get the `BufferId` then we should be able to guarantee that make_context will work, and we can just unwrap.

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +494,22 @@ impl<'a> EventContext<'a> {
         self.editor.borrow_mut().dec_revs_in_flight();
     }
 
+    /// Checks a newline needs to be added to the end of the file
+    pub(crate) fn pre_save_newline(&mut self) -> Rope {
```
cmyr: given that this is now returning the rope we will save, I'd call it something like `text_for_save`. The comment is also no longer up to date; it isn't really a check, it just returns the text to be saved, appending a newline if necessary.

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +494,22 @@ impl<'a> EventContext<'a> {
...
+        let editor = self.editor.borrow();
+        let mut rope = editor.get_buffer().clone();
+        let rope_len = rope.len();
+        if rope_len > 0
+            && (rope.byte_at(rope_len - 1) == b'\n' || rope.byte_at(rope_len - 1) == b'\r')
```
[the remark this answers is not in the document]
Cogitri: I'll just switch to LineEnding::parse_chunk.

rust/core-lib/src/tabs.rs
Cogitri: Alright

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +495,29 @@ impl<'a> EventContext<'a> {
         self.editor.borrow_mut().dec_revs_in_flight();
     }
 
+    /// Returns the text to be saved, appending a newline if necessary.
+    pub(crate) fn text_for_save(&mut self) -> Rope {
+        let editor = self.editor.borrow();
+        let mut rope = editor.get_buffer().clone();
+        let rope_len = rope.len();
+        let last_char = rope.byte_at(rope.len() - 1) as char;
```
cmyr: rope len can be zero; this would be an integer underflow.

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +495,29 @@ impl<'a> EventContext<'a> {
...
+
+        // Only add a new line if the user actually wrote something,
+        // there's no newline at the end of the file
+        // and the config option for it is true
+        if rope_len > 0
+            // If parse_chunk fails just play it safe and return Some so we don't add a superfluous newline
+            && LineEnding::parse_chunk(&last_char.to_string()).unwrap_or(Some(LineEnding::Lf)).is_none()
```
cmyr: this is ~pretty~ very confusing.

Can this ever fail? I... think it can't?

I think that the previous version, with just checking for `\n` at EOF is fine. However, if we want to do this _actually_ correctly, I would suggest the following approach:

```rust
let mut cursor = Cursor::new(&rope, rope.len());
let newline_at_eof = match cursor.get_leaf() {
    Some((last_chunk, _)) => last_chunk.ends_with(&self.config.line_ending),
    // there's no last chunk, which means rope is empty, but we checked that earlier?  ¯\_(ツ)_/¯
   // this is only true of course if we _do_ check for is_empty...
    None => unreachable!(),
};

if self.config.save_with_newline && newline_at_eof { ... }
```

Does that maybe make sense? We do the cursor business because it lets us efficiently get the actual string value for the last chunk of the rope.

Cogitri: > I think that the previous version, with just checking for \n at EOF is fine

Hum, if we already have LineEnding stuff we might as well use that to make sure we have everything in one place

Cursor still sort of is magic to me, will look into that tomorrow

Cogitri: Alright, I just use config.line_ending now, since it's not really relevant what kind of line ending is used here. Does this work if the user sets line ending x in the config and xi autodetects a different line ending in the file though?

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +494,32 @@ impl<'a> EventContext<'a> {
...
+
+        if rope_len < 1 {
+            return rope;
+        }
+
+        let cursor = Cursor::new(&rope, rope.len());
+        let has_newline_at_eof = match cursor.get_leaf() {
+            Some((last_chunk, _)) => last_chunk.ends_with(&self.config.line_ending),
+            // The rope can't be empty, since we would have returned earlier if it was
+            None => unreachable!(),
```
Cogitri: I guess we could just return rope here to handle this gracefully.

cmyr: If we ever reach this point something is fundamentally wrong with our assumptions, and we probably prefer to crash, I think?

This annotation is also a hint to the compiler; knowing that some point can't be reached can allow for additional optimizations.

rust/core-lib/src/event_context.rs
```diff
@@ -494,6 +494,32 @@ impl<'a> EventContext<'a> {
         self.editor.borrow_mut().dec_revs_in_flight();
     }
 
+    /// Returns the text to be saved, appending a newline if necessary.
+    pub(crate) fn text_for_save(&mut self) -> Rope {
+        let editor = self.editor.borrow();
+        let mut rope = editor.get_buffer().clone();
+        let rope_len = rope.len();
+
+        if rope_len < 1 {
```
cmyr: we can also return early if `self.config.save_with_newline == false`?

rust/core-lib/src/event_context.rs
Cogitri: Ah, alright.

rust/core-lib/src/event_context.rs
Cogitri: Ah, true

Cogitri: Pushed new version.

Cogitri:
No worries, thank you very much for your timely reviews and patience with me :)

PR #1108, opened 2019-02-06
Update GSoC 2019 project list

scholtzan:
Updated GSoC 2019 project list

### DIFF
--- a/docs/gsoc.md
+++ b/docs/gsoc.md
@@ -12,178 +12,77 @@ Good places to start contributing to Xi are [easy](https://github.com/xi-editor/
 
 --------
 
-List of project ideas for Google Summer of Code 2018:
- - [Xi-Mac Polish and UI improvements](#xi-mac-polish-and-ui-improvements)
- - [Language Server Plugin](#language-server-plugin)
- - [Improved Theming Support](#improved-theming-support)
- - [Theme / Syntax Loading](#theme--syntax-loading)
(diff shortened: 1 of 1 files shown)


scholtzan:
Not sure if this is necessary. There is one current version running here: https://xi-editor.io/gsoc.html

cmyr:
I think we're just using the other repo for web stuff. Maybe we should clean this up? @ryanbloom might have the best idea right now...

scholtzan:
We should definitely update the website URL on the Github page (https://github.com/xi-editor/xi-editor). Quite confusing to have two active websites ;)

cmyr:
Done!

ryanbloom:
This duplicated website thing isn't great. I think of `xi-editor/docs/docs` as the canonical documentation files, and `xi-editor.github.io` as the canonical rest-of-the-website. Eventually, I hope to remove all the jekyll stuff from this repo, keeping only the documentation in markdown. Then the website repo would auto-update whenever those files change. There's probably a way to do this, but it's not really what git was designed for.

I don't have a strong opinion on which repository GSoC (and for that matter, Rope Science) should live in. They're in the same section of the website as protocol docs, but not directly related to the code....

So yeah, it's a little bit messy right now. 🤷‍♂️

cmyr:
@ryanbloom okay, good explanation. Maybe open an issue for that remaining work, and I'll close this.

PR #1110, opened 2019-02-07
Keep existing file perms on saving

### LINKED ISSUE #1109: If saving to an existing file, keep the permissions
af02680d739311a6ec09904cba9b3d5edfbc6b25 has introduced atomic saving, but this has broken file permission keeping. This is rather annoying when editing executables.

# Expected behavior:

```
touch new
chmod 755 new
gxi new # <-- save here
ls -lah new
```
returns: 
-rwxr-xr-x 1 rasmus rasmus 1 Feb  6 22:21 new

# Actual behavior

```
touch new
chmod 755 new
gxi new # <-- save here
ls -lah new
```

-rw-r--r-- 1 rasmus rasmus 1 Feb  6 22:21 new

Cogitri:
fixes #1109

### DIFF
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -17,7 +17,7 @@
 use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
-use std::fs::{self, File};
+use std::fs::{self, File, Permissions};
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use std::str;
@@ -32,6 +32,8 @@ use crate::tabs::BufferId;
 use crate::tabs::OPEN_FILE_EVENT_TOKEN;
 #[cfg(feature = "notify")]
 use crate::watcher::FileWatcher;
+#[cfg(target_family = "unix")]
+use std::os::unix::fs::PermissionsExt;
 
 const UTF8_BOM: &str = "\u{feff}";
 
@@ -50,6 +52,8 @@ pub struct FileInfo {
     pub path: PathBuf,
     pub mod_time: Option<SystemTime>,
     pub has_changed: bool,
+    #[cfg(target_family = "unix")]
+    pub permissions: Option<u32>,
 }
 
 pub enum FileError {
@@ -134,13 +138,15 @@ impl FileManager {
     }
 
     fn save_new(&mut self, path: &Path, text: &Rope, id: BufferId) -> Result<(), FileError> {
-        try_save(path, text, CharacterEncoding::Utf8)
+        try_save(path, text, CharacterEncoding::Utf8, self.get_info(id))
             .map_err(|e| FileError::Io(e, path.to_owned()))?;
         let info = FileInfo {
             encoding: CharacterEncoding::Utf8,
             path: path.to_owned(),
             mod_time: get_mod_time(path),
             has_changed: false,
+            #[cfg(target_family = "unix")]
+            permissions: get_permissions(path),
         };
         self.open_files.insert(path.to_owned(), id);
         self.file_info.insert(id, info);
@@ -160,7 +166,8 @@ impl FileManager {
             return Err(FileError::HasChanged(path.to_owned()));
         } else {
             let encoding = self.file_info[&id].encoding;
-            try_save(path, text, encoding).map_err(|e| FileError::Io(e, path.to_owned()))?;
+            try_save(path, text, encoding, self.get_info(id))
+                .map_err(|e| FileError::Io(e, path.to_owned()))?;
             self.file_info.get_mut(&id).unwrap().mod_time = get_mod_time(path);
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


Cogitri:
Alright, this works for me, but I only have a Linux PC.

rust/core-lib/src/file.rs
```diff
@@ -171,22 +180,33 @@ fn try_load_file<P>(path: P) -> Result<(Rope, FileInfo), FileError>
 where
     P: AsRef<Path>,
 {
-    // TODO: support for non-utf8
+    // TODO: support for non-utf8fl
```
scholtzan: oops?

rust/core-lib/src/file.rs
```diff
@@ -134,13 +139,16 @@ impl FileManager {
     }
 
     fn save_new(&mut self, path: &Path, text: &Rope, id: BufferId) -> Result<(), FileError> {
-        try_save(path, text, CharacterEncoding::Utf8)
+        try_save(path, text, CharacterEncoding::Utf8, self.get_info(id).unwrap())
```
scholtzan: The `unwrap` here causes a panic when creating a new file in xi editor and saving it.

rust/core-lib/src/file.rs
```diff
@@ -209,6 +229,11 @@ fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result
 
     fs::rename(tmp_path, path)?;
 
+    #[cfg(target_family = "unix")]
+    fs::set_permissions(path, Permissions::from_mode(file_info.perms.unwrap()))?;
+
+    filetime::set_file_times(path, file_info.accessed_time.unwrap(), file_info.mod_time.unwrap())?;
```
scholtzan: I am not sure if there could be cases where using `unwrap` here might cause a panic? Couldn't cause any while playing around with it, so it might be save.

jansol: Those timestamps might not be available if the file is on a volume with a specific filesystem I think? Say, NTFS or Samba mounts.

Either way this behavior is wrong, since we obviously did access the file, and modified it as well so both those should be set to the current time. The only thing we have to copy over is the creation timestamp.

Cogitri: Is creation time stamp a thing? At least on Linux it doesn't exist

Cogitri: > Either way this behavior is wrong, since we obviously did access the file, and modified it as well so both those should be set to the current time.

Oops, right.

jansol: Windows and macOS appear to have creation time. Ext4 is also said to have it but it's not exposed on linux so not caring about it there is fine by me.

Cogitri: What about Fuchsia? Seems like it's also supported by xi

rust/core-lib/src/file.rs
Cogitri: Hm, should I guess I'll just optionalise FileInfo then, there's nothing to keep when we create a new file anyway.

rust/core-lib/src/file.rs
Cogitri: Hum, it doesn't seem like there's functionality in Rust/the filetime crate for setting creation time. Since I don't really have a clue of other platforms I'm not quite sure how to go on with this.

cmyr: hmm. This might be rabbit-hole territory (it seems reasonable that filetime would have this functionality.)

Fuchsia we can comfortably ignore for now.

rust/core-lib/src/file.rs
```diff
@@ -209,6 +226,14 @@ fn try_save(path: &Path, text: &Rope, encoding: CharacterEncoding) -> io::Result
 
     fs::rename(tmp_path, path)?;
 
+    if let Some(info_unwrapped) = file_info {
+        #[cfg(target_family = "unix")]
+        fs::set_permissions(
+            path,
+            Permissions::from_mode(info_unwrapped.permissions.unwrap_or(0o644)),
```
cmyr: hmm, maybe we should be using umask to get the default?

rust/core-lib/src/file.rs
Cogitri: Alright, what are we going to do about this then?

rust/core-lib/src/file.rs
```diff
@@ -244,6 +269,16 @@ where
     File::open(path).and_then(|f| f.metadata()).and_then(|meta| meta.modified()).ok()
 }
 
+#[cfg(target_family = "unix")]
```
cmyr: house style is to put compiler attributes etc _below_ doc comments. 😅

rust/core-lib/src/file.rs
```diff
@@ -244,6 +269,16 @@ where
...
+/// Returns the file permissions for the file at a given path on UNIXy systems,
+/// if present.
+fn get_permissions<P>(path: P) -> Option<u32>
+where
+    P: AsRef<Path>,
```
cmyr: I prefer to put generic constraints inline, when doing so doesn't produce overly long lines, e.g.

```rust
fn get_permissions<P: AsRef<Path>>(path: P) -> Option<u32> {
// ...
```
seems cleaner, here?

cmyr:
Also to elaborate on @scholtzan's comment about variable naming: while we have used lots of short names in the past, there has been a consensus that shorter names are more difficult for non-native english speakers, so where an abbreviation can be ambiguous, we prefer longer names. Perms for permissions is on the edge, though, and I think given the context it should be clear to most readers.

rust/core-lib/src/file.rs
cmyr: addressed in review. 🙂

rust/core-lib/src/file.rs
Cogitri: Sure, if that's what you prefer. I just always use where to be consistent.

rust/core-lib/src/file.rs
Cogitri: That does sound like a good idea, not quite sure how to get that in Rust though

rust/core-lib/src/file.rs
cmyr: yep I think that's fair, but the main perk I get out of doing all these reviews is being able to impose my personal stylistic preferences on people 😉

rust/core-lib/src/file.rs
cmyr: maybe we should just only set permissions when we have them stored? Otherwise we can just let rust use the defaults, which [apparently account for umask](https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html#tymethod.mode).

Cogitri: Alright

PR #1113, opened 2019-02-09
Avoid double copy when deserializing rope

### LINKED ISSUE #1111: Deserializing rope copies its contents twice
Looking at the [`Deserialize` implementation for `Rope`](https://github.com/xi-editor/xi-editor/blob/3169c041b8befff3384dc1cc8878128a9a837240/rust/rope/src/rope.rs#L399), it seems like deserializing a rope copies its contents twice: First to a `String`, then into the rope. It would be better to skip the `String` and avoid unnecessary copies, probably by making a custom implementation of [Visitor](https://docs.rs/serde/1.0.87/serde/de/trait.Visitor.html) which handles each visited type with the fewest copies possible for that type.

ahouts:
## Summary
Avoid double copy when deserializing a Rope.

This is my first PR, so let me know if I am missing anything.

## Related Issues
closes #1111

## Review Checklist

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -1267,7 +1267,8 @@ mod tests {
 
         let rope = Rope::from(test_str.as_str());
         let json = serde_json::to_string(&rope).expect("error serializing");
-        let deserialized_rope = serde_json::from_str::<Rope>(json.as_str()).expect("error deserializing");
+        let deserialized_rope =
+            serde_json::from_str::<Rope>(json.as_str()).expect("error deserializing");
         assert_eq!(rope, deserialized_rope);
     }
 }
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -1268,6 +1268,6 @@ mod tests {
         let rope = Rope::from(test_str.as_str());
         let json = serde_json::to_string(&rope).expect("error serializing");
         let deserialized_rope = serde_json::from_str::<Rope>(json.as_str()).expect("error deserializing");
-        assert_eq!(rope, rope2);
+        assert_eq!(rope, deserialized_rope);
     }
 }
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -1266,7 +1266,7 @@ mod tests {
         let test_str = TEST_LINE.repeat(n_seg);
 
         let rope = Rope::from(test_str.as_str());
-        let json = to_string(&rope).expect("error serializing");
+        let json = serde_json::to_string(&rope).expect("error serializing");
         let rope2 = from_str::<Rope>(json.as_str()).expect("error deserializing");
         assert_eq!(rope, rope2);
     }
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -413,7 +413,7 @@ impl<'de> Visitor<'de> for RopeVisitor {
     where
         E: de::Error,
     {
-        Rope::from_str(s).map_err(|_| de::Error::invalid_value(DeUnexpected::Str(s), &self))
+        Rope::from_str(s).map_err(|_| de::Error::invalid_value(de::Unexpected::Str(s), &self))
     }
 }
 
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -472,7 +472,7 @@ impl<'de> Deserialize<'de> for Delta<RopeInfo> {
             fn from(elem: RopeDeltaElement_) -> DeltaElement<RopeInfo> {
                 match elem {
                     RopeDeltaElement_::Copy(start, end) => DeltaElement::Copy(start, end),
-                    RopeDeltaElement_::Insert(r) => DeltaElement::Insert(r),
+                    RopeDeltaElement_::Insert(rope) => DeltaElement::Insert(rope),
                 }
             }
         }
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -396,7 +396,7 @@ impl<'de> Deserialize<'de> for Rope {
     where
         D: Deserializer<'de>,
     {
-        deserializer.deserialize_str(RopeVisitor {})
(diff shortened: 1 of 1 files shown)


rust/rope/src/rope.rs
```diff
@@ -1240,4 +1256,19 @@ mod tests {
...
+    fn serialize_and_deserialize() {
+        use serde_json::{from_str, to_string};
+        const TEST_LINE: &str = "test line\n";
+
+        // repeat test line enough times to exceed maximum leaf size
+        let n_seg = MAX_LEAF / TEST_LINE.len() + 1;
+        let test_str = TEST_LINE.repeat(n_seg);
+
+        let rope = Rope::from(test_str.as_str());
+        let json = to_string(&rope).expect("error serializing");
```
portstrom: I was wondering where `to_string` comes from and had to search to find the explanation. It would be clearer to call it `serde_json::to_string`, and as a bonus the `use` statement can be removed.
```suggestion
        let json = serde_json::to_string(&rope).expect("error serializing");
```

rust/rope/src/rope.rs
```diff
@@ -1240,4 +1256,19 @@ mod tests {
...
+        let rope2 = from_str::<Rope>(json.as_str()).expect("error deserializing");
```
portstrom: It's clearer to use words to describe things than to call them by numbers.
```suggestion
        let deserialized_rope = serde_json::from_str::<Rope>(json.as_str()).expect("error deserializing");
```

rust/rope/src/rope.rs
```diff
@@ -1240,4 +1256,19 @@ mod tests {
...
+        assert_eq!(rope, rope2);
```
portstrom:
```suggestion
        assert_eq!(rope, deserialized_rope);
```

rust/rope/src/rope.rs
```diff
@@ -29,7 +29,7 @@ use crate::tree::{Cursor, Leaf, Metric, Node, NodeInfo, TreeBuilder};
 
 use bytecount;
 use memchr::{memchr, memrchr};
-use serde::de::{Deserialize, Deserializer};
+use serde::de::{self, Deserialize, Deserializer, Unexpected as DeUnexpected, Visitor};
```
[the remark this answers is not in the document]
ahouts: Fair enough, probably clearer that way anyway.

ahouts:
@portstrom ty for review, all suggestions committed

rust/rope/src/rope.rs
```diff
@@ -396,8 +396,24 @@ impl<'de> Deserialize<'de> for Rope {
     where
         D: Deserializer<'de>,
     {
-        let s = String::deserialize(deserializer)?;
-        Ok(Rope::from(s))
+        deserializer.deserialize_str(RopeVisitor {})
+    }
+}
+
+struct RopeVisitor {}
```
cmyr: A struct with no fields (a 'unit-like struct') doesn't need any braces:

```rust
struct RopeVisitor;
```

rust/rope/src/rope.rs
```diff
@@ -443,7 +459,7 @@ impl<'de> Deserialize<'de> for Delta<RopeInfo> {
         #[serde(rename_all = "snake_case")]
         enum RopeDeltaElement_ {
             Copy(usize, usize),
-            Insert(String),
+            Insert(Node<RopeInfo>),
```
cmyr: hmm, I wonder if we can get rid of this whole custom deserialize impl now? The whole strategy of deserializing to internal types and then converting might not be necessary now that we have the visitor? I'm not totally sure if this is the case, since I forget exactly what the constraints were that inspired this originally, but it was something like this. Maybe worth playing around with for a few minutes?

ahouts: If we want to derive Serialize and Deserialize we would have to add those traits to the NodeInfo trait in DeltaElement and Delta, since we can't use derive for a concrete case like Delta<RopeInfo>

cmyr: Okay, thanks, let's keep it as is then.

rust/rope/src/rope.rs
```diff
@@ -396,8 +396,24 @@ impl<'de> Deserialize<'de> for Rope {
     where
         D: Deserializer<'de>,
     {
-        let s = String::deserialize(deserializer)?;
-        Ok(Rope::from(s))
+        deserializer.deserialize_str(RopeVisitor {})
```
cmyr: This `{}` is also now unnecessary. I'm actually surprised this compiled, but firing up a playground you can kind of see what's going on. With the three different struct types, the uncommented lines are valid constructors 😁 

```rust
struct Normal{}
struct Tuple();
struct Unit;

fn main() {
    // let n = Normal;
    // let n1 = Normal();
    let n2 = Normal{};
    
    let t = Tuple;
    let t1 = Tuple();
    let t2 = Tuple{};
    
    let u = Unit;
    // let u1 = Unit();
    let u2 = Unit{};
}
```

ahouts: I avoided the "constructors" without bracket/parenthesis previously because they look similar to traits, but I guess that doesn't matter now that dyn is a thing.

cmyr: Traits, even marker traits, require brackets, unless I'm misunderstanding?

```rust
trait MySync; // doesn't compile
trait MySend {} // compiles
```

ahouts: You are right, it was more of a familiarity thing for me.
```rust
// constructor, x an instance of Thing
let x = Thing::new();
// constructor, x an instance of Thing
let x = Thing{};
// x is a reference to the type Thing
let x = Thing;
```
Of course there is not actually an "instance" of Thing because it is optimized away, but all of the languages I have worked with would require an explicit constructor even for a type with no value.

cmyr: Gotcha. Yea, I guess since there aren't type-level functions or higher-kinded types it doesn't really make sense to have a reference to a type, at this point in Rust, but maybe at some point down the road?

Issue #698, opened 2018-06-08
Revisit page_up and page_down


cmyr:
(breaking out from #612)

We used to have RPCs for `page_up` and `page_down` that behaved differently from `scroll_page_up` and `scroll_page_down`. The distinction is in whether or not the cursor is moved. Scrolling by page without moving the cursor is something that we would want for, say, vi mode.

kanishkarj:
I would like to work on this.

jansol:
Sure, go ahead!

PR #1114, opened 2019-02-09
Add Page Up and Page Down

kanishkarj:
Addresses #698
- [x] Implement
- [ ] Test if working
- [ ] Write tests

I am not sure how to test this. Can someone guide me?

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -267,6 +267,7 @@ impl View {
             Replace { chars, preserve_case } => self.do_set_replace(chars, preserve_case),
             SelectionForReplace => self.do_selection_for_replace(text),
             SelectionIntoLines => self.do_split_selection_into_lines(text),
+            PageUpDown(movement) => self.page_up_down(movement),
         }
     }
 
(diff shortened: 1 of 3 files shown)


rust/core-lib/src/view.rs
```diff
@@ -351,6 +352,20 @@ impl View {
         self.set_selection(text, new_sel);
     }
 
+    pub fn page_up_down(&mut self, movement: Movement) {
+        let first_line: i64 = self.first_line as i64;
+        let scroll_height: i64 = self.scroll_height() as i64;
+        match movement {
+            Movement::DownPage => {
+                self.set_scroll(first_line - scroll_height, first_line);
```
scholtzan: Right now this only updates `first_line` and `height` in `View` but does not tell the frontend to actually scroll to that position. So for now nothing happens. 
To make it scroll you have to set `self.scroll_to = Some(<position>);`
There are some functions in `View` where this is already used, for example in `scroll_to_cursor`, which might be helpful.

kanishkarj:
But doesn't xi-editor already implement that? PageUp and PageDown are working on the `GXI` front-end. That's why I assumed I was supposed to implement ScrollPageUp/ScrollPageDown. I have attached a clip of what happens in GXI on PageUp and PageDown, take a look at it.

peek 2019-02-10 11-38

scholtzan:
Oh, I see. It seems that the xi-mac frontend doesn't implement it correctly. Yeah, then it makes sense to not change the selections when doing `PageUp`/`PageDown`. `ScrollPage` seems to move the cursor to the end/beginning of the visible line range.

kanishkarj:
Yes, so shall I go with implementing this properly ? Like without changing the selection.

scholtzan:
Yeah sure, go ahead.

kanishkarj:
I need some help with getting the cursor to not move. I tried messing around with selections and carets, but couldn't get much out of it.

jansol:
@scholtzan do you have any pointers for this?

PR #1118, opened 2019-02-17
Allow plugins to subscribe to core events

scholtzan:
# Allow plugins to subscribe to core events

This is more of an exploratory PR. It would be nice if plugins could subscribe to events happening in the core, for example to changes of selections, movements, gestures, etc.

Plugins can subscribe to `CoreNotification` and `EditNotification` by using the snake case name of the notification:

```rust
impl Plugin for ExamplePlugin {
    type Cache = ChunkCache;

    fn new_view(&mut self, view: &mut View<Self::Cache>) {
        view.subscribe("insert");
        view.subscribe("click");
    }
}
```

Subscribed events can be processed in the plugin:

```rust
fn subscribed_notification(
        &mut self,
        notification: CoreNotification) {

    match notification {
        CoreNotification::Edit(_) => {},
        _ => {}
    }
}
```

I am not sure if it's a good idea to subscribe to `CoreNotification`s or if instead xi should send some custom events such as `SelectionChange`, ... since for example in the case of selections there are several `EditNotifications` that cause changes (`SelectAll`, `Drag`, `FindAll`, ...). 
I'd be happy to discuss better potential solutions.

## Review Checklist

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -17,6 +17,8 @@ toml = "0.4"
 notify = { optional = true, version = "4.0" }
 regex = "1.0"
 memchr = "2.0.1"
+strum = "0.13.0"
+strum_macros = "0.13.0"
 
 xi-trace = { path = "../trace", version = "0.1.0" }
 xi-trace-dump = { path = "../trace-dump", version = "0.1.0" }
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -381,6 +381,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "heck"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "unicode-segmentation 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "humantime"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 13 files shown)


rust/core-lib/Cargo.toml
```diff
@@ -17,6 +17,8 @@ toml = "0.4"
 notify = { optional = true, version = "4.0" }
 regex = "1.0"
 memchr = "2.0.1"
+strum = "0.13.0"
+strum_macros = "0.13.0"
```
scholtzan: This allows to print an enum as string.

cmyr:
Cool, I think this is an interesting idea, and I think we will want something like it eventually; in particular I like that this requires plugins to explicitly opt in to various notification types, which will make messages low volume by default.

scholtzan:
Yeah, I think that from a plugin developers (and my personal) perspective this is a quite important feature to have. Right now the only changes that plugins are notified about are text changes. But most plugins will want to respond to other events besides those. (For example a word count plugin might want to count words in selections and needs to know when these change.)

rust/core-lib/Cargo.toml
Aareon: Would it not be feasible to, instead of adding another dependency, to print enums as strings in long form? I.e. [the answers to this SO question](https://stackoverflow.com/questions/28024373/is-there-a-way-to-print-enum-values)

scholtzan: This works for enums that don't have any fields. But if they have fields then their values would be part of the string:

```rust
#[derive(Debug)]
pub enum E {
    Delete,  // -> this works fine
    Edit(usize, usize) //  -> would print for example: "Edit(3, 4)"
}
```

Aareon: That's a fair point. I hadn't thought of that.

PR #1123, opened 2019-03-02
Add 'peek_next_codepoint' to rope cursor

### LINKED ISSUE #1120: [Question][xi-rope] peek next char from cursor
Hi everyone!

I am working on a parser lib and I use xi-rope as underlying data structure.
I need to be able to peek the next char, without advancing the cursor position. How can I do that?

For example - the code below is broken, because it is supposed to skip only whitespace, but it also consumes one extra char.

```rust

fn skip_whitespace(cursor: & mut Cursor<RopeInfo>) -> usize {
    let mut i: usize = 0;
    while let Some(c) = cursor.next_codepoint() {
        if !(c.is_whitespace()) {
            break
        } else {
            i+=1;
        }
    }
    return i;
}
...
#[test]
fn skip_whitespaces() {
    let r = Rope::from("  hello ");
    let mut cursor = Cursor::new(&r, 0);
    let spaces = skip_whitespace(&mut cursor);
    assert_eq!(spaces, 2);
}
```

I can rewind the cursor one char back, but I want to avoid that for efficiency sake.
I found that there is a way to get iterator over boundaries, but I don't know how to read chars from it.

Please help

cmyr:
This seems like a reasonable feature. 🤷‍♂️

Closes #1120

### DIFF
--- a/rust/rope/src/rope.rs
+++ b/rust/rope/src/rope.rs
@@ -772,6 +772,12 @@ impl<'a> Cursor<'a, RopeInfo> {
         }
     }
 
+    /// Get the next codepoint after the cursor position, without advancing
+    /// the cursor.
+    pub fn peek_next_codepoint(&self) -> Option<char> {
+        self.get_leaf().and_then(|(l, off)| l[off..].chars().next())
+    }
+
     pub fn next_grapheme(&mut self) -> Option<usize> {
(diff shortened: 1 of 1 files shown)


ngortheone:
@raphlinus  @cmyr  Thanks everyone!

PR #1128, opened 2019-03-04
Annotation support for plugins

scholtzan:
This adds annotation support for plugins.

## Review Checklist


rust/core-lib/src/editor.rs
```diff
@@ -900,6 +901,40 @@ impl Editor {
...
+        rev: RevToken,
+    ) {
+        let _t = trace_block("Editor::update_annotations", &["core"]);
+
+        for annotation_slice in annotation_slices {
+            let mut start = start;
+            let mut end_offset = start + len;
+            let mut annotations = annotation_slice.to_annotations(view, &self.text);
+
+            if rev != self.engine.get_head_rev_id().token() {
```
cmyr: it's a curious question what we want to do in this case, but this seems reasonable for now!

rust/core-lib/src/annotations.rs
```diff
@@ -78,6 +83,25 @@ impl AnnotationSlice {
             "n": self.ranges.len()
         })
     }
+
+    pub fn to_annotations(&self, view: &View, text: &Rope) -> Annotations {
+        let last_entry = self.ranges.last().unwrap_or(&[0, 0, 0, 0]);
+        let mut sb = SpansBuilder::new(last_entry[2] + last_entry[3]);
```
cmyr: this has taken me a while to figure out.

As far as I understand it this is equivalent to,

```rust
let span_len = self.ranges.last().map(|anno_range| anno_range.2 + anno_range.3).unwrap_or(0);
// ...
```

which might be clearer?

rust/core-lib/src/annotations.rs
```diff
@@ -44,14 +43,20 @@ impl AnnotationType {
 pub type AnnotationRange = Vec<[usize; 4]>;
```
cmyr: seeing this in use below now i wonder if it might make sense to make this a struct or something with named fields? Just thinking out loud. 🤷‍♂️

rust/core-lib/src/annotations.rs
cmyr: also it isn't totally clear that this logic is right, because it isn't clear that lines & cols are in the same unit space?

rust/core-lib/src/annotations.rs
scholtzan: Yeah, it's more readable as a struct

rust/core-lib/src/annotations.rs
scholtzan: Oh I forgot to commit and push the most recent change. Lines have to be translated into their offset.

scholtzan:
I did the suggested changes.
I also implemented invalidation for plugin annotations. So annotations that overlap with edits are invalidated and removed from the annotation store.

rust/core-lib/src/annotations.rs
```diff
@@ -122,8 +167,35 @@ impl AnnotationStore {
     }
 
     /// Invalidates and removes all annotations in the range of the interval.
-    pub fn invalidate(&mut self, _interval: Interval) {
-        // todo
+    pub fn invalidate(&mut self, interval: Interval) {
+        for val in self.store.values_mut() {
+            let mut annotations: Vec<Annotations> = Vec::new();
```
cmyr: I think it's worth rethinking this. This implementation is doing a linear scan of each annotation from each plugin, and I'm wondering if there's a better way. I would probably start by having a `fn invalidate(&mut self, interval: Interval)` function on `Annotations`, and then in there would have the logic for operating on each `Spans` object.

There, what we really want to be doing is something that takes advantage of the tree structure of `Spans`.

It might be easiest to add a new method to spans, with a name like `delete_intersecting(iv)`, and ideally we wouldn't be scanning the whole structure, but instead taking the leading valid section, the trailing valid section, and gluing them together with a new blank section.

scholtzan: You're right. I added `delete_intersecting` to `Spans` and cleaned up the invalidation logic in annotations.

scholtzan:
I cleaned up the invalidation logic and added some tests for annotations.

rust/core-lib/src/annotations.rs
```diff
@@ -89,34 +164,193 @@ pub trait ToAnnotation {
...
-        _iv: Interval,
-        _items: Spans<Value>,
-    ) {
-        // todo
+    pub fn update(&mut self, source: PluginId, interval: Interval, item: Annotations) {
+        let updated_items = item.clone();
+        self.store
+            .entry(source)
+            .and_modify(|e| {
+                let mut annotations = e
```
cmyr: I wonder if we can use Vec::retain here?

rust/core-lib/src/annotations.rs
```diff
@@ -89,34 +164,193 @@ pub trait ToAnnotation {
...
+                        annotations.push(updated_annotations);
+                    }
+                    None => {
+                        annotations.push(updated_items);
+                    }
+                }
+
+                *e = annotations;
+            })
+            .or_insert(vec![item]);
```
cmyr: I spent a long time staring at this function trying to figure out what was going on. As far as I can tell, it's equivalent to:

```rust
if !self.store.contains_key(&source) {
    self.store.insert(source, vec![item]);
    return;
}

let entry = self.store.get_mut(&source).unwrap();
if let Some(annotation) = entry.iter_mut().find(|a| a.annotation_type == item.annotation_type) {
    annotation.update(Interval::new(0, interval.end()), item.items);
} else {
    entry.push(item);
}
```

Does this seem true to you? I'm still confused about why we zero out interval.start.

scholtzan: Oh wow this is so much cleaner! Yeah I also cannot really wrap my head around the zero offset for the interval :/

scholtzan:
I cleaned up the code and added some short documentation

rust/core-lib/src/annotations.rs
cmyr: Okay, so I think this needs more thought: we're being given one interval but we're replacing it with another one. Why?

When we use spans elsewhere (such as for syntax highlighting) we ensure that the length of the total spans object is always equal to the length of the rope. In this case, the interval that we're sent by a plugin _probably_ refers to some specific region of the document, and zeroing that out would be incorrect?

This depends on the semantics of the API, so you'll have a better idea than I do at this point, but that's why this feels like it might be wrong.

rust/rope/src/spans.rs
```diff
@@ -300,6 +300,40 @@ impl<T: Clone> Spans<T> {
         }
         *self = b.build();
     }
+
+    /// Deletes all spans that intersect with `interval`.
+    pub fn delete_intersecting(&mut self, interval: Interval) {
+        let mut cursor = Cursor::new(self, interval.start());
+
+        while cursor.pos() > interval.start() && cursor.prev_leaf().is_some() {}
```
cmyr: this is a strange pattern, and I'm not totally sure what you're trying to do. 

This _is_ a tricky problem for interval trees. For instance, there may be an interval in the first leaf that spans the entire tree needs to be removed correctly. As a simple example, the following test case fails here:

```rust
    #[test]
    fn delete_intersecting_big_at_start() {
        let mut sb = SpansBuilder::new(10);
        sb.add_span(0..10, 0);

        let mut spans = sb.build();
        assert_eq!(spans.iter().count(), 1);

        spans.delete_intersecting(1..3);
        assert_eq!(spans.iter().count(), 0);
    }
```

In any case, happy to chat about this!

rust/core-lib/src/editor.rs
```diff
@@ -900,6 +901,42 @@ impl Editor {
...
+                if let Ok(delta) = self.engine.try_delta_rev_head(rev) {
+                    let mut transformer = Transformer::new(&delta);
+                    let new_start = transformer.transform(start, false);
+                    if !transformer.interval_untouched(Interval::new(start, end_offset)) {
+                        annotations.items =
+                            annotations.items.transform(start, end_offset, &mut transformer);
```
scholtzan: This step is unnecessary for `AnnotationSlice`

rust/core-lib/src/annotations.rs
scholtzan: Right, good point. When we update spans eg. for syntax highlighting then the interval of the spans are all relative to the interval to be updated.
However, when updating annotations `AnnotationSlice`s are sent which specify the absolute position of the annotations in the document.

Now I'm wondering if it would be more consistent to use `Spans` instead of `AnnotationSlice` in `UpdateAnnotations`? On the other hand, there is also the `ToAnnotation` trait which could be implemented for custom types and that one returns `AnnotationSlice`. 

If we decide to keep using `AnnotationSlice` then the current problem should be easily resolved by not doing the `transform` step [that I marked in the code](https://github.com/xi-editor/xi-editor/pull/1128/files#r266140348).

rust/rope/src/spans.rs
scholtzan: Looking at this again I am also not sure why I iterated over the leaves. 

I fixed the logic in that function, but it is a bit hairy now. Maybe any suggestions on a cleaner way to write this?

rust/core-lib/src/editor.rs
scholtzan: Oh no nvm, we need this

scholtzan:
I created a very simple plugin that uses the annotations for testing: https://github.com/scholtzan/xi-todo-highlight and added support to xi-mac on a separate branch https://github.com/scholtzan/xi-mac/tree/todo-annotations. The plugin highlights all the lines that contain "todo" or "fixme" or "note". I wrote the plugin mainly for testing purposes for the annotations.

scholtzan:
Something else I noticed: `line_of_offset` in `core-lib` is different from `line_of_offset` in `plugin-lib`. Core counts visual lines while the plugin library seems to count the logic lines. This results in problems if there is line wrapping, then annotations send from plugins are not shown at the correct positions.

cmyr:
yes I think plugins should be providing annotations as just ranges of the document; it's only when talking to the frontend that we care about line numbers.

scholtzan:
I added `DataSpan` to represent the updated annotations in `UpdateAnnotations`. Not sure if it's the best name for that though...

PR #1140, opened 2019-03-20
[WIP] Add behavior when xi-editor detects external changes on an edited file

### LINKED ISSUE #1126: Tell the frontend when a file has been changed since opening it
This would be a continuation of #538, I suppose. Right now xi only automatically reloads the file if no changes have been done by xi yet, but it'd be nice if xi-editor told the frontend that the file has changed and as such the user should reconsider saving.

mikaelmello:
## Summary

We send a 'content_changed' notification to the client

I guess that this will need more changes, but this is a first move to get feedback from the maintainers.

Is this the right direction?

## Related Issues

Closes #1126

## Review Checklist

### DIFF
--- a/rust/core-lib/src/client.rs
+++ b/rust/core-lib/src/client.rs
@@ -74,6 +74,15 @@ impl Client {
         self.0.send_rpc_notification("available_languages", &json!({ "languages": languages }))
     }
 
+    pub fn file_externally_changed(&self, view_id: ViewId) {
+        self.0.send_rpc_notification(
+            "file_externally_changed",
+            &json!({
+                "view_id": view_id,
+            }),
+        );
+    }
+
     pub fn theme_changed(&self, name: &str, theme: &ThemeSettings) {
         self.0.send_rpc_notification(
             "theme_changed",
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -718,18 +718,21 @@ impl CoreState {
         // determining if a file has been changed by another process.
         // A more robust solution would also hash the file's contents.
 
+        // this is ugly; we don't map buffer_id -> view_id anywhere
+        // but we know we must have a view.
+        let view_id = self
+            .views
+            .values()
+            .find(|v| v.borrow().get_buffer_id() == buffer_id)
+            .map(|v| v.borrow().get_view_id())
+            .unwrap();
+
         if has_changes && is_pristine {
             if let Ok(text) = self.file_manager.open(path, buffer_id) {
-                // this is ugly; we don't map buffer_id -> view_id anywhere
-                // but we know we must have a view.
-                let view_id = self
-                    .views
-                    .values()
-                    .find(|v| v.borrow().get_buffer_id() == buffer_id)
-                    .map(|v| v.borrow().get_view_id())
-                    .unwrap();
                 self.make_context(view_id).unwrap().reload(text);
             }
+        } else if has_changes {
+            self.peer.file_externally_changed(view_id);
         }
     }
 


rust/core-lib/src/client.rs
```diff
@@ -74,6 +74,15 @@ impl Client {
         self.0.send_rpc_notification("available_languages", &json!({ "languages": languages }))
     }
 
+    pub fn content_changed(&self, view_id: ViewId) {
+        self.0.send_rpc_notification(
+            "content_changed",
```
Cogitri: Not sure if this has been discussed elsewhere, but I personally feel this this doesn't express that the file _hasn't_ been changed by Xi but by something else. I'm really bad at naming things and the names should be short, but something like `file_externally_changed` would be so much clearer to me as frontend dev

rust/core-lib/src/client.rs
```diff
@@ -74,6 +74,15 @@ impl Client {
...
+            &json!({
+                "view_id": view_id,
```
Cogitri: Since we already capture the mtime of the file maybe we could also tell the frontend when the file has been changed? Not too relevant though, I suppose.

rust/core-lib/src/client.rs
mikaelmello: I agree, `file_externally_changed` does not seem to be too long.

rust/core-lib/src/client.rs
mikaelmello: How is the mtime retrieved from the front-end, is it too much additional work?

Also, is it relevant when notifying the user and asking how to proceed? The only use case I can think of is something along the lines of `The file has been changed outside of Xi {x minutes ago}` or `The file has been changed outside of Xi at {13:58 PM}`. But I don't know if that's good UX

Cogitri: > How is the mtime retrieved from the front-end, is it too much additional work?

Xi already has that info, it uses that to determine if the file has been changed (arguably it should use checksumming for that)

> But I don't know if that's good UX

Neither do I, although I'd personally find it interesting

cmyr: I'm not sure if this is something the frontend can do much about. These notifications are going to get sent more or less immediately, and the frontend isn't supposed to really be touching the file-system or anything like that, so I'd be inclined to leave it out.

Cogitri:
I guess this should also remove the `alarm` which is issued when we try to save to a file which has been changed externally? Tested this with gxi and the current changeset causes it to still display and error message about the file being changed externally when changing. Also, this isn't necessarily something which should already be implemented in this PR, but I guess we should get a `force_overwrite` param (or similar) for `save` to actually do something about the externally changed file?

Also, maybe it'd be worth updating the docs telling frontends what they should do about the notification?

I guess possibilities are:

* Ignore, continue editing.
* Force reload the file.

Right? Should frontends just keep a `has_changed_externally` bit or something if a `file_externally_changed` has been received for a view so it can handle this during saving (since right now Xi still refuses to save to a changed file and as such doesn't really leave me a choice other than reloading the file/saving to a different file and diffing)

cmyr:
> I guess this should also remove the `alarm` which is issued when we try to save to a file which has been changed externally? Tested this with gxi and the current changeset causes it to still display and error message about the file being changed externally when changing. Also, this isn't necessarily something which should already be implemented in this PR, but I guess we should get a `force_overwrite` param (or similar) for `save` to actually do something about the externally changed file?

Yes, I think we probably want something like this. The question is, right now, what do we expect a frontend to do with this notification? We're pretty limited by the current API. What changes would we need to make this useful?

Cogitri:
> What changes would we need to make this useful?

Hum, I guess it's easiest to show this with an example. In gxi I'd implement the following once this has landed:

* If we receive `context_changes` prompt the user that the file has been changed. Either let him reload the file, or not do anything
* On saving, prompt the user again that the file has been changed. Either let him force-overwrite the old file (I guess we'd need a `force` param for save for that?) or cancel to give him the option to save it elsewhere. IMHO if the user presses 'save' again the same dialog should pop up again.

cmyr:
Okay, so we will need some way to force reload, and we'll need some way to force overwrite? The reload probably has to be a new method, the force overwrite can be a flag on `save`. Maybe we want to add these changes first?

Cogitri:
Or we add them in here, but sure, those two things will have to be added. I think with those two things in place frontends can implement about any of the ways of handling this we've discussed in #1126 I think (at least if they keep track of us sending them `file_externally_changed`, maybe we should add a `tainted` flag to save too?)

cmyr:
well core will keep track of 'dirty' ness, and it won't be cleared until either 'force_reload' or 'save (force)' are sent. So I don't _think_ the frontend needs any special tracking of anything?

Cogitri:
> So I don't think the frontend needs any special tracking of anything?

Oh, it has to keep track of the dirtyness if it doesn't want to immediately notify the user about a change to the file, but only prompt him about that during saving and giving him options like `save|save as|overwrite|cancel` AFAICS

cmyr:
hmm. I basically try to avoid any situation where some piece of state exists in both the frontend and backend, because then we have to deal with keeping that state in sync.

To deal with the situation you describe, what makes the most sense to me is that the user tries to save, and then that fails, and _then_ they can figure out how to resolve that. To the user this would just look like 'try to save' -> 'get dialog about other changes' -> 'resolve conflict'.

mikaelmello:
> Oh, it has to keep track of the dirtyness if it doesn't want to immediately notify the user about a change to the file, but only prompt him about that during saving

Why do you think that is the best approach? I have no experience whatsoever with UX so sometimes I like to try and learn a little. All editors that I remember using notify on change, not saving, so I'm curious as to why you chose this behavior

Cogitri:
> Why do you think that is the best approach? I have no experience whatsoever with UX so sometimes I like to try and learn a little. All editors that I remember using notify on change, not saving, so I'm curious as to why you chose this behavior

Oh, I don't, as I've mentioned in a previous comment in here (sorry, can't link on mobile it seems?) I'll go with notify on change in gxi. However, I'd prefer to give other frontends the _possibility_ to have different behaviour if they prefer that.

> To deal with the situation you describe, what makes the most sense to me is that the user tries to save, and then that fails, and then they can figure out how to resolve that. 

That does seem a tiny bit less intuitive, but much easier to implement and I definitely don't mind going this route, sound good to me :)

cmyr:
Intuitive from the position of the developer or the user?

I do think we've muddied the waters a bit here. Is it worth backing up and thinking a little bit more about what we're trying to achieve and what a reasonable design is?

Cogitri:
> Is it worth backing up and thinking a little bit more about what we're trying to achieve and what a reasonable design is?

Sure, how would you suggest we do that? :)

cmyr:
If you'd like you could write a small RFC, and PR that? Alternatively we could continue discussion in #1126, or open a new discussion issue.

PR #1148, opened 2019-03-24
Use sha256 checksum instead of modtime to check if file has changed

Cogitri:
Kind of related to #1140

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -17,6 +17,9 @@ toml = "0.5"
 notify = { optional = true, version = "4.0" }
 regex = "1.0"
 memchr = "2.0.1"
+sha2 = "0.8"
+# Limited by digest 0.8
+generic-array = "0.12"
 
 xi-trace = { path = "../trace", version = "0.1.0" }
 xi-rope = { path = "../rope", version = "0.3" }
@@ -28,8 +31,7 @@ version = "3.0"
 default-features = false
 features = ["assets","dump-load-rs","dump-create"]
 
-[target."cfg(target_os = \"fuchsia\")".dependencies]
-sha2 = "0.7"
+#[target."cfg(target_os = \"fuchsia\")".dependencies]
 # Note: these can be available if needed, but the idea is that the
 # platform-specific code moves entirely to the binary wrapper that
 # includes xi-core-lib.
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -14,6 +14,8 @@
 
 //! Interactions with the file system.
 
+use generic_array::GenericArray;
+use sha2::{Digest, Sha256};
 use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
@@ -51,6 +53,7 @@ pub struct FileInfo {
     pub encoding: CharacterEncoding,
     pub path: PathBuf,
     pub mod_time: Option<SystemTime>,
+    pub sha_sum: Option<GenericArray<u8, generic_array::typenum::U32>>,
     pub has_changed: bool,
     #[cfg(target_family = "unix")]
     pub permissions: Option<u32>,
@@ -98,7 +101,10 @@ impl FileManager {
         if let Some(info) = self.file_info.get_mut(&id) {
             let mod_t = get_mod_time(path);
             if mod_t != info.mod_time {
-                info.has_changed = true
+                let sha = get_sha256(path);
+                if sha != info.sha_sum {
+                    info.has_changed = true
+                }
             }
             return info.has_changed;
         }
@@ -143,6 +149,7 @@ impl FileManager {
         let info = FileInfo {
             encoding: CharacterEncoding::Utf8,
             path: path.to_owned(),
+            sha_sum: get_sha256(path),
             mod_time: get_mod_time(path),
             has_changed: false,
             #[cfg(target_family = "unix")]
@@ -168,7 +175,9 @@ impl FileManager {
             let encoding = self.file_info[&id].encoding;
             try_save(path, text, encoding, self.get_info(id))
                 .map_err(|e| FileError::Io(e, path.to_owned()))?;
-            self.file_info.get_mut(&id).unwrap().mod_time = get_mod_time(path);
+            let file_info = self.file_info.get_mut(&id).unwrap();
+            file_info.sha_sum = get_sha256(path);
+            file_info.mod_time = get_mod_time(path)
         }
         Ok(())
     }
@@ -190,6 +199,7 @@ where
     let info = FileInfo {
         encoding,
         mod_time: get_mod_time(&path),
+        sha_sum: get_sha256(&path),
         #[cfg(target_family = "unix")]
         permissions: get_permissions(&path),
         path: path.as_ref().to_owned(),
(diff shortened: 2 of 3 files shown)


Cogitri:
Tested locally and seems to work fine on a FS without mtime enabled, but please do test yourself.

cmyr:
Just out of curiosity, was there a specific motivation for this work?

I forget the exact rationale for picking modtime originally, but I did consider using a checksum. I _think_ the reasoning was that modtime is going to provide some number of false positives but that we handle those further up the line (we diff the files after reloading) whereas calculating the checksum has a higher fixed cost. The real question is probably, 'how often in this call are we actually looking at a different file'?

Cogitri:
As you've mentioned modtime introduced some false positives for me, which caused the alarm that the file has been changed to not be sent reliably :/

> whereas calculating the checksum has a higher fixed cost. 

Yes, it does come at a cost, but IMHO that cost is tiny to the cost of rewriting a document because the modtime detection didn't work and you've overwritten a changed file. We could use a less expensive checksumming algo, like md5 or a smaller shasum if you want to.

> The real question is probably, 'how often in this call are we actually looking at a different file'?

Hm?

cmyr:
> The real question is probably, 'how often in this call are we actually looking at a different file'?

Agree that isn't clear. What I mean is: how often when we receive this notification (of a possible file change) are we dealing with an _actual_ file change?  I don't have the answer and don't expect you to, it just struck me as a reasonable guiding question.

Looking at some casual benchmarks sha256 looks only marginally slower than the alternatives, so don't mind that.

Cogitri:
> are we dealing with an actual file change? I don't have the answer and don't expect you to, it just struck me as a reasonable guiding question.

True that, I usually only hit this during `git rebase` (but that is like multiple times a week for me). I guess this heavily differs from person to person though, most people just doing simple editing won't ever hit this, I guess.

> Looking at some casual benchmarks sha256 looks only marginally slower than the alternatives, so don't mind that.

Alright, but I'm not sure how this plays out for big files

cmyr:
well we're going to be reading the file in anyway much of the time. idk. I was looking through the code, and I was _hoping_ that we computed the diff before we marked the file as dirty. We don't. That might be another approach, although I'm not sure it's better.

How about this: we keep modtime and add sha256, but we only bother computing the hash when modtime has changed? This way we use the cheapest option when possible.

Cogitri:
> How about this: we keep modtime and add sha256, but we only bother computing the hash when modtime has changed? This way we use the cheapest option when possible.

Sounds good, will look into that

rust/core-lib/src/file.rs
```diff
@@ -260,10 +260,14 @@ impl CharacterEncoding {
...
-/// if present.
-fn get_mod_time<P: AsRef<Path>>(path: P) -> Option<SystemTime> {
-    File::open(path).and_then(|f| f.metadata()).and_then(|meta| meta.modified()).ok()
+/// Returns the SHA256 checksum of a file at a given path, if present.
+fn get_sha256<P: AsRef<Path>>(path: P) -> Option<String> {
+    let mut hasher = Sha256::new();
+    File::open(path)
+        .and_then(|mut f| std::io::copy(&mut f, &mut hasher))
+        .map(|_| hasher.result())
+        .map(|s| format!("{:x}", s))
```
portstrom: Why would you hex encode it? Rust works very well with binary data.

Cogitri: Ah, was mainly just so I can print out the sha when it has changed, not quite sure if we want to do rhat though.

portstrom: If we want to do that, then let's add an implementation of the [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) trait to hex encode when printing.

Cogitri:
> How about this: we keep modtime and add sha256, but we only bother computing the hash when modtime has changed? This way we use the cheapest option when possible.

Actually, thinking about this again, we have to compute one out of the two hashes all the time anyway. Computing both hashes after the mod time has changed would be too late after all (since we'd compare the hash of the changed file with the hash of the changed file)

rust/core-lib/src/file.rs
Cogitri: Alright, I've just opted to not do that since we don't print it out anywhere yet.

portstrom: Converting `Vec` is also an unnecessary inefficiency. Better store the `GenericArray` you get from the `sha2` crate, which doesn't involve any pointers, lengths or allocations.

Cogitri: Alright, but to be honest I'm not exactly sure how we return a `GenericArray<u8, ArrayLength<u8>>` (since ArrayLength only is a trait and the generic-array docs haven't helped me and the sha2 docs also uses a Vec for this)

portstrom: It's `generic_array::GenericArray<u8, generic_array::typenum::consts::U32>`. It's also possible to use `[u8; 32]` directly without wrapping it in `GenericArray`, which is stored exactly the same way in memory, but then it won't be compatible with generic functions that require a generic array. Hopefully this distinction will eventually go away as [const generics](https://rust-lang.github.io/rfcs/2000-const-generics.html) are implemented.

cmyr: For our purposes I think that `[u8; 32]` is fine. I also don't really mind the vec, since this is being computed rarely and the performance impact will be ~nil.

rust/core-lib/Cargo.toml
```diff
@@ -28,8 +29,7 @@ version = "3.0"
 default-features = false
 features = ["assets","dump-load-rs","dump-create"]
 
-[target."cfg(target_os = \"fuchsia\")".dependencies]
-sha2 = "0.7"
+#[target."cfg(target_os = \"fuchsia\")".dependencies]
```
cmyr: oops?

rust/core-lib/src/file.rs
```diff
@@ -266,6 +275,17 @@ fn get_mod_time<P: AsRef<Path>>(path: P) -> Option<SystemTime> {
...
 }
 
+/// Returns the SHA256 checksum of a file at a given path, if present.
+fn get_sha256<P: AsRef<Path>>(path: P) -> Option<Vec<u8>> {
+    let mut hasher = Sha256::new();
+    let mut output = Vec::new();
+    File::open(path)
+        .and_then(|mut f| std::io::copy(&mut f, &mut hasher))
+        .map(|_| output.copy_from_slice(hasher.result().as_slice()))
+        .map(|_| output)
```
cmyr: this is weird... we're sort of hijacking the map to return something external. I would prefer,

```rust
File::open(path)
         .and_then(|mut f| std::io::copy(&mut f, &mut hasher))
         .map(|_| output.copy_from_slice(hasher.result().as_slice()))?;
Some(output)
```

but I'm not totally sure what the implication of failure is, here?

rust/core-lib/Cargo.toml
Cogitri: We hard depend on sha2 0.8 anyway, so we don't need to depend on 0.7 here. Crosscompiled to x86_64-fuchsia to test the change and it compiled just fine.

cmyr: Ah didn't notice this was related to sha2. We can just delete this line I think.

Cogitri: Ah, but there are some more commented lines about fuchsia below that, didn't want to destroy the context for thos.

rust/core-lib/src/file.rs
Cogitri: I don't really want to return a `Result<Option<_>, Error>` here :/

cmyr: well okay, but what if the file exists and we can't open it? In general I'm suspicious of discarding errors. If this file doesn't exist, can we handle that earlier, and only call this when we know we have a file? That would let us return a `Result` instead of an option.

Cogitri: Sure, I just tried to mimic get_mod_time here, I guess we should change it too? Anyway, I'm currently on holidays, so that'll take me bit :)

cmyr: Cool, enjoy!

dragostis: Is this the only blocker to this?

Cogitri: Kind of. Will work on this later today. If you feel like helping: Could you test if Xi sends you an alarm WITHOUT this PR when you open a file, then change it and then try to save to it? It doesn't for me, which suggests that modtime doesn't work for me for some reason :c

dragostis: @Cogitri, I don't know exactly what you mean by alarm, but xi seems to be working fine without this PR.

Cogitri: alert is a message from xi which usually opens an error dialogue in your frontend.

Cogitri:
I guess modtime cuts it for now

Issue #1139, opened 2019-03-19
Annotations selection range


NicolaiSoeborg:
- [X] I have searched existing issues and could not find my issue.
- [X] I have studied the documentation.

A few questions about annotations:

1) Wouldn't it make most sense if "annotations" in the very first "update"-RPC response was an empty list (instead of a selection of nothing?)

2) I think the AnnotationRange is wrong when using `add_selection_below`, e.g.:

Create a file with:
```
ABCD
EFGH
```
Now send events: `move_right_and_modify_selection`, `move_right_and_modify_selection`, `add_selection_below` and the annotation range will be: 
 * start_line=0, start_col=0, end_line=0, end_col=2
 * start_line=1, start_col=2, end_line=1, end_col=2

But I think that should be (i.e. all of first line + half of second line):
 * start_line=0, start_col=0, end_line=0, end_col=3
 * start_line=1, start_col=0, end_line=1, end_col=1

cmyr:
This does look suspicious. cc @scholtzan, thoughts?

scholtzan:
One thing to note here is that carets are also represented as annotations. So if we'd send an empty list of annotations in the first `update` then we would lose the caret which starts out to be at the beginning of the document (position `0,0,0,0`).

The `add_selection_below`/`add_selection_above`  seems to imitate the same behaviour as in Sublime. Which seems to be adding a new caret to the previous/next line that is in the same "column" as the original caret.

NicolaiSoeborg:
Okay, so I confused `add_selection_below` with `move_down_and_modify_selection`.

Would you like me to make a PR adding:
 * `add_caret_above` => `add_selection_above`
 * `add_caret_below` => `add_selection_below`
and "deprecating" `add_selection_above/below` (in the documentation) ?

And followup question: Would it make sense to differentiate between carets and selections? I think it would be more logical with an annotation with type `caret`

Also, should new frontends ignore the "cursor" property of lines? And only put caret at annotation with `start_line = end_line` and `start_col = end_col` ?

scholtzan:
One thing to keep in mind is that every change in the frontend protocol will break existing clients. Maybe @cmyr has some thoughts about renaming the methods?

Peltoche:
> And followup question: Would it make sense to differentiate between carets and selections? I think it would be more logical with an annotation with type `caret`

Hi everyone.
A big +1 for this proposition! IMHO the selection zone and the carets are different things and we can see it. 

[...]

PR #1149, opened 2019-03-25
Deprecate AddSelectionAbove/Below

NicolaiSoeborg:
Implement idea from issue 1139

Maybe we want to extend this even further by adding an annotation with type `caret` (and change a bit it the codebase to differentiate between carets and selections).

- [X] I have responded to reviews and made changes where appropriate.
- [X] I have tested the code with `cargo test --all` / `./rust/run_all_checks`.
- [X] I have updated comments / documentation related to the changes I made.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,3 +14,4 @@ Lê Viết Hoàng Dũng
 Markus Ast
 Raph Levien
 Hilmar Gústafsson
+Nicolai Søborg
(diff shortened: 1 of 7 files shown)


cmyr:
I agree that the current behaviour is slightly confusing given the name, but I'm also not convinced we want to introduce new terminology; it's possible this just needs a documentation fix. In any case, let's discuss in #1139.

PR #1152, opened 2019-03-27
Add more benchmarks

kanishkarj:
Added a few more benchmarks.


dsp:
@cmyr can you take alook. I don't have approval rights.

rust/rope/benches/edit.rs
```diff
@@ -87,6 +87,17 @@ fn benchmark_paste_into_line(b: &mut Bencher) {
...
 }
 
+#[bench]
+fn benchmark_large_paste_into_line(b: &mut Bencher) {
+    let mut text = Rope::from(build_short_lines(1_000_000));
+    let insertion = build_short_lines(50_000);
+    let mut offset = 25_000;
+    b.iter(|| {
+        text.edit(offset..=offset, &insertion);
+        offset += 150;
```
cmyr: Reading this is making me wonder if a bunch of these benchmarks (the existing ones) aren't poorly structured.

Specifically, everything in `b.iter(|| {` should be identical between each run, but we're regularly changing external state. This means that different runs are running different code, which sort of defeats the purpose of benchmarking.

dsp: Good catch. I removed my review. @kanishkarj do you think we can have a PR fixing the outlined issue from @cmyr, this means removing the change of state (e.g. offset += 150) in the iteration? We should fix this for the edit.rs benchmark cases and then rebase your PR on top of it.

kanishkarj: Consider it done. :)

PR #1155, opened 2019-03-29
Fix external state in benchmarking (edit.rs)

kanishkarj:
This is in reference to the discussion in PR #1152. This PR fixes the external state of the iterations in the benchmarks.


rust/rope/benches/edit.rs
```diff
@@ -72,7 +72,6 @@ fn benchmark_char_insertion_one_line_edit(b: &mut Bencher) {
     let mut offset = 100;
     b.iter(|| {
         text.edit(offset..=offset, "a");
-        offset += 1;
```
cmyr: maybe it still makes sense to do some loop inside the `iter`? we can just make it some fixed length.

kanishkarj:
Hmm, So you mean to say all the functions use the same offset update ? for eg 
```
offset += 1
```
would be the offset for all the benchmarks?

cmyr:
I would use whatever they were using before, but inside an explicit loop, and having `offset` declared fresh on each `b.iter(|| {`. So we want to be doing the exact same thing inside each `b.iter`. (We might also want to clone over the rope in `b.iter`, so we're using the same underlying rope each time.

cmyr:
@kanishkarj this is very close to mergeable if you want to update again!

PR #1157, opened 2019-03-29
Clarify Pronunciation

walston:
## Summary
Include IPA pronunciation and example of a standard english language rhyme

## Related Issues
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
   <a href="https://github.com/xi-editor/xi-editor">Xi Editor</a>
 </h1>
 
-<p align="center"><em>(pronounced "Zigh")</em></p>
+<p align="center"><em>pronounced "zaɪ̯"</em>; rhymes with "sigh", but starts with "z" from "zoo"</p>
 
 <h4 align="center">A modern editor with a backend written in Rust.</h4>
 


jeanm:
If Xi really is pronounced like _sigh_, then I don't think ⟨ɪ̯⟩ is correct. That would indicate a semi-vowel, and "igh" in _sigh_ isn't one. I would rather transcribe it as /zaɪ/, placed between slashes (cf. [the phonetic transcription of sigh on Wiktionary](https://en.wiktionary.org/wiki/sigh#Pronunciation).

raphlinus:
I agree with @jeanm here. The "inverted breve below" is overly pedantic, and according to [Wikipedia](https://en.wikipedia.org/wiki/Diphthong#Transcription) is really only indicated when there is a contrastive vowel sequence in the language, which there isn't here.

README.md
```diff
@@ -3,7 +3,7 @@
   <a href="https://github.com/xi-editor/xi-editor">Xi Editor</a>
 </h1>
 
-<p align="center"><em>(pronounced "Zigh")</em></p>
+<p align="center"><em>pronounced "zaɪ̯"</em>; rhymes with "sigh", but starts with "z" from "zoo"</p>
```
cmyr:
```suggestion
<p align="center"><em>pronounced</em> /zaɪ/; <em>rhymes with "sigh"</em></p>
```

walston: I dig this solution.

PR #1159, opened 2019-03-31
add save_traces utility function

bzm3r:
## Summary
Add a `save_traces` utility function, which saves trace output to a file which can be opened by Chrome's Trace Viewer. Location of output is determined by environment variable `TRACE_OUTPUT`.

## Related Issues
Currently trying to figure out where this function should go in druid-shell: https://github.com/xi-editor/druid/pull/35

Wondering if perhaps it should be a xi_trace utility?

## Review Checklist

### DIFF
--- a/rust/trace/src/chrome_trace_dump.rs
+++ b/rust/trace/src/chrome_trace_dump.rs
@@ -14,18 +14,18 @@
 
 #![cfg_attr(feature = "benchmarks", feature(test))]
 #![allow(
-clippy::if_same_then_else,
-clippy::needless_bool,
-clippy::needless_pass_by_value,
-clippy::ptr_arg
+    clippy::if_same_then_else,
+    clippy::needless_bool,
+    clippy::needless_pass_by_value,
+    clippy::ptr_arg
 )]
-
 
 #[cfg(all(test, feature = "benchmarks"))]
 extern crate test;
 
 use super::Sample;
 use serde_json;
+use std::fs::File;
 use std::io::{Error as IOError, Read, Write};
 
 #[derive(Debug)]
@@ -80,6 +80,42 @@ where
     R: Read,
 {
     serde_json::from_reader(input).map_err(Error::Json)
+}
+
+/// Save tracing data to path pointed to by the environment variable TRACE_OUTPUT, using the Trace
+/// Viewer format. Save path defaults to `./target/trace_output.trace`. Trace file can be opened
+/// with the Chrome browser by visiting the URL `about:tracing`. If `sorted_chronologically` is true
+/// then sort output traces chronologically by time of creation.
+pub fn save_traces(sorted_chronologically: bool) {
+    use std::env;
+
+    let traces = if sorted_chronologically {
+        super::samples_cloned_sorted()
+    } else {
+        super::samples_cloned_unsorted()
+    };
+
+    let trace_output_path = match env::var("TRACE_OUTPUT") {
+        Ok(output_path) => output_path,
+        Err(_) => {
+            println!("Environment variable TRACE_OUTPUT not set, defaulting to ./target/trace_output.trace");
+            String::from("./target/trace_output.trace")
+        }
+    };
+
+    let mut trace_file = match File::create(&trace_output_path) {
+        Ok(f) => f,
+        Err(_) => {
+            println!("Could not create trace output file at: {}.", &trace_output_path);
+            return;
+        }
+    };
+
+    if serialize(&traces, &mut trace_file).is_err() {
+        println!("Could not save trace file at: {}.", &trace_output_path);
+    } else {
+        println!("Saved trace file at: {}", &trace_output_path);
+    }
 }
 
 #[cfg(test)]
--- a/rust/trace/src/chrome_trace_dump.rs
+++ b/rust/trace/src/chrome_trace_dump.rs
@@ -25,14 +25,37 @@ extern crate test;
 
 use super::Sample;
 use serde_json;
-use std::fs::File;
-use std::io::{Error as IOError, Read, Write};
+use std::io::{Error as IOError, ErrorKind as IOErrorKind, Read, Write};
 
 #[derive(Debug)]
 pub enum Error {
     Io(IOError),
     Json(serde_json::Error),
     DecodingFormat(String),
+}
+
+impl From<IOError> for Error {
+    fn from(e: IOError) -> Error {
+        Error::Io(e)
+    }
+}
+
+impl From<serde_json::Error> for Error {
+    fn from(e: serde_json::Error) -> Error {
+        Error::Json(e)
+    }
+}
+
+impl From<String> for Error {
+    fn from(e: String) -> Error {
+        Error::DecodingFormat(e)
+    }
+}
+
+impl Error {
+    pub fn already_exists() -> Error {
+        Error::Io(IOError::from(IOErrorKind::AlreadyExists))
+    }
 }
 
 #[derive(Clone, Debug, Deserialize)]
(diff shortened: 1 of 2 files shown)


rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +82,42 @@ where
...
+    let traces = if sorted_chronologically {
+        super::samples_cloned_sorted()
+    } else {
+        super::samples_cloned_unsorted()
+    };
+
+    let trace_output_path = match env::var("TRACE_OUTPUT") {
+        Ok(output_path) => output_path,
+        Err(_) => {
+            println!("Environment variable TRACE_OUTPUT not set, defaulting to ./target/trace_output.trace");
```
cmyr: all of these `println` should probably be `eprintln`... don't think it's likely to be a problem in this case, but as xi uses stdout to communicate if this was used there it would cause a crash.

rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +82,42 @@ where
...
+            String::from("./target/trace_output.trace")
```
cmyr: I would discourage a fixed path name, which means you'll always overwrite any existing files. One compromise would be using the PID of the current process?

bzm3r: @cmyr or perhaps something that includes time + date?

cmyr: That's totallly an option; the downside is that it drags in another dependency.

bzm3r: @cmyr How about we let the calling application decide how they want to name their path? I have changed the function so that it now expects a path, and thus doesn't default to anything---thoughts?

cmyr: I think that is the best approach.

rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +83,36 @@ where
...
+/// using the Chrome browser by visiting the URL `about:tracing`. If `sorted_chronologically` is
+/// true then sort output traces chronologically by each trace's time of creation.
+pub fn save_traces(output_path: &Path, sorted_chronologically: bool) -> IOResult<()> {
+    let traces = if sorted_chronologically {
+        super::samples_cloned_sorted()
+    } else {
+        super::samples_cloned_unsorted()
+    };
+
+    match File::create(&output_path) {
```
cmyr: I would return the error directly (`File::create(&output_path)?`) and let the caller handle it.

rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +83,36 @@ where
...
+        Ok(f) => f,
+        Err(err) => {
+            eprintln!("Could not create trace output file at: {}.", &output_path);
+            return IOResult(err);
+        }
+    };
+
+    match serialize(&traces, &mut trace_file) {
```
cmyr: ooh, I didn't realize that this has a different result type. Luckily we actually already have a custom Error type in this crate, so we should be able to just use that as our return type. In this case as well we can let the caller handle the error, and we don't need to do any printing here.

rust/trace/src/chrome_trace_dump.rs
cmyr: Additionally, we should at least think about what the correct behaviour is in the case where the path exists.

rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +83,36 @@ where
     serde_json::from_reader(input).map_err(Error::Json)
 }
 
+/// Save tracing data to to supplied path, using the Trace Viewer format. Trace file can be opened
+/// using the Chrome browser by visiting the URL `about:tracing`. If `sorted_chronologically` is
+/// true then sort output traces chronologically by each trace's time of creation.
+pub fn save_traces(output_path: &Path, sorted_chronologically: bool) -> IOResult<()> {
```
cmyr: I think the correct type for `output_path` is `<P: AsRef<Path>>`. This has the advantage of working with `&Path`, `&PathBuf`, `PathBuf`, `String`, `str`, `OsStr`, etc.

rust/trace/src/chrome_trace_dump.rs
bzm3r: @cmyr I gave some thought to what correct behaviour should be in the case where the path (file) already exists:

- we could overwrite the file: this is dangerous, since an important file could be overwritten.
- we could rename the file name component of the path: in the end, in order to guarantee that we come up with a unique file name, we cannot rely on any methods that use randomness at some level to generate a file name (as there would always exist some possibility of failure). We must instead list every file name in the directory containing the file, come up with a file name (possibly through some random means), and ensure it is not in the already tabulated list of existing file.
- we could do nothing, and abort (return an error): this is robust, in the sense that `save_traces` will definitely not screw up in any important way, and it is up to the caller to ensure they are providing okay file names.

I have chosen to go with the last option, for now. Let me know what you think.

bzm3r:
@cmyr I think this process has been enlightening, because the save function we are left with is so barebones, after deciding what to relegate to the the caller, and what should be done by xi-trace, that I even wonder if the function is worth adding. Also, I am not quite sure how to elegantly deal with what Error type to return.

rust/trace/src/chrome_trace_dump.rs
```diff
@@ -82,6 +83,40 @@ where
...
+            if md.is_file() {
+                return Err(Error::Io(IOError::from(IOErrorKind::AlreadyExists)));
+            }
+        }
+        Err(err) => {
+            return Err(Error::Io(err));
+        }
+    }
+
+    let mut trace_file = match File::create(&output_path) {
```
cmyr: We should be able to write all (most) of these as, `let thing = Failable::call()?;` and the errors will be correctly propagated up.

cmyr:
Yea, finding the dividing line between a client and a library's responsibilities can be tricky. That said, small improvements to ergonomics are worth doing, and if this makes it easier to dump a trace to disk I think that makes sense.

In the interests of treating this as an exercise: one thing we might consider is more closely mimicking the api used elsewhere, where we have methods in an `impl` of the `Trace` type, and then we expose global functions that call these methods on the shared global instance?

So basically we could have `Trace::save(&self, path, sorted) -> Result<()>` and then we just expose that from a global function.

But that's just an idea.


In any case I think we can clean up this function a bit more now by using `?` to handle the error propagation, but otherwise I'm happy!

bzm3r:
@cmyr Do have a look when you have time, I hope to have addressed your most recent review comments.

rust/trace/src/lib.rs
```diff
@@ -819,6 +821,22 @@ impl Trace {
...
     }
+
+    pub fn save<P: AsRef<Path>>(
+        &self,
+        path: P,
+        sort: bool,
+    ) -> Result<(), chrome_trace_dump::Error> {
+        let traces = if sort { samples_cloned_sorted() } else { samples_cloned_unsorted() };
+
+        if fs::metadata(&path)?.is_file() {
```
cmyr: what if it's a directory?

In anycase I would just suggest `if path.exists() { .. }`.

Also, in these fns where we take `AsRef<Path>`, the first thing I normally do is `let path = path.into()` so that we are working concretely with a `&Path`, and can use `Path` fns.

rust/trace/src/lib.rs
```diff
@@ -819,6 +821,22 @@ impl Trace {
         samples.sort_unstable();
         samples
     }
+
+    pub fn save<P: AsRef<Path>>(
```
cmyr: I'd make this `pub(crate)` since we don't expect it to be used externally. One important reaeson to avoid `pub` when not necessary is that it makes it a semvar breaking change if we want to modify the api in the future.

rust/trace/src/lib.rs
```diff
@@ -819,6 +821,23 @@ impl Trace {
         samples.sort_unstable();
         samples
     }
+
+    pub(crate) fn save<P: AsRef<Path>>(
```
cmyr: I think this actually probably _is_ public API now? sorry 😬

bzm3r: no problem! is done :)

PR #1160, opened 2019-03-31
fix warnings in core-lib/src/files.rs

bzm3r:
## Summary
Fix the following warnings in core-lib/src/files.rs:

```
warning: unused import: `Permissions`
  --> core-lib\src\file.rs:20:27
   |
20 | use std::fs::{self, File, Permissions};
   |                           ^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

warning: unused import: `Permissions`
  --> core-lib\src\file.rs:20:27
   |
20 | use std::fs::{self, File, Permissions};
   |                           ^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

warning: unused variable: `info_unwrapped`
   --> core-lib\src\file.rs:229:17
    |
229 |     if let Some(info_unwrapped) = file_info {
    |                 ^^^^^^^^^^^^^^ help: consider using `_info_unwrapped` instead
    |
    = note: #[warn(unused_variables)] on by default

warning: unused variable: `info_unwrapped`
   --> core-lib\src\file.rs:229:17
    |
229 |     if let Some(info_unwrapped) = file_info {
    |                 ^^^^^^^^^^^^^^ help: consider using `_info_unwrapped` instead
    |
    = note: #[warn(unused_variables)] on by default
```

## Related Issues
None applicable.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -17,7 +17,7 @@
 use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
-use std::fs::{self, File, Permissions};
+use std::fs::{self, File};
 use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use std::str;
@@ -33,7 +33,7 @@ use crate::tabs::OPEN_FILE_EVENT_TOKEN;
 #[cfg(feature = "notify")]
 use crate::watcher::FileWatcher;
 #[cfg(target_family = "unix")]
-use std::os::unix::fs::PermissionsExt;
+use std::{fs::Permissions, os::unix::fs::PermissionsExt};
 
 const UTF8_BOM: &str = "\u{feff}";
 
@@ -198,6 +198,7 @@ where
     Ok((rope, info))
 }
 
+#[allow(unused)]
 fn try_save(
     path: &Path,
     text: &Rope,
@@ -226,12 +227,11 @@ fn try_save(
 
     fs::rename(tmp_path, path)?;
 
-    if let Some(info_unwrapped) = file_info {
-        #[cfg(target_family = "unix")]
-        fs::set_permissions(
-            path,
-            Permissions::from_mode(info_unwrapped.permissions.unwrap_or(0o644)),
-        )?;
+    #[cfg(target_family = "unix")]
+    {
+        if let Some(info) = file_info {
+            fs::set_permissions(path, Permissions::from_mode(info.permissions.unwrap_or(0o644)))?;
+        }
     }
 
     Ok(())


rust/core-lib/src/file.rs
```diff
@@ -226,12 +226,9 @@ fn try_save(
 
     fs::rename(tmp_path, path)?;
 
-    if let Some(info_unwrapped) = file_info {
+    if let Some(_info) = file_info {
```
Cogitri: That seems kinda hacky, it'd be best to only do the entire if clause on Unix.

cmyr: Agree, the solution should be to use the correct `#cfg[]` marker to skip the unused code on windows.

bzm3r: @Cogitri @cmyr but `#cfg[]` tags don't work on if statements?

cmyr: @bzm3r than we can wrap the whole `if` in a block?

bzm3r: @cmyr 🤦‍♂️ i'll do that!

raphlinus: Yeah, this isn't right - the underscore prefix should only be used when the variable is actually not used, which is clearly not the case here. Probably easier is just to set `#[allow(unused)]`.

rust/core-lib/src/file.rs
```diff
@@ -226,12 +227,11 @@ fn try_save(
 
     fs::rename(tmp_path, path)?;
 
-    if let Some(info_unwrapped) = file_info {
-        #[cfg(target_family = "unix")]
-        fs::set_permissions(
-            path,
-            Permissions::from_mode(info_unwrapped.permissions.unwrap_or(0o644)),
-        )?;
+    #[cfg(target_family = "unix")]
```
cmyr: is this necessary if we have the `#[allow(unused)]` annotation?

bzm3r: I believe so, because [`from_mode`](https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html#tymethod.from_mode) is a function which only exists for unix systems, and the field [`permissions` on `FileInfo`](https://github.com/xi-editor/xi-editor/blob/533a66aa91b019fdb69b7b4c3a90fc38c9c5f59a/rust/core-lib/src/file.rs#L49) only exists on unix systems too? see:

cmyr: okay, just checking. Basically: it's nice to have small patches, and if the previous version of the code now doesn't produce warnings that's great, but if this is required that's good too!

cmyr:
Annoyingly this is failing for other reasons on windows (https://travis-ci.com/xi-editor/xi-editor/jobs/190450527) but if you're happy @bzm3r then #yolo

bzm3r:
Yeah, I think I'm happy. That's just a nightly thing, right?

PR #1171, opened 2019-04-06
update leaves individually in delete_intersecting

### LINKED ISSUE #1167: delete_intersecting leaf-by-leaf
`delete_intersecting` in `Spans` currently iterates through _all_ spans every time it is called. Another option would be to go leaf-by-leaf, and check each leaf for whether or not it has any items in the interval; if they don't we keep them unchanged, otherwise we do this operation, but only within the leaf.

Related to https://github.com/xi-editor/xi-editor/pull/1162

ahouts:
## Summary
Instead of creating a completely new Spans in delete_intersecting, only update the parts that need to be updated.

## Related Issues
closes #1167 

## Review Checklist
- [X] I have tested the code with `cargo test --all` / `./rust/run_all_checks`.
- [X] I have updated comments / documentation related to the changes I made.

### DIFF
--- a/rust/rope/benches/spans.rs
+++ b/rust/rope/benches/spans.rs
@@ -1,0 +1,134 @@
+// Copyright 2019 The xi-editor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+#![feature(test)]
+
+extern crate test;
+extern crate xi_rope;
+
+use test::{black_box, Bencher};
+use xi_rope::interval::Interval;
+use xi_rope::spans::{Spans, SpansBuilder};
+
+// generate num_spans spans with a seeded random composition, values from 0-134456
+// for a given num_spans, it always generates the same spans
+fn gen_spans(num_spans: usize) -> Spans<usize> {
+    fn rng_gen() -> impl Fn(&mut usize) -> (Interval, usize) {
+        let pseudo_random = |seed: &mut usize| -> usize {
+            *seed = (8_121 * *seed + 28_411) % 134_456;
+            *seed
+        };
+
+        let gen_range = move |seed: &mut usize| -> (Interval, usize) {
+            let one = pseudo_random(seed);
+            let two = pseudo_random(seed);
+            let start = one.min(two);
+            let end = one.max(two);
+            (Interval::new(start, end), pseudo_random(seed))
+        };
+
+        gen_range
+    }
+
+    let mut seed = 123_456;
+    let gen_range = rng_gen();
+
+    let mut sb = SpansBuilder::new(134_456);
+    let mut to_add: Vec<_> = (0..num_spans).map(|_| gen_range(&mut seed)).collect();
+    to_add.sort_by(|p0, p1| p0.0.start().cmp(&p1.0.start()));
+    for (iv, data) in to_add {
+        sb.add_span(iv, data);
+    }
+    sb.build()
+}
+
+#[bench]
+fn test_delete_intersecting_100_tiny_range(b: &mut Bencher) {
+    let spans = gen_spans(100);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(10_000, 10_020)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_100_med_range(b: &mut Bencher) {
+    let spans = gen_spans(100);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(100, 10_000)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_100_huge_range(b: &mut Bencher) {
+    let spans = gen_spans(100);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(100, 134_456)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_10_000_tiny_range(b: &mut Bencher) {
+    let spans = gen_spans(10_000);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(10_000, 10_020)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_10_000_med_range(b: &mut Bencher) {
+    let spans = gen_spans(10_000);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(100, 10_000)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_10_000_huge_range(b: &mut Bencher) {
+    let spans = gen_spans(10_000);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(100, 134_456)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_1_000_000_tiny_range(b: &mut Bencher) {
+    let spans = gen_spans(1_000_000);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(10_000, 10_020)))
+    });
+}
+
+#[bench]
+fn test_delete_intersecting_1_000_000_med_range(b: &mut Bencher) {
+    let spans = gen_spans(1_000_000);
+    b.iter(|| {
+        let mut spans_copy = spans.clone();
+        black_box(spans_copy.delete_intersecting(Interval::new(100, 10_000)))
+    });
+}
+
+#[bench]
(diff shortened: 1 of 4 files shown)


cmyr:
@ahouts Cool, thanks for this! If possible, could you give a quick overview of your thinking & approach here? Will make it easier to do a reasonable review if I have a better high-level sense of how you're considering the problem.

It would also be cool to have some concrete benchmarks available to compare these two implementations?

ahouts:
Sure thing, while I was implementing the logic of walking the tree and modifying leaves I noticed that you can only really do one of three things to any given leaf: change it, remove it, or make no change. I also noticed that the logic of building a new tree based on these modifications was not dependent on what you are doing on each leaf.
As a result, I made the tree walking and construction a function of Node. It accepts a closure as an argument that performs whatever action to every leaf, returning what should be done with each leaf in the new tree.
Then in order to implement span removal, we just have to create a new leaf with the spans that we did not remove. Edge cases are we removed no spans so we return NoChange, or we removed every span so we return Remove.
One nice thing about this approach is it should only make changes if the mutate_leaf function makes changes, and only the dependent nodes will be changed. It will also flatten the tree as Nodes are removed.

ahouts:
I can't do benchmarking right now, but maybe I'll have time later today.

ahouts:
I had to make quite a few changes, as I didn't know there is information associated with the position of a Node in the tree. For example, the actual interval of a leaf is given by summing all previous leaves lengths together.
As a quick fix, I removed the `Remove` variant of `MutateResult` so the structure of the tree (and all implied information) will be preserved.
I also had to do a little refactoring so the closure given to `mutate_leaves` also takes the offset as an argument.

After all of that though, the benchmarks look promising. The "range" in the function name refers to the magnitude of the interval that is being deleted.
Original implementation
```
test test_delete_intersecting_10_000_huge_range ... bench:      23,764 ns/iter (+/- 175)
test test_delete_intersecting_10_000_med_range  ... bench:     149,115 ns/iter (+/- 17,102)
test test_delete_intersecting_10_000_tiny_range ... bench:     164,632 ns/iter (+/- 5,320)
```

New Implementation
```
test test_delete_intersecting_10_000_huge_range ... bench:      27,607 ns/iter (+/- 18,746)
test test_delete_intersecting_10_000_med_range  ... bench:      14,611 ns/iter (+/- 136)
test test_delete_intersecting_10_000_tiny_range ... bench:      18,331 ns/iter (+/- 91)
```

rust/rope/src/spans.rs
```diff
@@ -331,9 +349,10 @@ impl<'a, T: Clone> Iterator for SpanIter<'a, T> {
     type Item = (Interval, &'a T);
 
     fn next(&mut self) -> Option<(Interval, &'a T)> {
-        if let Some((leaf, start_pos)) = self.cursor.get_leaf() {
+        while let Some((leaf, start_pos)) = self.cursor.get_leaf() {
```
ahouts: The behavior was to bail on the first empty `leaf.spans`. Since we allow removing spans, it is possible for a leaf to have no spans but for spans to follow.

cmyr:
Cool, those numbers are certainly encouraging!

ahouts:
> I worry that this implementation is a bit _fancy_. That is, it's adding a bunch of code to `tree.rs` that I'm not sure is going to be used again, and some of which even if we do decide to pursue this pattern we might want to skip.
Fair enough, I had assumed that rebuilding all nodes except leaves every modification would be too expensive. In retrospect, there are relatively few such nodes compared to the number of leaves.
I also wasn't certain about adding a new method to Tree, but it does make sense to add that functionality there.

I'll take a look into it, that code certainly looks much simpler.

rust/rope/benches/spans.rs
```diff
@@ -0,0 +1,291 @@
+// Copyright 2018 The xi-editor Authors.
```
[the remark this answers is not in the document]
ahouts: Copy paste 🤦‍♂️

rust/rope/benches/spans.rs
```diff
@@ -0,0 +1,291 @@
...
+extern crate test;
+extern crate xi_rope;
+
+use test::{black_box, Bencher};
+use xi_rope::interval::Interval;
+use xi_rope::spans::SpansBuilder;
+
+#[bench]
+fn test_delete_intersecting_100_tiny_range(b: &mut Bencher) {
+    let mut seed = 123_456;
```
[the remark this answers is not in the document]
ahouts: The benchmarks were just something I hacked together in-between classes, definitely intended to clean that up before merging anything.

rust/rope/src/spans.rs
```diff
@@ -301,21 +301,30 @@ impl<T: Clone> Spans<T> {
...
+            }
+
+            // stop looking at leaves if we're past the end of the deletion interval
+            end_offset += node.len();
+            if end_offset > interval.end() {
+                break;
             }
         }
-        *self = builder.build();
+        b.push(self.subseq(interval.end..));
```
ahouts: In testing the subseq function doesn't seem to work as intended.
In the `delete_intersecting_empty` test `self.subseq(interval.end..)` returns a leaf with one span, iv=(2,3)

rust/rope/benches/spans.rs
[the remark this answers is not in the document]
ahouts: Oh yeah, I remembered the reason I did it this way. It has random looking results, but it gives the same values every time. This makes the tests reproducible, so checking the number of spans should give the same result as the previous implementation.

ahouts:
Benchmarks on this version are quite a bit better, so it looks like you had the right idea.
```
test test_delete_intersecting_100_huge_range       ... bench:         441 ns/iter (+/- 573)
test test_delete_intersecting_100_med_range        ... bench:         739 ns/iter (+/- 3)
test test_delete_intersecting_100_tiny_range       ... bench:         660 ns/iter (+/- 1)
test test_delete_intersecting_10_000_huge_range    ... bench:      28,893 ns/iter (+/- 12,396)
test test_delete_intersecting_10_000_med_range     ... bench:       6,387 ns/iter (+/- 103)
test test_delete_intersecting_10_000_tiny_range    ... bench:       9,756 ns/iter (+/- 35)
test test_delete_intersecting_1_000_000_huge_range ... bench:   3,629,148 ns/iter (+/- 28,344)
test test_delete_intersecting_1_000_000_med_range  ... bench:     436,883 ns/iter (+/- 8,711)
test test_delete_intersecting_1_000_000_tiny_range ... bench:   1,985,280 ns/iter (+/- 30,374)
```

`LeafIterator` returns `(Node, &Leaf)` because the `Node.get_leaf` method is private. If it was not private, `LeafIterator` could return just a `Node`, and be a real iterator.

ahouts:
One thing to note, all tests (including one with 10,000 spans) pass, but the test with 1,000,000 spans is off by 2. I don't have time to track it down right now though.

rust/rope/src/spans.rs
```diff
@@ -351,8 +364,40 @@ impl<'a, T: Clone> Iterator for SpanIter<'a, T> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    #[test]
 
+    // generate num_spans spans with a pseudo-random composition, values from 0-134456
+    // for a given num_spans it will always generate the same spans, making tests repeatable
+    fn gen_spans(num_spans: usize) -> Node<SpansInfo<usize>> {
```
cmyr: so thinking about this more:

With benchmarks, it's important that we're fully reproducible, but for testing there really is a benefit to randomness, although it is still important to reproduce failures.

I'm particularly interested in some randomness here because of known issues with `Spans`.

With this in mind I'd encourage using `rand` for these cases, but printing out the seed before each test; if we have a failure we can reproduce from the seed.

rust/rope/src/spans.rs
ahouts: `TreeBuilder::push_leaf` has no visibility into leaves other than what their length is. Since the length is the same as before, I believe it will still add leaves with no spans. We have to leave it this way unless we want to add the length to the previous leaf & omit it.

rust/rope/src/spans.rs
ahouts: Good idea, I'll use the `rand_pcg` crate for that.

rust/rope/src/tree.rs
```diff
@@ -421,6 +421,63 @@ impl<N: NodeInfo> Node<N> {
...
         m2 + M2::from_base_units(l, base)
     }
+
+    pub fn iter_leaf_nodes(&self) -> LeafIterator<N> {
+        LeafIterator::new(self)
+    }
+}
+
+/// iterator over all leaves in a Node
+pub struct LeafIterator<N: NodeInfo> {
```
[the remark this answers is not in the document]
ahouts: Good call

ahouts:
@cmyr I made the changes you suggested. I had to run `cargo update` to get `rand` to compile.

cmyr:
Okay cool. I've opened a PR that does a general dependency update, in #1174; wanted to figure out why we had three different versions of `rand` in the graph. Hopefully that will be cleaned up now, and we can rebase this onto that when it lands, which will be whenever it passes CI.

ahouts:
@cmyr Rebased onto master. What should we do about the test that fails?

cmyr:
I haven’t looked into it, tbh, but we should figure out what’s going on.

rust/rope/src/spans.rs
```diff
@@ -301,21 +301,33 @@ impl<T: Clone> Spans<T> {
...
-            // check if spans overlaps with interval
-            if iv.intersect(interval).is_empty() {
-                // keep the ones that are not overlapping
-                builder.add_span(iv, data.clone());
+        let mut b = TreeBuilder::new();
+        // the end of the last leaf seen
+        let mut end_offset = 0;
+        let mut leaf_iterator = self.iter_leaf_nodes();
+        while let Some((node, leaf)) = leaf_iterator.next() {
+            let SpansLeaf { len, spans } = leaf;
```
Cogitri: CI chokes on this

ahouts: Thanks! However this PR will not pass CI for now, there is a logical bug somewhere preventing this from working on extremely large sizes. IDK if it is a problem with spans or this particular feature.

PR #1175, opened 2019-04-14
Add the Vixi project to the list of frontends

Peltoche:
Hi!
Everything is in the title. I have just started a new side-project which consist into a  "Vim like" frontend from the xi-editor. I was hoping to add it to the frontend list inside the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ Here are some front-ends in various stages of development:
 
 * [xi-qt](https://github.com/sw5cc/xi-qt), a Qt front-end.
 
+* [vixi](https://github.com/Peltoche/vixi), a Vim like front-end in Rust.
+
 The following are currently inactive, based on earlier versions of the front-end
 protocol, but perhaps could be revitalized:
 


jansol:
Isn't there some other project with the same name and roughly the same goals already?

Peltoche:
Damn! I never had the idea to check! This could be problematic.

There is [jakalope/vixi](https://github.com/jakalope/vixi) but it doesn't seems active.

jansol:
Maybe you can join forces with @jakalope if they are still interested... Their code is MIT licensed so you could at least check if there's something you can reuse.

Anyway personally I don't have a strong opinion on which projects should be in the readme. If you can collaborate on this that'll just make everyone's lives easier IMHO.

Cogitri:
https://github.com/xi-editor/xi-editor.github.io/blob/master/frontends.md should most likely be changed as well.

Cogitri:
Sooo...what's the status on this?

cmyr:
oops. I'm happy to merge.

cmyr:
jakalope was there first but also hasn't been working on any of this stuff in ages, so let's prioritize the more active project.

PR #1179, opened 2019-04-20
fix prev line iteration

### LINKED ISSUE #1178: [xi-rope][bug] prev::<LinesMetric> does not work
It looks like `prev` is broken.  It moves cursor back once and that' s it.

```rust
    #[test]
    fn prev_line() {
        let rope = Rope::from_str("First line\nSecond line\r\nThird line\n").unwrap();
        let mut cursor = Cursor::new(&rope, 34);

        cursor.prev::<LinesMetric>();
        cursor.prev::<LinesMetric>();
        cursor.prev::<LinesMetric>();
        cursor.prev::<LinesMetric>();

        assert_eq!(cursor.pos(), 24);
}

```

ngortheone:
Fix #1178 

- [X] I have tested the code with `cargo test --all`


.gitignore
```diff
@@ -5,3 +5,4 @@ build/
 *.pyc
 __pycache__
 .cache/
+.idea/
```
cmyr: oops?

rust/rope/src/rope.rs
```diff
@@ -260,7 +260,7 @@ impl Metric<RopeInfo> for LinesMetric {
     }
 
     fn prev(s: &String, offset: usize) -> Option<usize> {
-        memrchr(b'\n', &s.as_bytes()[..offset]).map(|pos| pos + 1)
+        memrchr(b'\n', &s.as_bytes()[..offset - 1]).map(|pos| pos + 1)
```
cmyr: It might be worth adding something like, `debug_assert!( offset > 0, "caller is responsible for validating input")`, as a way of documenting our invariants?

.gitignore
ngortheone: will fix

PR #1182, opened 2019-04-23
[xi-rope] make compare_cursor_regex public

ngortheone:
I actively use xi-rope in my project https://github.com/org-rs/org-rs
and I need `compare_cursor_regex` function. It seems that it was forgotten to be made public as well as 
`compare_cursor_str_casei`

I used some of my examples as a cases for tests

- [X] I have tested the code with `cargo test --all`


cmyr:
Okay, this seems reasonable.

PR #1190, opened 2019-05-08
Apply clippy renames

Cogitri:
This fixes building with `cargo clippy -- -D warning`

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -7,7 +7,7 @@ task:
       matrix:
         - name: test (linux)
           container:
-            image: rust:1.31.0
+            image: rust:latest
             cpu: 4
             memory: 8G
           rustfmt_script: rustup component add rustfmt-preview && cd rust && cargo fmt --all -- --check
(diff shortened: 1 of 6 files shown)


cmyr:
Were you running this against 1.31? It's certainly reasonable for us to bump rustc, but I suspect this is why CI is failing.

Cogitri:
I think the CI failure is unrelated to this? Anyway, I tested against Rust 1.34 (CI) and 1.35 (beta. Only thing available for my platform (x86_64-unknown-linux-musl) right now).

cmyr:
Yea, CI runs against Rust 1.31, currently. We keep ourselves pinned a few versions behind stable, although this isn't a firm rule (It was more important when we were regularly keeping fuchsia up to date, but that isn't the case at the moment and fuchsia is still on nightly anyway, afaik)

Cogitri:
Hey @cmyr could you maybe look at the clippy failures? Clippy isn't yet available for my platform (it will be with the 1.35 release though, which should land soon-ish)

cmyr:
The only failure seems to be this:

```
  --> rpc/src/test_utils.rs:70:60
   |
70 |         let resp = raw.map_err(|e| e.to_string()).and_then(|r| r.into_response());
   |                                                            ^^^^^^^^^^^^^^^^^^^^^ help: remove closure as shown: `parse::RpcObject::into_response`
   |
```

Want to try that suggestion and see if that gets us passing?

Cogitri:
Sure, but I think it stops once it encounters a file with errors.

Cogitri:
This works on my machine now.

cmyr:
@Cogitri would you like to update this as per the zulip discission?

Cogitri:
Ah, right, will do

Cogitri:
Let's just use stable for now and (maybe) add 1.35.0 after 1.36.0 has been released

PR #1200, opened 2019-06-22
rope: optionalize serde dep

### LINKED ISSUE #1197: Make serialization/deserialization optional in xi-rope
As part of making xi more modular,  the `serde` dependency in xi-rope should become optional.

This requires adding a new feature to the `Cargo.toml` in xi-rope, and putting all the serde imports and definitions behind `#[cfg(feature = "serde")]` guards.

Cogitri:
fixes #1197

### DIFF
--- a/rust/rope/Cargo.toml
+++ b/rust/rope/Cargo.toml
@@ -10,11 +10,13 @@ edition = '2018'
 [dependencies]
 bytecount = "0.5"
 memchr = "2.0"
-serde = "1.0"
-serde_derive = "1.0"
+serde = { version="1.0", optional=true, features=["derive"] }
 unicode-segmentation = "1.2.1"
 regex = "1.0"
 
 [dev-dependencies]
 serde_test = "^1.0"
 serde_json = "1.0"
+
+[features]
+default = []
--- a/rust/rope/src/lib.rs
+++ b/rust/rope/src/lib.rs
@@ -27,10 +27,12 @@
 extern crate bytecount;
 extern crate memchr;
 extern crate regex;
-extern crate serde;
 extern crate unicode_segmentation;
+
+#[cfg(feature = "serde")]
 #[macro_use]
-extern crate serde_derive;
+extern crate serde;
+
 #[cfg(test)]
 extern crate serde_json;
 #[cfg(test)]
@@ -45,6 +47,7 @@ pub mod find;
 pub mod interval;
 pub mod multiset;
 pub mod rope;
+#[cfg(feature = "serde")]
 mod serde_impls;
 pub mod spans;
 #[cfg(test)]
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 xcuserdata/
 .DS_Store
 build/
+.idea
 
 *.pyc
 __pycache__
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -932,6 +932,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "serde"
 version = "1.0.94"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "serde_derive 1.0.94 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
 [[package]]
 name = "serde_derive"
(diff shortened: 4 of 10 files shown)


rust/rope/Cargo.toml
```diff
@@ -10,11 +10,15 @@ edition = '2018'
...
+serde_derive = { version="1.0", optional=true }
 unicode-segmentation = "1.2.1"
 regex = "1.0"
 
 [dev-dependencies]
 serde_test = "^1.0"
 serde_json = "1.0"
+
+[features]
+default = ["dese"]
```
Cogitri: We can't call the feature `serde`, Cargo complains that we can have a feature whose name is the same as one of our deps.

rust/rope/Cargo.toml
```diff
@@ -10,11 +10,15 @@ edition = '2018'
...
-serde = "1.0"
-serde_derive = "1.0"
+serde = { version="1.0", optional=true }
+serde_derive = { version="1.0", optional=true }
 unicode-segmentation = "1.2.1"
 regex = "1.0"
 
 [dev-dependencies]
 serde_test = "^1.0"
 serde_json = "1.0"
```
Cogitri: We can't make test deps optional, but I guess that's not too much of a problem.

rust/rope/src/lib.rs
```diff
@@ -27,10 +27,14 @@
 extern crate bytecount;
 extern crate memchr;
 extern crate regex;
-extern crate serde;
 extern crate unicode_segmentation;
+
+#[cfg(feature = "dese")]
+extern crate serde;
 #[macro_use]
+#[cfg(feature = "dese")]
```
cmyr: hmmm I think maybe we want this above `#[macro_use]`?

rust/rope/Cargo.toml
cmyr: right. Bikeshed, but maybe we call our feature `serde_impls`? Or otherwise something else where the name clearly describes the intent.

raphlinus: I would argue against this feature being default, and require its users to opt in explicitly. Otherwise you have to do no-default-features, which is not pleasant.

rust/rope/src/lib.rs
Cogitri: Sure

rust/rope/Cargo.toml
Cogitri: Hm, I personally prefer featues being enabled by default, but sure, I can make it disabled by default.

cmyr: Yea I can see both ways. The case against is that the serde stuff hurts compile times and code size and is only needed in use cases like `xi`, so we should give you the faster, smaller version by default. The case *for* is that we want to give you everything up front, and if you realize you don't need serde you can disable. I'm happy to defer to raph.

raphlinus: Also, I think this can be named "serde" by doing this:

    serde = { version="1.0", optional=true, features=["derive"] }
    serde_derive = { version="1.0", optional=true }

And not having a separate features section at all

Cogitri: Oh, that does work, thanks!

Cogitri:
I guess I'll have to make serde for tests optional (otherwise we'd need tests to always run with `--all-features`)

Cogitri:
Not quite sure if this CI failure is on this change or xi-trace being somewhat broken

cmyr:
Ah, this is something to do with xi-trace not liking `cargo test --all-features`.

Cogitri:
Should I revert this then?

cmyr:
I see two options: either dig into #1201 and get `xi_trace` to compile with `--all-features`, or set CI in this plugin to specifically turn on serde for testing.

The second of these might not be that bad; instead of `#[cfg(feature = serde_impls)]`, you might just need `#[cfg(any(feature = "serde_impls", feature = "test"))]`.

Cogitri:
Alright, I guwss I'll do the former in a few days, I won't have time for this tomorrow.

Cogitri:
Alright, I suppose this is good to go, mind taking a look, @cmyr ? :)

Cogitri:
:shipit:

PR #1205, opened 2019-06-26
Fix #1201

### LINKED ISSUE #1201: remove dict_payload feature from xi-trace
This feature has never been used, and causes compilation problems.

In the process of making this change, we will want to pay special attention to the documentation comments, and ensure that we also remove references to this feature.

Cogitri:


Cogitri:
Alright, I guess this should be OK, @cmyr ?

Issue #375, opened 2017-07-20
Getting Error on running cargo build command in UBUNTU


shubham1109:
Error is shown like this...

error: `pub(restricted)` syntax is experimental (see issue #32409)
   --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/zlib.rs:202:9
    |
202 |     pub(crate) fn read_from<R>(mut reader: R) -> io::Result<Self>
    |         ^^^^^

error: `pub(restricted)` syntax is experimental (see issue #32409)
  --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/gzip.rs:98:5
   |
98 | pub(crate) struct Trailer {
   |     ^^^^^

error: `pub(restricted)` syntax is experimental (see issue #32409)
   --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/gzip.rs:370:9
    |
370 |     pub(crate) fn read_from<R>(mut reader: R) -> io::Result<Self>
    |         ^^^^^

error: `pub(restricted)` syntax is experimental (see issue #32409)
  --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/deflate/mod.rs:29:5
   |
29 | pub(crate) mod symbol;
   |     ^^^^^

error: `pub(restricted)` syntax is experimental (see issue #32409)
  --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/non_blocking/deflate/decode.rs:72:9
   |
72 |     pub(crate) fn bit_reader_mut(&mut self) -> &mut TransactionalBitReader<R> {
   |         ^^^^^

error: struct field shorthands are unstable (see issue #37340)
   --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/non_blocking/deflate/decode.rs:117:60
    |
117 |                     DecoderState::ReadNonCompressedBlock { len }
    |                                                            ^^^

error: struct field shorthands are unstable (see issue #37340)
  --> /home/shubham/.cargo/registry/src/github.com-1ecc6299db9ec823/libflate-0.1.9/src/non_blocking/transaction.rs:15:34
   |
15 |         TransactionalBitReader { inner, savepoint }
   |                                  ^^^^^

[...]

PR #1209, opened 2019-07-01
Copy example preferences file when none found

nangtrongvuon:
Previously, xi did not copy the example preferences.xiconfig file when none was found, and simply created an empty file instead. This commit aims to fix that by creating a config file with the contents of preferences.xiconfig to the config folder if one exists.

## Related Issues
Closes [xi-mac #375](https://github.com/xi-editor/xi-mac/issues/375)

## Review Checklist

### DIFF
--- a/rust/core-lib/src/config.rs
+++ b/rust/core-lib/src/config.rs
@@ -16,7 +16,7 @@ use std::collections::HashMap;
 use std::error::Error;
 use std::fmt;
 use std::fs;
-use std::io::{self, Read};
+use std::io::{self, Read, Write};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
@@ -253,6 +253,21 @@ impl ConfigManager {
             extras_dir,
         }
     }
+
+    /// Copies the contents of `client_example.toml` to `preferences.xiconfig`. 
+    /// Used when an existing config file is not found.
+    pub fn create_example_preference_config(&self) -> Result<Option<PathBuf>, ConfigError> {
+        let client_example_config = include_str!("../assets/client_example.toml");
+        
+        if let Some(config_dir) = self.config_dir.as_ref() {
+            let config_file = config_dir.join("preferences.xiconfig");
+            let mut file = fs::File::create(config_file.as_path())?;
+            file.write_all(client_example_config.as_bytes())?;
+            Ok(Some(config_file))    
+        } else {
+            Ok(None)
+        }
+    } 
 
     /// The path of the user's config file, if present.
     pub(crate) fn base_config_file_path(&self) -> Option<PathBuf> {
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -202,8 +202,20 @@ impl CoreState {
     pub(crate) fn finish_setup(&mut self, self_ref: WeakXiCore) {
         self.self_ref = Some(self_ref);
 
+        // Loads existing config file if present, else copies the base example config.
         if let Some(path) = self.config_manager.base_config_file_path() {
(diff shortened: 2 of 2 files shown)


rust/core-lib/src/tabs.rs
```diff
@@ -235,7 +247,10 @@ impl CoreState {
         let _t = trace_block("CoreState::load_config_file", &["core"]);
         if let Some(domain) = self.config_manager.domain_for_path(path) {
             match config::try_load_from_file(&path) {
-                Ok(table) => self.set_config(domain, table),
+                Ok(table) => {
+                    self.set_config(domain, table);
+                    eprintln!("setting new config");
```
Cogitri: I don't think this is supposed to be here?

rust/core-lib/src/tabs.rs
```diff
@@ -202,8 +202,20 @@ impl CoreState {
...
+        } else {
+            // TODO: loads the client_example.toml here
+            eprintln!("no config present, trying to create an example config");
+            match self.config_manager.create_example_preference_config() {
+                Ok(config_path) => {
+                    if let Some(config_path) = config_path {
+                        self.load_file_based_config(&config_path);
+                    } 
+                },
+                Err(e) => eprintln!("Encountered error {:?} while creating an example config", e)
```
Cogitri: I think errors implement `Display`, no?

rust/core-lib/src/config.rs
```diff
@@ -254,6 +254,21 @@ impl ConfigManager {
...
 
+    /// Copies the contents of `client_example.toml` to `preferences.xiconfig`. 
+    /// Used when an existing config file is not found.
+    pub fn create_example_preference_config(&self) -> Result<Option<PathBuf>, ConfigError> {
+        let client_example_config = include_str!("../assets/client_example.toml");
+        
+        if let Some(config_dir) = self.config_dir.as_ref() {
+            let config_file = config_dir.join("preferences.xiconfig");
+            let mut file = fs::File::create(config_file.as_path())?;
+            file.write_all(client_example_config.as_bytes())?;
```
Cogitri: I'd _very_ much prefer "atomically" (well, -ish) writing the file here by writing it to a temp dir first and then moving it to the actual location

Cogitri: See https://github.com/xi-editor/xi-editor/blob/4107776ddc82340eadb30b53ccbcb3435fd7ba61/rust/core-lib/src/file.rs#L202

Cogitri:
Also, rustfmt doesn't seem to be happy with the changes :/

cmyr:
My opinion here is that copying over the template is the responsibility of the frontend. The design of the preferences system is such that there isn't always a preferences file; you can choose to manage all preferences via RPC. 

If this is currently not working in xi-mac, I think that's a regression in xi-mac; it was working when the preferences stuff was introduced.

Cogitri:
Fair enough, Tau did that when it still used the file based config (it has switched away from that now though). It doesn't seem like such a bad idea to copy the config over if `config_dir` is set in `client_started` to me though.

nangtrongvuon:
@Cogitri Thanks for reviewing! I didn’t think of atomically saving first, that makes a lot of sense.
@cmyr From what I can tell, xi-mac is supposed to use a config file right? I don’t think we’re doing preferences plists or anything at the moment, so probably the easiest solution is just to copy client_example.toml from the frontend then. Should I go ahead with that?

Cogitri:
> so probably the easiest solution is just to copy client_example.toml from the frontend then. Should I go ahead with that?

So I guess this can be closed?

cmyr:
yea, let's keep this to xi-mac.

PR #1210, opened 2019-07-02
Update notify 5.0

### LINKED ISSUE #1158: update to notify 5.0
A vague tracking issue: a major update to the notify crate is in the works, and we should investigate what will be involved in migrating to use it. I haven't looked into this at all, but there are [some docs here](https://github.com/passcod/notify/tree/v5-readme#notify).

sjoshid:
Closes #1158

This PR will change based on comments but gives a good idea of what's coming. 


rust/core-lib/src/watcher.rs
```diff
@@ -274,42 +291,17 @@ fn mode_from_bool(is_recursive: bool) -> RecursiveMode {
     }
 }
 
-// Debounced event does not implement clone
-// TODO: remove if https://github.com/passcod/notify/pull/133 is merged
-fn clone_event(event: &DebouncedEvent) -> DebouncedEvent {
```
sjoshid: `clone_event` not required as Events are cloneable.

rust/core-lib/src/watcher.rs
```diff
@@ -34,12 +34,12 @@
 //! - We are integrated with the xi_rpc runloop; events are queued as
 //! they arrive, and an idle task is scheduled.
 
-use notify::{watcher, DebouncedEvent, RecommendedWatcher, RecursiveMode, Watcher};
+use crossbeam::unbounded;
+use notify::{event::*, watcher, RecommendedWatcher, RecursiveMode, Watcher};
```
sjoshid: I have noticed `use` statements within methods. Is it because we dont want them to be loaded/linked in case we are not using notify?

Cogitri: > Is it because we dont want them to be loaded/linked in case we are not using notify?

Unlikely, you can put `#[cfg(feature = "notify")]` above normal use statements too. Not quite sure what that is about

Cogitri: Maybe it's due to naming conflicts

sjoshid: @Cogitri 
Im not sure either. I have mostly seen scoped `use` statements in testcases. By the way did you get a chance to look at https://github.com/xi-editor/xi-editor/issues/1158#issuecomment-507067284? I can provide more details if required.

sjoshid:
@cmyr and @Cogitri I think this is ready for review/merge. I have tested all kind of events on plugins, themes, configs and, of course, open files. The only worry is I was not able to "hit" `tabs.handle_plugin_fs_event`. 
Other than that there is one to one relationship between old events and new:

- Create => EventKind::Create(CreateKind::Any)
- NoticeWrite and Write => EventKind::Modify(ModifyKind::Any)
- chmod => EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any))
- NoticeRemove and Remove => EventKind::Remove(RemoveKind::Any)
- Rename => EventKind::Modify(ModifyKind::Name(RenameMode::Both))

Let me know what you guys think.

Cogitri:
Am on holidays right now, so I can't review this just yet.

cmyr:
@Cogitri are you interested in taking a first pass at this?

Cogitri:
Sure

rust/core-lib/src/tabs.rs
```diff
@@ -807,19 +815,25 @@ impl CoreState {
 
     /// Handles changes in theme files.
     #[cfg(feature = "notify")]
-    fn handle_themes_fs_event(&mut self, event: DebouncedEvent) {
-        use self::DebouncedEvent::*;
-        match event {
-            Create(ref path) | Write(ref path) => self.load_theme_file(path),
+    fn handle_themes_fs_event(&mut self, event: Event) {
+        use notify::event::*;
```
Cogitri: Same here about imports

rust/core-lib/src/tabs.rs
```diff
@@ -807,19 +815,25 @@ impl CoreState {
...
+        match event.kind {
+            EventKind::Create(CreateKind::Any) | EventKind::Modify(ModifyKind::Any) => {
```
Cogitri: I don't think we need to match `ModifyKind::Metadata` here, do we?

rust/core-lib/src/tabs.rs
```diff
@@ -807,19 +815,25 @@ impl CoreState {
...
+            }
+            EventKind::Modify(ModifyKind::Name(RenameMode::Both)) => {
+                let old = &event.paths[0];
+                let new = &event.paths[1];
                 self.remove_theme(old);
                 self.load_theme_file(new);
             }
-            Chmod(ref path) | Remove(ref path) => {
-                self.style_map.borrow_mut().sync_dir(path.parent())
+            EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any)) => {
```
Cogitri: Shouldn't `EventKind::Remove` be matched here too?

rust/core-lib/src/watcher.rs
Cogitri: Commented on that.

Cogitri:
To be honest I'm not quite sure if the usage of `Any` across this PR is correct as-is:

[From the docs](https://docs.rs/notify/5.0.0-pre.1/notify/event/enum.ModifyKind.html):

> Any should be used when more detail about the event is not known beyond the variant already selected. For example, AccessMode::Any means a file has been accessed, but that's all we know.

That sounds to me as if that in the case we _do_ know more details about the event (e.g. that a file has been renamed) `EventKind::Any` won't match

rust/core-lib/src/tabs.rs
```diff
@@ -697,12 +697,14 @@ impl CoreState {
...
     /// Handles a file system event related to a currently open file
     #[cfg(feature = "notify")]
-    fn handle_open_file_fs_event(&mut self, event: DebouncedEvent) {
-        use notify::DebouncedEvent::*;
-        let path = match event {
-            NoticeWrite(ref path) | Create(ref path) | Write(ref path) | Chmod(ref path) => path,
+    fn handle_open_file_fs_event(&mut self, event: Event) {
+        use notify::event::*;
+        let path = match event.kind {
+            EventKind::Create(CreateKind::Any)
```
Cogitri: If what I just asserted is true e.g. this wouldn't work since we want to match `CreateKind::File` and `CreateKind::Folder` here

rust/core-lib/src/tabs.rs
```diff
@@ -697,12 +697,14 @@ impl CoreState {
 
     /// Handles a file system event related to a currently open file
     #[cfg(feature = "notify")]
-    fn handle_open_file_fs_event(&mut self, event: DebouncedEvent) {
-        use notify::DebouncedEvent::*;
-        let path = match event {
-            NoticeWrite(ref path) | Create(ref path) | Write(ref path) | Chmod(ref path) => path,
+    fn handle_open_file_fs_event(&mut self, event: Event) {
+        use notify::event::*;
```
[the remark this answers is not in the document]
cmyr: to be fair this is originally my code; this is a pattern I've adopted to keep lines shorter when doing enum matching, the thinking being that putting the glob in the block makes it clear where the symbols come from. 🤷‍♂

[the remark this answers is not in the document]
Cogitri: Well, I don't want to force my way here, that's just the way I'd have done it :)

sjoshid:
Notify has gone through a lot of refactoring lately which introduced a new event classification module called `event`. You can find more details [here ](https://github.com/passcod/notify/wiki/The-Event-Guide). 
In notify, `CreateKind::File` , `CreateKind::Folder` and all other non-any or non-other events are called precise events. And the plan was to make such events configurable. But for some unknown reason this was not introduced in 5.0.0.pre1. (Actually it was introduced and then removed. See git [history](https://github.com/passcod/notify/blob/v5.0.0-pre.1/src/config.rs))
`CreateKind::File` , `CreateKind::Folder` and others are only declared but never used!! All the events used in this PR are actually emitted. 

So that begs the question: how does Xi know if an event is `CreateKind::File` when a file is created? Simple answer is Xi doesn't have to know. 4.0 version of notify never emitted precise events and Xi still worked. The same reasoning extends to `CreateKind::Any`. This does not mean Xi will never care for `CreateKind::File` or `CreateKind::Folder`  (I can think of open folder of VScode where we might need to know precise events). But for now Xi watchees watch for the `path` attribute to figure out event if it is a file or folder event. At this point, it doesnt matter if it's `CreateKind::File` or `CreateKind::Folder`. Hence `CreateKind::Any`.

5.0.0.pre1 is a step towards [next gen notify](https://github.com/passcod/notify/wiki/Presentation). And it includes this [issue ](https://github.com/passcod/notify/issues/146) which is required for another Xi [issue ](https://github.com/xi-editor/xi-editor/issues/922)I have been working on for a while.

Cogitri:
> At this point, it doesnt matter if it's CreateKind::File or CreateKind::Folder. Hence CreateKind::Any.

Yes, but does CreateKind::Any also include CreateKind::File and CreateKind::Folder? The docs sound to me as if they aren't included, so once notify starts emitting them we'd miss those events

rust/core-lib/src/tabs.rs
sjoshid: It's already been matched on line 826.

rust/core-lib/src/tabs.rs
```diff
@@ -697,12 +697,14 @@ impl CoreState {
...
+        let path = match event.kind {
+            EventKind::Create(CreateKind::Any)
+            | EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any))
+            | EventKind::Modify(ModifyKind::Any) => &event.paths[0],
```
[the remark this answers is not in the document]
sjoshid: It is simply wrong for notify to say something has changed but doesnt know the path. But you are right. If that is the case...at least Xi wont panic. Will add a check for empty `paths` vec.

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
sjoshid: Are you talking about namespaces @cmyr? That's why you moved it in method. If yes then having `CreateKind` in another other crate is highly unlikely. So maybe it makes sense to make it a global import?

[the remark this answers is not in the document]
Cogitri: Let's please not have global glob imports if possible :)

sjoshid:
Yes we would miss it. But as Xi currently stands it doesnt care about precise events. Also precise events are never emitted in 5.0.0.pre1. If I handle them in Xi...nothing would work.

Cogitri:
Alright, I guess we'll have to account for precise events later with newer notify releases then though?

sjoshid:
...only if we are interested in them. Good part is precise events can be turned on/off. So we can keep upgrading to latest and greatest version of notify without caring about precise events.

Cogitri:
Ah good, didn't know that. If we can turn them off then it's fine 👍

rust/core-lib/src/tabs.rs
sjoshid: We dont and it isnt. It's handled below in same method.

rust/core-lib/src/tabs.rs
```diff
@@ -735,16 +737,20 @@ impl CoreState {
...
-            Remove(ref path) if !path.exists() => self.remove_config_at_path(path),
-            Rename(ref old, ref new) => {
-                self.remove_config_at_path(old);
-                self.load_file_based_config(new);
+            EventKind::Remove(RemoveKind::Any) if !event.paths[0].exists() => {
+                self.remove_config_at_path(&event.paths[0])
+            }
+            EventKind::Modify(ModifyKind::Name(RenameMode::Both)) => {
+                self.remove_config_at_path(&event.paths[0]);
+                self.load_file_based_config(&event.paths[1]);
```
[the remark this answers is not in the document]
sjoshid: Implemented this.

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
sjoshid: Implemented this.

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
cmyr: I agree I'd like to see bounds checks before we do this indexing. @sjoshid did you maybe forget to push your change?

rust/core-lib/src/tabs.rs
```diff
@@ -786,8 +794,8 @@ impl CoreState {
                     self.do_start_plugin(ViewId(0), &new_plugin.name);
                 }
             }
-            Chmod(ref path) | Remove(ref path) => {
-                if let Some(plugin) = self.plugins.get_from_path(path) {
+            EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any)) => {
```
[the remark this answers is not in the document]
cmyr: Agree; this logic is maybe a bit unclear, but this is intended to handle the case where we get a `Remove` notification *but the file still exists*, which used to happen sometimes if you overwrote a file.

Is this clear from the code? Remove is only handled above if the path exists.

rust/core-lib/src/tabs.rs
cmyr: it has only been matched if thee path does not exist; we need to handle the case where the path does exist, see my earlier comment. (This is confusing, I understand).

rust/core-lib/src/watcher.rs
```diff
@@ -50,7 +50,7 @@ use xi_rpc::RpcPeer;
 pub const DEBOUNCE_WAIT_MILLIS: u64 = 50;
 
 /// Wrapper around a `notify::Watcher`. It runs the inner watcher
-/// in a separate thread, and communicates with it via an `mpsc::channel`.
+/// in a separate thread, and communicates with it via an `use crossbeam_channel::unbounded;`.
```
cmyr:
```suggestion
/// in a separate thread, and communicates with it via a [crossbeam channel].
/// [crossbeam channel]: https://docs.rs/crossbeam-channel
```

rust/core-lib/src/watcher.rs
```diff
@@ -213,22 +213,35 @@ impl FileWatcher {
...
-    fn wants_event(&self, event: &DebouncedEvent) -> bool {
-        use self::DebouncedEvent::*;
-        match *event {
-            NoticeWrite(ref p) | NoticeRemove(ref p) | Create(ref p) | Write(ref p)
-            | Chmod(ref p) | Remove(ref p) => self.applies_to_path(p),
-            Rename(ref p1, ref p2) => self.applies_to_path(p1) || self.applies_to_path(p2),
-            Rescan => false,
-            Error(_, ref opt_p) => opt_p.as_ref().map(|p| self.applies_to_path(p)).unwrap_or(false),
+    fn wants_event(&self, event: &Event) -> bool {
+        match &event.kind {
```
cmyr: Because the event no longer stores the paths in the same enum as the type, we can dramatically simplify this function, to something like:

```rust
fn wants_event(&self, event: &Event) -> bool {
    event.paths.iter().map(|p| self.applies_to_path(p)).any()
}
```

sjoshid:
I was on a mini vacation but Im back and will continue work on this PR today. 
As the PR currently stands, I have tested it thoroughly on Mac (XiMac) and Linux (Tau) without any issues. 
Tested three areas:
- [x] Pref changes: Change `word_wrap` and other settings. 
- [x] Disk changes on open file/s: Change a file from a different source when it is open in Xi. This includes renaming a file, changing permissions, deleting file. 
- [x] Theme changes: Added a custom theme and was able to select it from Debug -> Theme. Selected them was correctly applied. 

But I might have missed some gotchas. @nangtrongvuon if you find any such issues let me know.

rust/core-lib/src/watcher.rs
```diff
@@ -213,22 +213,35 @@ impl FileWatcher {
...
+                } else {
+                    false
+                }
+            }
+            EventKind::Modify(ModifyKind::Name(RenameMode::Both)) => {
+                if event.paths.len() == 2 {
+                    //There will be two paths. First is "from" and other is "to".
+                    self.applies_to_path(&event.paths[0]) || self.applies_to_path(&event.paths[1])
+                } else {
+                    false
```
sjoshid: We need exactly 2 to deterministically work with from and to paths. Any other len is not correct.

rust/core-lib/src/watcher.rs
```diff
@@ -213,22 +213,35 @@ impl FileWatcher {
...
+    fn wants_event(&self, event: &Event) -> bool {
+        match &event.kind {
+            EventKind::Create(CreateKind::Any)
+            | EventKind::Remove(RemoveKind::Any)
+            | EventKind::Modify(ModifyKind::Any)
+            | EventKind::Modify(ModifyKind::Metadata(MetadataKind::Any)) => {
+                if event.paths.len() == 1 {
+                    self.applies_to_path(&event.paths[0])
+                } else {
+                    false
```
sjoshid: This basically says "if path length is not 1 ...I dont want this event".

rust/core-lib/src/watcher.rs
sjoshid: Yeah. But Im taking advantage of `wants_event` to do path's length check. And the path's length will change based on event kind.

rust/core-lib/src/watcher.rs
Cogitri: Wouldn't
```
if let Some(path) = event.paths.get(0) {
    self.applies_to_path(path)
}
```

Be clearer?

rust/core-lib/src/watcher.rs
Cogitri: Same as above, if-let would be nicer here IMO

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
sjoshid: Ahh Rust got me. It's called Pattern Guard and match arm is not executed unless the pattern guard is true. Will make this change.

rust/core-lib/src/tabs.rs
sjoshid: Pattern guard here as well. Making this change.

rust/core-lib/src/watcher.rs
sjoshid: It is but not here. What if len is more than 1? Should we still get the first element? We cannot deterministically say first element is THE element we are looking for.

rust/core-lib/src/watcher.rs
cmyr: I think my code should work regardless of the length of `paths`.

rust/core-lib/src/watcher.rs
cmyr: The 'rustic' way to write this code, I think, would be:

```rust
event.paths.first().map(|p| self.applies_to_path(p)).unwrap_or(false)
```

rust/core-lib/src/watcher.rs
cmyr: To make this more clear: my code checks every path in `paths`, and returns `true` if `self.applies_to_path` is true for *any* of them. Do we need any more logic than this?

rust/core-lib/src/watcher.rs
Cogitri: Hm, that doesn't address @sjoshid 's concern if len is more than 1, does it?

rust/core-lib/src/watcher.rs
sjoshid: Sorry I maybe making this harder than it needs to be but I can think of two scenarios we need to avoid:
1. Imagine we get a rename event with ONLY from path. to is missing. Probably a bug in notify. 
This `event.paths.iter().map(|p| self.applies_to_path(p)).any()` might/ might not return true. If it does, we are guaranteed to panic down the line. If at this point, `wants_event` returns false...we can avoid panic. 
2. For other events (create, modify, remove) what if paths len is not 1? Other suggestion `event.paths.first().map(|p| self.applies_to_path(p)).unwrap_or(false)` will work even if path > 1. But we need to return false here too. 

So to avoid index out of bound panics, we should check
1. For create, modify and remove events we check it they have exactly one path. 
2. For rename events, check if it has exactly two. 

If there is a more rustic way to do this....I would be happy to look in to it.

rust/core-lib/src/watcher.rs
sjoshid: yeah. @cmyr I commented in other conversation.

rust/core-lib/src/watcher.rs
cmyr: Okay, I see some of what you're saying. That said, I think the larger problem is that we are unsure of the API guarantees of the `notify` crate. Is it *possible* for a `Rename` event to arrive with less than 2 paths? If it is, then it is totally reasonable for other code of ours to want to receive that event and handling it; if other code is assuming there are always 2 paths, I think that's a mistake.

I think it might be okay if we ignore events with 0 paths, but it might be worth logging if this is happening.

So basically: I think we need to correctly handle a variable number of paths, if that is something that the API provides under normal conditions.

sjoshid: > Is it possible for a Rename event to arrive with less than 2 paths?

To be specific we use `RenameMode::Both` and I dont think so. See [this](https://docs.rs/notify/5.0.0-pre.1/notify/event/enum.RenameMode.html#variant.Both). For all other events notify doesnt say they will have exactly 1 path. So we do that check. 

> If it is, then it is totally reasonable for other code of ours to want to receive that event and handling it; 

This means putting conditions all over. We can avoid this. 

> if other code is assuming there are always 2 paths, I think that's a mistake.

Other code can safely assume there are 2 paths because `wants_event` gave us the guarantee. If you agree to this..then I dont think we need more changes to `wants_event`. Changes might be requried to make it more rustic. Correct me if Im wrong. 

> I think it might be okay if we ignore events with 0 paths, but it might be worth logging if this is happening.

Agree. Making this change.

Cogitri:
So, what's the roadmap for this? Do we want to merge this already even though notify is still in pre-release?

sjoshid:
Although it says "pre" it is quite stable. Pre-release [description ](https://github.com/passcod/notify/tree/v5.0.0-pre.1#why-a-prerelease)says so too. .

But Im sad to find out notify is abandoned. See [note](https://github.com/passcod/notify#notify-is-abandoned).
Does this mean we need to stop using it in Xi or somehow include it as part of Xi ie no dependency?

Cogitri:
Well, it also mentions that there are multiple people with commit and release bits, so I'd guess that someone will keep it going

sjoshid:
I hope so.

nangtrongvuon:
I've been doing some general notes writing and stuff with xi-mac + this PR, and I haven't found anything unusual as of 3 hours. So far this seems good to me™️.

sjoshid:
Thanks, @nangtrongvuon. 
@cmyr what do you think?

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
cmyr: Bump this comment; this will crash if one of the paths is missing, which I believe is possible; for instance if the file was moved somewhere that we don't have permission to access.

cmyr:
I think this is mostly good, but I'm still uncomfortable with doing indexing without checking bounds. I do not think we can assume that `paths` contains any items, and I don't think we should ever index into `paths` without checking the length first.

rust/core-lib/src/tabs.rs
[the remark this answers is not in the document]
sjoshid: Agree. But that will never happen. `wants_event` will catch such invalid events.

rust/core-lib/src/watcher.rs
```diff
@@ -495,15 +483,26 @@ mod tests {
         assert_eq!(
             events,
             vec![
-                (2.into(), DebouncedEvent::NoticeWrite(tmp.mkpath("adir/dir2/file"))),
-                (2.into(), DebouncedEvent::Write(tmp.mkpath("adir/dir2/file"))),
+                //(2.into(), Event::NoticeWrite(tmp.mkpath("adir/dir2/file"))),
```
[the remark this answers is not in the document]
cmyr: @sjoshid can these comments be removed?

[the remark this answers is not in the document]
sjoshid: Yeah. Sorry. I see you commented a while ago.

sjoshid:
@cmyr 
What should we do about this PR? Any more changes you think I need to make? I hope I answered all your concerns.

Cogitri:
Ah, that may very well be on my system being odd though, I think we should merge this and and afterwards I'll work on upping the error reporting for inotify

sjoshid:
> Note to whoever merges: I'd like to squash this, because there's a whole bunch of extra merge commits.
Yup. I work on multiple machines so I commit every little thing I do. But will keep squashing them henceforth. 

> Thanks for your patience @sjoshid!
Anytime. 

Thank you both for reviewing this!!

PR #1218, opened 2019-08-02
core-lib: send plugin_started in EventContext::finish_init instead of CoreState::plugin_connect

### LINKED ISSUE #1217: Sometimes (at start?) plugin_started isn't sent
So, after some digging I found out that `plugin_started` is usually called in `connect_plugin`. The problem is in line 908 in core_lib/src/tab.rs:

```
self.iter_groups().for_each(|mut cx| cx.plugin_started(&plugin);
```

At startup we don't have any views yet, so this isn't executed. Should we wait for views here, or maybe we should send plugin_started for a new view when it's created anyway? I only tested this with the syntect plugin.

CC @cmyr

Cogitri:
This way we can be sure that if a plugin knows about a view, the view also knows
about that plugin. Without this commit `plugin_started` wouldn't be sent to new
views but only to the ones which exist at the point the plugin has been started.

fixes #1217

### DIFF
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -387,7 +387,10 @@ impl<'a> EventContext<'a> {
     pub(crate) fn finish_init(&mut self, config: &Table) {
         if !self.plugins.is_empty() {
             let info = self.plugin_info();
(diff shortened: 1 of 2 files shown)


Cogitri:
Thanks for taking a look :)

PR #1223, opened 2019-08-23
core-lib: do not set a buffer's default language to "" but to "Plain Text"

### LINKED ISSUE #1194: language_changed can send `""` as  new language
I'm honestly not sure if this is expected behavior, but shouldn't it send 'Plain Text' instead? At least in gxi I just redirect `""` to `Plain Text`.

Cogitri:
Otherwise we might send "" in the language_changed RPC to the frontend, which
isn't a valid syntax.

fixes #1194


Cogitri:
Thanks for having a look :)

PR #1225, opened 2019-08-26
core-lib: don't create empty file in FileManager::open

### LINKED ISSUE #1092: 'new_view' creates the file specified by 'file_path'
# Xi-Editor Issue Template
- [X] I have searched existing issues and could not find my issue.
- [X] I have studied the documentation.

## Details

using xi-editor rev 97a60f6f6591931877e0b48c8df0340b142616cd

### Expected

```
{"id"=0, "method": new_view, "params":{"file_path":"/tmp/file"}}
```

xi-editor creates a new view but only creates `/tmp/file`  after `save` has been called for that view.

### Actual

```
{"id"=0, "method": new_view, "params":{"file_path":"/tmp/file"}}
```

xi-editor creates an empty file `/tmp/file` upon creating the new view. This is irritating for users, as they might end up with empty files if they typo in the file name etc. even though they've never called save.

Cogitri:
Not quite sure if this isn't too simplistic, but at least for me this
seems to work fine.

fixes #1092

### DIFF
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -107,7 +107,7 @@ impl FileManager {
 
     pub fn open(&mut self, path: &Path, id: BufferId) -> Result<Rope, FileError> {
         if !path.exists() {
-            let _ = File::create(path).map_err(|e| FileError::Io(e, path.to_owned()))?;
+            return Ok(Rope::from(""));
         }
 
         let (rope, info) = try_load_file(path)?;


Cogitri:
I've had this running with Tau for the last few days and it seems to work just fine.

rust/core-lib/src/file.rs
```diff
@@ -105,9 +105,9 @@ impl FileManager {
         false
     }
 
-    pub fn open(&mut self, path: &Path, id: BufferId) -> Result<Rope, FileError> {
+    pub fn open(&mut self, path: &Path, id: BufferId) -> Result<Option<Rope>, FileError> {
         if !path.exists() {
-            let _ = File::create(path).map_err(|e| FileError::Io(e, path.to_owned()))?;
+            return Ok(None);
```
cmyr: We could also just return an empty rope here? I don't have a strong opinion, returning an empty rope is a smaller change. It's possible that at some point we will want to have some special behaviour if the file doesn't exist, but since we don't do that now I think we can do whatever is easiest.

Cogitri: Alrighty, that can work too, but felt a bit dirty to me. I'm fine with this though.

Cogitri: Pushed new version.

PR #1229, opened 2019-09-20
core-lib: don't error if file permissions can't be recovered in save

Cogitri:
Some file systems don't support this (typically /run in Linux), which cause
error messages along the line of:

`[2019-09-20][16:28:39][xi_core_lib::tabs][ERROR] File error: "Function not implemented (os error 38). File path: \"/run/user/1000/doc/8f43530e/loremipsum.txt\""`

Yet saving suceeds, it only fails to set the permissions. As such only throw
a warning if this goes wrong.


Cogitri:
Thanks for taking a look :)

PR #1230, opened 2019-09-20
deps: update, bump bytecount from 0.5 to 0.6

Cogitri:


rust/Cargo.toml
```diff
@@ -1,3 +1,4 @@
+
```
cmyr: whitespace

rust/Cargo.toml
```diff
@@ -38,3 +39,8 @@ members = [
 
 [profile.bench]
 lto = true
+
+# Avoid libllvm/libclang dep
```
cmyr: What's the rationale here?

Cogitri: This is currently a PR but upstream seems kind of dead :/

Without this we'll pull in libclang and libllvm (both pretty heavy deps) and also pull in clap (which bloats binary size by ~1.5MB). I hope this will be merged soon.

rust/Cargo.toml
Cogitri: Oh right. Thanks for noticing

rust/Cargo.toml
cmyr: Can we include a link to the PR in the code, with a note to remove this when it's merged? This will greatly benefit future archaeologists. 🕵️‍♀️

Cogitri: Right, sounds fair

Cogitri: Done

PR #1231, opened 2019-09-28
lsp: support sync-kind from SyncOptions

yonilevy:
## Summary
Prior to this, the LSP plugin was ignoring language servers specification of the way they want to be synced if they provided it through `TextDocumentSyncOptions`, defaulting to a full sync.

Relevant protocol spec:

> interface ServerCapabilities {
> 	/**
> 	 * Defines how text documents are synced. Is either a detailed structure defining each notification or
> 	 * for backwards compatibility the TextDocumentSyncKind number. If omitted it defaults to `TextDocumentSyncKind.None`.
> 	 */
> 	textDocumentSync?: **TextDocumentSyncOptions** | number;
> 

This means, when working with a language server that makes use of that field (such as [pyls](https://github.com/palantir/python-language-server)), every edit to the document would be synced by sending over the entire file contents.

This small change fixes that.

### DIFF
--- a/rust/lsp-lib/src/language_server_client.rs
+++ b/rust/lsp-lib/src/language_server_client.rs
@@ -284,6 +284,10 @@ impl LanguageServerClient {
     pub fn get_sync_kind(&mut self) -> TextDocumentSyncKind {
         match self.server_capabilities.as_ref().and_then(|c| c.text_document_sync.as_ref()) {
             Some(&TextDocumentSyncCapability::Kind(kind)) => kind,
+            Some(TextDocumentSyncCapability::Options(options)) => match options.change {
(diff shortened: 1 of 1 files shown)


yonilevy:
the build failures seem to be rustfmt diffs that aren't related to my change, not sure why they're triggered now (and not on the previous build)

Cogitri:
Possibly due to new lints in newer versions of rustfmt.

cmyr:
I'm going to close this. If you're interested in reopening and can get CI to pass I'd be happy to merge small fixes, but I consider lsp-lib abandoned at this point, and I don't really expect it to be picked up in this form again. (If we do support it again, we would probably start from scratch)

PR #1233, opened 2019-10-01
Replace idle call to finalize_new_views() with immediate call

### LINKED ISSUE #1115: Panics when sending RPC too early (?)
## Details

- Tested on Linux
- Using Rustc 1.32.0, installed via rustup
- Tested in gxi frontend master

I've just added more threading to gxi's SharedQueue (which holds rpc from xi-editor until gxi handles it) in https://github.com/Cogitri/gxi/commit/dd8a00559509596d1d80ead20e9171cd34c5363f. This causes xi-editor to panic in non-release mode on big files due to https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/linewrap.rs#L676, I guess it doesn't like receiving some RPC before it has finished loading the file? Letting the thread sleep for ~100 msec fixes it, so I guess this is some timing problem.

See https://gist.github.com/Cogitri/c8f29a4b3a37530c88cb702277e71756 for the backtrace (and gxi logging).

### LINKED ISSUE #1232: 'new_view' + 'set_theme' cause an assertion failure in linewrap.rs
Hi,

I have encountered a data race between finalize_new_views() and do_set_theme() in CoreState, which results in a debug assertion failure in linewrap.rs.

This data race is triggered by the following simple sequence of rpcs:

1. client_started
2. new_view with params: {"file_path":"/path/to/some/file"} (file should be nonempty)
3. set_theme

See test update in c4eater/xi-editor@7dd6b40d36cbcac3aba3e990931a87aa5ff7a644 that demonstrates the issue. An attempt to run the updated test "test_state" in core_lib (in a **debug** build) will result in the following output:

```
thread 'test_state' panicked at 'assertion failed: `(left == right)`
  left: `935`,
 right: `0`', core-lib/src/linewrap.rs:685:9
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: xi_core_lib::linewrap::MergedBreaks::new
             at core-lib/src/linewrap.rs:685
   8: xi_core_lib::linewrap::Lines::iter_lines
             at core-lib/src/linewrap.rs:192
   9: xi_core_lib::view::View::send_update_for_plan
             at core-lib/src/view.rs:796
  10: xi_core_lib::view::View::render_if_dirty
             at core-lib/src/view.rs:866
  11: xi_core_lib::event_context::EventContext::render
             at core-lib/src/event_context.rs:371
  12: xi_core_lib::event_context::EventContext::render_if_needed
             at core-lib/src/event_context.rs:357
  13: xi_core_lib::tabs::CoreState::notify_client_and_update_views::{{closure}}
             at core-lib/src/tabs.rs:463
  14: core::iter::traits::iterator::Iterator::for_each::{{closure}}
             at /usr/src/rustc-1.34.2/src/libcore/iter/traits/iterator.rs:604
  15: core::iter::traits::iterator::Iterator::fold::{{closure}}
             at /usr/src/rustc-1.34.2/src/libcore/iter/traits/iterator.rs:1684
  16: core::iter::traits::iterator::Iterator::try_fold
             at /usr/src/rustc-1.34.2/src/libcore/iter/traits/iterator.rs:1572
  17: core::iter::traits::iterator::Iterator::fold
             at /usr/src/rustc-1.34.2/src/libcore/iter/traits/iterator.rs:1684
  18: core::iter::traits::iterator::Iterator::for_each
             at /usr/src/rustc-1.34.2/src/libcore/iter/traits/iterator.rs:604
  19: xi_core_lib::tabs::CoreState::notify_client_and_update_views
             at core-lib/src/tabs.rs:458
  20: xi_core_lib::tabs::CoreState::do_set_theme
             at core-lib/src/tabs.rs:449
  21: xi_core_lib::tabs::CoreState::client_notification
             at core-lib/src/tabs.rs:319
  22: <xi_core_lib::core::XiCore as xi_rpc::Handler>::handle_notification
             at core-lib/src/core.rs:115
  23: <xi_rpc::RpcLoop<W>>::mainloop::{{closure}}


<...etc>
```

Debugging shows the following:

RPC "new_view" launches do_new_view(). In the end, this function initiates an asynchronous idle call to finalize_new_views() and returns.
Then do_set_theme kicks in, and the execution flow ends up in  inewrap::MergedBreaks::new().
An assert in this new() method expects the line breaks to be initialized, which is not true in this particular case: they are initialized in finalize_new_views() which has not been called yet.

I don't have a strong opinion on how to fix this in the right way, but a straightforward immediate call to finalize_new_views() in the end of do_new_view() seems to do the job (see c4eater/xi-editor@1e60c5ae18748cd972a3cd5f9970395386167c12).

c4eater:
## Summary
This change adds a new (failing) test demonstrating #1232 and then a fix, which includes making a synchronous call to finalize_new_views.

## Related Issues
Closes #1232.
Closes #1115.

## Checklist

### DIFF
--- a/rust/core-lib/src/event_context.rs
+++ b/rust/core-lib/src/event_context.rs
@@ -384,9 +384,17 @@ impl<'a> EventContext<'a> {
 /// requires access to particular combinations of state. We isolate such
 /// special cases here.
 impl<'a> EventContext<'a> {
+    pub(crate) fn view_init(&mut self) {
+        let wrap_width = self.config.wrap_width;
+        let word_wrap = self.config.word_wrap;
+
+        self.with_view(|view, text| view.update_wrap_settings(text, wrap_width, word_wrap));
+    }
+
     pub(crate) fn finish_init(&mut self, config: &Table) {
         if !self.plugins.is_empty() {
             let info = self.plugin_info();
+
             self.plugins.iter().for_each(|plugin| {
                 plugin.new_buffer(&info);
                 self.plugin_started(plugin);
@@ -402,7 +410,16 @@ impl<'a> EventContext<'a> {
 
         self.client.config_changed(self.view_id, config);
         self.client.language_changed(self.view_id, &self.language);
-        self.update_wrap_settings(true);
+
+        // Rewrap and request a render.
+        // This is largely similar to update_wrap_settings(), the only difference
+        // being that the view is expected to be already initialized.
+        self.rewrap();
+
+        if self.view.borrow().needs_more_wrap() {
+            self.schedule_rewrap();
+        }
+
         self.with_view(|view, text| view.set_dirty(text));
         self.render()
     }
@@ -731,6 +748,7 @@ mod tests {
             let recorder = RefCell::new(Recorder::new());
             let harness = ContextHarness { view, editor, client, core_ref, kill_ring,
                              style_map, width_cache, config_manager, recorder };
+            harness.make_context().view_init();
             harness.make_context().finish_init(&config);
             harness
 
--- a/rust/core-lib/src/tabs.rs
+++ b/rust/core-lib/src/tabs.rs
@@ -379,10 +379,15 @@ impl CoreState {
 
         let config = self.config_manager.add_buffer(buffer_id, path.as_ref().map(|p| p.as_path()));
 
-        //NOTE: because this is a synchronous call, we have to return the
-        //view_id before we can send any events to this view. We mark the
-        // view as pending and schedule the idle handler so that we can finish
-        // setting up this view on the next runloop pass, in finalize_new_views.
+        // NOTE: because this is a synchronous call, we have to initialize the
+        // view and return the view_id before we can send any events to this
+        // view. We call view_init(), mark the view as pending and schedule the
+        // idle handler so that we can finish setting up this view on the next
+        // runloop pass, in finalize_new_views.
+
+        let mut edit_ctx = self.make_context(view_id).unwrap();
+        edit_ctx.view_init();
+
         self.pending_views.push((view_id, config));
         self.peer.schedule_idle(NEW_VIEW_IDLE_TOKEN);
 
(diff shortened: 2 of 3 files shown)


Cogitri:
Seems like this currently breaks opening files during startup, e.g. if you open tau with a file as argument xi-core crashes now.

c4eater:
> Seems like this currently breaks opening files during startup, e.g. if you open tau with a file as argument xi-core crashes now.

OK you were probably trying my earlier (invalid) implementation (c4eater/xi-editor@1e60c5a)...
I just rebased against the xi-editor's master and got a green CI status. Have no pb opening a file (with a different frontend) as well. Can you please check it again?

Cogitri:
Just tried again, seems to still be happening :/

Cogitri:
Ah, so it seems like without this the cmd flow is like this:

1) tau -> xi: client_started & config_changed
2) xi -> tau: available_themes & available_languages
3) tau -> xi: set_theme and language

c4eater:
Aha, I get it. My fix moved "EventContext::update_wrap_settings" (which in turn could ask for "measure_width") from the idle handler into an immediate call to do_new_view. This is probably legal but I don't need to change the behavior that dramatically to fix the bug. I've updated my PR once again and I believe that the behavior change now is truly minimalistic - only view::update_wrap_settings is now called immediately.

Cogitri:
This also fixes #1115 too apparently, could you add that to the commit msg please? :)

LGTM otherwise

c4eater:
OK, updated the commit msg.

Cogitri:
I think `(fixes #1232, #1115)` doesn't work, you need to do `fixes #1232, fixes #1115`

c4eater:
Interesting... Just played around with the commit msg and found that the issue page and the presence of "maybe fixed by..." probably reacts on what is written in the PR description, not in the commit (s). I've updated the commit msg to "fixes #x, fixes #y" but that didn't affect 1115's description page. It picked up the changes I updated the PR text though.

Anyway, looks like the comment is consistent now.

Cogitri:
Thank you!

PR #1234, opened 2019-10-02
chore(deps): bump

Cogitri:
Since there has been a force-push to rust-onig this also fixes the build.


cmyr:
looks like we need a rustfmt run, if you want to add that to this PR that works for me 🤷‍♂

Cogitri:
Done

PR #1236, opened 2019-10-05
Bugfix: rendering and updating annotations when scrolling

scholtzan:
## Summary

While hacking on the [git diff plugin](https://github.com/scholtzan/xiff) I've been working on, I ran into two issues:

1. Existing annotations were incorrectly invalidated. Only annotations that intersected with some edit region were invalidated/deleted. However, that caused some weird behaviour, for example, when deleting some text region that was followed by annotations. In that case all annotations that followed had incorrect offsets. This PR fixes this problem and invalidates all annotations that follow some edit region.

2. Currently, annotations are not cached in the frontend, so every time the visible region changes all annotations of that region need to be sent to the frontend. Before, updates of the view (which annotations are part of) were only sent if the text or selections had changed since frontends cache this information. This resulted in annotations disappearing when scrolling to different regions. This PR fixes this behaviour.

## Review Checklist

### DIFF
--- a/rust/core-lib/src/annotations.rs
+++ b/rust/core-lib/src/annotations.rs
@@ -98,7 +98,7 @@ impl Annotations {
 
     /// Remove annotations intersecting `interval`.
     pub fn invalidate(&mut self, interval: Interval) {
-        self.items.delete_intersecting(interval);
+        self.items.delete_after(interval);
     }
 }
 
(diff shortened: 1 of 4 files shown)


c4eater:
Anna,

Sorry to inform that this change introduces a bug in the line cache notification flow, which results in a frontend crash on attempt to open virtually any file. The bug shows up in "xrl" based frontends (xi-term, tau) as an xrl::LineCache consistency error, but I believe it affects every frontend. A sample panic notification:

> thread 'tokio-runtime-worker-1' panicked at 'cache update failed', xrl-0.0.8/src/cache.rs:232:9

From a quick look, I can see that xi-core sends 2 completely similar "update" notifications (for an initially empty cache) in a row, which would definitely break the consistency of any good client-side cache.

The duplicating update is (irrelevant details omitted):

op = ins, n=12, lines = "line1\n" "line2\n"... // blah
op = invalidate, n=998
op = invalidate, n=185

scholtzan:
@c4eater I see. One workaround would be to copy all the lines to the "new cache state". I created a PR that implements that behaviour: https://github.com/xi-editor/xi-editor/pull/1238
But if you can think of something better then let me know.

PR #1237, opened 2019-10-19
Implement moving to soft beginning/end of line

nangtrongvuon:
## Summary
This adds a new movement type: `BeginningOfLine` and `EndOfLine`. When a movement of either one of these types is performed, the caret is moved to the soft beginning/end of line, which is defined as the first/last offset where a non-whitespace character appears. If the caret is already on this position, the caret is then moved to the hard beginning/end of line.

## Review Checklist


Cogitri:
Just to clarify: You'd use this to move to the beginning of the line + the indention basically?

nangtrongvuon:
Yeah. What's currently in xi right now is "move to hard beginning of line", which does the same thing as `Ctrl + A` on `emacs`. When trying to edit code, having to move past the indentation again is annoying, which is mainly the reason for this PR.

side note: I don't know why CI is failing due to the weird build artifacts? I ran rustfmt a while ago but it's all messy for some reason.

Cogitri:
Travis CI failure seems legit, maybe you're using an older rustfmt?

nangtrongvuon:
Whoops, my bad, you were right. I updated my toolings and the local checks seem to pass now.

Cogitri:
Thanks. Will look at it in a bit :)

rust/core-lib/src/movement.rs
```diff
@@ -226,6 +230,37 @@ pub fn region_movement(
             }
             (offset, None)
         }
+        Movement::BeginningOfLine => {
+            let line = view.line_of_offset(text, r.end);
+            let mut bol_offset = view.offset_of_line(text, line);
+
+            let mut word_cursor = WordCursor::new(text, bol_offset);
```
cmyr: Okay I've thought about this a lot, and I think there's a simpler solution:

- move to the beginning of the line (as you have, here)
- create a WordCursor
- move to the _next_ word
- move to the _previous_ word
- ensure that you are in the correct line

and then for 'end of line' you can just flip the next/prev calls.

This also gives us another thing we can call this; it can be `MoveToLeftmostWordInLine`.

nangtrongvuon: I've actually considered this approach, and I couldn't go with it because `move to next word`  considers punctuation and special characters, so you might have to call `next word` twice instead of just simply one time.

Consider this simple line:
`let mut something = 0;`

Move to end of line, prev word then next word would have the cursor like this:
`let mut something = 0|;`

PR #1238, opened 2019-10-24
Copy lines when updating annotations

scholtzan:
This PR attempts to fix the issue described in https://github.com/xi-editor/xi-editor/pull/1236#issuecomment-545373431

@c4eater 

## Review Checklist

### DIFF
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -750,7 +750,9 @@ impl View {
             .collect::<Vec<_>>();
 
         if !self.lc_shadow.needs_render(plan) {
-            let update = Update { ops: Vec::new(), pristine, annotations };
+            let total_lines = self.line_of_offset(text, text.len()) + 1;
+            let update =
+                Update { ops: vec![UpdateOp::copy(total_lines, 1)], pristine, annotations };
             client.update_view(self.view_id, &update);
             return;
         }


Cogitri:
I'm a bit (:D) late to the party, but I just got around to updating Tau to this and unfortunately it seems like with #1236 applied even with this patch there are still problems in the linecache when doing big jumps - e.g. when using find I'm getting error messages like:

```
[2020-05-27T16:59:28Z ERROR tau_linecache::linecache] line 0+9=9, a copy source is none
[2020-05-27T16:59:28Z ERROR tau_linecache::linecache] line 0+66=66, a copy source is none
[2020-05-27T16:59:28Z ERROR tau_linecache::linecache] line 0+67=67, a copy source is none
[2020-05-27T16:59:28Z ERROR tau_linecache::linecache] line 0+68=68, a copy source is none
```

From my linecache, meaning it wasn't able to copy a line because the line it was supposed to copy isn't in Tau's linecache yet. This might very well be a bug in Tau's linecache, but again, this doesn't happen without #1236 applied. Not really sure what do about that since I don't know the internals of xi-editor too well, so I figured reporting it here would make sense.

c4eater:
I'm also aware of this problem and this is a different issue, unrelated to annotations. I think I have mentioned it in [this](https://xi.zulipchat.com/#narrow/stream/147924-xi-editor/topic/invalidating.20a.20line.20in.20the.20middle.20of.20the.20linecache.3F) zulip topic. This issue exists for a long time and it is a xi-core's problem. In my zulip reply, I demonstrated how to reproduce it with raw XML-RPCs sent to xi-core:

client_started
new_view
scroll [0,34]
insert "line 1"
insert newline
insert "line 2"
insert newline
// (...etc... insert lot of lines; 200 is enough)
// (Instead of inserting 200 lines, can just open a precreated file with that many lines)
goto_line 1
goto_line 200

I didn't open any issue on github yet because I'd like to create a bug report with an ironclad proof that this is a xi-core issue (current suspect is the core's line cache shadow). I'm actually investigating this (veeeery slowly though) because this issue is a blocker for an important xrl update (switching to hashmap). I can provide more details if anyone wants to contribute.

PR #1241, opened 2019-11-23
Tail toggle v2

sjoshid:
Closes https://github.com/xi-editor/xi-editor/issues/922

## Review Checklist

### DIFF
--- a/rust/core-lib/src/client.rs
+++ b/rust/core-lib/src/client.rs
@@ -235,6 +235,17 @@ impl Client {
     pub fn schedule_timer(&self, timeout: Instant, token: usize) {
         self.0.schedule_timer(timeout, token);
     }
+
+    /// Let client know tail was successfully enabled/disabled
+    pub fn toggle_tail_config_changed(&self, view_id: ViewId, is_tail_enabled: bool) {
+        self.0.send_rpc_notification(
+            "toggle_tail_config_changed",
+            &json!({
+                "view_id": view_id,
+                "is_tail_enabled": is_tail_enabled,
+            }),
+        );
+    }
 }
 
 #[derive(Debug, Serialize)]
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -55,7 +55,7 @@ pub struct FileInfo {
     #[cfg(target_family = "unix")]
     pub permissions: Option<u32>,
     #[cfg(feature = "notify")]
-    pub tailing_enabled: Option<u64>,
+    pub tail_position: Option<u64>,
 }
 
 pub enum FileError {
@@ -107,6 +107,19 @@ impl FileManager {
         false
     }
 
+    pub fn tail(&mut self, path: &Path, id: BufferId) -> Result<Rope, FileError> {
+        if !path.exists() {
+            return Ok(Rope::from(""));
+        }
+
+        //sj_todo handle unwrap.
+        let existing_file_info = self.file_info.get(&id).unwrap();
+        let (rope, info) = try_tailing_file(path, existing_file_info)?;
+
+        self.file_info.insert(id, info);
+        Ok(rope)
+    }
+
     pub fn open(&mut self, path: &Path, id: BufferId) -> Result<Rope, FileError> {
         if !path.exists() {
             return Ok(Rope::from(""));
@@ -150,7 +163,7 @@ impl FileManager {
             #[cfg(target_family = "unix")]
             permissions: get_permissions(path),
             #[cfg(feature = "notify")]
-            tailing_enabled: None,
+            tail_position: None,
         };
         self.open_files.insert(path.to_owned(), id);
         self.file_info.insert(id, info);
@@ -189,16 +202,52 @@ impl FileManager {
         let end_position =
             f.seek(SeekFrom::End(0)).map_err(|e| FileError::Io(e, path.to_owned()))?;
 
-        existing_file_info.tailing_enabled = Some(end_position);
+        existing_file_info.tail_position = Some(end_position);
         Ok(())
     }
 
     #[cfg(feature = "notify")]
     pub fn disable_tailing(&mut self, id: BufferId) -> Result<(), FileError> {
         let existing_file_info = self.file_info.get_mut(&id).unwrap();
-        existing_file_info.tailing_enabled = None;
+        existing_file_info.tail_position = None;
         Ok(())
     }
+}
+
+/// When tailing a file, instead of reading file from beginning, we need to get changes from the end.
+/// This method does that.
+#[cfg(feature = "notify")]
+fn try_tailing_file<P>(path: P, file_info: &FileInfo) -> Result<(Rope, FileInfo), FileError>
+where
+    P: AsRef<Path>,
+{
+    let mut f =
+        File::open(path.as_ref()).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+    let end_position =
+        f.seek(SeekFrom::End(0)).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+
+    //sj_todo unwrap needs to be handled for None.
+    let current_position = file_info.tail_position.unwrap();
+    let diff = end_position - current_position;
+    let mut buf = vec![0; diff as usize];
+    f.seek(SeekFrom::Current(-(buf.len() as i64))).unwrap();
+    f.read_exact(&mut buf).unwrap();
+
+    let mod_time =
+        f.metadata().map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?.modified().ok();
+
+    let encoding = file_info.encoding;
+    let permissions = file_info.permissions;
+    let rope = try_decode(buf, encoding, path.as_ref())?;
+    let info = FileInfo {
+        encoding,
+        mod_time,
+        path: path.as_ref().to_owned(),
+        has_changed: false,
+        permissions,
+        tail_position: Some(end_position),
+    };
+    Ok((rope, info))
 }
 
 fn try_load_file<P>(path: P) -> Result<(Rope, FileInfo), FileError>
@@ -221,9 +270,8 @@ where
         permissions: get_permissions(&path),
         path: path.as_ref().to_owned(),
         has_changed: false,
-        //sj_todo tailing_enabled should be set from existing FileInfo.
         #[cfg(feature = "notify")]
-        tailing_enabled: None,
+        tail_position: None,
     };
     Ok((rope, info))
 }
--- a/rust/core-lib/src/file.rs
+++ b/rust/core-lib/src/file.rs
@@ -18,7 +18,7 @@ use std::collections::HashMap;
 use std::ffi::OsString;
 use std::fmt;
 use std::fs::{self, File};
-use std::io::{self, Read, Write};
+use std::io::{self, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
 use std::str;
 use std::time::SystemTime;
@@ -54,6 +54,8 @@ pub struct FileInfo {
     pub has_changed: bool,
     #[cfg(target_family = "unix")]
     pub permissions: Option<u32>,
+    #[cfg(feature = "notify")]
+    pub tailing_enabled: Option<u64>,
 }
 
 pub enum FileError {
@@ -147,6 +149,8 @@ impl FileManager {
             has_changed: false,
             #[cfg(target_family = "unix")]
             permissions: get_permissions(path),
+            #[cfg(feature = "notify")]
+            tailing_enabled: None,
         };
         self.open_files.insert(path.to_owned(), id);
         self.file_info.insert(id, info);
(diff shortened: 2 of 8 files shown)


sjoshid:
@cmyr 
Tailing works with this branch. Since this goes hand in hand with xi mac....you wont be able to actually tail the file (maybe you can simulate rpc call??). Facing some conflicts with xi mac branch so I'll clean it up and open PR there. 
This PR is still not cleaned up. I have a bunch of custom todos which I'll clean up. Also clippy fails with 
```
error: value assigned to `is_tail_event` is never read
   --> core-lib/src/tabs.rs:737:17
    |
737 |         let mut is_tail_event = false;
    |                 ^^^^^^^^^^^^^
    |
    = note: `-D unused-assignments` implied by `-D warnings`
    = help: maybe it is overwritten before being read?
```
But these arent issues that will stop you from reviewing. 

If there is something you want me to do make it easier for you to review...let me know.

rust/core-lib/src/tabs.rs
```diff
@@ -741,6 +795,41 @@ impl CoreState {
         }
     }
 
+    #[cfg(feature = "notify")]
+    fn handle_open_file_tail_fs_event(&mut self, event: Event) {
+        use notify::event::*;
+
+        let path = match event.kind {
+            EventKind::Modify(ModifyKind::Any) => &event.paths[0],
```
sjoshid: I think I need to handle other kinds of event here too. Im playing around with this.

rust/core-lib/src/tabs.rs
```diff
@@ -708,7 +748,21 @@ impl CoreState {
...
-            | EventKind::Modify(ModifyKind::Any) => &event.paths[0],
+            | EventKind::Modify(ModifyKind::Any)
+                if event.flag().is_none() =>
+            {
+                &event.paths[0]
+            }
+            EventKind::Modify(ModifyKind::Any) if event.flag().is_some() => {
+                let flag_kind = event.flag().unwrap();
+
+                if Flag::Ongoing.eq(flag_kind) {
```
sjoshid: `Ongoing` events should not be handled here.

rust/core-lib/src/view.rs
```diff
@@ -869,8 +882,30 @@ impl View {
         let height = self.line_of_offset(text, text.len()) + 1;
         let plan = RenderPlan::create(height, self.first_line, self.height);
         self.send_update_for_plan(text, client, styles, style_spans, &plan, pristine);
-        if let Some(new_scroll_pos) = self.scroll_to.take() {
-            let (line, col) = self.offset_to_line_col(text, new_scroll_pos);
+        if !self.is_tail_enabled {
```
sjoshid: This check makes sure that when user scrolls a tailing file, we dont auto scroll to cursor location. Mimics `tail` command.

rust/core-lib/src/watcher.rs
```diff
@@ -98,7 +102,11 @@ impl FileWatcher {
         let state = Arc::new(Mutex::new(WatcherState::default()));
         let state_clone = state.clone();
 
-        let inner = watcher(tx_event, Duration::from_millis(100)).expect("watcher should spawn");
+        let mut inner =
+            watcher(tx_event, Duration::from_millis(100)).expect("watcher should spawn");
+        inner
+            .configure(Config::OngoingEvents(Some(Duration::from_millis(50))))
```
sjoshid: Any mills less than 100 would made sense here.

rust/core-lib/src/client.rs
```diff
@@ -235,6 +235,17 @@ impl Client {
     pub fn schedule_timer(&self, timeout: Instant, token: usize) {
         self.0.schedule_timer(timeout, token);
     }
+
+    /// Let client know tail was successfully enabled/disabled
```
Cogitri: What are the reasons for a client to disable this?

rust/core-lib/src/file.rs
```diff
@@ -105,6 +106,17 @@ impl FileManager {
         false
     }
 
+    pub fn tail(&mut self, path: &Path, id: BufferId) -> Result<Rope, FileError> {
+        if !path.exists() {
+            return Ok(Rope::from(""));
+        }
+
+        let existing_file_info = self.file_info.get(&id).expect("File info must be present");
```
Cogitri: Using `expect` here is kind of ugly.

rust/core-lib/src/file.rs
```diff
@@ -172,6 +190,79 @@ impl FileManager {
...
+        if let Some(file_info) = self.file_info.get_mut(&buffer_id) {
+            let path = &file_info.path.to_owned();
+
+            let mut f = File::open(path).map_err(|e| FileError::Io(e, path.to_owned()))?;
+            let end_position =
+                f.seek(SeekFrom::End(0)).map_err(|e| FileError::Io(e, path.to_owned()))?;
+
+            file_info.tail_position = Some(end_position);
+            self.watcher.unwatch(path, OPEN_FILE_EVENT_TOKEN);
+            self.watcher.watch(path, false, OPEN_FILE_TAIL_EVENT_TOKEN);
```
Cogitri: I don't know inotify too well, but do you always need to do this and not just when you didn't have tailing enabled prefviously?

rust/core-lib/src/file.rs
```diff
@@ -172,6 +190,79 @@ impl FileManager {
...
+where
+    P: AsRef<Path>,
+{
+    let mut f =
+        File::open(path.as_ref()).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+    let end_position =
+        f.seek(SeekFrom::End(0)).map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?;
+
+    let current_position = file_info.tail_position.expect(
+        "tail_position is None. Since we are tailing a file, tail_position cannot be None.",
```
Cogitri: Again, `expect` here isn't nice

rust/core-lib/src/file.rs
```diff
@@ -172,6 +190,79 @@ impl FileManager {
...
+    );
+    let diff = end_position - current_position;
+    let mut buf = vec![0; diff as usize];
+    f.seek(SeekFrom::Current(-(buf.len() as i64))).unwrap();
```
Cogitri: Isn't buf.len() == diff here?

Cogitri: Also, it'd be nice if we could avoid unwrap

rust/core-lib/src/file.rs
```diff
@@ -172,6 +190,79 @@ impl FileManager {
...
+    f.read_exact(&mut buf).unwrap();
+
+    let mod_time =
+        f.metadata().map_err(|e| FileError::Io(e, path.as_ref().to_owned()))?.modified().ok();
```
Cogitri:
```suggestion
        f.metadata().map_err(|e| FileError::Io(e, path.clone()))?.modified().ok();
```

Cogitri: (Not tested though since this currently doesn't apply to master)

rust/core-lib/src/tabs.rs
Cogitri: What other events do you have in mind?

rust/core-lib/src/tabs.rs
```diff
@@ -741,6 +795,41 @@ impl CoreState {
...
+            }
+        };
+
+        let buffer_id = match self.file_manager.get_editor(path) {
+            Some(id) => id,
+            None => return,
+        };
+
+        let has_changes = self.file_manager.check_file(path, buffer_id);
+        let is_pristine = self.editors.get(&buffer_id).map(|ed| ed.borrow().is_pristine()).unwrap();
```
Cogitri: Again, unwrapping isn't nice (unless None is a programmer error in which case the called code should probably `panic!` anyway)

rust/core-lib/src/view.rs
Cogitri: I suppose it'd be worth adding a comment to the source file about this.

rust/core-lib/src/client.rs
sjoshid: If client doesnt want to tail the file anymore, it'll disable tailing.

rust/core-lib/src/file.rs
sjoshid: Tailing is enabled only when user clicks "Tail" option under "Debug" menu. Just because a file is opened does not mean it will always be tailed. `disableTailing` is inverse of this. When tailing is disabled, the state goes back to the way it was before we started tailing.

rust/core-lib/src/tabs.rs
sjoshid: None actually. I forgot to update this comment.

rust/core-lib/src/view.rs
sjoshid: Will do.

rust/core-lib/src/file.rs
sjoshid: Will make this change and test it.

rust/core-lib/src/tabs.rs
sjoshid: Agree but this code is copied as is from original.

rust/core-lib/src/client.rs
sjoshid: If client doesnt want to tail file anymore.

rust/core-lib/src/file.rs
sjoshid: Because when a file is opened, doesnt mean it will always be tailed. This is inverse of `disableTailing` which resets the state to the way it was when file was opened.

rust/core-lib/src/file.rs
sjoshid: > Isn't buf.len() == diff here?

yes.

rust/core-lib/src/file.rs
sjoshid: Gives error
```
error[E0599]: no method named `clone` found for type `P` in the current scope
   --> core-lib/src/file.rs:252:56
    |
252 |         f.metadata().map_err(|e| FileError::Io(e, path.clone()))?.modified().ok();
    |                                                        ^^^^^ method not found in `P`
    |
    = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following trait defines an item `clone`, perhaps you need to restrict type parameter `P` with it:
    |
234 | fn try_tailing_file<P: std::clone::Clone>(path: P, file_info: &FileInfo) -> Result<(Rope, FileInfo), FileError>
    |                     ^^^^^^^^^^^^^^^^^^^^

```

rust/core-lib/src/client.rs
Cogitri: Ah, but wouldn't disabling tailing imply that the whole file is reloaded when something changes then (since inotify is still enabled)? I don't see the advantage of disabling tailing

rust/core-lib/src/file.rs
Cogitri: So it'd make more sense to just use diff here directly I think

rust/core-lib/src/file.rs
Cogitri: Ah, fair enough then

rust/core-lib/src/tabs.rs
Cogitri: Hm, I guess this returning None doesn't really happen then? @cmyr

Cogitri:
Could you please rebase this on master?

rust/core-lib/src/file.rs
Cogitri: Yes, but wouldn't it be sufficient to only call this when we actually enable/disable tailing then instead of calling this every time we update the position?

sjoshid: This method is only called when user enables tailing. Not every time file is updated.

rust/core-lib/src/file.rs
sjoshid: Ok. Will make the change.

sjoshid:
I have already rebased this on master.

rust/core-lib/src/client.rs
sjoshid: That is good point. So if i open up a log file and later on it gets constantly updated, should Xi also update the contents? Without explicitly giving the option to enable/disable, the end user experience would not be great as the file is constantly changing even though it might not be desirable. Also tailing mimics the `tail` command. Just like in in `tail` when you are at the bottom of file it'll keep scrolling, when you slightly scroll up, it stops. This behavior needs to be controllable too I think.
Having said that, I still think the way file is loaded currently is not efficient. Instead of loading delta, we load the whole file. I think this is what you are talking about. I havent changed existing things.

rust/core-lib/src/file.rs
Cogitri: Oh, got it, sorry.

rust/core-lib/src/client.rs
Cogitri: > I think this is what you are talking about. I havent changed existing things.

Ah no, I not exactly and it's entirely fine that you're not touching that. I was just curious: If you disabling tailing, won't file reloading be enabled instead and as such the editor will keep loading new contents added to the file? The only differences I see between having tailing enabled/disabled:

* With tailing enabled you'll scroll if you're at the end of the file
* Reloading the entire file is less efficient

So wouldn't just having a switch to disable the scrolling do the trick? You'd have efficient adding of content without scrolling. But maybe I just misunderstand something :)

sjoshid: Ok I see your point now. Thing is when tailing a file, I **APPEND** delta to the file. When tailing is disabled, all bets are off. A change in file (when tailing is disabled) might or might not be a delta change at the end of file. In that case it is harder to update the file I think. And for this reason, I think, we load the entire file when tailing is off.

Cogitri: Ah, alrighty. I had expected that if I have a file:

```
line 1
line 2
line 3
line 4
```

and I add a new line between line 2 and 3, line 2.5 that we'd only reload the file from line 2 then, but I  suppose that was wrong and I don't know inotify too well :)

Sounds good to me then.

sjoshid: Yeah. It gets a lot more complicated here. Besides, notify will only tell us something changed. It doesnt tell where what or where the change occurred.

Issue #1242, opened 2019-11-27
During `cargo install --path rust`, xi-core-lib fails to compile, "mismatched types"


ndrewtl:
# Xi-Editor Issue Template
- [x] I have searched existing issues and could not find my issue.
- [x] I have studied the documentation.

## Details

OS: Fedora 31

```sh
$ rustc --version
rustc 1.39.0 (4560ea788 2019-11-04)
```
rustc installed via rustup
Not using frontend

```sh
user@host:~/xi-editor % cargo install --path rust
  Installing xi-core v0.3.0 (/home/user/xi-editor/rust)
    Updating git repository `https://github.com/kornelski/rust-onig`
    Updating crates.io index
warning: Patch `onig v5.0.0 (https://github.com/kornelski/rust-onig?branch=default-bindgen#c9f5b5be)` was not used in the crate graph.
Patch `onig_sys v69.2.0 (https://github.com/kornelski/rust-onig?branch=default-bindgen#c9f5b5be)` was not used in the crate graph.
Check that the patched package version and available features are compatible
with the dependency requirements. If the patch has a different version from
what is locked in the Cargo.lock file, run `cargo update` to use the new
version. This may also occur with an optional dependency that is not enabled.
   Compiling xi-core-lib v0.3.0 (/home/user/xi-editor/rust/core-lib)
error[E0308]: mismatched types
   --> core-lib/src/watcher.rs:101:29
    |
101 |         let inner = watcher(tx_event, Duration::from_millis(100)).expect("watcher should spawn");
    |                             ^^^^^^^^ expected struct `crossbeam_channel::channel::Sender`, found a different struct `crossbeam_channel::channel::Sender`
    |
    = note: expected type `crossbeam_channel::channel::Sender<std::result::Result<notify::Event, notify::Error>>`
               found type `crossbeam_channel::channel::Sender<_>`
note: Perhaps two different versions of crate `crossbeam_channel` are being used?
   --> core-lib/src/watcher.rs:101:29
    |
101 |         let inner = watcher(tx_event, Duration::from_millis(100)).expect("watcher should spawn");
    |                             ^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: failed to compile `xi-core v0.3.0 (/home/user/xi-editor/rust)`, intermediate artifacts can be found at `/home/user/xi-editor/rust/target`

Caused by:
  could not compile `xi-core-lib`.

To learn more, run the command again with --verbose.
[...]

PR #1243, opened 2019-11-30
deps: update, only allow crossbeam-channel ~> 0.3

### LINKED ISSUE #1240: Cargo install fails because of inconsistent crossbeam versions
# Details
Running `cargo run install -f .` on master fails with the message below (I think this was caused by upgrading notify in this [commit](https://github.com/xi-editor/xi-editor/commit/0b15cfe11c24d0b5f6f4fe74956f7e7308a28fdd), I can build it successfully 1 commit before).

```
error[E0308]: mismatched types
   --> core-lib/src/watcher.rs:101:29
    |
101 |         let inner = watcher(tx_event, Duration::from_millis(100)).expect("watcher should spawn");
    |                             ^^^^^^^^ expected struct `crossbeam_channel::channel::Sender`, found a different struct `crossbeam_channel::channel::Sender`
    |
    = note: expected type `crossbeam_channel::channel::Sender<std::result::Result<notify::Event, notify::Error>>`
               found type `crossbeam_channel::channel::Sender<_>`
```
Environment
```
Ubuntu 18.04
rustc 1.38.0 (625451e37 2019-09-23)
```

Cogitri:
notify-0.5 only supports crossbeam-channel 0.3 as of now.

fixes #1240 #1242

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -14,17 +14,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "anymap"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "arrayref"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "arrayvec"
-version = "0.4.11"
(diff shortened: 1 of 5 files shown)


Cogitri:
Thanks

PR #1247, opened 2020-01-03
Rustfmt for 1.40

cmyr:
🤷‍♀


cmyr:
merging this as-is; we also need clippy fixes but that can be separate work.

Issue #1199, opened 2019-06-22
Move editing functions out of editor.rs


cmyr:
Similar to #1198, we would like to be able to use editing functions independently of an `Editor` or a runloop.

For more detail, see #1187. For an illustration of how this could be done, see the [`edit_ops` mod in the `xi-modal` branch](https://github.com/cmyr/xi-editor/blob/xi-modal/rust/core-lib/src/editor.rs#L955). (Note that having this be in a submodule of editor.rs isn't desirable long term).

A goal of this work should be to make it much easier to write simple tests of various edit operations.

KeeganMyers:
I am a relative beginner to rust and completely new to the code base. Would this be a decent 1st PR opportunity or should I look into another issue? If so I can probably start looking at this over the weekend.

cmyr:
This is definitely a little bit of a project, but it isn't crazy, and the linked resources will be useful. It would certainly be a good way to become more familiar with the codebase.

Iamshankhadeep:
I am new to this community and rust, Can I work on this issue?

nangtrongvuon:
@Iamshankhadeep Sure, feel free to dive into this and ask around on the xi zulipchat if you need help!

Iamshankhadeep:
@nangtrongvuon sorry I am a little bit busy with something so I will start working on it this Sunday . Is it okay?

TDecki:
I think this issue can be closed now.

LiHRaM:
Yep!

Issue #1198, opened 2019-06-22
Move selection modification out of view.rs


cmyr:
A user of the `xi-core` library should be able to use and modify selections without needing to use a `View`. 

The main challenge here is that we need to introduce some abstraction for representing the things that are currently provided by the view, such as navigation up and down by arrow keys, which depends on line breaks.

My inclination is to introduce a trait that describes this behaviour.

For more details see #1187; also a version of this exists in my `xi-modal` branch (used for the rust playground): [see here](https://github.com/cmyr/xi-editor/blob/xi-modal/rust/core-lib/src/view.rs#L64).

PR #1263, opened 2020-04-27
Add an edit_ops.rs file

TDecki:
## Summary
This PR does the following:
* It moves central editing operation away from editor.rs, closing #1199.
* Said operations no longer require a `View` to operate, but take a `&[SelRegion]` instead.
   * This will help with #1198, as someone working with raw selections no
      longer has to construct a `View` to use them for these functions.
   * The operations used the `View` not just to get selections, but also for  functions
      like `offset_of_line` to get the lines. In case of `View` these are _visual_ lines, not logical ones.
      Activating word wrap thus changed their behaviour, with special mention to
      `duplicate_line`, which was extremly broken. By making them think in logical lines,
      many previously undiscovered bugs have been fixed.
* It also applies the changes demanded by clippy 0.0.212.

### DIFF
--- a/rust/core-lib/src/backspace.rs
+++ b/rust/core-lib/src/backspace.rs
@@ -16,7 +16,7 @@
 use xi_rope::{Cursor, Rope};
 
 use crate::config::BufferItems;
-use crate::line_offset::{DefaultLineOffset, LineOffset};
+use crate::line_offset::{LineOffset, LogicalLines};
 use crate::selection::SelRegion;
 use xi_unicode::*;
 
@@ -26,7 +26,7 @@ pub fn offset_for_delete_backwards(region: &SelRegion, text: &Rope, config: &Buf
         region.min()
     } else {
         // backspace deletes max(1, tab_size) contiguous spaces
-        let (_, c) = DefaultLineOffset.offset_to_line_col(&text, region.start);
+        let (_, c) = LogicalLines.offset_to_line_col(&text, region.start);
 
         let tab_off = c % config.tab_size;
         let tab_size = config.tab_size;
--- a/rust/core-lib/src/backspace.rs
+++ b/rust/core-lib/src/backspace.rs
@@ -16,14 +16,13 @@
 use xi_rope::{Cursor, Rope};
 
 use crate::config::BufferItems;
-use crate::line_offset::LineOffset;
+use crate::line_offset::{LineOffset, DefaultLineOffset};
 use crate::selection::SelRegion;
 use crate::view::View;
 use xi_unicode::*;
 
 #[allow(clippy::cognitive_complexity)]
 pub fn offset_for_delete_backwards(
-    view: &View,
     region: &SelRegion,
     text: &Rope,
     config: &BufferItems,
@@ -32,7 +31,7 @@ pub fn offset_for_delete_backwards(
         region.min()
     } else {
         // backspace deletes max(1, tab_size) contiguous spaces
-        let (_, c) = view.offset_to_line_col(&text, region.start);
+        let (_, c) = DefaultLineOffset.offset_to_line_col(&text, region.start);
 
         let tab_off = c % config.tab_size;
         let tab_size = config.tab_size;
--- a/rust/core-lib/src/backspace.rs
+++ b/rust/core-lib/src/backspace.rs
@@ -16,16 +16,12 @@
 use xi_rope::{Cursor, Rope};
 
 use crate::config::BufferItems;
-use crate::line_offset::{LineOffset, DefaultLineOffset};
+use crate::line_offset::{DefaultLineOffset, LineOffset};
 use crate::selection::SelRegion;
 use xi_unicode::*;
 
 #[allow(clippy::cognitive_complexity)]
-pub fn offset_for_delete_backwards(
-    region: &SelRegion,
-    text: &Rope,
-    config: &BufferItems,
-) -> usize {
+pub fn offset_for_delete_backwards(region: &SelRegion, text: &Rope, config: &BufferItems) -> usize {
     if !region.is_caret() {
         region.min()
     } else {
--- a/rust/core-lib/src/backspace.rs
+++ b/rust/core-lib/src/backspace.rs
@@ -18,7 +18,6 @@ use xi_rope::{Cursor, Rope};
 use crate::config::BufferItems;
 use crate::line_offset::{LineOffset, DefaultLineOffset};
 use crate::selection::SelRegion;
-use crate::view::View;
 use xi_unicode::*;
 
 #[allow(clippy::cognitive_complexity)]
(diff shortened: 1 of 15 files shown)


cmyr:
sorry I didn't notice this and fixed the various lints separately, want to rebase and I'll take a look?

TDecki:
Rebased.

rust/core-lib/src/line_offset.rs
```diff
@@ -12,15 +12,21 @@
...
+use crate::linewrap::Lines;
+use crate::selection::SelRegion;
+
 /// A trait from which lines and columns in a document can be calculated
 /// into offsets inside a rope an vice versa.
 pub trait LineOffset {
     // use own breaks if present, or text if not (no line wrapping)
 
-    /// Returns the byte offset corresponding to the given visual line.
+    /// Returns the byte offset corresponding to the given line.
```
cmyr, line 29: `visual` is an important distinction, here; it indicates a distinction from 'logical' lines, which are just lines that end in a newline character. a 'visual' line may not end in a line break, and may be the result of word wrap.

I guess this whole comment is slightly out of date, having been moved from `view.rs`?

rust/core-lib/src/tabs.rs
```diff
@@ -19,6 +19,8 @@
 //! This file is called 'tabs' for historical reasons, and should probably
 //! be renamed.
 
+#![allow(clippy::clone_on_copy)]
```
cmyr, line 22: what's the rationale for this? I'd prefer to fix whatever the problem was.

rust/rope/src/delta.rs
```diff
@@ -16,6 +16,8 @@
 //! It's useful to explicitly represent these operations so they can be
 //! shared across multiple subsystems.
 
+#![allow(clippy::many_single_char_names)]
```
cmyr, line 19: I would prefer this to be on a specific function of that is possible?

rust/core-lib/src/line_offset.rs
```diff
@@ -12,15 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::linewrap::Lines;
+#![allow(clippy::range_plus_one)]
```
cmyr, line 15: again I would prefer us to fix the lint rather than allow it, unless there's a good reason not to.

rust/core-lib/src/backspace.rs
```diff
@@ -16,23 +16,17 @@
...
-    text: &Rope,
-    config: &BufferItems,
-) -> usize {
+pub fn offset_for_delete_backwards(region: &SelRegion, text: &Rope, config: &BufferItems) -> usize {
     if !region.is_caret() {
         region.min()
     } else {
         // backspace deletes max(1, tab_size) contiguous spaces
-        let (_, c) = view.offset_to_line_col(&text, region.start);
+        let (_, c) = DefaultLineOffset.offset_to_line_col(&text, region.start);
```
cmyr, line 29: this seems to be changing the behaviour? by not using the view we may get different results here when word wrap is enabled. I'm not sure if this is a problem in practice, but it feels fishy.

rust/core-lib/src/editor.rs
```diff
@@ -691,171 +425,156 @@ impl Editor {
...
-                    let word = self.text.slice_to_cow(start..end);
-
-                    // first letter is uppercase, remaining letters are lowercase
-                    let (first_char, rest) = word.split_at(1);
-                    let capitalized_text =
-                        [first_char.to_uppercase(), rest.to_lowercase()].concat();
-                    builder.replace(interval, Rope::from(capitalized_text));
+            Transpose => {
+                let delta = edit_ops::transpose(&self.text, regions);
+                if !delta.is_identity() {
```
cmyr, line 472: we need to set `EditType::Transpose` here. It is probably worth doing another close read and make sure the edit type is correct in other places as well?

rust/core-lib/src/editor.rs
```diff
@@ -691,171 +425,156 @@ impl Editor {
...
-        let mut builder = DeltaBuilder::new(self.text.len());
-        for region in view.sel_regions() {
-            let mut cursor = WordCursor::new(&self.text, region.end);
-            let (mut start, end) = cursor.select_word();
-
-            // if the word begins with '-', then it is a negative number
-            if start > 0 && self.text.byte_at(start - 1) == (b'-') {
-                start -= 1;
+        match cmd {
+            Delete { movement, kill } => {
```
cmyr, line 446: reading through this patch, I think my personal inclination would be to keep using the separate `do_$thing` methods, and then just in those methods call out to `edit_ops`; a major advantage of this approach is that it makes this diff easier to read and verify.

TDecki: I've reversed this.
I didn't want multiple related functions with the same name doing different things when I made this. For now, the functions in `editor.rs` that call `edit_ops.rs` functions have the `do_`-prefix, while the functions in `edit_ops.rs` do not have that prefix.

rust/core-lib/src/editor.rs
TDecki: Fixed. There was indeed another case, but I belive that got them all.

(Are we testing undo and redo behaviour?)

rust/core-lib/src/backspace.rs
TDecki: `view.line_of_offset` calls `lines.visual_line_of_offset`, so different behaviour on word wrap was the default and the change a bugfix (see PR summary).
I've renamed `DefaultLineOffset` to `LogicalLines`, which should make the change more obvious (and `DefaultLineOffset` more descriptive).

rust/rope/src/delta.rs
TDecki: Fixed.

rust/core-lib/src/line_offset.rs
TDecki: I've copied this line together with the `get_line_range` function from `view.rs`.

rust/core-lib/src/tabs.rs
TDecki: Me getting annoyed with clippy. Fixed.

rust/core-lib/src/line_offset.rs
TDecki: (I hope this is a good reason.)

rust/core-lib/src/line_offset.rs
TDecki: `LineOffset` deals with both visual lines and logical lines, so this comment is indeed out of date.

I don't think that one trait coming with two behaviours is ideal in the long run,
but perhaps there is a way out of this.

cmyr: Right, so this was an intentional choice, to the best of my recollection; the idea is that in the absence of line breaks, the hard breaks *are* the visual lines. This means that all edit ops can work with just a buffer, but things 'just work' when you add a view?

It may be that this isn't helpful, but that was the original thinking.

rust/core-lib/src/editor.rs
cmyr: There are a few tests around recording that exercise undo/redo, but overall not really. 😒

TDecki:
I don't have any further tweaks to make. Please go ahead and merge.

cmyr:
Thanks!

PR #1265, opened 2020-04-28
Merging selections now preserves the direction of the most recent

### LINKED ISSUE #323: multi selection polish

shtanton:
## Summary
Merging 2 selections if one is forward and one is backward, now follows the direction of the more recent instead of always forward.

## Related Issues
Related to #323

closes #323

## Review Checklist

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -7,3 +7,4 @@ build
 
 # Editor configs
 .idea/
+.vscode/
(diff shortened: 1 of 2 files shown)


cmyr:
you may need to rebase, I just pushed some clippy fixes.

PR #1280, opened 2020-06-26
Fix panic in is_simple_delete

TDecki:
In `is_simple_delete`, it is possible that `els.is_empty()` is true while `self.base_len > 0` is false - it can occur when the buffer is empty and the user tries to delete something.
In that case the statement in the `if let` below will then try to access a nonexisting element, causing a panic.

This PR fixes this.

### DIFF
--- a/rust/rope/src/delta.rs
+++ b/rust/rope/src/delta.rs
@@ -90,7 +90,7 @@ impl<N: NodeInfo> Delta<N> {
 
     /// Returns `true` if this delta represents a single deletion without
     /// any insertions.
-    /// 
+    ///
     /// Note that this is `false` for the trivial delta, as well as for a deletion
     /// from an empty `Rope`.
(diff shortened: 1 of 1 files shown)


rust/rope/src/delta.rs
```diff
@@ -91,8 +91,8 @@ impl<N: NodeInfo> Delta<N> {
     /// Returns `true` if this delta represents a single deletion without
     /// any insertions. Note that this is `false` for the trivial delta.
```
cmyr, line 92:
```suggestion
    /// Returns `true` if this delta represents a single deletion without
    /// any insertions.
    ///
    /// Note that this is `false` for the trivial delta, as well as for a deletion
    /// from an empty `Rope`.
    
```

TDecki: Fixed.

cmyr:
ugh sorry that's annoying

PR #1284, opened 2020-06-26
Improve memory efficency of rope diffing

### LINKED ISSUE #946: improve memory efficiency of rope diffing

TDecki:
Fixes #946 . I've used the first approach using a `RopeSlice`-struct. I've tried to stay as close as possible to the original code,
and my measurements confirm a reduced memory consumption for `test_larger_diff`.

### DIFF
--- a/rust/rope/src/diff.rs
+++ b/rust/rope/src/diff.rs
@@ -14,14 +14,15 @@
 
 //! Computing deltas between two ropes.
 
-use std::borrow::Cow;
 use std::collections::HashMap;
+use std::hash::{Hash, Hasher};
+use std::ops::Range;
 
 use crate::compare::RopeScanner;
 use crate::delta::{Delta, DeltaElement};
 use crate::interval::Interval;
 use crate::rope::{LinesMetric, Rope, RopeDelta, RopeInfo};
-use crate::tree::{Node, NodeInfo};
+use crate::tree::{Cursor, Node, NodeInfo};
 
 /// A trait implemented by various diffing strategies.
 pub trait Diff<N: NodeInfo> {
@@ -65,6 +66,18 @@ impl Diff<RopeInfo> for LineHashDiff {
             return builder.to_delta(base, target);
         }
 
+        // if a continuous range of text got deleted, we're done
+        if target.len() < base.len() && start_offset + diff_end == target.len() {
+            builder.copy(base.len() - diff_end, target_end, diff_end);
+            return builder.to_delta(base, target);
+        }
+
+        // if a continuous range of text got inserted, we're done
+        if target.len() > base.len() && start_offset + diff_end == base.len() {
+            builder.copy(base.len() - diff_end, target_end, diff_end);
+            return builder.to_delta(base, target);
+        }
+
         let line_hashes = make_line_hashes(&base, MIN_SIZE);
 
         let line_count = target.measure::<LinesMetric>() + 1;
@@ -74,10 +87,13 @@ impl Diff<RopeInfo> for LineHashDiff {
         let mut prev_base = 0;
 
         let mut needs_subseq = false;
-        for line in target.lines_raw(start_offset..target_end) {
+        for line in SliceIter::new(base, start_offset) {
+            let len = line.range.end - line.range.start;
             let non_ws = non_ws_offset(&line);
-            if line.len() - non_ws >= MIN_SIZE {
-                if let Some(base_off) = line_hashes.get(&line[non_ws..]) {
+            if len - non_ws >= MIN_SIZE {
+                if let Some(base_off) = line_hashes
+                    .get(&RopeSlice { rope: base, range: line.range.end + non_ws..line.range.end })
+                {
                     let targ_off = targ_line_offset + non_ws;
                     matches.push((start_offset + targ_off, *base_off));
                     if *base_off < prev_base {
@@ -86,7 +102,7 @@ impl Diff<RopeInfo> for LineHashDiff {
                     prev_base = *base_off;
                 }
             }
-            targ_line_offset += line.len();
+            targ_line_offset += len;
         }
 
         // we now have an ordered list of matches and their positions.
@@ -192,8 +208,23 @@ fn longest_increasing_region_set(items: &[(usize, usize)]) -> Vec<(usize, usize)
 }
 
 #[inline]
-fn non_ws_offset(s: &str) -> usize {
-    s.as_bytes().iter().take_while(|b| **b == b' ' || **b == b'\t').count()
+fn non_ws_offset(s: &RopeSlice) -> usize {
+    let mut cursor = Cursor::new(s.rope, s.range.start);
+
+    let mut result = 0;
+    let mut looping = true;
+    while looping {
+        if let Some(ch) = cursor.next_codepoint() {
+            if ch == ' ' || ch == '\t' {
+                result += 1;
+            } else {
+                looping = false;
+            }
+        } else {
+            looping = false;
+        }
+    }
+    result
 }
 
 /// Represents copying `len` bytes from base to target.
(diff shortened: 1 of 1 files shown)


rust/rope/src/diff.rs
```diff
@@ -245,21 +276,85 @@ impl DiffBuilder {
...
+                    looping = c1 == c2;
+                } else {
+                    return false;
+                }
+            } else {
+                return it2.next().is_none();
+            }
+        }
+
+        false
```
Menschenkindlein, line 307: This way it is possible to avoid resetting `looping` variable.
```rust
loop {
    let c1 = it1.next();
    let c2 = it2.next();

    if c1 != c2 {
        return false;
    }

    if c1.is_none() {
        return true;
    }
}
```

rust/rope/src/diff.rs
```diff
@@ -192,8 +208,23 @@ fn longest_increasing_region_set(items: &[(usize, usize)]) -> Vec<(usize, usize)
...
+    let mut looping = true;
+    while looping {
+        if let Some(ch) = cursor.next_codepoint() {
+            if ch == ' ' || ch == '\t' {
+                result += 1;
+            } else {
+                looping = false;
+            }
+        } else {
+            looping = false;
```
Menschenkindlein, line 224: `break` would do the job.

rust/rope/src/diff.rs
TDecki: Fixed.

rust/rope/src/diff.rs
TDecki: Fixed. This loop also used `while let` now.

Tiggilyboo (approved):

rust/rope/src/diff.rs
```diff
@@ -65,6 +66,18 @@ impl Diff<RopeInfo> for LineHashDiff {
...
+        if target.len() < base.len() && start_offset + diff_end == target.len() {
+            builder.copy(base.len() - diff_end, target_end, diff_end);
+            return builder.to_delta(base, target);
+        }
+
+        // if a continuous range of text got inserted, we're done
+        if target.len() > base.len() && start_offset + diff_end == base.len() {
+            builder.copy(base.len() - diff_end, target_end, diff_end);
+            return builder.to_delta(base, target);
+        }
```
cmyr, line 79: It might make sense to have this be a separate patch, since it doesn't seem to be really related to the other changes here, and it might be worth having anyway? It would be nice to measure and see that this is a reasonable speedup in these cases (it makes sense that it would be, but it's nice to measure this sort of code).

rust/rope/src/diff.rs
```diff
@@ -74,10 +87,13 @@ impl Diff<RopeInfo> for LineHashDiff {
         let mut prev_base = 0;
 
         let mut needs_subseq = false;
-        for line in target.lines_raw(start_offset..target_end) {
+        for line in SliceIter::new(base, start_offset) {
+            let len = line.range.end - line.range.start;
```
cmyr, line 91: I would probably just add a `len()` method to this type.

rust/rope/src/diff.rs
```diff
@@ -74,10 +87,13 @@ impl Diff<RopeInfo> for LineHashDiff {
...
             let non_ws = non_ws_offset(&line);
-            if line.len() - non_ws >= MIN_SIZE {
-                if let Some(base_off) = line_hashes.get(&line[non_ws..]) {
+            if len - non_ws >= MIN_SIZE {
+                if let Some(base_off) = line_hashes
+                    .get(&RopeSlice { rope: base, range: line.range.end + non_ws..line.range.end })
```
cmyr, line 95: this should be `line.range.start+non_ws`? When I make this change though I notice that the deltas that are generated are different from what were previously generated, and it looks like there are a lot of empty `Copy` operations getting included; something else to investigate.

rust/rope/src/diff.rs
```diff
@@ -192,8 +208,18 @@ fn longest_increasing_region_set(items: &[(usize, usize)]) -> Vec<(usize, usize)
...
+    let mut cursor = Cursor::new(s.rope, s.range.start);
+
+    let mut result = 0;
+    while let Some(ch) = cursor.next_codepoint() {
+        if ch == ' ' || ch == '\t' {
+            result += 1;
+        } else {
+            break;
+        }
+    }
```
cmyr, line 221: performance matters here, and doing codepoint calculations adds a bunch of expense; since we're only looking for acsii characters it's much cheaper to work with bytes.

rust/rope/src/diff.rs
```diff
@@ -245,21 +271,82 @@ impl DiffBuilder {
...
+#[derive(Debug, Clone)]
+pub struct RopeSlice<'a> {
+    rope: &'a Rope,
+    range: Range<usize>,
+}
+
+impl PartialEq for RopeSlice<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        let mut it1 =
+            self.rope.iter_chunks(Interval::from(self.range.clone())).flat_map(|x| x.chars());
```
cmyr, line 284: ditto here, and `chars`; we want to use `bytes`.

I would just add a method to `RopeSlice`, like

```rust
impl<'a> RopeSlice<'a> {
    fn iter_bytes(&'a self) -> impl Iterator<Item=u8> + 'a {
            self.rope.iter_chunks(self.range.clone()).flat_map(|chunk| chunk.bytes())
        }
}
```

rust/rope/src/diff.rs
```diff
@@ -245,21 +271,82 @@ impl DiffBuilder {
...
+        let mut it2 =
+            other.rope.iter_chunks(Interval::from(other.range.clone())).flat_map(|x| x.chars());
```
cmyr, line 286: also we can omit the `Interval::from` here and just pass a `Range`.

rust/rope/src/diff.rs
```diff
@@ -245,21 +271,82 @@ impl DiffBuilder {
...
+        }
+    }
+}
+
+impl Eq for RopeSlice<'_> {}
+
+impl Hash for RopeSlice<'_> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        let iter =
+            self.rope.iter_chunks(Interval::from(self.range.clone())).flat_map(|x| x.chars());
```
cmyr, line 308: ditto chars being slow here.

It might be worth looking into what the contract is around the hash impls of `std` types; I suspect we can actually just hash the chunks returned by `iter_chunks` one by one and get the same hash for a given string, regardless of how it is broken up.

cmyr (changes requested):
I just spent a while with this, and I'm actually a bit confused; the issue referenced in #946 was actually addressed in #1137, a little over a year ago.

So: I'm not totally sure where this patch fits. Running the benchmarks, it appears that we're looking at a significant slowdown (~2x) over the existing implementation, and I think memory savings would have to be pretty significant for us to want to take that tradeoff. If you can included any measurements that would be helpful; as-is I'm not sure this is buying us much 😞

I've reviewed this as I would if we were going to merge it, and there are some little bugs; it's quite possible that performance will be improved significantly when those are addressed.

rust/rope/src/diff.rs
TDecki: I've separated this in #1295. It _is_ an improvement.

PR #1291, opened 2020-09-04
[xi-unicode] Explicit breaks at end of text

### LINKED ISSUE #1289: xi-unicode: detecting break at end of string
`xi_unicode::LineBreakIterator` always reports a hard-break at the end of its input. Fine, but I need to know (for `kas-text`; also affects @alexheretic's `glyph-brush-layout`) whether my document ends with an explicit line-break or not. Is there a clean way to do this?

The `linebreak_property` fn is exposed but with no documentation as to its return value (and it's not easy to tell from the code's state machine look-up).

(There is also the question of whether one can even tell by looking at only the last char — though `\r\n` is essentially equivalent to `\n`.)

Perhaps the easiest solution is to match the last char against any of the [BK, CR, LF or NL classes](https://www.unicode.org/reports/tr14/#Table1)?

raphlinus:
Instead of always indicating end of text as a hard break, report whether
the text ends with an explicit line break.

Document the behavior.

Document the line break properties, and also make the ordering
consistent with ICU.

This is a breaking change but hopefully should be useful and not cause
too much trouble.

Fixes #1289


rust/unicode/src/lib.rs
```diff
@@ -23,6 +23,10 @@ use core::cmp::Ordering;
 
 use crate::tables::*;
 
+/// The Unicode line breaking property of the given code point.
+///
+/// This is given as a numeric value which matches the ULineBreak
+/// enum value from ICU.
```
cmyr, line 29: Is it worth having an actual enum for this? could be `#[repr(u8)]`.

rust/unicode/src/lib.rs
```diff
@@ -37,8 +41,11 @@ pub fn linebreak_property(cp: char) -> u8 {
...
 }
 
-// Return property, length
-// May panic if ix doesn't point to a valid character in the string
+/// The Unicode line breaking property of the given code point.
+///
+/// Look up the line breaking property for the first code point in the
+/// string. Return the property as a numeric value, and also the utf-8
+/// length of the codepoint, for convenience.
 pub fn linebreak_property_str(s: &str, ix: usize) -> (u8, usize) {
```
cmyr, line 49: If this is legitimately a public API I would return a struct with named fields. 🤷

rust/unicode/src/lib.rs
```diff
@@ -85,13 +97,15 @@ impl<'a> Iterator for LineBreakIterator<'a> {
                 Ordering::Equal => {
                     // LB3, break at EOT
                     self.ix += 1;
-                    return Some((self.s.len(), true));
+                    let i = (self.state as usize) * N_LINEBREAK_CATEGORIES;
+                    let new = LINEBREAK_STATE_MACHINE[i];
+                    return Some((self.s.len(), new >= 0xc0));
```
cmyr, line 102: although not part of this PR, I also dislike that this iterator returns a tuple; again I think a struct with named fields would make for a more ergonomic API.

rust/unicode/src/tables.rs
```diff
@@ -1497,15 +1498,15 @@ pub const LINEBREAK_STATE_MACHINE: [u8; 3827] = [
...
     128, 129, 130, 131, 132, 133, 6, 135, 8, 137, 10, 11, 140, 141, 142, 143,
     16, 17, 146, 147, 148, 149, 150, 151, 152, 153, 26, 27, 28, 29, 30, 159,
     160, 161, 162, 163, 36, 165, 166, 167, 168, 169, 170,
-    // state 87: SP+ EB
+    // state 87: SP+ EM
     128, 129, 130, 131, 132, 133, 6, 135, 8, 137, 10, 11, 140, 141, 142, 143,
     16, 17, 146, 147, 148, 149, 150, 151, 152, 153, 26, 27, 28, 29, 30, 159,
     160, 161, 162, 163, 36, 165, 166, 167, 168, 169, 170,
-    // state 88: SP+ EM
+    // state 88: SP+ ZWJ
```
cmyr, line 1509: can you summarize what the table changes represent?

cmyr (approved):
This looks reasonable. I have some API notes you're free to ignore as well as a question about the changes to the tables, which I do not understand.

rust/unicode/src/lib.rs
raphlinus: I agree this would be better, but didn't take it on in this PR because it's a fairly major change, especially with all the autogeneration.

These are definitely expert level APIs. Also, as an evolution, it's likely that property value lookup will be moving to icu4x. So I'm not sure how much I want to invest in this long term.

rust/unicode/src/lib.rs
raphlinus: This is basically following the tradition of char_indices :) But I hear you.

rust/unicode/src/tables.rs
raphlinus: Yes, when I verified that the numeric values were the same as ICU ULineBreak, I found that I had gotten the order wrong, so this just renumbers those properties. The table has the same semantics, just permuted a bit.

raphlinus:
I think I'll merge this now, and take on your suggestions if and when I rev the API in the future. There's other work to be done, like updating to Unicode 13, which is not trivial. Thanks for the review!

dhardy:
Any indication of when 0.3.0 is going to be released?

raphlinus:
I can push a release today. Any other concerns from anyone before I do?

raphlinus:
Ok, 0.3.0 is published. Let me know if there are any issues.

PR #1293, opened 2020-09-08
Expose push_subseq method on TreeBuilder

raphlinus:
It's cleaner to have this method on TreeBuilder than Tree. It's also a
useful an important method; the general template for any incremental
algorithm is to use a TreeBuilder to assemble its result, and I think we
should encourage its use.

### DIFF
--- a/rust/rope/src/tree.rs
+++ b/rust/rope/src/tree.rs
@@ -377,38 +377,10 @@ impl<N: NodeInfo> Node<N> {
         M::measure(&self.0.info, self.0.len)
     }
 
-    pub(crate) fn push_subseq(&self, b: &mut TreeBuilder<N>, iv: Interval) {
-        if iv.is_empty() {
-            return;
-        }
-        if iv == self.interval() {
-            b.push(self.clone());
-            return;
-        }
-        match self.0.val {
-            NodeVal::Leaf(ref l) => {
-                b.push_leaf_slice(l, iv);
-            }
-            NodeVal::Internal(ref v) => {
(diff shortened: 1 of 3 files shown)


raphlinus:
One question for discussion: I preserved the name "subseq", which was chosen at the time I think largely to accommodate the needs of span builders, but I think perhaps `push_slice` might be a cleaner name, and more consistent with `push_leaf_slice` already exists.

rust/rope/src/tree.rs
```diff
@@ -569,6 +541,35 @@ impl<N: NodeInfo> TreeBuilder<N> {
         }
     }
 
+    /// Push a subsequence of a rope.
```
cmyr, line 544: This should be clearer about what the `iv` argument is.

cmyr (commented):
I do think this is a good time to get the naming and docs right, here.

I think `push_slice` or `push_interval` probably makes sense. I would also clarify the behaviour in the docs, since I think the arguments are not self-explanatory.

raphlinus:
Changed the name and added some docstring.

There's a bunch of potential cleanup to do here. For one, I'm not sure `Interval` is pulling its weight, it should probably be `Range`.

Aside from cleanup, for use as an immutable vector in Druid there's functionality we should add, especially in-place mutation. But overall I want to be conscious of time. If there's a person interested in taking on xi-rope improvements, it's something I could mentor.

cmyr (approved):
Docs are good, thanks!

PR #1302, opened 2020-11-16
Implement UpdateOp::update and fix line invalidation in the middle of the cache

### LINKED ISSUE #1228: Frontend protocol: why cursor updates are bundled with text updates?
Hi,

I think I probably found a behavioral issue in the frontend protocol.

The question is, why does the protocol bundle the cursor updates with text updates?
For instance, here are the notifications sent by xi-core when I move the cursor down from line "mmm" to line "nnn". (The client is, quite expected, using a line cache that is storing a contiguous vector of recently rendered lines printed to the client window)

 1. copy n=12 (instructs to copy 12 lines from the old cache before "mmm". 12 is just some irrelevant # of lines before the cursor)
   
 2. ins n=2 (instructs to insert 2 new lines to the cache)
   
  * line: mmm
  * line: nnn cursor=[0]

 3. skip n=2 (instructs to skip 2 lines from the old cache, which I guess are "mmm"+cursor and "nnn")

 4. copy n = 33 (copy remaining lines from the old cache)

Effectively, this bunch of operations notifies the client of a cursor move. The text itself remains intact.
But there is no direct way for the client to know that the text hasn't been changed, unless the client starts explicitly comparing the skipped old lines with the incoming lines (which is ineffective).

Btw the frontend spec has an "update" operation which can move the cursor, but says it's currently unused. Why?

I'd add that my client is a console client, thus the cursor redrawing is performed as an individual operation after redrawing the text itself, and is done by simple termion::cursor::Goto. Thus, it's quite natural to expect cursor motion notifications as individual updates. Here is what I would expect from xi-core:

 1. copy n=45 (instructs to copy 45 lines from the old cache - this is the whole screen)
   
 2. update (imaginary operator instructing to reposition the cursor)
   
Thanks in advance. If that is needed I'll write a simple test that reproduces the aforementioned workflow.

### LINKED ISSUE #1301: xi-core invalidates a line in the middle of the line cache
(This is a duplicate of the correspondent [zulip topic](https://xi.zulipchat.com/#narrow/stream/147924-xi-editor/topic/invalidating.20a.20line.20in.20the.20middle.20of.20the.20linecache.3F). I'm creating it for the bookkeeping.)

The problem is: given a specially constructed  sequence of XML-RPC commands (see below) from the frontend, xi-core can generate a vector of UpdateOps that has an "invalidate" op in the middle of the sequence. This may be not expected by the xi frontends; eg. xrl explicitly [assumes](https://github.com/xi-frontend/xrl/blob/0.0.8/src/cache.rs#L6) that the invalidated lines occur only in the start or in the end of the cache.

I have created a simple xi-core test c4eater/xi-editor@cebb4ce that demonstrates the problematic sequence of commands:
```
client_started
new_view
insert "line 1"
insert newline
insert "line 2"
insert newline
// (...etc...)
insert "line 19"
insert newline

goto_line 1
goto_line 19
```

Which results in the following vector of ops in the end:
```
copy          n=1
invalidate    n=1 // oops
skip          n=1
copy          n=5
copy          n=11
ins           n=1
ins           n=1
```

To reproduce:
```
git clone git://github.com/c4eater/xi-editor
git co -f 140c588
cd xi-editor/rust/core-lib
cargo test --test rpc test_invalidate
```

Further investigation showed why this happens:

- View::send_update_for_plan() iterates over the line cache shadow segments and emits series of UpdateOps according to the segment's RenderTactic and position. In particular, if this iteration encounters a segment with the "Preserve" tactic and a tainted validity (anything different from line_cache_shadow::ALL_VALID), it emits UpdateOp::invalidate, the main troublemaker. As it turns out, it is possible to have such segment in the middle of the cache shadow.
- Xi-core's default render plan adds small 2-line "slops" (SCROLL_SLOP) before and after the visible line range. These slops use the "Preserve" RenderTactic. Let's choose a special line on such slop so that this line is not the starting or the ending line of the cache (eg slop on lines 0-1, choose line 1). If we jump on such slop line via ViewEvent::GotoLine and then jump elsewhere, the validity of the visited line in the cache will be (TEXT_VALID STYLES_VALID  !CURSOR_VALID), which makes it non-ALL_VALID. And thus, RenderTactic::Preserve + bad validity = UpdateOp::invalidate. As this line is not first/last, there will be other UpdateOps before/after this invalidate op, resulting in "invalidate" in the middle of the UpdateOps vector. This is exactly what is done in test_invalidate.

Bottom line: the culprit "invalidate" results from the cursor jumping on the slop line and then jumping away. The code of send_update_for_plan() actually has a [TODO comment](https://github.com/xi-editor/xi-editor/blob/v0.3.0/rust/core-lib/src/view.rs#L765) saying "in the case where it's ALL_VALID & !CURSOR_VALID, and cursors are empty, could send update removing the cursor".

Implementing UpdateOp::update should be the solution. This op is documented in the frontend protocol spec, but it is currently unused by xi-core. I have an implementation of this op that I'll shortly submit as a PR.

Should note that xi-core is still not giving any guarantees on the position of "invalidate" op in the ops vector. Having UpdateOp::update sent on cursor movement fixes the specific issue described above, but formally, the original behavior of xi-core is perfectly correct too; frontends should not make any assumptions on the precedence of UpdateOps.

Thanks

c4eater:
## Summary
This change introduces a new UpdateOp, "update", in addition to "copy", "insert", "invalidate" and "skip".

This UpdateOp was outlined in the design docs before me but was never actually implemented.

UpdateOp::update will be sent by View::send_update_for_plan() in case of an update of a line cache shadow segment if this update only moves the cursor (!CURSOR_VALID) and leaves the text and styles intact (TEXT_VALID && STYLES_VALID).

As a side effect, this change fixes the problem with an "invalidate" op showing up in the middle of the vector of ops.

It also fixes a couple minor bugs in line/region overlaps calculation in View::render_line().

This change requires the frontends to implement the support for UpdateOp::update (if not done already).

Because of the incompatible change in behavior, xi-core version is bumped to 0.4.

## Related Issues
Fixes #1301 
Fixes #1228 

## Review Checklist

### DIFF
--- a/rust/core-lib/Cargo.toml
+++ b/rust/core-lib/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "xi-core-lib"
-version = "0.3.0"
+version = "0.4.0"
 license = "Apache-2.0"
 authors = ["Raph Levien <raph@google.com>"]
 description = "Library module for xi-core"
--- a/rust/core-lib/src/view.rs
+++ b/rust/core-lib/src/view.rs
@@ -615,7 +615,7 @@ impl View {
     // Encode a single line with its styles and cursors in JSON.
     // If "text" is not specified, don't add "text" to the output.
     // If "style_spans" are not specified, don't add "styles" to the output.
-    fn render_line(
+    fn encode_line(
         &self,
         client: &Client,
         styles: &StyleMap,
@@ -670,7 +670,7 @@ impl View {
             result["text"] = json!(text.slice_to_cow(start_pos..pos));
         }
         if let Some(style_spans) = style_spans {
-            result["styles"] = json!(self.render_styles(
+            result["styles"] = json!(self.encode_styles(
                 client,
                 styles,
                 start_pos,
@@ -689,7 +689,7 @@ impl View {
         result
     }
 
-    pub fn render_styles(
+    pub fn encode_styles(
         &self,
         client: &Client,
         styles: &StyleMap,
@@ -699,7 +699,7 @@ impl View {
         hls: &Vec<Vec<(usize, usize)>>,
         style_spans: &Spans<Style>,
     ) -> Vec<isize> {
-        let mut rendered_styles = Vec::new();
+        let mut encoded_styles = Vec::new();
         assert!(start <= end, "{} {}", start, end);
         let style_spans = style_spans.subseq(Interval::new(start, end));
 
@@ -709,26 +709,26 @@ impl View {
         // same span exists in both sets (as when there is an active selection)
         for (index, cur_find_hls) in hls.iter().enumerate() {
             for &(sel_start, sel_end) in cur_find_hls {
-                rendered_styles.push((sel_start as isize) - ix);
-                rendered_styles.push(sel_end as isize - sel_start as isize);
-                rendered_styles.push(index as isize + 1);
+                encoded_styles.push((sel_start as isize) - ix);
+                encoded_styles.push(sel_end as isize - sel_start as isize);
+                encoded_styles.push(index as isize + 1);
                 ix = sel_end as isize;
             }
         }
         for &(sel_start, sel_end) in sel {
-            rendered_styles.push((sel_start as isize) - ix);
-            rendered_styles.push(sel_end as isize - sel_start as isize);
-            rendered_styles.push(0);
+            encoded_styles.push((sel_start as isize) - ix);
+            encoded_styles.push(sel_end as isize - sel_start as isize);
+            encoded_styles.push(0);
             ix = sel_end as isize;
         }
         for (iv, style) in style_spans.iter() {
             let style_id = self.get_or_def_style_id(client, styles, &style);
-            rendered_styles.push((iv.start() as isize) - ix);
-            rendered_styles.push(iv.end() as isize - iv.start() as isize);
-            rendered_styles.push(style_id as isize);
+            encoded_styles.push((iv.start() as isize) - ix);
+            encoded_styles.push(iv.end() as isize - iv.start() as isize);
+            encoded_styles.push(style_id as isize);
             ix = iv.end() as isize;
         }
-        rendered_styles
+        encoded_styles
     }
 
     fn get_or_def_style_id(&self, client: &Client, style_map: &StyleMap, style: &Style) -> usize {
@@ -830,12 +830,12 @@ impl View {
                             // !CURSOR_VALID; update cursors
                             let start_line = seg.our_line_num;
 
-                            let rendered_lines = self
+                            let encoded_lines = self
                                 .lines
                                 .iter_lines(text, start_line)
                                 .take(seg.n)
                                 .map(|l| {
-                                    self.render_line(
+                                    self.encode_line(
                                         client,
                                         styles,
                                         l,
@@ -848,7 +848,7 @@ impl View {
 
                             let logical_line_opt =
                                 if logical_line == 0 { None } else { Some(logical_line + 1) };
-                            ops.push(UpdateOp::update(rendered_lines, logical_line_opt));
+                            ops.push(UpdateOp::update(encoded_lines, logical_line_opt));
                         }
                         b.add_span(seg.n, seg.our_line_num, seg.validity);
                         line_num = seg.their_line_num + seg.n;
@@ -858,12 +858,12 @@ impl View {
                             b.add_span(seg.n, 0, 0);
                         } else if seg.tactic == RenderTactic::Render {
                             let start_line = seg.our_line_num;
-                            let rendered_lines = self
+                            let encoded_lines = self
                                 .lines
                                 .iter_lines(text, start_line)
                                 .take(seg.n)
                                 .map(|l| {
-                                    self.render_line(
+                                    self.encode_line(
                                         client,
                                         styles,
                                         l,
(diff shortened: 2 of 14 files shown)


c4eater:
@Cogitri can you pls review this or propose a reviewer? (See the linked zulip topic for details)

rust/core-lib/src/view.rs
```diff
@@ -612,28 +612,40 @@ impl View {
...
+            //
+            //  || (region.is_upstream() && c == pos)
+                || (region.is_upstream() && pos == region.start)
+            // TODO: Not clear why this case is special.
+            // The line overlap with the region should be 100% determined by
+            // region start/end, region affinity, and line start/end.
+            // No special cases here and no dependency on "text".
+            // Comment out the condition for now.
+            //
+            //  || (c == pos && c == text.len() && self.line_of_offset(text, c) == line_num)
```
Cogitri, line 648: Have you talked about these two with @cmyr or @raphlinus ? I have to admit I don't know the inner workings of xi too well, so not sure if this doesn't have unintended effects

rust/core-lib/src/view.rs
```diff
@@ -662,14 +674,17 @@ impl View {
...
-        let mut result = json!({
-            "text": &l_str,
-            "styles": styles,
-        });
+        let mut result = json!({});
 
+        if let Some(text) = text {
+            let l_str = text.slice_to_cow(start_pos..pos);
+            result["text"] = json!(&l_str);
+        }
```
Cogitri, line 682: Untested, but imho something like this would be nicer than having the temporary variable `l_str`:

```suggestion
        if let Some(text) = text.map(|s| s.slice_to_cow(start_pos..pos)) {
            result["text"] = json!(&text);
        }
```

rust/core-lib/src/view.rs
```diff
@@ -790,51 +805,112 @@ impl View {
...
+                                    self.render_line(
+                                        client, styles, l, /* text = */ None,
+                                        /* style_spans = */ None,
+                                    )
+                                })
+                                .collect::<Vec<_>>();
+
+                            let logical_line_opt =
+                                if logical_line == 0 { None } else { Some(logical_line + 1) };
+                            ops.push(UpdateOp::update(rendered_lines, logical_line_opt));
```
Cogitri, line 902: Isn't this the sme as on line 827? Could we factor this out into a seperate function or similiar so we don't have the same code twice?

Cogitri (changes requested):
Looks pretty good overall, just a few nits. I'll try to test this at the weekend (feel free to ping me in case I forget :)

rust/core-lib/src/view.rs
c4eater: AFAIK cmyr is busy and he cannot participate in code reviews.
This if clause is written by Raph but looks like he isn't coming in the comments.
I will try PMing him or notifying him on zulip.

I tested my whole change in a real xi-term session and didn't see any issues with moving over EOLs (which is what is covered by this clause). But there is still a chance that I'm missing something.

rust/core-lib/src/view.rs
c4eater: Correct. The same code duplication existed in the match arms before my change. This could have been done intentionally (for simplicity). But I will try to factorize it out if that doesn't make the call graph too complex.

c4eater:
> Looks pretty good overall, just a few nits. I'll try to test this at the weekend (feel free to ping me in case I forget :)

OK... From my side, I live-tested this change on xi-term. Don't see any problems so far.

Please note that you need a modified xrl if you want a field testing (in an xrl based editor). The upstream xrl (v0.0.8) has 2 problems preventing it from handling UpdateOp::update correctly: it does not allow empty "styles" in the "line" json (despite the frontend protocol spec), and it does not handle the "ln" (which is needed in the update op just as it is needed in the copy op if the line numbers are shifted).

I have prepared a dedicated branch with all the fixes needed: https://github.com/c4eater/xrl/tree/v0.1.0, just check it out and relink your xi client to this updated xrl.

rust/core-lib/src/view.rs
c4eater: OK, Raph provided some explanations on zulip and I also gave a second look on selection.rs myself. Bottom line: these checks are legal (save for a minor issue: can remove the line_of_offset check as it always passes) and the last one is indeed a special case.

So I'm taking back my edits here, and will not change the behavior of interval matching in render_line(). I'll only make a few cosmetic changes.

rust/core-lib/src/view.rs
c4eater: I'll inline it into `json!(...)`. Combining "if let" and map() can be harder for understanding.

rust/core-lib/src/view.rs
Cogitri: Okie :+1:

Cogitri (approved):

rust/core-lib/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "xi-core-lib"
-version = "0.3.0"
+version = "0.4.0"
```
Cogitri, line 3: Ah, just one more thing: Do you plan to do additional changes? If so, it might be worth it to delay 0.4 until after you made those.

c4eater: No, I'm not planning any other changes at the moment.

PR #1309, opened 2021-01-10
Mention maintenance status in readme

cmyr:
I feel like it's worth including something like this in the readme, at this point.


raphlinus (approved):
Good idea, thanks for the patch.

raphlinus:
I'm ok with this being a force-push, as it's a doc only change, and chasing clippy is a good example of the kinds of things we no longer want to spend time on.

PR #1318, opened 2021-08-24
warnings + clippy

TDecki:


rust/rope/src/delta.rs
```diff
@@ -673,7 +673,7 @@ impl<'a, N: NodeInfo> Iterator for DeletionsIter<'a, N> {
 
     fn next(&mut self) -> Option<Self::Item> {
         let mut result = None;
-        while let Some(elem) = self.els_iter.next() {
+        for elem in &mut self.els_iter {
```
Cogitri, line 676: Does this really have the same semantics? AFAIC the for loop will loop through all elements that are Some while the while loop will abort on the first element that isn’t Some

TDecki: Yes. The [official documentation](https://doc.rust-lang.org/nightly/std/iter/#for-loops-and-intoiterator) which provides pseudocode for a desugared for-loop explicitly terminates the loop once it encounteres a None.

This is also done on behalf of clippy, which is also a good indicator that these snippets are semantically equivalent.

cmyr (approved):
yep this all looks reasonable!

cmyr:
Thanks!

PR #1328, opened 2022-05-24
Add Lapce

ricklamers:
Not affiliated with Lapce, the description should probably be approved by https://github.com/dzhou121

## Summary
Add Lapce as one of the frontends to the README.

## Review Checklist


jansol:
AFAIK Lapce is not really a xi-core frontend at this point. It does use xi-rope and possibly some other parts of xi, though.

PR #1339, opened 2022-12-09
Add discontinuation notice to the Xi editor

PoignardAzur:
I know we like our "we may work again on this in the future" notices and to keep possibilities open, but as discussed on Zulip, we're extremely unlikely to actually do any meaningful work on this project in the next few years (or ever, arguably), and having a note that implies the opposite feels a bit dishonest.

This also adds a link to the Lapce project, which *is* actively maintained.


cmyr (approved):
Thanks!

